PR #160, opened 2015-04-30
Of/bulk bifrost delete

stevendanna:
From @oferrigni in chef/oc_bifrost#45:

> Adds the resource "bulk/delete" which accepts post with a request body:
> {
> "type" : "(actor|group)",
> "collection": ["authz_ids"]
> }
> 
> and returns error in the event of a server error and ok with body
> {
> "count": num_deleted
> }
> in the event of success. Num_deleted may be different than the length of the collection if some authz_ids are missing at time of request.

### DIFF
--- a/src/oc_bifrost/apps/bifrost/priv/dispatch.conf
+++ b/src/oc_bifrost/apps/bifrost/priv/dispatch.conf
@@ -26,6 +26,7 @@
 
 %% POST for bulk requests
 {["bulk"], bifrost_wm_bulk_resource, []}.
+{["bulk", "_delete"], bifrost_wm_bulk_delete_resource, []}.
 
 %% GET and DELETE of existing entities
 {["actors", id], bifrost_wm_named_resources, [{request_type, actor}]}.
--- a/src/oc_bifrost/apps/bifrost/priv/pgsql_statements.config
+++ b/src/oc_bifrost/apps/bifrost/priv/pgsql_statements.config
@@ -125,6 +125,14 @@
  <<"SELECT actor_has_bulk_permission_on AS authz_id
     FROM actor_has_bulk_permission_on($1, $2, $3, $4)">>}.
 
+%% Deleting all actors in list
+{bulk_delete_actor,
+ <<"DELETE FROM auth_actor WHERE authz_id = ANY($1)">>}.
+
+%% Deleting all groups in list
+{bulk_delete_group,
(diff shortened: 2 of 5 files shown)


marcparadise:
Once we add test coverage, this seems eminently merge-able.

stevendanna:
@oferrigni @marcparadise This PR is now very old (it was old before I moved it over from the old repo).  Do we remember who the consumer of this API was supposed to be.  Is it still needed?  I agree the implementation code looks good but without tests or a known user of the endpoint and given the lack of movement on the PR, my inclination is to close this.

stevendanna:
I rebased this against master.

markan:
IIRC I think the reason for this was for some object cleanup code Oliver had built; deleting things on the server side could be done with one SQL call (say for everything in an org), but required many calls to  bifrost, and we were overloading things.

markan:
(Capturing some tribal knowledge)
I should note that this is a start towards solving a real issue we have, but haven't had a chance to get to. AFAIK it doesn't land on anyone's roadmap.

We've seen some long running instances (or users with certain operational patterns) end up with a huge amount of garbage in their bifrost databases, sometimes exceeding the oc_erchef database by 10-20x. Probably the biggest issue is org deletion, where we don't clean up any of the stuff in the deleted org; it just lingers around. 

This can become painful quickly. A convenient pattern for disaster recovery is to copy orgs from one chef server instance to a live backup; next time around the old org is deleted and a new one created. This can run frequently and create tons of undeleted crud.

stevendanna:
@markan So, this still merges and was presumably correct at some point, should we maybe just review and merge it?  Bifrost is an internal service so we aren't tying ourself to a long-term API decision.

markan:
@stevendanna I'm good with merging

marcparadise:
:+1: to merge.

stevendanna:
I'm not sure why I didn't merge this in January.  I'll note, however, that we did end up tackling the issue that @markan discussed: https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb

Overall, the bigger issue when trying to solve that problem was figuring out /which/ IDs to delete, rather than doing the actual deletes.  

While this still seems fine, I'm just going to close it but leave the branch around.  If we haven't needed it in the last year, I don't really see a pressing need to merge it now.

PR #162, opened 2015-04-30
Reindex from a hidden node.

stevendanna:
From @christophermaier in chef/oc_erchef#19:

> Not super-essential now, since Erchef isn't a distributed application,
but who knows what the future might bring?  This seems like a good bit
of hygiene for maintenance / support script, though, which really
shouldn't be considered an active participant in the node / cluster to
which it is connecting.

> Erlang campfire ghost story: "THE CALL WAS COMING FROM INSIDE THE
CLUSTER!" :)

### DIFF
--- a/src/oc_erchef/rel/files/reindex-opc-organization
+++ b/src/oc_erchef/rel/files/reindex-opc-organization
@@ -120,7 +120,14 @@ get_org_id(Context, OrgName) ->
 init_network() ->
     net_kernel:start([?SELF, longnames]),
     erlang:set_cookie(node(), ?ERCHEF_COOKIE),
-    pong = net_adm:ping(?ERCHEF).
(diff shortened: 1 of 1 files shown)


stevendanna:
Discussion on the previous PR between @manderson26 and @christophermaier  centered around the following points:

- Whether to use hidden_connect/1 or hidden_connect_node/1 based on http://erlang.2086793.n4.nabble.com/Escript-and-hidden-nodes-td2101459.html.  The consensus was hidden_connect_node/1.

- Whether a library for utility functions like this in scripts would be useful, and

- Whether we should use a magical header (`%%! -hidden`) should be used instead and, if so, how to handle the cookie and node name.

stevendanna:
Nothing has moved on this for a while.  If we actually move to distributed erlang, we have a lot of issues to figure out and can handle this at that time.

PR #163, opened 2015-04-30
Ma/add read only group

stevendanna:
From @manderson26 in chef/oc_erchef#99:

> Here's a quick hack to add read only groups. WIP, Needs pedant tests and probably some migration tooling to coerce existing orgs into having this group.

### DIFF
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_org_creator.erl
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_org_creator.erl
@@ -103,7 +103,9 @@
            {add_acl, [{container, policy_groups}], [read], [{group, clients}]},
            {add_acl, [{container, cookbook_artifacts}], [read], [{group, clients}]},
            {add_acl, [{container, data}], [read], [{group, clients}]},
-           {add_acl, mk_tl(container, [cookbooks, environments, roles]), [read] , [{group, clients}]}
+           {add_acl, mk_tl(container, [cookbooks, environments, roles]), [read] , [{group, clients}]},
+           {add_acl, [{container, policies}], [read], [{group, clients}]},
+           {add_acl, [{container, cookbook_artifacts}], [read], [{group, clients}]}
           ]
          }
         ]).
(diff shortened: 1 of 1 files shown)


src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_org_creator.erl
```diff
@@ -96,6 +96,12 @@
            {add_acl, [mk_tl(container, [groups, containers]), {organization}], [read], [{group, users}]},
            {add_acl, [{container, sandboxes}], [create], [{group, users}]},
 
+           %% read only users
+           {add_acl,
+            [mk_tl(container, [cookbooks, data, nodes, roles, environments, policies, cookbook_artifacts, clients, groups, containers])],
```
stevendanna: Do we need policy_groups here as well?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_org_creator.erl
```diff
@@ -103,7 +109,9 @@
            {add_acl, [{container, policy_groups}], [read], [{group, clients}]},
            {add_acl, [{container, cookbook_artifacts}], [read], [{group, clients}]},
            {add_acl, [{container, data}], [read], [{group, clients}]},
-           {add_acl, mk_tl(container, [cookbooks, environments, roles]), [read] , [{group, clients}]}
+           {add_acl, mk_tl(container, [cookbooks, environments, roles]), [read] , [{group, clients}]},
+           {add_acl, [{container, policies}], [read], [{group, clients}]},
+           {add_acl, [{container, cookbook_artifacts}], [read], [{group, clients}]}
```
stevendanna: I'm a bit confused by this bit.  Doesn't this duplicate the lines above (where line 108==113 and line 110==114)?

stevendanna:
@chef/lob I think we are likely to have a release with a number of ACL policy changes in the near future.  It might be nice to get this in as well for that release and then we can put together a blog post with the policy changes and why we made them.

Overall I like this idea.  I've left a few comments.  Are we going to want to do a mover migration as well to add such a group to existing organizations?

stevendanna:
Let's get this merged.  I think we can simply forgo a mover migration for now.  I have some questions above re adding policyfiles to it but otherwise this looks good.

tylercloke:
Looks like we just have a few issues you brought up on this? Still need pedant tests?

marcparadise:
Is this still relevant with the read-only changes that landed a little while back?

markan:
I'll take on looking at this after we sort out the #886 work

stevendanna:
I'm going to to close this out as it is very old.  I'll leave the branch in place.  Anyone should feel free to resubmit it if they have time to rebase it make make sure it still works as expected.

PR #164, opened 2015-04-30
use jiffy via chef_json instead of ejson everywhere

stevendanna:
From @marcparadise in chef/oc_erchef#116:

> Let's just use chef_json so we don't need to pull in ejson anymore. ejson has an old version of mochiweb pinned which is causing minor trouble in upstream projects (chef-mover).

> Running through tests now, locally it all looks good.

> This won't be completable until we remove couchbeam - our fork depends on a fork of ejson.


stevendanna:
Closing since we've merged a0d686660c9d1d41de4d48a178cb95de87ad0f1a

PR #165, opened 2015-04-30
Add experimental config options to allow webmachine connection properties

stevendanna:
From @manderson26 in chef/oc_erchef#145:

> This is experimental.


stevendanna:
I agree these could be useful, but I'm going to close this for now as it hasn't moved in a long time and the tests are failing.  @markan feel free to open a new PR if we still want this in.

PR #166, opened 2015-04-30
Dt/status

stevendanna:
From @doubt72  in chef/oc-chef-pedant#51:

> Some basic tests for _status endpoint


stevendanna:
According to @sdelano, this branch may need to be tested against the keygen_stats addition to this endpoint.

PR #169, opened 2015-04-30
Use the strip-version sofwware from Delivery to reduce package size

stevendanna:


stevendanna:
cc @chef/lob 

This saves us about 50MB in the final 10.04 artifiacts.

@marcparadise Previous gave a :+1: 

@manderson26 Had concerns re our ability to use advanced debugging tools like systemtap given this change.

marcparadise:
I think I'm going to reverse my +1 on this in light of Mark's thoughts - If the savings here is just 50MB, I'm not sure it's worth the tradeoff

stevendanna:
Let's close this for now.  We now know that it *works* more or less.  If we ever want to make a concerted effort to decrease the artifact size, we can look into this further then.

PR #170, opened 2015-04-30
Allow user to specify remote SQL host and credentials

stevendanna:


mmzyk:
Travis build failed on a timeout from oc-id to rubygems. Rebuilding.

omnibus/files/private-chef-ctl-commands/open_source_chef12_upgrade.rb
```diff
@@ -359,21 +359,33 @@ def make_dir(dir, permissions)
...
-    sql_host = running_config['chef_server']['postgresql']['vip']
-    sql_port = running_config['chef_server']['postgresql']['port']
-    sql_user = running_config['chef_server']['postgresql']['sql_user']
-    sql_password = running_config['chef_server']['postgresql']['sql_password']
-    [sql_host, sql_port, sql_user, sql_password]
+    if !(sql_host && sql_port && sql_pass && sql_user)
+      log "Reading SQL configuration from /etc/chef-server/chef-server-running.json"
+      begin
+        running_config = JSON.parse(File.read("/etc/chef-server/chef-server-running.json"))
+        sql_host ||= running_config['chef_server']['postgresql']['vip']
```
mmzyk: Why the ||= ? We should only have entered this block if all of these values are not set and presumably if that is the case we want to set them all to what is in running.json, so we should just set them with =.

It might be possible we call this block twice, so we don't reset them I guess with ||=.

omnibus/files/private-chef-ctl-commands/open_source_chef12_upgrade.rb
```diff
@@ -359,21 +359,33 @@ def make_dir(dir, permissions)
...
-    # This code pulled from knife-ec-backup and adapted
-    log "Pulling open source Chef 11 database credentials"
-    if !File.exists?("/etc/chef-server/chef-server-running.json")
-      log "Failed to find /etc/chef-server/chef-server-running.json"
-      exit 1
-    end
+    sql_host = @options.sql_host
+    sql_port = @options.sql_port
+    sql_pass = @options.sql_pass
+    sql_user = @options.sql_user
```
mmzyk: I assume if the options.sql_* wasn't given this will just set the values to nil?

omnibus/files/private-chef-ctl-commands/open_source_chef12_upgrade.rb
mmzyk: What happens if I leave off one of the sql_* options but specify 2 or 3 on the command line? If my reading of the if statement is right, we enter the block to set the options from the running.json. But with the ||= we won't over write any that were set on the command line, so we'll end up with mixed options. This might work, or fail spectacularly.

It seems like we should make this all or nothing. You set all of them, or we pull all of them from running.json.

omnibus/files/private-chef-ctl-commands/open_source_chef12_upgrade.rb
```diff
@@ -71,7 +71,7 @@ def download_chef11_data_setup
 
   def download_chef11_data_download(chef11_data_dir, key_file)
 
-    wait_for_ready_server("Chef 11")
+    wait_for_ready_server("Chef 11") unless @options.skip_setup
```
mmzyk: This change appears unrelated to adding the sql options. Was this an oversight when the skip setup option was added?

mmzyk:
Travis build passed after it was kicked.

omnibus/files/private-chef-ctl-commands/open_source_chef12_upgrade.rb
stevendanna: Yup, see the second commit message.

omnibus/files/private-chef-ctl-commands/open_source_chef12_upgrade.rb
stevendanna: Right, the current behavior is such that you can have a mix of command line options or config file options with the command line taking precedence.  I think at the time my thinking was that someone might want to use the running json file to handle the credentials which are a bit large to pass on the command line.  However, I can easily change it to be all-or-nothing.

omnibus/files/private-chef-ctl-commands/open_source_chef12_upgrade.rb
mmzyk: duh, that makes sense.

omnibus/files/private-chef-ctl-commands/open_source_chef12_upgrade.rb
mmzyk: I am just trying to think what will be least surprising to the user. I'd rather that we do the option that fails early as to the one that fails more cryptically later. I think that's probably the all or nothing option, but I could probably be persuaded otherwise.

irvingpop:
@stevendanna can this be closed?

stevendanna:
I'm going to close this.  I think it is still a *correct* change, but we haven't had another user ask for it and it seems risky to make unnecessary modifications to the upgrade code at this time.

PR #171, opened 2015-04-30
[OC-11648] EC11: Undefined Behavior with Full Log or Data Disk

stevendanna:
From @sean-horn in chef/opscode-omnibus#297:

> I would appreciate review on this. I have tested it against a working EC11.1.3 HA install with no issues through keepalived shutdown failovers, recoveries of disk space so that the cluster member comes out of FAULT state and enters BACKUP. Also, if both backend cluster members enter FAULT state, things look good(no member has either the DRBD volume or the backend_VIP) and either one can recover and resume MASTER status.

I plan to parameterize the obvious settings and allow for multiple vrrp_scripts.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/cluster.sh.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/cluster.sh.erb
@@ -63,7 +63,9 @@ function transition
       fi
 
       ;;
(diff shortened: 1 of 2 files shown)


stevendanna:
Previous maintainer review of this PR was positive given the following changes:

- make the %-free where this takes effect configurable

PR #172, opened 2015-04-30
Runit should wait for drdb before starting backend services

stevendanna:
From @jmink in chef/opscode-omnibus#638:

> Background info here: https://chef.leankit.com/Boards/View/111432846/146112858
> 
> In short the backend services have historically tried to start before the HA backing store was mounted and then got horribly confused. This prevents the backend ha services from starting until the backing datastore is mounted and passing basic tests.
> 
> It relies on the ha_backend_storage status command and is therefore backing storage level agnostic.
> 
> I tested it by hand on standalone (ensuring it built correctly and pedant passed) and with HA on EC2 (ebs) in which I ensured the services started correctly (with pedant tests passing) and then changed the backend storage status script to return 1 and saw that the services were waiting.
> 
> Wilson run: http://wilson.ci.opscode.us/job/chef-server-12-trigger-ad_hoc/231/

### DIFF
--- a/files/private-chef-cookbooks/private-chef/recipes/runit_setup.rb
+++ b/files/private-chef-cookbooks/private-chef/recipes/runit_setup.rb
@@ -1,0 +1,25 @@
+#
+# Author:: Jess Mink (<jmink@chef.io>)
+# Copyright:: Copyright (c) 2014 Opscode, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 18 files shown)


stevendanna:
Previous summary of maintainer review by @irvingpop:

> Notes from Irving/Steven discussion.

> To get this merged and feel good about it:

> - fix the whitespace assumption in is_drbd_mounted()
- consolidate all of the include_recipe "private-chef::runit_setup" from individual recipes (inconsistent) to default.rb
- remove the changes from sv-keepalived-run.erb
- add support for the mounted argument to aws-ha ha_backend_storage - and refine any documentation about the expected arguments to that script

stevendanna:
Unfortunately, the ball got dropped on this right after that conversation between @irvingpop and I.  This is still the *right thing* for the DRDB stack.  While the code change required are minor, the testing burden of DRBD changes is high, so this work should probably have a dedicate ticket if we want it to continue: @charlesjohnson thoughts?

charlesjohnson:
Let's prioritize work that makes chef-backend usage more awesome over anything related to DRBD right now.

PR #208, opened 2015-05-01
[WIP] Add more caching to speed up builds

smith:
:construction: don't merge yet.

* Cache local Lua Rocks
* Use built-in RVM instead of installing our own
* Enable apt caching

Not sure if there are reasons it wasn't doing some of this or if people
don't want this setup but I'm trying it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,10 @@ branches:
   only:
     - master
 cache:
-  bundler: true
+  - apt
+  - directories:
+    - vendor/bundle
+    - $HOME/.luarocks/rocks
 # Turn this back on once luarocks can be installed
 # via the add-ons
 # sudo: false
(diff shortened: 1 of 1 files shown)


smith:
Apparently the `rvm: 2.1.4` only works if the `language: ruby` is there:

```
Gem::InstallError: ohai requires Ruby version >= 2.0.0.
```

stevendanna:
@smith Yeah, that is why we went with the:

```
+  - rvm install --binary 2.1.4
+  - rvm use 2.1.4
```

I'm :+1: on the caching.  We had it in place for oc_erchef, but my patch inadvertently dropped it.

stevendanna:
@smith I merged something similar to this into master; however, if I am reading the documentation correctly, it won't take effect until we move over to the container-based Travis infrastructure.

The only blocker for that currently is getting luarocks added to the apt package whitelist.

I've filed a request to get it added to the whitelist here:

https://github.com/travis-ci/travis-ci/issues/3716

If this is hurting us enough, we could consider turning off the lua tests temporarily.

smith:
Thanks @stevendanna! For those keeping score #223 contains those changes.

PR #207, opened 2015-05-01
Ra/vi 75

stevendanna:
@rmoshier said:

>This is not ready to be merged, but I've made some progress and would like to check in.

>When a user navigates to /id/profile, he or she sees the identity management components that used to be in account manage. (I did not remove those components from account manage, rather they have been duplicated here.) I created a profile controller to keep these components separate from the applications section of the app.

>Per @cnunciato 's suggestion, I omitted the Gravatar component and the 'Community Profile' and 'Need Help' boxes that are currently on the profile page in account manage.

>Next steps/questions:
1. Per @smith 's initial guidance, the forms are all currently static (including the profile, change password, and generate key forms). Is hooking up the forms within the scope of this ticket?
2. The show method in the profile controller is currently empty. Should anything else be included here?
3. Anything else I'm missing?

>General feedback is very welcome, as usual. :smile: Thank you!

PR #209, opened 2015-05-01
Revert "Ra/vi 75"

smith:
Reverts chef/chef-server#207

(Since this was moved over from the oc-id repo, the files got put at the root instead of in src/oc-id. We'll need to rename and re-merge to get things in the right places.)


sdelano:
:+1:

PR #215, opened 2015-05-04
remove SquareSerif font, comm-503 page, and associated resources

smith:
The comm-503 page is not used now. It looks like this:

screenshot

This removes that page and all of the assets it (and only it) uses, including the Square Serif font.


sdelano:
:+1:, merging.

PR #216, opened 2015-05-04
Dialyzer fixes in various from_json functions

joedevivo:


stevendanna:
:+1: Looks good.

stevendanna:
For the permanent record, the only failing tests were in oc-id because of rubygems timeouts.  :+1: on just merging until we merge bundler caching.

PR #217, opened 2015-05-04
Repair incorrect DTD

juliandunn:
This has always bothered me!


stevendanna:
:+1: but...can we go further?  Do we need these?  Why would the core API ever return HTML errors? Seems like we should return JSON errors.

smith:
@juliandunn why not just make it HTML 5 with `<!DOCTYPE html>`?

@stevendanna from here it looks like only 404 and index.html are used: https://github.com/chef/chef-server/blob/ad878ecf61c1e40171f7499a946e649a2d4e6629/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb#L29-L34

juliandunn:
@smith I could certainly do that... if we're sticking with HTML for the error pages.
@stevendanna I can definitely just convert these to JSON errors if you can give me an idea of what should correctly go in that payload?

smith:
I think we already have JSON versions: https://github.com/chef/chef-server/blob/ad878ecf61c1e40171f7499a946e649a2d4e6629/omnibus/files/private-chef-cookbooks/private-chef/files/default/html/409.json, et al.

index.html and 404.html should stay for when somebody's requesting pages from a browser and doesn't have Manage installed.

stevendanna:
So, it seems like:

- [] remove all but index.html and 404.html
- [] fix index.html and 404.html using `<!DOCTYPE html>` is the todo

I think serving a 404.json when the content-type requested is application/json can be done separately as it will likely take some nginx config changes.

juliandunn:
There's also `503.json` that I need to leave behind because of this:

`    error_page 503 =503 /503.json;`

smith:
:+1: on this; it looks like we keep the same behavior but delete a bunch of stuff which is the best.

The right thing might be to serve html or json depending on the accept header and have a 404/503/json/html. This gist shows an example of how that might be done: https://gist.github.com/lextoumbourou/d6221deb818da4f342ea

stevendanna:
:+1:

PR #218, opened 2015-05-04
Remove nested directories from log rotation template

ryancragun:
nginx, redis_lb, and opscode-solr4 are the current services that use the
logrotate.erb template for their corresponding logrotate entries.  None
of those services actually have nested log directories which results in
an error that prevent logrotate from running correctly:

error: stat of /var/log/opscode/nginx/*/*.log failed: No such file or
directory

Removing unnecessary nested directory directive resolves this problem

### DIFF
--- a/omnibus/CHANGELOG.md
+++ b/omnibus/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Remove install message from postinst package script
 * Update chef-server-ctl key commands to use chef-client's Chef::Key object.
 * New gather-log script gathers a lot more debugging information.
+* Remove nested directories from log rotation template
 
 ## 12.0.8 (2015-04-20)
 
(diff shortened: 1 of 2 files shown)


stevendanna:
@ryancragun :+1: This looks good to me.  

Since this fixes a bug would you mind adding a short:

`ChangeLog-Entry: DESCRIPTION`

line to your commit message?

Out of curiostiy, is there a particular problem you were seeing this fail on?  I ask because that line subdirectory glob has been in place since the start of Chef Server 11.

ryancragun:
@stevendanna I added an entry to the CHANGELOG.md and updated the commit message.  As per the problem, when the incorrect glob was there the file would get rotated (eg: moved to access.log.0) but nginx would keep an open file handle on old log and continue writing to it even though it was renamed.  Strangely not even a USR1 to nginx would fix it.

stevendanna:
:+1: MWR

sdelano:
@ryancragun - Please rebase next time before merging. :heart:

ryancragun:
awh crap

PR #220, opened 2015-05-05
[perf] Use list:append instead of list:flatten in chef_s3:make_key

stevendanna:
This is a minor performance improvement as the list:append function is
faster than list:flatten.  We can substitute list:append for
list:flatten here because we know that we only have one level of
nesting.

For more information see:

http://www.erlang.org/doc/efficiency_guide/listHandling.html#id66880

In local testing with pathelogically large cookbooks, I was able to
get a 10% improvement in request times of the cookbook_versions
endpoint with this change in place.


stevendanna:
cc @chef/lob

joedevivo:
Good find! I'd be interested in seeing how `++` and `io_lib:format/2` compare speedwise.

Since this gives a confirmed improvement, :+1:

Issue #180, opened 2015-05-01
Upgrade jetty in solr


stevendanna:
From @jessehu in chef/opscode-omnibus#725:

> As https://dev.eclipse.org/mhonarc/lists/jetty-announce/msg00069.html said "This year will mark the end of our open source support for Jetty 7 and Jetty 8." So Jetty 6/7/8 is end of life by 2014.
> 
> Can we update solr to newer version which uses higher version of jetty ? Chef Server 11 nows uses jetty 6.1.3.
> 
> ll /opt/chef-server/embedded/service/chef-solr/jetty/lib/total 740
> -rw-r--r-- 1 root root 472383 Oct 17 01:50 jetty-6.1.3.jar
> -rw-r--r-- 1 root root 128058 Oct 17 01:50 jetty-util-6.1.3.jar
> drwxr-xr-x 2 root root 4096 Mar 12 03:09 jsp-2.1
> -rw-r--r-- 1 root root 131998 Oct 17 01:50 servlet-api-2.5-6.1.3.jar
> 
> related code: https://github.com/chef/opscode-omnibus/blob/master/config/software/opscode-solr4.rb
> 
> Our product uses Chef 11 and jetty 9.2.9 is required by the security strategry.

stevendanna:
As mentioned in the previous discussion, it looks like it would be rather easy to move us to Jetty 8 as there are prebuilt Solr packages that include Jetty 8.  

Assigning to myself.

stevendanna:
@jessehu Just want to let you know that we've opened a PR https://github.com/chef/chef-server/pull/221 for upgrading solr and jetty.  That still doesn't put us on jetty 9, but I wanted to give you a status update.

jessehu:
Thanks @stevendanna. I will try Chef Server 12 to check it.

PrajaktaPurohit:
I am sorry we could not get to this sooner. Closing since we are replacing Solr with Elasticsearch from the next stable release onwards (14.x)

PR #221, opened 2015-05-05
Upgrade to Solr 4.9.1 and Jetty 8

stevendanna:
This move Chef Server onto the latest version of Solr, which includes
Jetty 8 rather than Jetty 6.

### DIFF
--- a/omnibus/config/software/opscode-solr4.rb
+++ b/omnibus/config/software/opscode-solr4.rb
@@ -15,10 +15,10 @@
 #
 
 name "opscode-solr4"
-default_version "4.5.1"
+default_version "4.9.1"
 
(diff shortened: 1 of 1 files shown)


stevendanna:
Passing wilson build: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/500/downstreambuildview/

Per https://github.com/chef/chef-server/issues/180 Jetty 8 is still old and we need to move to Jetty 9 eventually. I'm taking a look at the changes needed to do that, but since moving to 8 appears to be a drop-in upgrade, I figured it was worth getting closer to correct in the near term.

cc @chef/lob

stevendanna:
@chef/lob I've also tested an 11->12 solr->solr4 migration manually.

PR #224, opened 2015-05-05
Mp/dvm/coverage support

marcparadise:
I still need to test behavior of oc_bifrost and bookshelf with sync enabled at dev-time, so this isn't ready for merge.    I would welcome feedback on enabling coverage support and usage of it though. 

This also includes some enhancements to rpc calls into the erlang node, so that it will wait for the call to complete before returning. 

ping @chef/lob /redundant-cc @joedevivo

### DIFF
--- a/dev/dvm/lib/dvm/project/erlang.rb
+++ b/dev/dvm/lib/dvm/project/erlang.rb
@@ -2,7 +2,7 @@ require_relative "erlang_dep"
 
 module DVM
   class ErlangProject < Project
-    attr_reader :rebar_config_path, :project_dir, :relpath, :libpath, :relname
+    attr_reader :rebar_config_path, :project_dir, :relpath, :libpath, :relname, :node
     def initialize(project_name, config)
       super
       # TODO use .lock if available, else use .config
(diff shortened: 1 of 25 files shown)


dev/dvm/lib/dvm/project/erlang.rb
```diff
@@ -165,42 +166,70 @@ def do_build
...
-  Dirs = file:wildcard("/host/#{project_name}/apps/*/ebin\") ++ ["/host/#{project_name}/ebin"]
-  [cover:compile_beam_dir(Dir) || Dir <- Dirs]
-end,
-spawn(#{node}, F).
-EOM
-      run_command("#{erl_command} -eval \"#{eval_cmd}\" -s erlang halt", "Cover-compiling modules on the erlang node.")
-    end
-    def dump_cover(modules = nil)
-      # TODO
+    def cover(action, modulename, options)
```
stevendanna: What do you think of breaking this up a bit?  Perhaps something like

```ruby
def cover(action, modulename, options)
  runargs, msg = case action
                 when 'enable'
                   erl_rpc_enable_args(modulename, options)
                 when 'report'
                   erl_rpc_report_args(modulename, options)
                 [...SNIP...]
  erl_rpc("user_default", "cov_#{action}", runargs, msg)
end

marcparadise: I like that, thanks. I found myself missing pattern matching in fun heads .... :D

stevendanna: Every day I want functions of different arities and pattern matching in Ruby.

PR #226, opened 2015-05-06
[oc_erchef] [oc-chef-pedant] [cleanup] WIP - remove couch and associated deps

marcparadise:
Moved over from oc_erchef and rebased atop latest master.  Needs work on tests and dialyzer, but is passing pedant. 

* remove usage of couchbeam and chef_otto
* remove otto context data from context records
* eliminate darklaunch behaviors couch v sql
* use jiffy via chef_json instead of ejson everywhere
* remove app references to oauth and ejson
    * oauth was pulled in because of couchdbeam which is no longer a dep.
    * ejson was used but we have since standardized on chef_json->jiffy.
* deprecate x-ops-api-version response header
 * remove ruby- and chef server 11- specific behaviors from pedant

TODO: 
* [ ] move reindex script into omnibus 
* [ ] any lingering chef_otto health checks in status endpoint?
* [x] walk through CT/eunit and bring them up to date with removed functions  
   and smaller signatures
* [ ] split this into multiple PRs? *most* of the changes do go together - x-op-api-version is the only one that comes to mind as cleanly separable -  but it's a lot of modified files
* [x] dialyze it!

ChangeLog-Entry: This update fully removes couch support from oc_erchef


src/oc_erchef/apps/chef_db/src/chef_db.erl
```diff
@@ -285,33 +275,10 @@ fetch_org_id(Context, OrgName) ->
...
-        true ->
-            cdb_fetch_org_metadata(Context, OrgName);
-        false ->
-            sql_fetch_org_metadata(Context, OrgName)
-    end.
-
-%% For couchdb-based orgs, fetch information from the orgname -> metadata cache
-cdb_fetch_org_metadata(#context{reqid = ReqId,
-                                otto_connection = Server}, OrgName) ->
-    case chef_cache:get(org_metadata, OrgName) of
```
stevendanna: Should we also rip out the org_guid cache code altogether?  At this point enabling it would be a no-op.  Alternatively, we could see if there are benefits of changing the cahce to support pgsql and start using it again, might be interesting to see what benefits we reap.

src/oc_erchef/apps/chef_db/src/chef_otto.erl
```diff
@@ -1,754 +0,0 @@
...
-                      'authz_role'.
-
--type http_port() :: non_neg_integer().
--type db_key() :: binary() | string().
-
--define(gv(Key, PList), proplists:get_value(Key, PList)).
-
--spec connect() -> couch_server().
-connect() ->
-    Host = envy:get(chef_db, couchdb_host, string),
```
stevendanna: It looks like we might also be able to clean up our configuration file template:

omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb

and some of our omnibus attributes:

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb

src/oc_erchef/relx.config
```diff
@@ -3,12 +3,7 @@
 %% ex: ts=4 sw=4 ft=erlang et
 {release,{oc_erchef,"1.8.3"},[oc_erchef, syntax_tools, compiler]}.
 {extended_start_script,true}.
-{overlay_vars,"rel/reltool.config"}.
+% Do we need etc/keys? Seems that could live in omnibus?
 {overlay,[{mkdir,"log/sasl"},
           {mkdir,"lib/patches"},
-          {mkdir,"etc/keys"},
-          {copy,"schema","."},
```
stevendanna: I think this might break the omnibus build as the schema won't be in the erchef release, where omnibus is expecting it.  If we don't want this in the erchef release anymore, I think we can do this; it just means we also need to update the omnibus build configuration for the schema.

src/oc_erchef/apps/chef_db/src/chef_db.erl
marcparadise: I think I'll plug it into the sql lookups.  The amount of data is small, and I can't see a case in which it would hurt to db calls to find data  that can't change.

src/oc_erchef/apps/chef_db/src/chef_otto.erl
marcparadise: Thanks, I thought I'd done this. Done now!

src/oc_erchef/apps/chef_db/src/chef_db.erl
stevendanna: One disadvantage of the old cache was that if a customer asked for their org to be deleted, and then some other user created an org of the same name, you got some crazy results until the cache was cleared.

src/oc_erchef/relx.config
marcparadise: I think we should move this, but if it causes trouble I can defer this part of the  change for another time.

marcparadise:
ping @chef/lob

oc-chef-pedant/pedant_config.rb
```diff
@@ -195,24 +195,6 @@
...
-ruby_container_endpoint_in_sql? true
-ruby_group_endpoint?       false
-ruby_acl_endpoint?         false
-ruby_system_recovery_endpoint? false
-ruby_org_acl_endpoint?     false
-ruby_org_assoc?            false
-ruby_organizations_endpoint? false
-chef_12?                   true
-policies?                  true
-
```
sdelano: We also render a pedant config in `omnibus/files/cookbooks/private-chef/templates`. We should clean that up too.

sdelano:
Why are we deprecating the `x-ops-api-version` and how is that related to removing couchdb libraries?

src/oc_erchef/apps/chef_db/src/chef_db.erl
sdelano: I'm :+1: on removing it altogether. Now that we're down to one lookup / request and it's not causing a huge scaling problem like it was on CouchDB, then we're good removing that.

marcparadise:
@sdelano I'm going to split that and a few other things into logical PRs.

oc-chef-pedant/pedant_config.rb
marcparadise: Yeah, it looks like all of my omnibus changes didn't survive the change in repos.  Makes  me think I only had them locally to start with.

src/oc_erchef/apps/chef_db/src/chef_db.erl
marcparadise: Removed in latest push

stevendanna:
Can we close this master PR now that we have it split up?

marcparadise:
Yep

PR #229, opened 2015-05-06
dvm tweaks

mmzyk:
I tried my hand at using dvm, but it wasn't happy initially. I made some modifications to fix that.

Commit messages contain rationale and thoughts behind why I made the changes I did.

### DIFF
--- a/dev/Vagrantfile
+++ b/dev/Vagrantfile
@@ -9,7 +9,7 @@ Vagrant.configure("2") do |config|
   # Vagrant will auto resolve the url to download from Atlas
   config.vm.box = "ubuntu/trusty64"
   config.vm.hostname = "api.chef-server.dev"
-  config.vm.network "private_network", ip: "33.33.33.100"
+  config.vm.network "private_network", ip: "192.168.33.100"
 
   begin
     custom_config = YAML.load_file("config.yml")
--- a/dev/Vagrantfile
+++ b/dev/Vagrantfile
@@ -5,8 +5,9 @@ require "yaml"
 
 Vagrant.configure("2") do |config|
   provisioning, installer, installer_path, attributes = prepare()
-  config.vm.box     = "oc_erchef-ubuntu-14.04"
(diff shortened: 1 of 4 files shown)


dev/Vagrantfile
```diff
@@ -5,10 +5,11 @@ require "yaml"
...
 Vagrant.configure("2") do |config|
   provisioning, installer, installer_path, attributes = prepare()
-  config.vm.box     = "oc_erchef-ubuntu-14.04"
-  config.vm.box_url= "https://opscode-vm-bento.s3.amazonaws.com/vagrant/virtualbox/opscode_ubuntu-14.04_chef-provisionerless.box"
+  # Use the official Ubuntu 14.04 box
+  # Vagrant will auto resolve the url to download from Atlas
+  config.vm.box = "ubuntu/trusty64"
   config.vm.hostname = "api.chef-server.dev"
-  config.vm.network "private_network", ip: "33.33.33.100"
+  config.vm.network "private_network", ip: "192.168.33.100"
```
danielsdeleo: Just curious, what does this do?

marcparadise:
:+1: this looks good - MWR , and thanks again

dev/Vagrantfile
mmzyk: Sets up the IP address the vm uses. The reason for the change is that when working locally we really should use a private ip, which 192.168.x.x is. 33.33.33.x isn't. (whois tells me the DoD Network Information Center owns it). The use of 33.33.33.10 came from an early vagrant recommendation that was a mistake. Later versions of vagrant fixed it, but we've continued to propagate the 33.33.33.x ip in our dev tooling. There is a small risk when using a non-private ip that we could send data to the actual public ip. In this case, I can't tell that there is actually anything at 33.33.33.x, but we still shouldn't use it.

danielsdeleo: Ah, okay, I couldn't remember off-hand if 33 was some kind of link-local thing that theoretically would be okay to use, like the 169 addresses.

PR #230, opened 2015-05-07
Cache timezone information in local file

stevendanna:
Rather than requiring sudo on every invocation, this change caches the
timezone information in a .cached_tz file. Most developer workstations
change timezone infrequently.  As such, I believe this will be a bit
more user friendly than asking for the sudo password all the time.

### DIFF
--- a/dev/.gitignore
+++ b/dev/.gitignore
@@ -1,2 +1,3 @@
 .vagrant
 testdata
+.cached_tz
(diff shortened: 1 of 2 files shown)


joedevivo:
+1 I hate typing my password! @stevendanna for president!

marcparadise:
do eet!

PR #233, opened 2015-05-07
Implemented X-Ops-Server-API-Version Response Header

tylercloke:
Ping @chef/lob 

Related: https://github.com/chef/chef-rfc/pull/123

[Build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/505/downstreambuildview/)

##### Response Header

In the response header, `X-Ops-Server-API-Version` will return json of the format:

```
{
  "min_version":"<non-negative integer>",
  "max_version":"<non-negative integer>",
  "request_version":"<integer>",
  "response_version":"<integer>"
}
```

Where:
+ `min_version` is the minimum API version the server supports.
+ `max_version` is the maximum API version the server supports.
+ `request_version` is an integer representing the desired API version from the client passed via X-Ops-Server-API-Version
defaulting to 0 if X-Ops-Server-API-Version was not sent by the client. If X-Ops-Server-API-Version sent by the client contained
an invalid value (not an integer), then -1 is returned in `request_version`.
+ `response_version` is an integer representing the API version used by the server to process the request.
It either matches what the client requested in X-Ops-Server-API-Version or is -1 if a 406 occurred (which happens
when X-Ops-Server-API-Version sent by the client was invalid or not in the supported range of the server).

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
@@ -84,9 +84,11 @@ service_available(Req, #base_state{reqid_header_name = HeaderName} = State) ->
 
 %% @doc Validate the requested X-Ops-Server-API-Version value and populate base_state,
 %% and reply with it as or reply with error if it's not valid.
+%%
+%% If X-Ops-Server-API-Version is not sent, the server assumes an API version of 0.
 -spec server_api_version(undefined|string()) -> api_version() | {error| string()}.
 server_api_version(undefined) ->
-    ?API_MIN_VER;
+    0;
 server_api_version(RequestedVersion) ->
     try list_to_integer(RequestedVersion) of
         Version when Version < ?API_MIN_VER orelse
@@ -106,7 +108,15 @@ invalid_server_api_version_response(Req, State, RequestedVersion) ->
                                              {<<"message">>, Message},
                                              {<<"min_version">>, ?API_MIN_VER},
                                              {<<"max_version">>, ?API_MAX_VER}]}),
-    {{halt, 406}, Req2, State#base_state{darklaunch = no_header, log_msg = LogMsg}}.
+    %% if the RequestedVersion was a valid integer, pass that along so we can parse it in the header response,
+    %% else, pass undefined so we know something invalid was passed.
+    ParsedRequestedVersion = try list_to_integer(RequestedVersion) of
+                                 IntegerRequestedVersion -> IntegerRequestedVersion
+                             catch
+                                 _:_ ->
+                                     bad_value_requested
+                             end,
+    {{halt, 406}, Req2, State#base_state{darklaunch = no_header, log_msg = LogMsg, server_api_version = ParsedRequestedVersion}}.
 
 set_req_contexts(Req, #base_state{reqid = ReqId, server_api_version = ApiVersion} = State) ->
     {GetHeader, State1} = chef_wm_util:get_header_fun(Req, State),
@@ -746,11 +756,15 @@ finish_request(Req, #base_state{reqid = ReqId,
         case Code of
             500 ->
                 % Sanitize response body
-                ErrReq = create_500_response(AnnotatedReq, State),
-                {true, ErrReq, State};
+                ErrReq0 = create_500_response(AnnotatedReq),
+                ErrReq1 = add_api_info_header(ErrReq0, State, undefined),
+                {true, ErrReq1, State};
+            406 ->
+                AnnotatedReq1 = add_api_info_header(AnnotatedReq, State, not_acceptable),
+                {true, AnnotatedReq1, State};
             _ ->
-                 AnnotatedReq1 = add_api_info_header(AnnotatedReq, State),
-                 {true, AnnotatedReq1, State}
+                AnnotatedReq1 = add_api_info_header(AnnotatedReq, State, undefined),
+                {true, AnnotatedReq1, State}
         end
     catch
         X:Y ->
@@ -845,15 +859,15 @@ malformed_request(Req, #base_state{resource_mod=Mod,
             {true, NewReq, State1#base_state{log_msg = Why}}
     end.
 
-
-create_500_response(Req, State) ->
+-spec create_500_response(#wm_reqdata{}) ->
+                                 #wm_reqdata{}.
+create_500_response(Req) ->
     %% sanitize response body
     Msg = <<"internal service error">>,
     Json = chef_json:encode({[{<<"error">>, [Msg]}]}),
     Req1 = wrq:set_resp_header("Content-Type",
                                "application/json", Req),
-    Req2 = add_api_info_header(Req1, State),
-    wrq:set_resp_body(Json, Req2).
+    wrq:set_resp_body(Json, Req1).
 
 %% @doc Extract information from `State' needed to generate the X-Ops-API-Info header value.
 api_info(#base_state{api_version = ApiVersion,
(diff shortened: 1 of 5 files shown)


stevendanna:
This looks good to me.  I'm wondering where we set the RFC-required `X-Ops-Server-API-Version:` in the response?  That isn't part of this pull request, but by grepping around I can't find it.  Do we need to add that as well.

tylercloke:
@stevendanna Hrm great question! I don't think that header is being returned. Isn't that the header you _pass_ to the server? I don't think the server returns that:

```
DEBUG: x-ops-api-min-version: 0
DEBUG: x-ops-api-max-version: 1
DEBUG: x-ops-api-info: flavor=cs;version=12.0.0;oc_erchef=1.8.3
```

joedevivo:
+1 I saw the whole thing go down!

stevendanna:
The RFC says:

```
If $version is valid or is not specified, the API response will contain the headers:

X-Ops-Server-API-Version: $actual_version
Vary: X-Ops-Server-API-Version`
$actual_version MUST indicate the API version level of the response. The Vary header is included in the event it becomes desirable to support caching of specific responses.
```

marcparadise:
Wow, nice catch @stevendanna - I thought I removed that from the final version of the RFC, so didn't implement :D

marcparadise:
Given that means we have three version-related header components (ignoring th one that's going to get retired), would it make sense to use a single X-Ops-Server-API: { min_version: $min, max_version: $max, request_version: $actual or 'none' if 406}.

tylercloke:
Implemented Marc's comment, except it doesn't return `-1` if 406, but only if `X-Ops-API-Min-Version` header was not sent to server, that makes the most sense to me. It can still 406 with a non `-1` API version sent to the server.

marcparadise:
:+1:

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -874,7 +875,19 @@ api_info_header_value(#base_state{}=State) ->
 %% version information (useful for maintaining back-compatibility) as well as OTP version
 %% information (more useful for debugging purposes).
 add_api_info_header(Req, State) ->
-    wrq:set_resp_header("X-Ops-API-Info", api_info_header_value(State), Req).
+    APIVersion = case wrq:get_req_header("X-Ops-Server-API-Version", Req) of
+                     undefined ->
+                         <<"-1">>;
```
marcparadise: So the intent here (at least as written in the RFC) was that we'd report back the API version level used to service the request - which would only not exist if the request was not serviced b/c requested version was invalid.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -84,9 +84,11 @@ service_available(Req, #base_state{reqid_header_name = HeaderName} = State) ->
 
 %% @doc Validate the requested X-Ops-Server-API-Version value and populate base_state,
 %% and reply with it as or reply with error if it's not valid.
+%%
+%% If X-Ops-Server-API-Version is not sent, the server assumes an API version of 0.
 -spec server_api_version(undefined|string()) -> api_version() | {error| string()}.
 server_api_version(undefined) ->
-    ?API_MIN_VER;
+    0;
```
tylercloke: Note: this is biggest functional change. We will default to zero if no header passed!

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -870,11 +884,53 @@ api_info_header_value(#base_state{}=State) ->
...
+-spec add_api_info_header(#wm_reqdata{}, #base_state{}, atom()) ->
+                                 #wm_reqdata{}.
+add_api_info_header(Req, #base_state{server_api_version = APIVersion} = State, undefined) ->
+    Req0 = wrq:set_resp_header("X-Ops-API-Info", api_info_header_value(State), Req),
+    HeaderBody = build_api_info_header_body(APIVersion, APIVersion),
+    wrq:set_resp_header("X-Ops-Server-API-Version", HeaderBody, Req0);
+add_api_info_header(Req, #base_state{server_api_version = RequestedVersion} = State, not_acceptable) ->
+    %% indicates the request wasn't processed due to a 406
+    ActualAPIVersion = -1,
+    RequestedAPIVersion = case RequestedVersion of
```
marcparadise: We can move this into the build_api_header_body fun head, like: 
```
build_api_info_header_body(bad_value_requested, ActualAPIVersion) -> 
    build_api_info_header_body(-1, ActualAPIVersion); 
```
That should let us remove the specific not_acceptable head of `add_api_info_header` as well.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
marcparadise: :+1:

marcparadise:
Just the one comment, once that's done LGTM

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
tylercloke: :+1: Context switching between Ruby and Erlang is hard!

tylercloke: Er actually, the arity is 3 for line 914 actually. We want to respond with `-1`as the `response_version` in the 406 case, so I'm not sure how moving that `case` statement would help with that.

marcparadise:
Alright, on read-through of the file (Now With Context!) yep, I see you can't get rid of the change in arity. But I *think* it can remove the case+multiple heads in add_api_info_header (untested and it's late, so my internal VM stack may be corrupt :D): 

```
-spec add_api_info_header(#wm_reqdata{}, #base_state{}, undefined|not_acceptable) ->
                                 #wm_reqdata{}.
add_api_info_header(Req, #base_state{server_api_version = APIVersion} = State, RequestedVersionInfo) ->
    Req0 = wrq:set_resp_header("X-Ops-API-Info", api_info_header_value(State), Req),
    HeaderBody = build_api_info_header_body(RequestedVersionInfo, APIVersion),
    wrq:set_resp_header("X-Ops-Server-API-Version", HeaderBody, Req0).

-spec build_api_info_header_body( undefined | not_acceptable | integer(), integer()) ->
                                 list().

build_api_info_header_body(undefined, ActualAPIVersion) ->
    build_api_info_header_body(ActualAPIVersion, ActualAPIVersion);
build_api_info_header_body(not_acceptable, ActualAPIVersion) ->
    build_api_info_header_body(-1, ActualAPIVersion);
build_api_info_header_body(RequestedAPIVersion, ActualAPIVersion) ->
    binary_to_list(chef_json:encode({[
                                      {min_version, integer_to_binary(?API_MIN_VER)},
                                      {max_version, integer_to_binary(?API_MAX_VER)},
                                      {request_version, integer_to_binary(RequestedAPIVersion)},
                                      {response_version, integer_to_binary(ActualAPIVersion)}
                                     ]})).
```

tylercloke:
The above isn't quite the correct logic. Its missing the part where `RequestedAPIVersion` is set to `-1` in the `bad_value_requested` case, which simply means the original request contained a value that was not parseable as an API version (aka not an integer). 

I think the way the code is laid out currently is actually the most clear way of displaying the logic, basically:

```
if 406 then
   set ActualAPIVersion` to -1
   if #base_state.server_api_version == bad_value_requested then
the client passed something unparsable (non-integer)

```

tylercloke:
@marcparadise ^ I'll hold of on merging until you comment on the above, I feel like I'm missing something fundamental about what your comment is getting at.

stevendanna:
@marcparadise @tylercloke I'm :+1: on merging but I'm wondering if this gets cleaner if we add a new field to the base state and just set these two values to their expected values way up in invalid_server_api_version_response():

```erlang
invalid_server_api_version_response(Req, State, RequestedVersion) ->
    LogMsg = {invalid_server_api_version_requested, RequestedVersion},
    Message = iolist_to_binary([ <<"Specified version ">>, RequestedVersion, <<" not supported">>]),
    Req2 = chef_wm_util:set_json_body(Req, {[{<<"error">>, <<"invalid-x-ops-server-api-version">>},
                                             {<<"message">>, Message},
                                             {<<"min_version">>, ?API_MIN_VER},
                                             {<<"max_version">>, ?API_MAX_VER}]}),
    %% if the RequestedVersion was a valid integer, pass that along so we can parse it in the header response,
    %% else, pass undefined so we know something invalid was passed.
    ParsedRequestedVersion = try list_to_integer(RequestedVersion) of
                                 IntegerRequestedVersion -> IntegerRequestedVersion
                             catch
                                 _:_ ->
                                     -1
                             end,
    {{halt, 406}, Req2, State#base_state{darklaunch = no_header, log_msg = LogMsg, 
                                         server_api_version = ParsedRequestedVersion,
                                         response_api_version = -1}}.
```

That is the core logic @tylercloke mentioned, but now it is contained in one place. And then these functions only need to worry about the case where response_api_version is undefined, which will happen when all was well (also the additional switching in finish request related to this could also be removed):

```erlang
-spec add_api_info_header(#wm_reqdata{}, #base_state{}, atom()) -> #wm_reqdata{}.
add_api_info_header(Req, #base_state{server_api_version = RequestedVersion,
                                     api_response_version = ActualAPIVersion} = State) ->
    Req0 = wrq:set_resp_header("X-Ops-API-Info", api_info_header_value(State), Req),
    HeaderBody = build_api_info_header_body(RequestedAPIVersion, ActualAPIVersion),
    wrq:set_resp_header("X-Ops-Server-API-Version", HeaderBody, Req0).

-spec build_api_info_header_body(integer()|undefined, integer()) -> list().
build_api_info_header_body(RequestedAPIVersion, undefined) ->
    build_api_info_header_body(RequestedAPIVersion, RequestedAPIVersion);
build_api_info_header_body(RequestedAPIVersion, ActualAPIVersion) ->
    binary_to_list(chef_json:encode({[
                                      {min_version, integer_to_binary(?API_MIN_VER)},
                                      {max_version, integer_to_binary(?API_MAX_VER)},
                                      {request_version, integer_to_binary(RequestedAPIVersion)},
                                      {response_version, integer_to_binary(ActualAPIVersion)}
                                     ]})).
```

PR #234, opened 2015-05-07
use relx tooling to not start all apps. sync versions on chef-server version

marcparadise:
use relx tooling to not start dev apps in normal runs 
make our erlang project  versions the same as the chef-server version. 

ping @chef/lob


stevendanna:
:+1:

PR #235, opened 2015-05-07
fix oc_id recipe so that it doesn't try to parse log messages as json

sdelano:
Since we changed the #command method enough, I decided that it was
best to remove it and inline it into the calling function so that
nobody else tries to use it. This was the only place where that
function was called to begin with.

/cc @chef/lob @schisamo @PrajaktaPurohit


sdelano:
also /cc @smith @mmzyk

stevendanna:
:+1: for a pragmatic solution

schisamo:
:+1:

marcparadise:
:+1:

PR #238, opened 2015-05-08
Only clean depsolver bundle when clean target is run

stevendanna:
Using the new CLEAN_HOOK, the Makefile now only cleans out the bundle
when told to. This should avoid re-installing the bundle on local
invocations of make.

Also, remove the extra call to eunit in the travis target.


stevendanna:
cc @chef/lob

marcparadise:
:+1:

PR #240, opened 2015-05-08
Add tests for returning a 403 when removing admin group from any ACE

mmzyk:
This work isn't done yet, but these are the tests that will show
when it is done.
For now the tests are marked pending so they can go down into master.

### DIFF
--- a/oc-chef-pedant/spec/api/groups_acl_spec.rb
+++ b/oc-chef-pedant/spec/api/groups_acl_spec.rb
@@ -1,0 +1,59 @@
+require 'pedant/rspec/common'
+require 'pedant/acl'
+
+describe "Groups ACL", :acl do
+  include Pedant::ACL
+
+  describe "returns Forbidden when trying to remove the admin group from any grant ACE" do
+    # For now testing with groups, but potentially also want to test with other objects
+    %w(admins clients users).each do |test_group|
+      context "for #{test_group}" do
+        let(:admin_username) { platform.admin_user.name }
+        let(:acl_url) { "#{platform.server}/groups/#{test_group}/_acl" }
+        let(:read_acl_url) { api_url("groups/#{test_group}/_acl") }
+        let(:write_acl_url) { api_url("groups/#{test_group}/_acl/grant") }
+
+        let(:actors) { ["pivotal"] }
+        let(:groups) { [] }
+
+        # Both the actors and groups keys must be present when doing a PUT
+        # or else the server rejects the request
+        let(:request_body) {{
+          "grant" => {
+            "actors" => actors,
+            "groups" => groups
+          }
+        }}
+
+        # Since we're playing with permissions, let's make sure we restore them
+        # If these tests pass, this shouldn't be needed, but if they fail we
+        # might be in a bad state, and then the restore is needed
+        before :each do
+          original_state = get(read_acl_url, superuser)
+          # If we can't get the original state, die, because then we can't restore it
+          original_state.should look_like({:status => 200})
+          # The response is returned as a string that is valid JSON (in this case at least)
+          # Parsing it turns it into a usable hash
+          state_hash = JSON.parse original_state
+          # We're only changing the grant ACE, so pull that out to be able to use it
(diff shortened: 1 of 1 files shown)


manderson26:
I'd also test that pivotal cannot be removed; fixup is harder though since the pivotal user is the usual fixup route.. Maybe try removing admins, and fixing up with pivotal, then removing pivotal then fixing up with a user in admins is the way to go for tests.

mmzyk:
@manderson26 Ideally, if the tests pass, fixup wouldn't be needed, but I see what you're saying. What are the feelings on pivotal? Treat it like Admins and it can never be removed from the grant ACE as well?

manderson26:
I think that pivotal needs to be treated the same way; right now things break if it is removed.

oc-chef-pedant/spec/api/groups_acl_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        end
+
+        after :each do
+          payload = {"grant" => @original_grant_perms}
+          # If the original permissions don't restore, die, as we have an issue
+          put(write_acl_url, superuser,
+              :payload => payload).should look_like({:status => 200})
+        end
+
+        it "returns Fordidden trying to remove grant ace from #{test_group} group", :pending => true do
```
smith: Will there be a spec for the actual words of the error message? It would be nice to have them be good words so we can forward them directly to the client.

mmzyk: We should have a good error message yes. I'm not sure it's something we need a test around, as that strikes me as the most brittle kind of test. Not sure I know a better way though. With the way our system is now error messages are very intertwined with the code.

stevendanna:
Part of me wonders whether we are going about this change and other changes related to permissions-related behaviors in the right way.  Namely, we seem to have a set of behavior we want to enforce, in this case:

- admin users should always be able to edit objects inside an organization
- the superuser ("pivotal") should always be able to edit objects across the system

But rather than implementing those behaviors as part of the authentication flow, we are arranging permissions such that they are true by default and then implementing restrictions that prevent a state change away from that default.

Another similar restriction that we've talked about recently is making the pre-created groups un-deletable.  One way to implement such a change would be to prevent the deletion of those groups via some policy in the named_group module.  Another way would be to ensure that nobody was in the DELETE ACE for those groups and then prevent anyone from adding actors or groups into that ACE.  The former seems like the more straightforward implementation to me, but a lot of our recent decisions would point at the latter.

I am not asking to NAK this change, but rather to make sure that I understand our reasoning here.  I think that we likely have a number of these type of behaviors we will want to implement over the next quarter as people are focusing on sanding the edges of our ACL system and I think it is important that we make sure we don't end up with lots of competing special-case code in various subsystems.

mmzyk:
@ssd No worries about NAKing anything. This is precisely the debate we need to be having. Part of why I've been doling out such small changes is to encourage this. The ACL system is complicated and I'm still learning it, so I don't presume to know all the nuances and assume I've likely missed something.

I think once we make one of the three changes you're talking about (starting with the admin restriction on the grant ace) I think it's likely that the other two will follow. I agree that we should try to have it as all centrally located as possible so it's as easy to understand as possible. That will save future pain.

If you look at https://github.com/chef/chef-server/pull/248 where I'm trying to implement code for these tests, you'll see that already in the code comments I have questions on how the implementation plays out. Maybe that tells us this approach isn't right and we need to re-evaluate. Perhaps the answer to put something like this in bifrost? That is one option I haven't explored, although from my understanding bifrost doesn't understand groups, which made me put this change higher up in the oc_chef_wm_acl* path and oc_chef_authz path, since that is the common entry point for all of this.

I don't really understand the pros/cons of the approaches you bring up with regards to the removal of the default groups - using the named_group module vs. modifying ACEs. Care to elaborate there?

mmzyk:
Closing this out, since at this point https://github.com/chef/chef-server/pull/248 is the place all the work is happening and contains this code as well.

PR #241, opened 2015-05-08
use correct authz id, add policy cleanup migration

marcparadise:
1. Add new migration 'mover_policies_cleanup_callback' for use in
cleaning up migrations that occurred with a bug in
mover_policies_containers_creation_callback.
2. Fixed bug in mover_policies_containers_creation_callback that had
us using the wrong superuser authz id.
3. Added logging on failure in mv_oc_chef_authz and mv_oc_chef_authz_http

ping @danielsdeleo @chef/lob


danielsdeleo:
:+1: LGTM

PR #244, opened 2015-05-11
Allow users to read other user objects when both users share an org

stevendanna:
This changes the creation policy for newly created organizations and
provides a migration for existing organizations to ensure that the
users group is added to the ORGNAME_global_admins group.

The ORGNAME_global_admins group has one purposes inside the Chef
Server: to provide READ access to users that are associated with
ORGNAME. Previously only admins were granted this READ access.
However, users of chef-vault would like ordinary users to be able to
encrypt items for other users, thus promping this change in behavior.

The following API requests that would have previously failed, will now
succeed if two users are members of the same organization:

- GET /users/USERNAME
- GET /users/USERNAME/organizations
- GET /users/USERNAME/keys
- GET /users/USERNAME/keys/KEYNAME

### DIFF
--- a/src/chef-mover/src/mover_global_admins_user_addition_callback.erl
+++ b/src/chef-mover/src/mover_global_admins_user_addition_callback.erl
@@ -1,0 +1,106 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
+%% ex: ts=4 sw=4 et
+%% @author Steven Danna <steve@chef.io>
+%% @copyright 2015 Chef Software, Inc.
+%%
+%% The global_admins_user_addition addition callback adds the user
+%% group from a given organization into the ORGNAME_global_admins
+%% group for that organization.
+%%
+%% The ORGNAME_global_admins group has a signle purpose: providing
+%% READ access to other user objects in ORGNAME.  Thus, the only
+%% change that this migration produces is that users that share an org
+%% can now see each others user objects.
+%%
+%% This change is necessary to facilitate the use of chef-vault.
+%%
+-module(mover_global_admins_user_addition_callback).
+
+-export([
+         migration_init/0,
+         migration_complete/0,
+         migration_type/0,
+         supervisor/0,
+         migration_start_worker_args/2,
+         migration_action/2,
+         next_object/0,
+         error_halts_migration/0,
+         reconfigure_object/2,
+         needs_account_dets/0
+        ]).
+
+-include("mover.hrl").
+
+-record(org, {name, id, authz_id}).
+-record(group, {name, id, authz_id}).
+
+migration_init() ->
+    mv_oc_chef_authz_http:create_pool(),
+    mover_transient_migration_queue:initialize_queue(?MODULE, all_orgs()).
+
+migration_action(#org{name = OrgName} = Org, _) ->
+    GAGroup = global_admin_group(Org),
+    UserGroup = user_group(Org),
+    case add_group_to_group(UserGroup, GAGroup) of
+        ok ->
+            ok;
+        {error, Error} ->
+            lager:warn("Organization ~p failed during group addition.", [OrgName]),
+            Error
+    end.
+
+migration_complete() ->
+    mv_oc_chef_authz_http:delete_pool().
+
+global_admin_query() ->
+    <<"SELECT name, id, authz_id FROM groups WHERE name = $1">>.
+
+users_group_query() ->
+    <<"SELECT name, id, authz_id FROM groups WHERE name = 'users' AND org_id = $1">>.
+
+all_orgs_query() ->
+    <<"SELECT name, id, authz_id FROM orgs">>.
+
+all_orgs() ->
+    %% TODO: Will this be Bad(TM) in Hosted?
+    {ok, Orgs} = sqerl:select(all_orgs_query(), [], rows_as_records, [org, record_info(fields, org)]),
+    Orgs.
+
+global_admin_group(Org) ->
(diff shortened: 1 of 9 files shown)


stevendanna:
cc @chef/lob @charlesjohnson @manderson26 

I think the only change not previously discussed here is the /users/USERNAME/organization endpoint that pedant revealed is also allowed for anyone with READ access to the user.  Do we think that organization membership is confidential data?  If so, we could try changing that to only being allowed when `user == requesting_user` or when the requesting user is the superuser.

marcparadise:
I could see that being a problem in a Hosted environment. What about filtering it to orgs in common unless superuser? It would be a new query based on the person accessing.

stevendanna:
@marcparadise Yeah, we could also filter the results to only orgs in common. Perhaps something like:

```
SELECT user1orgs.org_id FROM 
     (SELECT * FROM org_user_associations WHERE user_id = $1) AS user1orgs, 
     (SELECT * FROM org_user_associations WHERE user_id = $2) AS user2orgs 
WHERE user1orgs.org_id = user2orgs.org_id
```

tylercloke:
I like where this is going, I wish we had tests for mover / migrations in general.

`/me thinks about what to do there`

stevendanna:
Added result filtering (and tests to cover it)

src/chef-mover/src/mover_global_admins_user_addition_callback.erl
```diff
@@ -0,0 +1,106 @@
...
+        ]).
+
+-include("mover.hrl").
+
+-record(org, {name, id, authz_id}).
+-record(group, {name, id, authz_id}).
+
+migration_init() ->
+    mv_oc_chef_authz_http:create_pool(),
+    mover_transient_migration_queue:initialize_queue(?MODULE, all_orgs()).
```
marcparadise: Thinking of our hosted run, we might need to give some thought to how we initialize this. If we have to pause and rerun, we'll be doing a lot of duplicated attempts here. (On the other hand, that may be OK.)

marcparadise:
Just as a heads up, as part of 'make migration path happy' work, I'm looking at some significant changs here, probably to the point where we'll be running mover.old for all existing migrations and then an entropy-resistant, environment-aware mover.new for newer stuff.  I don't think it affects this though, because mover.old can be declared final  at any time, and include this migration.

src/chef-mover/src/mover_global_admins_user_addition_callback.erl
```diff
@@ -0,0 +1,106 @@
...
+    <<"SELECT name, id, authz_id FROM groups WHERE name = $1">>.
+
+users_group_query() ->
+    <<"SELECT name, id, authz_id FROM groups WHERE name = 'users' AND org_id = $1">>.
+
+all_orgs_query() ->
+    <<"SELECT name, id, authz_id FROM orgs">>.
+
+all_orgs() ->
+    %% TODO: Will this be Bad(TM) in Hosted?
```
stevendanna: This seems OK in preproduction.

src/chef-mover/src/mover_global_admins_user_addition_callback.erl
stevendanna: Migrating involves:

- 2 reads from the opscode_chef database
- 1 write to the bifrost API.  The API request is a no-op if it has already been completely previously.

Thus, I think that re-running the full migration should be safe.  That said, it will be slower than necessary and cause extra load on the database. The only way I can think of to get the "unmigrated orgs" is almost as much work as the migration since we have to:

- look up the authz_id of the ORGNAME_global_admins group
- look up the authz_id of the "users" group for ORGNAME
- ask biforst if the users group is a member of the ORGNAME_global_admins group 

Since the only write operation in the migration is a no-op if the users group is already in the ORGNAME_global_admins group, I think that rerunning the migration is actually cheaper than figuring out which orgs haven't been migrated. Perhaps there is some shorter path that I'm missing.

In general, I think we need to fix up the migration state tracking functionality.  As far as I can tell, none of that works now that orgs are in sql.  It is also a bit suboptimal as it requires 1 row per org per migration.  Perhaps something closer to how sqitch tracks migrations but on a per-org basis would be better.

tiago-cruz-movile:
Hello,

In which version of Chef Server I can test this feature? :+1:

stevendanna:
@tiago-cruz-movile None yet.  It is going to be merged after the 12.1 release, hopefully for 12.2.  We held off on this change since it required additional data migrations and we already had a number of migrations in 12.1.

tylercloke:
Sounds good, maybe we can add some info about when requests will and will not fail to the API docs once I get this branch merged :) I have all affected endpoint documented except `/orgs/org/users` so we should try to keep them up to date.

https://github.com/chef/chef-server/pull/369/files

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
```diff
@@ -226,24 +226,35 @@ to_json(Req, #base_state{ resource_args = Args,
...
 
+superuser_or_self(UserId, UserId, _Req) ->
+    true;
+superuser_or_self(_UserId, _RequestorID, Req) ->
+    % The is_superuser function only knows how to deal with usernames.  Thus, we pull the
+    % username from request rather than using the RequestorID:
+    UserName = list_to_binary(wrq:get_req_header("x-ops-userid", Req)),
+    oc_chef_wm_base:is_superuser(UserName).
+
+orgs_to_ej(Orgs) ->
```
joedevivo: This was a good refactor. makes things cleaner for sure. I'm wondering if chef_object should have callbacks for to and from ej as a future enhancement

joedevivo:
+1, unless you want to rebase once #369 is merged and update the docs @tylercloke mentioned.

PR #246, opened 2015-05-11
Mp/mover policy cleanup 2

marcparadise:
A couple of corrections here: 
* force 'superuser' for authz deletion - because we never properly set up 'pivotal' with access due to the earlier issue that this cleans up, using pivotal fails to delete the resource.
* add logic to reprocess records from the previous demigration to ensure authz data is cleaned up.
* don't delete DB records if authz delete fails
* also log 500s from mv_oc_chef_authz_http 
* add failure logging in the cleanup migration itself 

ping @chef/lob @danielsdeleo


danielsdeleo:
Heh, so obvious in hindsight. :+1:

marcparadise:
I'd also like to understand why bifrost is returning a very slow (5s) 500 instead of a 400, when internally it does log that permission fails and that it should be a 400. But that's for a different day.

PR #248, opened 2015-05-11
ACL Restrictions - Prevent admins group removal from grant ACE

mmzyk:
Add in stubb for performing acl integrity checks

This commit changes nothing about how the current code operates. However, it does add a new module and call to that module that gives a hook for performing acl integrity checks, such as if we are removing the admin group from the grant ace, which is something we desire not to be possible.

This branch is branched from https://github.com/chef/chef-server/pull/240.

The desire at this point is that both of these branches could be merged to master and would essential be no-ops as the desired functionality is built on top of these changes.

### DIFF
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl_constraints.erl
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl_constraints.erl
@@ -20,15 +20,20 @@
 
 -module(oc_chef_authz_acl_constraints).
 
+-include_lib("ej/include/ej.hrl").
+
 -ifdef(TEST).
 -compile([export_all]).
 -endif.
 
 -export([check_acl_constraints/4]).
 
+-spec check_acl_constraints(binary(), atom(), binary(),tuple()) -> ok | [atom(),...].
 check_acl_constraints(AuthzId, Type, AclPerm, Ace) ->
   check_acl_constraints(AuthzId, Type, AclPerm, Ace, acl_checks()).
 
+
+-spec check_acl_constraints(binary(),atom(),binary(),tuple(), [fun()]) -> ok | [atom(),...].
 check_acl_constraints(AuthzId, Type, AclPerm, Ace, AclChecks) ->
   case lists:filtermap(fun(Check) -> Check(AuthzId, Type, AclPerm, Ace) end, AclChecks) of
     [] ->
@@ -37,11 +42,14 @@ check_acl_constraints(AuthzId, Type, AclPerm, Ace, AclChecks) ->
       Failures
   end.
 
+-spec acl_checks() -> [fun()].
 acl_checks() ->
   [
     fun check_admins_group_removal_from_grant_ace/4
   ].
 
+-spec check_admins_group_removal_from_grant_ace(binary(),atom(),binary(),tuple())
+      -> false | {true, attempted_admin_group_removal_grant_ace}.
 check_admins_group_removal_from_grant_ace(AuthzId, Type, AclPerm, NewAce) ->
   %% It is necessary to pull the current ace and compare to the new ace.
   %% This is because there are some groups that don't have the admin
@@ -65,6 +73,7 @@ check_admins_group_removal_from_grant_ace(AuthzId, Type, AclPerm, NewAce) ->
       false
   end.
 
+-spec check_admins_group_removal([binary()], [binary()]) -> 'not_removed' | 'removed'.
 check_admins_group_removal(CurrentGroups, NewGroups) ->
   %% Check if the CurrentGroups contains the admin group. If it doesn't, there
   %% is nothing to do. If it does, then check if the admin group is present in
@@ -81,6 +90,7 @@ check_admins_group_removal(CurrentGroups, NewGroups) ->
       end
   end.
 
+-spec contains_admins_group([binary()]) -> boolean().
 contains_admins_group(Groups) ->
   case lists:filter(fun(X) -> X =:= <<"admins">> end, Groups) of
     [] ->
@@ -89,6 +99,8 @@ contains_admins_group(Groups) ->
         true
     end.
 
+%% json_object is a type defined in ej;
+-spec extract_acl_groups(binary(), json_object()) -> [binary()].
 extract_acl_groups(AclPerm, Ace) ->
       ActorsAndGroups = ej:get({AclPerm}, Ace),
       ej:get({<<"groups">>}, ActorsAndGroups).
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl_constraints.erl
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl_constraints.erl
@@ -1,0 +1,94 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
+%% ex: ts=4 sw=4 et
+%% @author Mark  Mzyk <mm@chef.io>
+%% Copyright 2015 Chef Software, Inc.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+
+-module(oc_chef_authz_acl_constraints).
+
+-ifdef(TEST).
+-compile([export_all]).
+-endif.
+
+-export([check_acl_constraints/4]).
+
+check_acl_constraints(AuthzId, Type, AclPerm, Ace) ->
+  check_acl_constraints(AuthzId, Type, AclPerm, Ace, acl_checks()).
+
+check_acl_constraints(AuthzId, Type, AclPerm, Ace, AclChecks) ->
+  case lists:filtermap(fun(Check) -> Check(AuthzId, Type, AclPerm, Ace) end, AclChecks) of
+    [] ->
+      ok;
+    Failures ->
+      Failures
+  end.
+
+acl_checks() ->
+  [
+    fun check_admins_group_removal_from_grant_ace/4
+  ].
+
+check_admins_group_removal_from_grant_ace(AuthzId, Type, AclPerm, NewAce) ->
+  %% It is necessary to pull the current ace and compare to the new ace.
+  %% This is because there are some groups that don't have the admin
+  %% group by default, such as billing-admins. This will have the effect
+  %% that if a group doesn't have the admin group, but then it is later added,
+  %% the admin group will never be able to be removed.
+  case AclPerm of
+    <<"grant">> ->
+      NewGroups = extract_acl_groups(AclPerm, NewAce),
+      CurrentAce = oc_chef_authz_acl:fetch(Type, AuthzId),
+      CurrentGroups = extract_acl_groups(AclPerm, CurrentAce),
+      case check_admins_group_removal(CurrentGroups, NewGroups) of
+        not_removed ->
+          false;
+        removed ->
+          {true, attempted_admin_group_removal_grant_ace}
+      end;
+    _Other ->
+      ok
+  end.
+
+check_admins_group_removal(CurrentGroups, NewGroups) ->
+  %% Check if the CurrentGroups contains the admin group. If it doesn't, there
+  %% is nothing to do. If it does, then check if the admin group is present in
+  %% the new group.
+  case contains_admins_group(CurrentGroups) of
+    false ->
+      not_removed;
+    true ->
+      case contains_admins_group(NewGroups) of
+        true ->
+          not_removed;
+        false ->
+          removed
+      end
+  end.
+
+contains_admins_group(Groups) ->
+  case lists:filter(fun(X) -> X =:= <<"admins">> end, Groups) of
+    [] ->
+        false;
+    _NonEmpty ->
+        true
+    end.
+
+extract_acl_groups(AclPerm, Ace) ->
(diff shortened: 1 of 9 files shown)


mmzyk:
Nice, travis failed on dialyzer. Guess this isn't ready for merge then.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl_constraints.erl
```diff
@@ -0,0 +1,25 @@
...
+%% ex: ts=4 sw=4 et
+%% @author Mark  Mzyk <mm@chef.io>
+%% Copyright 2015 Chef, Inc. All Rights Reserved.
+
+-module(oc_chef_authz_acl_constraints).
+
+-export([check_acl_constraints/4]).
+
+check_acl_constraints(AuthzId, Type, AclPerm, Ace) ->
+  check_admin_group_removal_from_grant_ace(AuthzId, Type, AclPerm, Ace).
```
stevendanna: If we use true|false instead of ok, what would you think of the following:

```erlang
acl_checks() ->
    [
     check_admin_group_removal_from_grant_ace/4
    ].

check_acl_constraints(AuthzId, Type, AclPerm, Ace) ->
    lists:all(apply_check_fun(AuthzId, Type, AclPerm, Ace), 
              mandatory_checks()).

apply_check(AuthzId, Type, AclPerm, Ace) ->
    fun(Check) -> 
            apply(Check, [AuthzId, Type, AclPerm, Ace])
    end.
```

which might make it easier to add new checks in the future.

stevendanna:
What other ACL checks do we envision?  Part of me wonders whether

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl_constraints.erl
mmzyk: Yes, that makes sense. I hadn't gotten to that level of generalization yet, but was moving in the direction of using true|false. 

Good feedback, thanks.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
```diff
@@ -1,7 +1,8 @@
 %% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92-*-
 %% ex: ts=4 sw=4 et
 %% @author Douglas Triggs <doug@chef.io>
-%% Copyright 2014 Chef Software, Inc. All Rights Reserved.
+%% @auther Mark Mzyk <mm@chef.io)
```
smith: "@author"

smith: "<mm@chef.io>" not "<mm@chef.io)"

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
```diff
@@ -1,7 +1,8 @@
...
+%% Copyright 2014-2015 Chef Software, Inc. All Rights Reserved.
```
smith: "All Rights Reserved." should be removed. This contradicts the Apache 2.0 license below.

stevendanna: Heh, :+1: but I've largely given up.  Seems no one can agree on it.

mmzyk: Outside of changing names, this was just a copy paste and I didn't think about the details, so I'll update. Thanks for catching this.

mmzyk:
It appears this change has had some ripple effects with some other pedant tests, so I'm working on sorting those out.

Also, I should add in dialyzer hints were appropriate.

While I work on those tasks, we should debate if this is the direction we want to go or not.

@stevendanna, @manderson26 and the rest of @chef/lob, as well as any other interested parties.

stevendanna:
cc @chef/lob

mmzyk:
@stevendanna I almost wonder if y'all shouldn't make me an honorary member so I can ping you. 

Also, cc @chef/chef-visual-interaction so they can follow along and learn them some erlang for great good.

stevendanna:
@manderson26 and I talked a bit about this at the Chef Server triage meeting last week.  

To shortly summarize my original concern:

This and other changes are in pursuit of some higher-level policy such as "admins can always modify objects".  However, instead of creating a system that directly expresses that desire, we have implemented "by default, permissions are such that an admin can always modify objects, and we are going to make it hard to alter permissions in a way that would make that untrue".

My worry was that we were describing our desired policy obtusely, possibly making it unclear to future maintainers and making it fragile.

However, there are advantages to this particular implementation:

- we currently don't have a good place to express higher-level authorization policy in oc_erchef.  We could make one, but it would possible require a high level of churn in the auth code, which is a lot of risk to swallow at once.

- expressing some of that policy in oc_erchef would require additional bifrost/database calls per request

- it is even harder to express that policy in oc_bifrost since bifrost has none of the context needed (ie, it has no idea what an admin group is).

- we have other systems that might be doing permissions lookups to bifrost.  expressing these policies as a set of default bifrost state with change restrictions make it so that the policy is shared across those services.

@manderson26 feel free to add something I missed.

Those reasons quelled my initial concerns.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -839,6 +839,10 @@ malformed_request(Req, #base_state{resource_mod=Mod,
             lager:info("json too large (~p)", [Msg]),
             Req3 = wrq:set_resp_body(chef_json:encode({[{<<"error">>, Msg}]}), Req),
             {{halt, 413}, Req3, State1#base_state{log_msg = too_big}};
+        throw:{acl_constraint_violation, Violation} ->
+            Msg = chef_wm_malformed:malformed_request_message(Violation, Req, State),
+            Req3 = wrq:set_resp_body(chef_json:encode(Msg), Req),
+            {{halt, 403}, Req3, State1#base_state{log_msg = Msg}};
```
stevendanna: minor: I think a 400 makes more sense here.  The problem isn't that you didn't have permission to make the change, the problem is that the document you tried to post was malformed per our validation rules.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
```diff
@@ -69,13 +70,26 @@ validate_request('PUT', Req, #base_state{chef_db_context = DbContext,
...
+    %% anyway, so go ahead and do so
+    {Req1, State1 = #base_state{resource_state = #acl_state{
+          authz_id = AuthzId}}} =
+                                  oc_chef_wm_acl:validate_authz_id(Req, State,
+                                                                   AclState#acl_state{acl_data = Ace},
+                                                                    Type, OrgId,  DbContext),
+
+    %% Check if we're violating any constraints around modifying ACLs
+    %% i.e. deleting default groups, etc.
+    case oc_chef_authz_acl_constraints:check_acl_constraints(AuthzId, Type, Part, Ace) of
```
stevendanna: Should the superuser be exempt from these checks?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl_constraints.erl
```diff
@@ -0,0 +1,94 @@
...
+check_admins_group_removal_from_grant_ace(AuthzId, Type, AclPerm, NewAce) ->
+  %% It is necessary to pull the current ace and compare to the new ace.
+  %% This is because there are some groups that don't have the admin
+  %% group by default, such as billing-admins. This will have the effect
+  %% that if a group doesn't have the admin group, but then it is later added,
+  %% the admin group will never be able to be removed.
+  case AclPerm of
+    <<"grant">> ->
+      NewGroups = extract_acl_groups(AclPerm, NewAce),
+      CurrentAce = oc_chef_authz_acl:fetch(Type, AuthzId),
```
stevendanna: This will make an extra bifrost call per ACL update.  Since ACL updates are rare, this is fine.  I wonder if this will be common enough to make it worth it to pass in the old document as an argument to the check functions rather than the AuthzID?

stevendanna:
@mmzyk Is this still work in flight?

mmzyk:
@ssd yes. It got slightly derailed by a week in Seattle + a week of PTO. I'm returning to it this week to finish it up. I believe it is code complete, so any feedback is appreciated. However, last I checked some pedant tests failed in the code that I was unsure if they were related to these changes, so I need to investigate those failures.

mmzyk:
^ rebase that. Kicking the development of this back into gear (not a fixie)

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl_constraints.erl
mmzyk: That's certainly a change worth considering. I'll get it fully polished as is and then we can contemplate making that change.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
mmzyk: Good question. With great power comes great responsibility and all that jazz, but I'm dubious if we always need to give the superuser the power to route around everything in the system.

That said, we probably should be consistent. Do we allow the superuser to by pass other checks like this?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
mmzyk: I thought about 400 vs. 403. It comes down to how you want to parse things - you can say that it's malformed via our rules, or you can do as I did: it's a legal request that was sent across the wire just fine, but we're going to refuse to honor it, hence 403 forbidden (403 has a wider range than just you aren't allowed to access something, per my reading of the definition: http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html).

I'm not wedded to 403 and could easily go with 400. Anyone else have opinions for consideration?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
marcparadise: Not with any consistency, but we do in places.
The question becomes whether there's any repair/maintenance/support scenario which would make this action necessary for a superuser to perform?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
marcparadise: I think 403 fits better - the request is not invalid, we're just refusing to take the action.

mmzyk:
All the pedant tests now pass with latest updates. Superuser bypass has been implemented. This is almost ready to go, pending final review on it all.

There will be at least one more commit cleaning up the superuser check that will eliminate some copy/paste I found around the code base while researching how to implement the check.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl_constraints.erl
mmzyk: So I looked into if we could pass in the current state, but I'm not seeing anywhere that we'd already have it before this point. Since this endpoint only allows a put, we have the new ACL state, but we don't appears to have the existing ACL state. 

The acl_state record that might contain this data is updated in the oc_chef_wm_acl_permission code to hold the new ACL, so that doesn't have what we're looking for.

mmzyk:
Added the change log entry to the last commit.

mmzyk:
@stevendanna, @marcparadise, the rest of LOB team and @manderson26 , I believe this is ready to go.

Code is complete, all pedant tests pass, unit tests pass, dialyzer is happy, changelog entry line added to last commit.

A couple of questions:

Does this work need to incorporate/change the API version in anyway? I'm not sure how we're handling that. I assume no, but want to check.

This work will have the side effect of if you add the admins group to the grant ace of an object where it was not there before you will be unable to remove it, unless you use the superuser. I want to make sure this is okay. I believe it to be, since the admins group should likely always be in the grant ace, but if this isn't okay, speak up now.

Should this go out in the next server release after it goes down, or do we need to wait for another release (Chef server 13? 12.2? something else?)?

I am also going to test out how this change plays out in Manage to ensure it doesn't have unintended effects there.

Answers to these questions and any other feedback on this welcome.

mmzyk:
Per discussion on this issue in chef-manage-issues https://github.com/chef/chef-manage-issues/issues/54 the side effect of not being able to remove the admins group from the grant ace if it was added to a grant ace when it wasn't there to begin with should be fine, as we currently have the expectation it should be there.

mmzyk:
Tested a server with this change and Chef Manage to see how the UI behaved. Manage states there is an error while saving when the Grant box is unchecked for the admins group and amusingly puts the check back showing it's enabled. Makes me think of HAL refusing a command. Perhaps not the greatest UI in the world, but it functionally works and shouldn't stop us from shipping this.

smith:
Should we make a separate issue for having the UI handle this gracefully?

mmzyk:
@smith Once it's clear this code is going to be merged and this is the behavior we're going with, yes.

mmzyk:
Remain questions, other than general review on this PR:

Does this work need to incorporate/change the API version in anyway? I'm not sure how we're handling that. I assume no, but want to check.

Should this go out in the next server release after it goes down, or do we need to wait for another release (Chef server 13? 12.2? something else?)?

marcparadise:
:+1: these changes look good, though I think n

Re: release -  let's land this in 12.2 (current master, it's the next minor release out the door)  and make it as clear as possible in release notes and posts.  This prevents a significant cause of customer pain, and introducing it is a net gain for keeping orgs in a maintainable state.  

Re versioning:   In terms of the actual change, it's a minor alteration to ACLs API behavior that places an additional restriction on it, but it alters neither the inputs nor outputs of the API, nor does it change the effect of a successful call. Given that, I'd say let's not place this behavior under a new API version.

mmzyk:
@marcparadise thanks for the review. I shall get this merged then.

PR #251, opened 2015-05-12
Rename ORGNAME_global_admins groups to ORGNAME_read_access_group

stevendanna:
The ORGNAME_global_admins group has one purpose inside the Chef
Server. To provide READ access to users who join an organization.  Its
name, however, implies great power. This change renames the group to
something more appropriate by:

- changing the organization creation code to create a
  ORGNAME_read_access_group rather than a global_admins group,

- providing a migration for existing global_admins groups,

- changing oc_chef_authz to transparently handle either group

### DIFF
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_db.erl
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_db.erl
@@ -28,17 +28,18 @@
 
 -export([container_record_to_authz_id/2,
          fetch_container/3,
-         make_global_admin_group_name/1,
-         fetch_global_group_authz_id/3,
+         make_read_access_group_name/1,
+         fetch_read_access_group/2,
          fetch_group_authz_id/3,
+         fetch_global_group_authz_id/3,
          fetch_group/3,
          make_context/3,
          statements/1
         ]).
 
-%-ifdef(TEST).
+-ifdef(TEST).
 -compile([export_all]).
-%-endif.
+-endif.
 
 -include("../../include/oc_chef_authz.hrl").
 -include("oc_chef_authz_db.hrl").
@@ -92,14 +93,10 @@ fetch_container(#oc_chef_authz_context{reqid = ReqId}, OrgId, ContainerName) ->
             {error, Error}
     end.
 
-make_global_admin_group_name(OrgName) ->
-  lists:flatten(io_lib:format("~s_global_admins", [OrgName])).
-
 %% TODO: the only global groups are global admins groups and this should only be used for those
 fetch_global_group_authz_id(Ctx, OrgName, GroupName) ->
     RealGroupName = list_to_binary(lists:flatten(io_lib:format("~s_~s", [OrgName, GroupName]))),
     fetch_group_authz_id(Ctx, ?GLOBAL_PLACEHOLDER_ORG_ID, RealGroupName).
-
 
 %% @doc Retrieve the authz ID for a given group in an organaization.
 %%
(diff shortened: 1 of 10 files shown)


stevendanna:
cc @chef/lob 

Note this is based off of ssd/global-admins-migration, as it is a bit pointless until that is merged.

manderson26:
Looks reasonable. I'll have to modify fixie to handle this.

marcparadise:
To confirm - we're going to hold off on this until 12.1 RC is out?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_db.erl
```diff
@@ -550,17 +531,34 @@ fetch_group_authz_id_sql(#oc_chef_authz_context{server_api_version = ApiVersion,
 %%  * in oc_chef_authz_groups
 %%  * in oc_chef_organization_policy
 %%
-fetch_global_admins(#oc_chef_authz_context{otto_connection=_Server,
-                                           darklaunch = Darklaunch} = Ctx,
-                    OrgName) ->
-    GlobalGroupName = make_global_admin_group_name(OrgName),
+fetch_read_access_group(#oc_chef_authz_context{} = Ctx, OrgName) ->
```
joedevivo: probably should change the comment above to no longer reference 'global admins group'

also, I see that this function name was changed, but it's not exported, and I don't see any calls in this module that have changed to call this function. What am I missing?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_db.erl
```diff
@@ -550,17 +531,34 @@ fetch_group_authz_id_sql(#oc_chef_authz_context{server_api_version = ApiVersion,
...
+    ReadAccessGroupName = make_read_access_group_name(OrgName),
+    case fetch_global_group(Ctx, ReadAccessGroupName) of
+        {not_found, authz_group} ->
```
joedevivo: Maybe there should be a note in here that should be removed at some point when compatibility no longer matters?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_associations.erl
```diff
@@ -147,33 +147,33 @@ deprovision_delete_usag(ok, #context{usag = USAG,
...
     {error, {error_removing_from_org_user_group, Error}}.
 
-deprovision_fetch_org_global_admins(ok, #context{authz_context = AuthzContext,
-                                                 org_name = OrgName} = Context) ->
-    Result = oc_chef_authz_db:fetch_global_group_authz_id(AuthzContext, OrgName, "global_admins"),
-    deprovision_remove_global_org_admin_ace(Result, Context) ;
-deprovision_fetch_org_global_admins(Error, Context) ->
+deprovision_fetch_org_read_access_group(ok, #context{authz_context = AuthzContext,
+                                                     org_name = OrgName} = Context) ->
+    Result = oc_chef_authz_db:fetch_read_access_group(AuthzContext, OrgName),
```
joedevivo: so here's that function I mentioned before. I just don't see the export that lets us call it here.

stevendanna: Thanks for the careful review.  This was only working because the -ifdef around an export_all was commented out:

https://github.com/chef/chef-server/blob/ssd/global-admin-group-rename/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_db.erl#L38-L40

tylercloke:
Much more descriptive name :+1:

joedevivo:
:+1:

joedevivo:
for posterity, this code was merged as part of #376

PR #253, opened 2015-05-13
Update sqerl to latest version to address increased transaction rate

stevendanna:
Recent changes to sqerl resulted in users seeing double their previous
transaction rate.  This was caused by a `SET statement_time` command
being sent before every query.  The newest version of sqerl only sends
the command on startup.

ChangeLog-Entry: Eliminate unnecessary postgresql transactions caused
by new version of sqerl.


sdelano:
:+1:

PR #254, opened 2015-05-14
Update folsom and bear to avoid badarith errors

stevendanna:
The newest folsom pulls in the newest bear, which has fixes for
divide-by-zero errors. These errors are likely culprits for the
badarith errors that have been crashing the folsom_graphite_workers.

If enough folsom_graphite_workers fail quickly, the supervisor also
fails, bringing down the application (not a bug, this is the configure
OTP supervision tree behavior).


stevendanna:
cc @chef/lob

sdelano:
@stevendanna was there something here that necessitated the removal of `warnings_as_errors`?

marcparadise:
@sdelano beat me to the removal of warnings_as_errors comment

stevendanna:
Yup. Check out the commit message for that change.

marcparadise:
Cool. The 'clean up all the things' branch  have in flight (and will be returning to shortly) removes both of those deps.

PR #255, opened 2015-05-14
Update the bulk fetch cookbook query to return less data

stevendanna:
The only use of bulk_fetch_cookbook_verions is in
chef_wm_depsolvers. All cookbooks it returns pass through
chef_cookbook_version:minimal_cookbook_ejson, which only returns a
subset of the cookbook record. This change removes any data not used
in minimal_cookbook_ejson from the initial query to avoid doing extra
work in the database.

To wit, this change removes:

- checksums (and the subquery required to get them)
- id
- major
- minor
- patch
- meta_attributes
- meta_long_desc

from the return of the bulk fetch query.

To make the change clear, the query has been renamed to
bulk_fetch_minimal_cookbook_versions.

### DIFF
--- a/src/oc_erchef/apps/chef_db/priv/pgsql_statements.config
+++ b/src/oc_erchef/apps/chef_db/priv/pgsql_statements.config
@@ -361,13 +361,10 @@
 %% Used in implementation of environemnts/ENVNAME/cookbook_versions endpoint.
 %%
 %% Depends on the cbv type existing in the schema
-{bulk_fetch_cookbook_versions,
+{bulk_fetch_minimal_cookbook_versions,
  <<"WITH inputs AS ( select * from unnest($2::cbv[]) )
-  SELECT v.id,  v.major, v.minor, v.patch, v.frozen, v.meta_attributes,
-         v.meta_deps, v.meta_long_desc, v.metadata, v.serialized_object,
-         v.last_updated_by, v.created_at, v.updated_at, c.authz_id, c.org_id, c.name,
-         (SELECT ARRAY(SELECT checksum FROM cookbook_version_checksums cvc
-                       WHERE org_id = $1 AND cvc.cookbook_version_id = v.id)) checksums
+  SELECT v.frozen, v.meta_deps, v.metadata, v.serialized_object,
+         c.authz_id, c.org_id, c.name
     FROM inputs,
          cookbooks as c,
          cookbook_versions as v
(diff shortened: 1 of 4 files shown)


stevendanna:
cc @chef/lob 

Using test data of 20 cookbooks each with 1000s of files, this cuts rdbms_time substantially.

Before:
``` 
rdbms_time=26;
```

After:
```
rdbms_time=10;
```

Clearly my test data isn't representative of the pathological request times we are seeing.  I'm working on getting better test data now.  However, given that pedant still passes and that the new query is strictly less work, I am optimistic that this will help.

NB: When we deploy this in Hosted, my testing has shown we are likely to seen an initial spike in rdbms_time as I think we might miss the query cache.

danielsdeleo:
In chat you said you tested it manually, so :+1: from me. Just a word of caution that the depsolver spec, AFAICT, only creates empty cookbooks with no files. https://github.com/chef/chef-server/blob/5055f0ecfe5cdfd9f15716ee04aa16f41287572a/oc-chef-pedant/spec/api/depsolver_spec.rb

`make_cookbook` in there only creates empty cookbooks. It's only pedant code like this that creates cookbooks with files in them: https://github.com/chef/chef-server/blob/5055f0ecfe5cdfd9f15716ee04aa16f41287572a/oc-chef-pedant/spec/api/cookbooks/delete_spec.rb#L80-L98

stevendanna:
@danielsdeleo Good test.  I did manually test the return values of cookbook_version calls with cookbooks with lots of files.  But based on the fact that our pedant tests seem laking, I'm going to double check with some end-to-end testing that uses real cookbooks and runs chef-client.

danielsdeleo:
Cool, though if you can upgrade our test coverage in a similar amount of time that'd be boss.

stevendanna:
Using the following call before and after the change:

```
knife exec -E 'puts api.post("environments/_default/cookbook_versions", :run_list => ["cookbook1", "cookbook2", "cookbook3", "cookbook4", "cookbook5", "cookbook6", "cookbook7", "cookbook8", "cookbook9", "cookbook10", "cookbook11", "cookbook12", "cookbook13", "cookbook14", "cookbook15", "cookbook16", "cookbook17", "cookbook18", "cookbook19", "cookbook20"]).to_json' > FILENAME.json
```

I get exactly the same response:

```
root@api:~# md5sum before1.json after1.json
96c6bbae7243602098169d872422c517  before1.json
96c6bbae7243602098169d872422c517  after1.json
```

And nginx sees exactly the same number of bytes sent.

PR #256, opened 2015-05-14
[dvm] fix for dep loading to work again in the new location

marcparadise:
Ths unbreaks (to an extent) project dep loading, like 'dvm load oc_erchef sqerl' and 'dvm unload omnibus private-chef-cookbooks'


joedevivo:
:+1: This fixed the problem I was having.

PR #258, opened 2015-05-14
use orgid we already have instead of looking it up again

marcparadise:
ping @chef/lob 

ChangeLog-Entry: [perf] reduce database calls by using data we already have.

### DIFF
--- a/src/oc_erchef/apps/chef_db/src/chef_db.erl
+++ b/src/oc_erchef/apps/chef_db/src/chef_db.erl
@@ -35,7 +35,6 @@
 
          create_name_id_dict/3,
 
-         fetch_org_id/2,
          fetch_org_metadata/2,
 
          %% temporary hack for reindex script until
@@ -52,8 +51,6 @@
          %% user ops
          node_record_to_authz_id/2,
 
-         %% role ops
-         fetch_roles/2,
          %% role_record_to_authz_id/2,
 
          %% environment ops
@@ -274,12 +271,6 @@ update(ObjectRec, #context{reqid = ReqId}, ActorId) ->
 darklaunch_from_context(#context{darklaunch = Darklaunch}) ->
     Darklaunch.
 
--spec fetch_org_id(#context{}, binary()) -> not_found | binary().
-fetch_org_id(Context, OrgName) ->
-    case fetch_org_metadata(Context, OrgName) of
-        not_found -> not_found;
-        {OrgId, _OrgAuthzId} -> OrgId
-    end.
 
 %% Fetch the id and authz_id for a given organization name
 -spec fetch_org_metadata(#context{}, binary() | ?OSC_ORG_NAME) -> not_found | {binary(), binary()}.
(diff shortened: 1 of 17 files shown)


src/oc_erchef/apps/chef_db/src/chef_db.erl
```diff
@@ -52,8 +51,6 @@
          %% user ops
          node_record_to_authz_id/2,
 
-         %% role ops
-         fetch_roles/2,
```
manderson26: Interesting; we don't use that anywhere? Nice catch.

manderson26:
I'm very supportive of threading org id through the code rather than fetching it again. I'm a little bit concerned about removing orgname from the signature as it can be helpful in tracing and debugging to have a human readable name available in a way that is easy to access.

marcparadise:
In practice I've found that org name is usually captured elsewhere in the stack or nearby in the log - particularly since most crashes manifest with a 500 in the request log. 

In the worst case, we can still correlate it via the request ID which remains in the arg list via db context.

stevendanna:
:+1: This branch passes pedant both locally for me and in CI:

http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/509/downstreambuildview/

src/oc_erchef/apps/oc_chef_wm/itest/mocks/chef_otto.erl
```diff
@@ -6,12 +6,9 @@
 -module(chef_otto).
 
 -export([
-         connect/0,
-         fetch_org_id/2
+         connect/0
```
tylercloke: Do we need to be concerned about upgrades needing this code?

tylercloke:
Generally LGTM, are there any itest or unit tests we should add, since this seems like a pretty impactful change?

src/oc_erchef/apps/oc_chef_wm/itest/mocks/chef_otto.erl
stevendanna: In chef-mover that shouldn't be an issue since we lock the versions of all the dependency.  But I'm not sure if there is other upgrade machinery that would require it.  I can't think of any.

marcparadise: I searched mover for this and other modified functions and didn't find any references to it - either at chef_db or chef_otto level.

marcparadise:
@tylercloke Not that I can think of in the short term. (recap of in-channel conversation for posterity) . The offending fetch_org_id was fully removed, specifically so that things would crash badly if they tried to use it.  

However, I am in the process of verifying via coverage checks that we hit each of the changed functions in our pedant run.

manderson26:
A little late, but it was helpful to be able to redbug match the org name when tracing things in production. Now we have to redbug match the org guid, or figure out how to destructure a very large record with redbug. So we've lost something here I think.

mmzyk:
@manderson26 Perhaps that speaks to us needing to devise a way to better trace requests through the system, as this change seems important for us to keep load on the db down. 

Perhaps we should consider a change to standardize logging and tag all requests with a guid for the life of a request? Maybe that doesn't hit on exactly what you're looking for, but that would likely help us make the logs more useful and might reduce the need for redbug tracing, while also making ops life and our customer's ops teams happier.

(I've got Mark Imbriaco's advice on my brain: https://www.dropbox.com/s/rroy3wegh8x2awc/OperableSystems.pdf?dl=0)

PR #259, opened 2015-05-15
make the sqitch install use local cache for faster builds

marcparadise:
Looks like the change to cache the cpanm sqitch install and save a few minutes of travis build time got lost along the way, this re-adds it. 

cc @chef/lob 

Will merge after this passes.

### DIFF
--- a/src/oc_erchef/custom.mk
+++ b/src/oc_erchef/custom.mk
@@ -55,9 +55,11 @@ bundle:
 
 install:
 	@./rebar get-deps -C rebar.config.lock
+	@cpanm --local-lib=~/perl5 local::lib && eval $(perl -I ~/perl5/lib/perl5/ -Mlocal::lib)
 	@cpanm --notest --quiet App::Sqitch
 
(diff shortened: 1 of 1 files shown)


stevendanna:
:+1: But I don't think the cache is actually effecitve until we can re-enable containerized builds which we can't do until luarocks is whitelisted.

marcparadise:
The cache is still maintained - basically anything you declare as cached will get tar gz'd up locally, then extracted the next time the vm (or container) is spun up.

There's another issue here, though - need to tweak the make file further to actually ensure that the env vars exposed in the 'eval' statement get passed on to cpanm install of sqitch; in this build they don't.

stevendanna:
Not sure about that.  This page:

http://docs.travis-ci.com/user/caching/

says:

> Caching Dependencies and Directories

> The features described here are currently only available for private repositories on travis-ci.com and our new container-based infrastructure.

marcparadise:
obsoleted

PR #262, opened 2015-05-18
Fix chef-server-ctl install bugs

stevendanna:
This PR fixes 2 bugs in chef-server-ctl install:

- Stale node state was not being removed before
  the install command, leading to multiple packages
  to be installed in some cases.

- The opscode-analytics add-on should not be installed
  on the same machine as your chef-server

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/cleanup.rb
+++ b/omnibus/files/private-chef-ctl-commands/cleanup.rb
@@ -4,13 +4,11 @@
 add_command_under_category "cleanup", "general", "Perform post-upgrade removal of now-obsolete data, configuration files, logs, etc.  Add the '--no-op' flag to see what *would* be removed.", 2 do
   use_why_run_mode = ARGV.include?("--no-op")
 
-  # Our cleanup process is really just a special chef run
-  command = ["chef-client -z",
-             "--config #{base_path}/embedded/cookbooks/solo.rb",
-             "--json-attributes #{base_path}/embedded/cookbooks/post_upgrade_cleanup.json",
-             "--log_level fatal"] # yes, that's an underscore in log_level
-  command << "--why-run" if use_why_run_mode
+  attributes_path = "#{base_path}/embedded/cookbooks/post_upgrade_cleanup.json"
 
-  status = run_command(command.join(" "))
+  chef_args = "-l fatal"
+  chef_args << " --why-run" if use_why_run_mode
+
+  status = run_chef(attributes_path, chef_args)
   exit!(status.success? ? 0 : 1)
 end
--- a/omnibus/files/private-chef-ctl-commands/install.rb
+++ b/omnibus/files/private-chef-ctl-commands/install.rb
@@ -4,7 +4,6 @@
 KNOWN_ADDONS = %w(
   chef-ha
   chef-sync
-  opscode-analytics
   opscode-manage
   opscode-push-jobs-server
   opscode-reporting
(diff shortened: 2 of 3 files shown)


omnibus/files/private-chef-ctl-commands/install.rb
```diff
@@ -4,7 +4,6 @@
 KNOWN_ADDONS = %w(
   chef-ha
   chef-sync
-  opscode-analytics
```
stevendanna: This was requested by support.  Will it break any of the existing demo frameworks?

stevendanna:
cc @chef/lob

stevendanna:
CI Build: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/513/downstreambuildview/

omnibus/files/private-chef-ctl-commands/install.rb
mmzyk: Most likely. I know I've seen chatter about demos being done on a single machine even through we stopped supporting that for actually running it in production. That said, we should probably break the demos and find a solution there.

omnibus/files/private-chef-ctl-commands/cleanup.rb
```diff
@@ -4,13 +4,10 @@
...
   use_why_run_mode = ARGV.include?("--no-op")
 
-  # Our cleanup process is really just a special chef run
-  command = ["chef-client -z",
-             "--config #{base_path}/embedded/cookbooks/solo.rb",
-             "--json-attributes #{base_path}/embedded/cookbooks/post_upgrade_cleanup.json",
-             "--log_level fatal"] # yes, that's an underscore in log_level
-  command << "--why-run" if use_why_run_mode
+  chef_args = ""
+  chef_args << "--why-run" if use_why_run_mode
```
mmzyk: We seem to have lost setting the log_level to fatal. Not sure if that was intentional or not. Looking at the run_chef code in omnibus ctl I'm not seeing the log level set anywhere, so I'm assuming it's just using the default in this case. I'm not sure what the default is however.

omnibus/files/private-chef-ctl-commands/install.rb
```diff
@@ -31,6 +24,7 @@
     exit 1
   end
 
+  attributes_path = "#{base_path}/embedded/cookbooks/install_params.json"
```
mmzyk: Same comment on log_level fatal. At this point I assume it's removal is purposeful.

stevendanna: log_level fatal is the default of the run_chef method.

stevendanna: d'oh, no it isn't, good catch

stevendanna: restored the log level configuration

mmzyk:
I'm :+1: on this, assuming the updated omnibus-ctl version has already been pulled in (I didn't see any changes around that in this PR, but I also know we're trying to float off of master for dependencies like omnibus-ctl).

Some comments around log-level, but I think this change was intentional and I'm just missing some detail.

As far as demo setups, we might want to reach out to the pre-sales engineering team, who's name escapes me, b/c this likely matters to them. Also Michael Ducy asked about this relatively recently, so it would be good to reach out to him and see what they need in terms of setup so we can assist them. At that said, that should not stop the merging of this, because we should get this behavior right and then worry about making demos easy.

omnibus/files/private-chef-ctl-commands/install.rb
mmzyk: Just for consistency I'd add a chef_args var here and pass it in, since the same is done in the cleanup command, but that's more stylistic for consistency than functionality.

stevendanna: done

mmzyk:
I kicked the travis build as it failed for seemingly unrelated reasons failing to install some deps.

Once that passes, I'm :+1:

PR #264, opened 2015-05-18
[oc_erchef] update lockfile for sqerl

marcparadise:
This updates to the latest sqerl, which pipelines the bind & execute of a request instead of performing them separately.


marcparadise:
Proceeding with merge. Current oc-id failure is due to rubygems blip.

PR #269, opened 2015-05-20
[omnibus] add internal attribute for sever api version in use

marcparadise:
Just like it says on the label.  This is currently used for pedant config, and will also be used once we move the reindex script into a template, since it needs to be version-aware.

ping @chef/lob


stevendanna:
:+1:

PR #271, opened 2015-05-20
Provide pooler bounded-queuing configuration for all pooler controlled queues

stevendanna:
Previously, the only connection pooler with user-controllable queuing settings was
the sqerl pool in oc_erchef.  Now the queueing can be added to the following pools:

- sqerl in oc_erchef,
- depsolvers in oc_erchef,
- oc_chef_authz in oc_erchef,
- sqerl in oc_bifrost

By default, queueing is diabled because the relevant pooler_timeouts
are set to 0.  The queue max for sqerl is left at 20, while all others
are set to 50 which is the existing default in pooler.

As mentioned queueing is only used when a timeout > 0 is given.  This
also requires that pooler:take_member/2 is used rather than
pooler:take_member/1.  Thus, one callsite of pooler:take_member/1 has
been converted.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -180,11 +180,40 @@ default['private_chef']['opscode-erchef']['port'] = 8000
 default['private_chef']['opscode-erchef']['auth_skew'] = '900'
 default['private_chef']['opscode-erchef']['authz_pooler_timeout'] = '0'
 default['private_chef']['opscode-erchef']['bulk_fetch_batch_size'] = '5'
+default['private_chef']['opscode-erchef']['udp_socket_pool_size'] = '20'
+# Pool configuration for postgresql connections
+#
+# db_pool_size - the number of pgsql connections in the pool
+#
+# db_pool_queue_max - the maximum number of pgsql requests to queue up
+# if all connections are busy
+#
+# db_pooler_timeout - the maximum amount of time a request should wait
+# in the queue before timing out.  Request queueing is only effective
+# if db_pooler_timeout > 0
 default['private_chef']['opscode-erchef']['db_pool_size'] = '20'
 default['private_chef']['opscode-erchef']['db_pooler_timeout'] = '0'
 default['private_chef']['opscode-erchef']['sql_db_timeout'] = 5000
-default['private_chef']['opscode-erchef']['udp_socket_pool_size'] = '20'
 default['private_chef']['opscode-erchef']['db_pool_queue_max'] = '20'
+# Pool configuration for depsolver workers
+#
+# depsolver_worker_count - the number of depselector workers.  This is
+# a CPU bound task, thus setting this over the number of CPUs is not
+# advised.
+#
+# depsolver_pool_queue_max - the number of depsolve requets to queue
+# if all workers are busy.
+#
+# depsolver_pooler_timeout - the time in ms to wait before a queued
+# request times out.  Requesting queueing is only effective if
+# db_pooler_timeout > 0
+#
+# depsolver_tiemout - the amount of time to wait in ms before a
+# request to a depsolver worker is abandoned.
+default['private_chef']['opscode-erchef']['depsolver_pooler_timeout'] = '0'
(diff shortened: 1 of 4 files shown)


marcparadise:
:+1: LGTM

omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
```diff
@@ -223,6 +225,7 @@
                     [{name, chef_depsolver},
                      {max_count, <%= @depsolver_worker_count %>},
                      {init_count, <%= @depsolver_worker_count %>},
+                     {queue_max, <%= @depsolver_pool_queue_max %>}],
```
stevendanna: typo here, that "]" is no bueno

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -182,11 +182,40 @@
...
+# if all workers are busy.
+#
+# depsolver_pooler_timeout - the time in ms to wait before a queued
+# request times out.  Requesting queueing is only effective if
+# db_pooler_timeout > 0
+#
+# depsolver_tiemout - the amount of time to wait in ms before a
+# request to a depsolver worker is abandoned.
+default['private_chef']['opscode-erchef']['depsolver_pooler_timeout'] = '0'
+default['private_chef']['opscode-erchef']['depsolver_pool_queue_max'] = '50'
```
sdelano: Why are we, in two places, allowing a queue of 50 but setting a timeout of zero? It won't queue if the timeout is set to zero.

stevendanna: @sdelano Correct.  All the values I've used maintain the defaults that currently exist in the application.  Or at least, that was the intent.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -441,6 +471,9 @@
 ####
 default['private_chef']['oc_chef_authz']['http_init_count'] = 25
 default['private_chef']['oc_chef_authz']['http_max_count'] = 100
+# The queue max is only effective if authz_pooler_timeout (in the
```
sdelano: Does `authz_pooler_timeout` actually work like this? I'm going to go into the `opscoderl_httpc` library to check that out. I'm a bit surprised that we're not surfacing more tunables here and that they're not exposed via the main `pooler` config.

stevendanna: I believe it does.  That timeout configurable feeds this in the app config:

https://github.com/chef/chef-server/blob/ssd/swimsuits-and-sunblock/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb#L182-L184

which is read here and used in the take_member/2 function:

https://github.com/chef/opscoderl_httpc/blob/master/src/oc_httpc.erl#L115
https://github.com/chef/opscoderl_httpc/blob/master/src/oc_httpc.erl#L135-L145

The rest of the pool config is read in here:

https://github.com/chef/chef-server/blob/ssd/swimsuits-and-sunblock/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_http.erl#L201-L203

which is passed to oc_httpc:create_pool here:

https://github.com/chef/chef-server/blob/ssd/swimsuits-and-sunblock/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_http.erl#L191-L194

which explicitly calls the pooler:new_pool function here:

https://github.com/chef/chef-server/blob/ssd/swimsuits-and-sunblock/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_http.erl#L191-L194

I too find it odd that we didn't just use the pooler config for this.

stevendanna:
After much gnashing of teeth, this decided to build and pass pedant: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/526/downstreambuildview/

stevendanna:
Marking for review now that both travis and wilson are passing: @chef/lob

marcparadise:
:+1: This looks good to me. We may want to consider changing our defaults down the road, but htis makes sense for now by exposing it without changing anything.

PR #272, opened 2015-05-20
[oc_erchef] replace ejson calls with chef_json

marcparadise:
In preparation for removing couchdb, remove all calls to ejson
which is only pulled in via the couchdb dependency.

ping @chef/lob

### DIFF
--- a/src/oc_erchef/apps/chef_index/src/chef_index.app.src
+++ b/src/oc_erchef/apps/chef_index/src/chef_index.app.src
@@ -23,7 +23,6 @@
   {applications, [
                   kernel,
                   stdlib,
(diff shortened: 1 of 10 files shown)


stevendanna:
:+1: chef_json is a wrapper around jiffy which also tends to be faster than ejson (although it appears only one of the changes is in an actual module, so I doubt we'll see much impact :).

marcparadise:
The only components left will get wrapped up when we remove the actual couchbeam dependency which needs  ejson, in order to avoid breaking dialyzer and build: 

* chef_db/Makefile 
* reltool.config 
* rebar.config
* rebar.config.lock

PR #273, opened 2015-05-20
[oc_erchef] couch removal prep - remove chef_cache

marcparadise:
Remove `chef_cache`, which was used for org metadata caching.  It was useful with couch, but  changing to use  it under postgres  introduces its own potential issues with stale data. 

ping @chef/lob


src/oc_erchef/apps/chef_db/src/chef_db.erl
```diff
@@ -287,18 +287,7 @@ fetch_org_metadata(#context{darklaunch = Darklaunch} = Context, OrgName) ->
...
-                not_found ->
-                    not_found;
-                {OrgGuid, OrgAuthzId} ->
-                    chef_cache:put(org_metadata, OrgName, {OrgGuid, OrgAuthzId}),
-                    {OrgGuid, OrgAuthzId}
-            end;
-        {ok, {OrgGuid, OrgAuthzId}} ->
-            {OrgGuid, OrgAuthzId}
-    end.
+    ?SH_TIME(ReqId, chef_otto, fetch_org_metadata, (Server, OrgName)).
```
sdelano: I'm assuming this function is removed in a later PR, correct?

sdelano:
:+1:

src/oc_erchef/apps/chef_db/src/chef_db.erl
marcparadise: Yep

PR #274, opened 2015-05-20
[oc_erchef] move authz sql statements to pgsql_statements.config

marcparadise:
For consistency with chef_sql, move the statements previously
in oc_chef_authz_db.erl  to oc_chef_authz/priv/pgsql_statements.config

A further step here will be to consolidate both sql files, once we move
oc_chef_authz object modules into chef_objects.

ping @chef/lob


sdelano:
:+1:

PR #278, opened 2015-05-20
level up travis build caching and use system gecode for faster builds

marcparadise:
* install sqitch to cache 
* install chef-dk 
* modify travis build to use chefk embedded libgecode instead of building with each run
* remove a couple of files that were causing our cache to get repackaged on every build 

ping @chef/lob

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@
 # luarocks is added to the apt package whitelist:
 #   https://github.com/travis-ci/travis-ci/issues/3716
 #
+sudo: false
 branches:
   only:
     - master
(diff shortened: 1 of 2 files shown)


stevendanna:
Heads up, I merged the "enable containers" portion of this here:

https://github.com/chef/chef-server/pull/282

I think the libgecode caching is probably the next biggest win in terms of test time.  Have we tried caching the entire bundle rather than building libgecode ourselves?

marcparadise:
@stevendanna  I'm reluctant to cache components that are part of the build output - particularly if we want to look at building artifacts on travis at any point.

stevendanna:
@mp I see, that makes sense.

marcparadise:
Now updated with more accurate description.

marcparadise:
Odd, github isn't picking up on the fact that this was merged.

PR #285, opened 2015-05-21
Erlang 17.5 Compatibility

joedevivo:
This PR allows all projects to build in erlang 17.5 except for `chef-mover`.

`chef-mover` needs some attention to get that done, which it should be receiving shortly.

Since that's the case, these changes were all made to be compatible with R16B03-1 as well, and omnibus continues to build it that way.

### DIFF
--- a/src/bookshelf/rebar.config
+++ b/src/bookshelf/rebar.config
@@ -1,6 +1,6 @@
 %% -*- mode: erlang; -*-
 
-{require_otp_vsn, "17.*"}.
+{require_otp_vsn, "R16|17.*"}.
 
 {erl_dep_retries, 10}.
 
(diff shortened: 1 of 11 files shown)


src/bookshelf/rebar.config
```diff
@@ -31,7 +31,7 @@
 
 {erl_opts, [debug_info,
             {parse_transform, lager_transform},
-            warnings_as_errors}].
+            warnings_as_errors]}.
```
marcparadise: How was that working before?

stevendanna: My guess was that it wasn't but we didn't notice because rebar.config.lock gets used instead for nearly all builds.

joedevivo: @stevendanna is right. rebar.config.lock was giving us a pass on this. Although I have no idea how the last person to generate it was successful

stevendanna:
:+1:

PR #287, opened 2015-05-21
Explicitly call fetch_org_metadata in chef_wm_object_identifiers

stevendanna:
For typical requests, organization metadata is populated into the
base_state in malformed_request callback. The object_identifiers
module is an unauthenticated internal-use-only module that does not
export malformed_request. Thus, fetch_org_metadata was never being
called. This was resulting in the reporting add-on getting 404s for
valid node objects. This adds an explicit call to fetch_org_metadata
to avoid this problem.


stevendanna:
@sdelano gave the :+1: here: https://github.com/chef/chef-server/pull/286

PR #289, opened 2015-05-21
reindex into omnibus 2

marcparadise:
* Move reindex script into priv/ instead of rel/. 
* have reindex script ask erchef for the correct version to run via rpc
* add function to oc_erchef_app to request version.

ping @chef/lob


sdelano:
:+1: Thanks! Is `oc_erchef_app` the best place for this? My only other thought would be in `chef_wm_util`.

marcparadise:
I figured version was global enough (not really just wm) that it made sense at the highest level.

sdelano:
@marcparadise that works for me.

PR #290, opened 2015-05-22
Cleanup remnants of the couch removal

stevendanna:


stevendanna:
cc @chef/lob

marcparadise:
Thanks @stevendanna , it appears these got lost in the PR splits. :+1:

marcparadise:
Actually this is strange. These were all in the follow up PR to the couch usage removal, which got merged into that PR.

PR #297, opened 2015-05-22
cache deps and erlang plt files to avoid regen/redownload

marcparadise:
the cache mechanism used by container builds re-extracts a complete bz2-compressed tarball of the cache for each named item, resulting in a linearly increasing constant time operation for each line item that we add to the cache.  

This works around that by pulling cache content out of a single travis-cache item, .realcache/, using rsync to preserve dates/times/etc.   After the build it then uses rysnc to push changed things into .realcache/. 

Testing has shown that this does also prevent mtime updates to .realcache unless something has changed - so travis doesn't reupload the cache unnecessarily.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,18 +42,18 @@ env:
     - CHEFDK_GECODE_PATH=/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/dep-selector-libgecode-1.0.2/lib/dep-selector-libgecode/vendored-gecode
     - LUALIB=~/.luarocks/lib/lua/5.2
   matrix:
-  #- COMPONENT=omnibus
+  - COMPONENT=omnibus
   - COMPONENT=src/oc_erchef
   - COMPONENT=src/oc-id
   - COMPONENT=src/chef-mover
 install:
   - mkdir -p $HOME/.realcache
-  # Some errors expected but not fatal -- thus ||true
+    # Some warning-type errors expected . THis will give a non-zero return
+    # code and stop the build, thus ||true
   - rsync -azr --no-R --stats /home/travis/.realcache/ / || true
-    # TODO - luarocks local install of lpeg is failing, need to look at why.
-  #- luarocks install --local lpeg
-  #- luarocks install --local lua-cjson
-  #- eval $(luarocks path)
+  - luarocks install --local lpeg
+  - luarocks install --local lua-cjson
+  - eval $(luarocks path)
   - rvm install --binary 2.1.4
   - rvm use 2.1.4
   - cpanm --notest --quiet --local-lib=$HOME/perl5 local::lib
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,15 @@ branches:
 cache:
   apt: true
   directories:
-    - src/oc-id/vendor/bundle
-#    - $HOME/.luarocks/rocks
-    - $HOME/.cpanm
-    - $HOME/.cpan
-    - $HOME/perl5
+    - $HOME/.realcache
+
+
 before_cache:
   # Prevent build log from changing cache cand causing repackage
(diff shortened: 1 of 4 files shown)


marcparadise:
Seems to be something where the cache mechanism is creating a placeholder file for the named .plt files when no cache exists -  which causes dialyzer to fail since they're  not valid plts.  I'll keep this open for now to look into a bit more later.

marcparadise:
ping @chef/lob this is more-or-less ready, I"m just doing some final experiments to determine whether it's worth including rvm in the cache.

.travis.yml
```diff
@@ -39,23 +40,28 @@ env:
     - PERL5LIB=~/perl5/lib/perl5/x86_64-linux-gnu-thread-multi:~/perl5/lib/perl5:/etc/perl:/usr/local/lib/perl/5.14.2:/usr/local/share/perl/5.14.2:/usr/lib/perl5:/usr/share/perl5:/usr/lib/perl/5.14:/usr/share/perl/5.14:/usr/local/lib/site_perl
     - USE_OMNIBUS_FILES=0
     - CHEFDK_GECODE_PATH=/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/dep-selector-libgecode-1.0.2/lib/dep-selector-libgecode/vendored-gecode
-
-#    - LUALIB=~/.luarocks/lib/lua/5.2
+    - LUALIB=~/.luarocks/lib/lua/5.2
   matrix:
-    # - COMPONENT=omnibus
+  #- COMPONENT=omnibus
```
stevendanna: I think we can turn this back on now that you've re-enabled luarocks?

.travis.yml
```diff
@@ -39,23 +40,28 @@ env:
...
   - COMPONENT=src/chef-mover
 install:
-#  - sudo apt-get install luarocks
-#  - luarocks install --local lpeg
-#  - luarocks install --local lua-cjson
-#  - eval $(luarocks path)
+  - mkdir -p $HOME/.realcache
+  # Some errors expected but not fatal -- thus ||true
+  - rsync -azr --no-R --stats /home/travis/.realcache/ / || true
+    # TODO - luarocks local install of lpeg is failing, need to look at why.
```
stevendanna: oh i see :(

marcparadise:
Some crazytown experimentation going on here, not ready for merge with this latest commit.

.travis.yml
```diff
@@ -39,23 +40,27 @@ env:
     - PERL5LIB=~/perl5/lib/perl5/x86_64-linux-gnu-thread-multi:~/perl5/lib/perl5:/etc/perl:/usr/local/lib/perl/5.14.2:/usr/local/share/perl/5.14.2:/usr/lib/perl5:/usr/share/perl5:/usr/lib/perl/5.14:/usr/share/perl/5.14:/usr/local/lib/site_perl
     - USE_OMNIBUS_FILES=0
     - CHEFDK_GECODE_PATH=/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/dep-selector-libgecode-1.0.2/lib/dep-selector-libgecode/vendored-gecode
-
-#    - LUALIB=~/.luarocks/lib/lua/5.2
-  matrix:
```
stevendanna: Is removing the matrix build strictly necessary?  I'm concerned that (1) it will slow down the build (note that if we are just comparing build time on the reported total, the reported total is incorrect when using matrix builds as it sums all the builds even when they happen in parallel), and (2) that the feedback we get from builds will be much less clear.  Right now it is very easy to see which component failed its tests.

stevendanna:
Now that we are on rebar3, I believe that if we still want to do this, the easiest way is to add:

`$HOME/.cache/rebar3/` 

to this list of directories: https://github.com/chef/chef-server/blob/master/.travis.yml#L8

PR #301, opened 2015-05-26
Generate 2048 bit dhparam on first reconfigure

stevendanna:
We were previously using the default nginx/openssl values for dhparam.
This means that we were using a 1024 bit prime for Diffie-Hellman.
This is problematic because:

- it is smaller than our certificate size which means users of
  non-elliptic DHE were getting, in some sense, weaker security than
  those using less advanced ciphers, and

- recent research shows that 1024 bit primes are vulnerable to
  pre-computation attacks by nation-state level actors.


marcparadise:
:+1:

PR #304, opened 2015-05-27
[wip] return 500 when an unexpected error occurs during authn

sdelano:
Still a WIP:

**TODO**:
* [x] passing pedant / omnibus CI
* [x] needs unit / ct tests
* [x] updated dialyzer specs

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
@@ -426,7 +426,8 @@ is_authorized(Req, State) ->
         {false, ReqOther, StateOther} ->
             %% FIXME: the supported version is determined by the chef_authn application
             %% also, see: https://wiki.corp.chef.io/display/CORP/RFC+Authentication+Version+Negotiation
-            {"X-Ops-Sign version=\"1.0\" version=\"1.1\"", ReqOther, StateOther}
+            {"X-Ops-Sign version=\"1.0\" version=\"1.1\"", ReqOther, StateOther};
+        {{halt, _Code}, _Req, _State} = Halt -> Halt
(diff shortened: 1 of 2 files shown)


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -953,7 +954,7 @@ verify_request_signature(Req,
              State#base_state{log_msg = {not_found, user_or_client}}};
         {error, Error} ->
             Msg = verify_request_message(error_finding_user_or_client, Name, OrgName),
-            {false, wrq:set_resp_body(chef_json:encode(Msg), Req),
+            {{halt, 500}, wrq:set_resp_body(chef_json:encode(Msg), Req),
```
stevendanna: I'm wondering if we want a case specifically for no_connections where we return 503 so that we reserve 500 for "something unexpected went wrong".

marcparadise: :+1: to @stevendanna's suggestion

sdelano: That sounds good to me. I was just whipping something up really quick without tests (and even dialyzer specs) to see if it would pass pedant (it did). I deferred the decision on 500/503 because there are so many other places where we could catch the `no_connections` and return 503. Having said that, it's not blocking us from doing so here.

sdelano:
No longer WIP. 

> When unexpected errors occur during authentication, returning a 401
> has the effect of stopping the chef-client run because chef-client
> does not retry 401 errors. chef-client will retry 500 and 503 errors.

/ping @chef/lob

marcparadise:
:+1:

PR #309, opened 2015-05-28
ensure that postgres shared buffers are calculated correctly

### LINKED ISSUE #44: postgres shared_buffers calculation is broken because values have different units
A system with 64GB of memory couldn't just install CS12 and have it work because the `postgres['shared_buffers']` value was being set too high. This kept postgresql from starting because it was trying to `shmget` more space than `shmmax` allowed.

The problem is that everything in the calculation code linked below uses values that have different units. Some are in kB and some are in bytes.

1. the ohai value in `node['memory']['total'].to_i` is in kB
2. shmmax value is bytes
3. shared_bytes is assumed to be in kB

My recommendation is that every value should be converted to bytes and then proceed with calculations.

https://github.com/opscode/opscode-omnibus/blob/12.0.0/files/private-chef-cookbooks/private-chef/attributes/default.rb#L377-L385

thommay:
cc @mmzyk 
closes #44 

this causes both shmmax and the nodes memory to be handled in kilobytes, rather than shmmax being in bytes and memory in KB.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -4,7 +4,7 @@
 # All Rights Reserved
 #
 
-BYTES_IN_GIG = 1048576
+KB_IN_GIG = 1048576
 
 ###
 # Set a project-name for the enterprise-chef-common cookbook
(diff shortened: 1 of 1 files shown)


marcparadise:
If you rebase on latest master, this build should pass travis. Has it gone through wilson CI as well?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -426,15 +426,15 @@
 #  https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server
 # They generally recomend 25% of avalable memory, and we reserve 15% of shared memory for other uses
 # We limit to 2GB less than shmmax as exessive amounts of shared memory is unhelpful.
+shmkb = node['private_chef']['postgresql']['shmmax'] / 1024
```
stevendanna: It would be nice to have a comment about the units of these various values.

thommay:
@marcparadise yes, that's gone through a full wilson cycle quite happily.

stevendanna:
:+1: rebase and merge when ready

sdelano:
:+1: rebase and merge - also please update the CHANGELOG to reflect the changes. The CHANGELOG is in `omnibus/CHANGELOG`

thommay:
merged and pushed.

PR #315, opened 2015-05-29
not for merge: PoC - support external postgres instances 

marcparadise:
ping @chef/lob to take a look  and offer feedback. 

This is a preliminary PoC to demonstrate allowing chef server to use a remote postgres instance that it can connect to/run migrations against purposes of creating schemas and users, but does not attempt to otherwise own it. 

This PR needs some splitting up and cleanup, and there are more tasks/questions/tests  that I'll write up separately , but overall we don't have far to go from this point. 

Local pedant passes with --all using a remote host. I'm going to ship it off to CI to make sure I didn't break the default configuration.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -390,6 +390,7 @@ default['private_chef']['postgresql']['version'] = "9.2"
 # we'll be using these directories to determine what versions we have installed and
 # whether we need to run pg_upgrade.
 default['private_chef']['postgresql']['enable'] = true
+default['private_chef']['postgresql']['remote'] = false
 default['private_chef']['postgresql']['ha'] = false
 default['private_chef']['postgresql']['dir'] = "/var/opt/opscode/postgresql/#{node['private_chef']['postgresql']['version']}"
 default['private_chef']['postgresql']['data_dir'] = "/var/opt/opscode/postgresql/#{node['private_chef']['postgresql']['version']}/data"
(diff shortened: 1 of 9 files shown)


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -389,6 +389,7 @@
 # we'll be using these directories to determine what versions we have installed and
 # whether we need to run pg_upgrade.
 default['private_chef']['postgresql']['enable'] = true
+default['private_chef']['postgresql']['remote'] = false
```
stevendanna: Is there any way we could infer this automatically?

marcparadise: That's the path I started down, and I think it's possible but opted for the short path for now. 

We could create a helper method that checks for ipv4/ipv6 link-local addresses as well as iterating the bound IPs on available interfaces.   The  risk here is that it's possible someone will want to manage their own postgres, but still do it on the same box - in which case we'd get a false negative.

marcparadise:
Almsot forgot. Here's what needed in chef-server.rb to run with this branch: 

```
postgresql['remote'] = true 

postgresql['enable'] = false

postgresql['vip'] = 'some-remote-ip-or-host'

# Username of someone with supahpowahs 
postgresql['username'] = 'remote-username'

 # Can be any value if trust auth is enabled, but should be specified regardless
postgresql['password'] = 'remote-superuser-password'
```

marcparadise:
This now works with both embedded and remote postgres. 

Leaving the PR open for a bit to gather any thoughts or feedback to include in the 'next things' document. 

Another thing for the list: 'cleanse' will need to be modified to actually talk to the DB for deletions if remote is configured, since we can't just nuke the data directory.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
mivok: How about checking for postgresql['enable'] == false to avoid the false negative. That way, if postgresql['vip'] evaluates to a remote machine, or if it evaluates to a local address, but postgres is disabled, then treat it as remote.

PR #316, opened 2015-05-29
omnibus: build chef last

sdelano:
We build chef from git master, so it changes frequently. To avoid
dirtying the omnibus git cache on each new build / nightly, this
commit puts chef and the end of the build. This should speed up
rebuilds significantly.

/cc @chef/lob 

TODO:
* [x] passing [ci](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/542/downstreambuildview/)
* [x] verification that chef actually builds last


tylercloke:
:+1:

PR #318, opened 2015-05-30
Allow file specificity to be null

coderanger:
This happens when you put templates or cookbook files directly in the base folder (added in Chef 12.0.0). I have no idea if this patch is correct, just trying to reconstruct based on reading the existing source and ej. Feel free to chuck it and write a better patch.

### DIFF
--- a/src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
+++ b/src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
@@ -275,7 +275,7 @@ file_list_spec() ->
     {array_map, {[{<<"name">>, string},
                   {<<"path">>, string},
                   {<<"checksum">>, string},
-                  {<<"specificity">>, string}
+                  {<<"specificity">>, {any_of, {string, null}}}
                  ]}}.
 
 -define(COOKBOOK_SEGMENTS, [<<"attributes">>,
--- a/src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
+++ b/src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
@@ -275,7 +275,7 @@ file_list_spec() ->
     {array_map, {[{<<"name">>, string},
                   {<<"path">>, string},
                   {<<"checksum">>, string},
-                  {<<"specificity">>, {any_of, {string, null}}}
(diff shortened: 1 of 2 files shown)


coderanger:
To clarify after a bit more research, this is only triggered by Berkshelf. Knife correctly populates that field as `'root_default'`. Going to fix this on the berks side too, but would be good to make this field optional as we move towards deprecating it.

marcparadise:
Thanks for thePR Noah.  I'd like to get some pedant coverage on this as well, and make it deprecated in api v1 - if you have some time tomorrow, we can discuss in irc?

coderanger:
Sure, I'll be around. Ridley side of this fix is tracked in https://github.com/reset/ridley/pull/309

coderanger:
Also I see in the ej code there is an `opt` flag that maybe I should be using instead?

marcparadise:
I'm currently waiting for a flight to take off, but will take a closer look at the full change tonight or early am

stevendanna:
> Also I see in the ej code there is an opt flag that maybe I should be using instead?

Even with the opt flag, if the key exists ej:valid requires that it pass the type check, so we would still need the change you've included here.

```
7> ej:valid({[{{opt, <<"foo">>}, string}]}, {[{}]}).
ok
8> ej:valid({[{{opt, <<"foo">>}, string}]}, {[{<<"foo">>, null}]}).
{ej_invalid,json_type,<<"foo">>,null,null,string,undefined}
```

If we allow that key to be optional on the way in, my concern is whether we have clients that will expect it in results such that we'd have to add it ourselves either on the way in or the way out.

>  I'd like to get some pedant coverage on this as well, and make it deprecated in api v1

+1 on the pedant tests, but I think at this point the deprecation would need to be in v2 since v1 is released?  I guess it depends on what you mean by "deprecated" since we don't have a good way to single that to API users currently.

stevendanna:
OK. This is nearly 2 years old.  @thommay @coderanger Is this still a relevant change?  If so, @juliandunn @dmccown can we get this assigned to POOL to add whatever tests we think we need.

thommay:
I think it's not - implementations should always send a specificity and we should reject null.

stevendanna:
OK, closing based on @thommay's comment. We can reopen in the future if we need to.

PR #320, opened 2015-06-01
Bypasses a memory leak in Erlang's ssl_session_cache

joedevivo:
For a particular load scenario, Erlang's ssl_session_cache would use up
too much memory and then the CPU would start churning, hoping to GC all
that memory, but it was never able to catch up. By replacing
ssl_manager's callback module (originally ssl_session_cache) to our
noop_session_cache, we essentially do nothing, and therefore leak no
memory.

Reproducing this error, often got CPU usage up to at least 300% in the
short term, and over 700% in 30 minutes or less. With this fix, I have
yet to see CPU usage go over 200%.

P.S. This is terrible branch name for what this actually did.

ping @chef/lob and everybody!

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -236,6 +236,7 @@ default['private_chef']['opscode-erchef']['keygen_cache_workers'] = :auto
 default['private_chef']['opscode-erchef']['keygen_timeout'] = 1000
 default['private_chef']['opscode-erchef']['keygen_key_size'] = 2048
 default['private_chef']['opscode-erchef']['strict_search_result_acls'] = false
+default['private_chef']['opscode-erchef']['ssl_session_caching']['enabled'] = false
 
 ###
 # Legacy path (required for cookbok migration)
--- a/src/chef-mover/test/mover_eredis_sup_test.erl
+++ b/src/chef-mover/test/mover_eredis_sup_test.erl
@@ -12,5 +12,5 @@
 %% the start_link path from being followed.
 start_link_should_start_client() ->
     process_flag(trap_exit, true),
-    ?assertEqual({error, {connection_error, {connection_error,nxdomain}}}, mover_eredis_sup:eredis_start_link(host, port)).
-            
+    ?assertMatch({error, {connection_error, {connection_error,_}}}, mover_eredis_sup:eredis_start_link("host", 1234)).
+
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
@@ -227,6 +227,11 @@
                      {start_mfa, {chef_depsolver_worker, start_link, []}}]]},
            {metrics_module, folsom_metrics}
           ]}
+<% if !node['private_chef']['opscode-erchef']['ssl_session_caching']['enabled'] -%>
+,  {ssl, [
+      {session_cb, noop_session_cache}
+  ]}
+<% end -%>
 <% if node['private_chef']['folsom_graphite']['enabled'] -%>
 , {folsom_graphite, [
                      {graphite_host, "<%= @helper.normalize_host(node['private_chef']['folsom_graphite']['host']) %>"},
--- a/src/oc_erchef/rebar.config.lock
+++ b/src/oc_erchef/rebar.config.lock
@@ -98,7 +98,7 @@
                               "fd40c960c7912193631d948fe962e1162a8d1334"}},
        {erlware_commons,".*",
                         {git,"https://github.com/erlware/erlware_commons.git",
-                             "ef0d252b11c863f9c228af2fe93a4e42fba2f7f3"}},
+                             "2e23e43079686ddb68bfca772d37f78dfe4dd95e"}},
        {darklaunch,".*",
                    {git,"https://github.com/opscode/opscode-darklaunch-erlang",
                         "05881cb04e9393ab42b6fac3b22803130ef2701c"}},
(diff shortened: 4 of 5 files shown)


src/oc_erchef/rebar.config
```diff
@@ -54,7 +54,7 @@
         {jiffy, ".*",
          {git, "https://github.com/joedevivo/jiffy", {branch, "master"}}},
         {ibrowse, ".*",
-         {git, "https://github.com/opscode/ibrowse", {tag, "v4.0.1.1"}}},
+         {git, "https://github.com/cmullaparthi/ibrowse", {branch, "master"}}},
```
sdelano: Do we have to pull in the new version of ibrowse for this to work? Considering how much we all love ibrowse around here, it seems like quite a large undertaking to swallow ~2.5 years worth of change from that project.

src/chef-mover/test/mover_eredis_sup_test.erl
```diff
@@ -12,5 +12,5 @@
 %% the start_link path from being followed.
 start_link_should_start_client() ->
     process_flag(trap_exit, true),
-    ?assertEqual({error, {connection_error, {connection_error,nxdomain}}}, mover_eredis_sup:eredis_start_link(host, port)).
-            
+    ?assertEqual({error, {connection_error, {connection_error,timeout}}}, mover_eredis_sup:eredis_start_link("host", 1234)).
```
sdelano: This change fails travis. You had mentioned earlier that it was related to a new version of the `eredis` library, but I don't see a new version getting pulled in here.

sdelano:
The only other concern here is that we'll be having to re-do ssl session negotiation for every connection. This ssl option is global, so we can't turn it on or off for a certain upstream.

joedevivo:
We can take a look at doing our own implementation of the behavior

stevendanna:
Nice find!

> The only other concern here is that we'll be having to re-do ssl session negotiation for every connection.  This ssl option is global, so we can't turn it on or off for a certain upstream.

We may be already.

Did we confirm whether sessions were actually getting reused at all?  I because the nginx documentation (http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_session_cache) claims that the default for `ssl_session_cache` is `none` and I can't see anywhere in omnibus'  nginx configuration where we change that:

https://github.com/chef/opscode-omnibus/blob/master/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb#L6-L16

(we set the timeout, but I don't think that matters if we haven't enable the cache). Compare that too Hosted's configuration (private, sorry):

https://github.com/chef/opscode-platform-cookbooks/blob/master/cookbooks/opscode-lb/templates/default/nginx/api.conf.erb#L77-L78

If I'm right here, it means we also aren't using ssl session caching for client connections to the frontends.  If I'm wrong here, it still means that the server is timing out those sessions at 5 minutes.  Cookbook uploads are rare events for most servers so it is likely that for most users, most new cookbook operations would have to renegotiate.  In the default config, we only talk to bookshelf over TLS, so no other frontend->backend requests will keep those sessions alive.

stevendanna:
Brief update: In our current configuration, it looks like we support session resumption only for tls implementations that support session tickets but not for those that only support session resumption by session-id.

stevendanna:
After playing around in the erl shell and reading a bit of the code, I'm pretty confident that in our current nginx configuration, we weren't actually able to take advantage of the ssl session cache because it doesn't support session tickets.

As such, I don't think this patch will cause any additional ssl negotiations than we already do.

:+1: from me; if we have time I'd be interested in seeing if enabling session-id-based-session-resumption in nginx and leaving the session cache on in erchef also resolves the problem.  However, given what we have seen, I wouldn't be surprised if that just delays the problem.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -236,6 +236,7 @@
 default['private_chef']['opscode-erchef']['keygen_timeout'] = 1000
 default['private_chef']['opscode-erchef']['keygen_key_size'] = 2048
 default['private_chef']['opscode-erchef']['strict_search_result_acls'] = false
+default['private_chef']['opscode-erchef']['ssl_session_caching']['enabled'] = true
```
stevendanna: This feels a bit backwards.  This should be `false` and the conditional in the template should be `if !`

sdelano:
> If the connections aren't being reused, I also wonder if that is part of the erlang problem. Perhaps there is a path where sessions aren't being invalidated properly, so the session list keeps growing.

The default expiry time (also doesn't appear configurable) for the client-side (erlang) ssl session cache is 24h.

sdelano:
Also :+1: on this once it's passing CI

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
joedevivo: you're right.

joedevivo:
CI Passing. Last step is just to deploy on dev-vm and make sure `reconfigure` turns ssl_caching on and off.

joedevivo:
```
default['private_chef']['opscode-erchef']['ssl_session_caching']['enabled'] = false
```

```
root@api:~# tail -n 6 /var/opt/opscode/opscode-erchef/sys.config
           {metrics_module, folsom_metrics}
          ]}
,  {ssl, [
      {session_cb, noop_session_cache}
  ]}
].
root@api:~# echo "opscode_erchef['ssl_session_caching'] =  {'enabled' => true}" >> /etc/opscode/chef-server.rb
root@api:~# private-chef-ctl reconfigure
...
opscode Reconfigured!
root@api:~# tail -n 6 /var/opt/opscode/opscode-erchef/sys.config
                     {init_count, 5},
                     {queue_max, 50},
                     {start_mfa, {chef_depsolver_worker, start_link, []}}]]},
           {metrics_module, folsom_metrics}
          ]}
].

```

joedevivo:
TODO: This setting should be documented.

PR #322, opened 2015-06-02
Remove more couch-related dead code

stevendanna:


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_routes.erl
```diff
@@ -280,19 +280,5 @@ make_args(Item, _, OrgName) ->
...
-%% item is under the 'raw_data' key of the wrapper. When data_bag_items are stored in SQL,
-%% just the item is stored. This helper function extracts the id of the data_bag_item for
-%% either type. We could have used darklaunch here, but since there are two cases and it is
-%% easy to test for, we just inspect the data at hand. This can go away once data_bag_items
-%% are only in SQL.
-%% TODO: when data bags are no longer in couchdb, clean this
 data_bag_item_id(Item) ->
-    case ej:get({<<"id">>}, Item) of
-        undefined ->
-            %% we must have a structure coming out of couchdb wrapped in data_bag_item cruft
```
marcparadise: I had left this one in on my branch because I figured it was possible we still had old data out there that fit it - but looking closer , I think removing it is correct, since this would have been changed/fixed during the original migration.

stevendanna: Old data is definitely the only question I had.  I'm wondering if we can check hosted for such data.  If such data exists, Hosted is likely to have some of it.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
```diff
@@ -410,40 +396,15 @@ encode_results(Results, Prefix, Acc) ->
...
-%% This function knows how to deal with gzip binary from SQL and with EJSON data coming
-%% straight from couch. If the data has come from couch, this is where couch cruft keys _id
-%% and _rev are removed.
 encode_result_rows([Item|_Rest]=Items) when is_binary(Item) ->
     ItemList = << <<(chef_db_compression:decompress(Bin))/binary, ",">> || Bin <- Items >>,
     %% remove trailing "," from binary
     binary:part(ItemList, {0, size(ItemList) - 1});
 encode_result_rows(Items) ->
-    %% ensure no couchdb cruft leaks out
-    CleanItems = [ {remove_couchdb_keys(Doc)} || {Doc} <- Items ],
```
marcparadise: As with the other change, I had left this one in in my original branch.  I'm less sure about removing it - is it possible that any of these keys remain in migrated data?

marcparadise:
So it looks like this and the changes from your earlier PR got lost when I did a local merge of the branch containing them intto its parent branch, but then forgot to push the parent branch before merging it into master.

marcparadise:
@stevendanna is this good to merge?

stevendanna:
I've opened a new PR here: https://github.com/chef/chef-server/pull/529 with just the .hrl removal.  I think there is enough of a question about the possibility of bad data somewhere that we can just leave the other code paths.

PR #324, opened 2015-06-02
Changed runit command for oc_id to use unicorn by default

paulmooring:
We were installing unicorn but not using it, this change leaves the same default port but changes the rack server to unicorn.  There was already a unicorn config in place which is now used, but wasn't modified.


tylercloke:
Ping @chef/lob

raskchanky:
Doesn't running `bundle exec rails server` start Unicorn already?  This is what I see locally:

2015-06-02 at 3 52 pm

Unless runit works differently than this?

smith:
Yeah @paulmooring, I think @raskchanky is right. What server was it using before you made this change?

paulmooring:
@smith @raskchanky you're correct, there was another problem with my config causing some strangeness.

PR #326, opened 2015-06-03
erlang 17.4 travis support

marcparadise:
1. Use TRAVIS_OTP_VERSION to grab the correct plt file (pulled in via `concrete update` 
2. Now that the 17.4 PLT has been uploaded, change the project back to using 17.4


marcparadise:
ping @chef/lob 

Just pending seal of approval from travis.

PR #328, opened 2015-06-03
Use custom context configuration for solr jetty

### LINKED ISSUE #323: Chef Server 12.0.8 -> 12.1.0-rc.1 doesn't upgrade Solr correctly
On my recently upgraded system, Solr still reports itself as 4.5.1, and still has the old jars deployed:
```
root@ip-ub-backend1:/opt/opscode/embedded/service/opscode-solr4/jetty# find . -type f -name "*solr*"
./solr-webapp/webapp/img/solr-ico.png
./solr-webapp/webapp/img/solr.png
./solr-webapp/webapp/WEB-INF/lib/solr-core-4.5.1.jar
./solr-webapp/webapp/WEB-INF/lib/solr-solrj-4.5.1.jar
```

The *new* solr war does live `/opt/opscode/embedded/service/opscode-solr4/jetty/webapps/solr.war` but it appears that Jetty is not auto-deploying it.   To correct this, I had to:

1.  stop solr:  `chef-server-ctl stop opscode-solr4`
2.  remove the solr-webapp directory:  `rm -rf /opt/opscode/embedded/service/opscode-solr4/jetty/solr-webapp`
3. start solr:  `chef-server-ctl start opscode-solr4`

now this shows the correctly jar versions:
```
root@ip-ub-backend1:/opt/opscode/embedded/service/opscode-solr4/jetty# find . -name "*solr*"
./solr-webapp
./solr-webapp/webapp/img/solr-ico.png
./solr-webapp/webapp/img/solr.png
./solr-webapp/webapp/WEB-INF/lib/solr-solrj-4.9.1.jar
./solr-webapp/webapp/WEB-INF/lib/solr-core-4.9.1.jar
```

stevendanna:
Previously, the default context configuration was setting
tempDirectory in the context configuration for solr. In Jetty 8, this
means that jetty will initially extract the WAR to the specified
directory and only re-extract it when the mtime on the WAR is newer
than the mtime on the directory.  Unfortunately, this leads to cases
where solr will not get updated on chef-server upgrades if the current
version was installed after the new version was built.

To avoid custom logic to update the WARs mtime, we configure jetty
without tempDirectory, leading it to extract the WAR on every restart.
To avoid depending on /tmp, we also set java.io.tmpdir.

Fixes #323


stevendanna:
cc @chef/lob @irvingpop 

There are many ways to solve the deployment problem.  This is one.  The pros are that we don't have to upgrade jetty and don't have to maintain custom logic in the recipe related to the mtime of the WAR.  The downsides are that we extract the WAR on every restart and we have moved java.io.tmpdir which solr may use for other things.  The WAR extraction is very fast from what I've seen.  Using `inotifywait` I observed /tmp through a full pedant run and didn't see solr make any other uses of it, but it could be that it starts using temporary files when handling data above the sizes we see in pedant.  This could be a performance issue if, for instance, those temporary files were previously on a tempfs and now are on a disk-backed filesystem.

Some alternatives:

- This same change but without the change to java.io.tmpdir, this would introduce another use of /tmp, which we have been trying to avoid but would have the benefit of unattended consequences of changing java.io.tmpdir.

- Move to Jetty 9 and use the persistTempDirectory setting to always re-extract the war.  This would require managing jetty ourselves rather than depending on the version shipped with solr.  Moving to Jetty 9 has been requested by at least one user.

- Custom logic in the cookbook to update the mtime of solr.war when an upgrade is needed.  This will force jetty to re-extract it without changing the current configuration.  

Overall, the relative advantages of these aren't really that clear to me even after a bit of research into jetty 8 and jetty 9.

irvingpop:
:+1: this seems totally reasonable to me to make it work with Jetty 8 for now, rather than bump Jetty during an RC cycle

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-solr4.rb
```diff
@@ -90,6 +93,15 @@
   notifies :restart, 'runit_service[opscode-solr4]' if is_data_master?
 end
 
+template File.join(solr_jetty_dir, "contexts", "solr-jetty-context.xml") do
+  owner OmnibusHelper.new(node).ownership['owner']
+  group OmnibusHelper.new(node).ownership['group']
+  mode "0644"
+  source "solr4/solr-jetty-context.xml.erb"
+  notifies :restart, 'runit_service[opscode-solr4]' if is_data_master?
+end
```
sdelano: Why is this a template and not a file resource? Nothing looks configurable here.

sdelano:
@stevendanna other than the template / file comment above, has this been tested as an upgrade?

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-solr4.rb
stevendanna: Mostly because I have the expectation that we'll want to manage more of this file in the future if we upgrade to jetty 9.

stevendanna:
Re upgrade testing.  I tested upgrades during development but I'm running through it again now.

src/chef-mover/src/mover.app.src
```diff
@@ -7,10 +7,19 @@
                   stdlib,
                   sasl,
                   crypto,
+                  envy,
                   inets,
                   ibrowse,
                   lager,
                   bcrypt,
+                  moser,
+                  depsolver,
```
stevendanna: A lot of these are actually transitive dependencies of moser.  We should probably declare them there.

danielsdeleo:
LGTM

PR #329, opened 2015-06-03
Add policy groups list API

danielsdeleo:
Add API for `GET $url_root/policy_groups`

Note that this discloses more information about individual "sub-objects" than we do for other APIs, but I did not do bulk AuthZ. I'm open to adding that if we consider it a problem. (Check the pedant tests for more info). Also, this API involves a bit more database effort than our standard list APIs do. On the other hand, the UI for this in ChefDK needs all this data, so the only other option would be to "spider" the API which I assume would be a lot more work than one join query.

More API endpoints to follow.

@chef/lob

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policy_groups.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policy_groups.erl
@@ -1,0 +1,113 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92-*-
+%% ex: ts=4 sw=4 et
+%% @author Daniel DeLeo <dan@chef.io>
+%% Copyright 2015 Chef Software, Inc. All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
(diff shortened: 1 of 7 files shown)


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policy_groups.erl
```diff
@@ -0,0 +1,113 @@
...
+
+build_nested_base_ejson([{[PolicyGroupName], URI}|Rest], EJSON) ->
+    NewEJSON = ej:set_p({PolicyGroupName, "uri"}, EJSON, URI),
+    build_nested_base_ejson(Rest, NewEJSON);
+build_nested_base_ejson([], EJSON) ->
+    EJSON.
+
+build_nested_list_data([Row|Rest], EJSON) ->
+    {PolicyGroupName, PolicyName, RevisionID} = Row,
+    NewEJSON = ej:set_p({PolicyGroupName, "policies", PolicyName, "revision_id"}, EJSON, RevisionID),
```
marcparadise: Is it possible to have more than one revision ID for a policy group name + policy name? If so, we should have a test for that as well -- I think this will capture only one of them as it is.

marcparadise:
Looks good other than the one question.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policy_groups.erl
danielsdeleo: Nope, revision is is unique per (policy name, policy group)

Issue #303, opened 2015-05-26
Customers want to be able to specify root level attributes in Chef Search


jeremiahsnapp:
Customers want to be able to specify root level attributes in Chef Search to avoid the problems caused by extracting [nested fields](http://docs.chef.io/chef_search.html#nested-fields) to the top level.

>When nested fields are present in a JSON structure, the chef-client [actually opscode-expander right?] will extract those nested fields to the top-level, flattening them into compound fields that support wildcard search patterns.

Some very important attributes are at the top level such as chef_environment. Customers have cookbooks or other tools that rely on Chef Search to return all nodes in a particular chef_environment but all it takes to cause unexpected results (and a production outage) is having another chef_environment attribute with a different value nested anywhere at all in the node data.

I discussed this briefly with one of the developers and thought it could be possible (maybe not even hard) to enable this functionality for users but that a feature request for it would likely get triaged as a low priority. I would like to see this kind of improvement in reliability be made a much higher priority. Especially if it isn't even that hard to do.

I've seen time and time again that two separate Chef Server features can give completely correct results and yet cause customers serious trouble when those results aren't at all what they expected. The two features are Cookbook Depsolver and Chef Search. Currently all we can say is the server is "right" even though you're not getting what you expect. And then we have to train the customer to learn how to do their best to avoid these problems in the future.

I love the direction Policyfiles is taking to make it easier for customers to explicitly declare what set of recipes will be run on a node ahead of time rather than being bitten by the complexity of cookbook depsolving during a deploy. I think adding the ability to specify root level attributes in a Chef Search would be another great win in the battle to make things more sane for customers.

jrwesolo:
Here is an example to reproduce the issue with nested attributes named the same as root level attributes:
https://github.com/jrwesolo/chef_search_woes

rlaveycal:
I'd like to add a voice for this. We've just had an issue where we deleted the wrong VM on AWS because the `knife ec2 server delete -N` command uses search to resolve the instance_id from the given node name. We had additional `name` attributes with values corresponding to valid nodes deep within our attribute data. Consequently the search returned multiple answers and the ec2 plugin chose the first one. The user didn't spot this.

This is a ticking time bomb and really needs fixing so that everything that uses search can start to be migrated to safe searches. I appreciate that the default will have to remain as-is for some considerable time.

kamaradclimber:
[...]

PR #331, opened 2015-06-03
Add ability to specify top level node attributes in Chef Search

jeremiahsnapp:
Suggested fix for https://github.com/chef/chef-server/issues/303

### DIFF
--- a/src/oc_erchef/apps/chef_objects/src/chef_node.erl
+++ b/src/oc_erchef/apps/chef_objects/src/chef_node.erl
@@ -148,6 +148,8 @@ ejson_for_indexing(#chef_node{name = Name, environment = Environment}, Node) ->
     DefaultNormalOverride = chef_deep_merge:merge(DefaultNormal, Override),
     {Merged} = chef_deep_merge:merge(DefaultNormalOverride, Automatic),
     RunList = ej:get({<<"run_list">>}, Node, []),
+    ExtractedRecipes = extract_recipes(RunList),
+    ExtractedRoles = extract_roles(RunList),
     %% We transform to a dict to ensure we override the top-level keys
     %% with the appropriate values and don't introduce any duplicate
     %% keys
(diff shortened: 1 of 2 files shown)


src/oc_erchef/apps/chef_objects/src/chef_node.erl
```diff
@@ -159,7 +159,12 @@ ejson_for_indexing(#chef_node{name = Name, environment = Environment}, Node) ->
                                    {<<"chef_environment">>, Environment},
                                    {<<"recipe">>, extract_recipes(RunList)},
                                    {<<"role">>, extract_roles(RunList)},
-                                   {<<"run_list">>, RunList}]),
+                                   {<<"run_list">>, RunList},
+                                   {<<"cheftoplevel_name">>, Name},
```
marcparadise: Maybe something like "#environment", #name" etc, to make it less typing to use?

src/oc_erchef/apps/chef_objects/src/chef_node.erl
```diff
@@ -159,7 +159,12 @@ ejson_for_indexing(#chef_node{name = Name, environment = Environment}, Node) ->
                                    {<<"chef_environment">>, Environment},
                                    {<<"recipe">>, extract_recipes(RunList)},
```
marcparadise: Let's extract roles/recipes once and assign, then use it in both places instead of extracting twice.

marcparadise:
Could you add some tests for this? I can help out with where/how if you're unfamiliar (and the search tests are ... challenging to navigate).

marcparadise:
Looks like some updates (and additions) to eunit will be needed, as this changed the expected values: 

https://travis-ci.org/chef/chef-server/jobs/65301220#L1292

src/oc_erchef/apps/chef_objects/src/chef_node.erl
jeremiahsnapp: Avoiding the longer name would be nice but I tested your suggestion and I got the following error.

```
knife search "#chef_environment:prod" -i
ERROR: knife search failed: invalid search query: '#chef_environment:prod'
```

The other reason for the longer name was to make it less likely that a node would have an identically named attribute.

jeremiahsnapp:
@mp Thanks a lot for the feedback. I don't know how to write tests at all let alone tests in an area that is challenging to navigate. I don't mind learning if you can devote time to teach me. If it's going to be too time consuming though it might be best to have you or someone else just add the coverage.

src/oc_erchef/apps/chef_objects/src/chef_node.erl
```diff
@@ -157,9 +159,16 @@ ejson_for_indexing(#chef_node{name = Name, environment = Environment}, Node) ->
...
-                                   {<<"run_list">>, RunList}]),
+                                   {<<"recipe">>, ExtractedRecipes},
+                                   {<<"role">>, ExtractedRoles},
+                                   {<<"run_list">>, RunList},
+                                   {<<"cheftoplevel-name">>, Name},
+                                   {<<"cheftoplevel-chef_environment">>, Environment},
+                                   {<<"cheftoplevel-recipe">>, ExtractedRecipes},
+                                   {<<"cheftoplevel-role">>, ExtractedRoles},
+                                   {<<"cheftoplevel-run_list">>, RunList},
+                                   {<<"cheftoplevel-recipes">>, ej:get({"recipes"}, Automatic)},
```
stevendanna: What will the result here be if "recipes" doesn't exist in the node json?  This is the sort of case that would be good to have a test around so that the behavior doesn't change in the future.

tylercloke:
@jeremiahsnapp If you aren't comfortable diving into the tests alone we should set up some time to pair with someone in the "pool" so that you'll know for the future! Good to spread the knowledge.

marcparadise:
@jeremiahsnapp I didn't see your earlier reply (ouch - in June '15), but if you're still interested in getting this in, I'd be happy to pair up with you  and work through getting some tests added.

stevendanna:
@jeremiahsnapp Apologies that I'm closing this after so long.  Thanks for taking the time to put this together, we do appreciate it even though we are ultimately closing this PR.  Feel free to ping me directly if you have any questions.

PR #336, opened 2015-06-04
Add ability to specify top level node attributes in Chef Search

jeremiahsnapp:
A more comprehensive fix for https://github.com/chef/chef-server/issues/303

This allows users to search any top level attribute they are interested in as opposed to only the Chef specific top level attributes.

This would be helpful to users but it will have a greater impact on the server than PR https://github.com/chef/chef-server/pull/331.

The flattened object will be double its usual size which will impact memory, network and disk usage. This might be enough reason to avoid implementing but I'll leave that up for discussion.

### DIFF
--- a/src/opscode-expander/lib/opscode/expander/flattener.rb
+++ b/src/opscode-expander/lib/opscode/expander/flattener.rb
@@ -29,6 +29,7 @@ module Opscode
 
         @item.each do |key, value|
           flatten_each([key.to_s], value)
+          flatten_each(['toplevel-' + key.to_s], value)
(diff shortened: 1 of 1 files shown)


stevendanna:
My inclination is :-1: as this will effectively double the amount of expanding work we need to do and the amount of indexing in solr.  I think #331 is a more pragmatic solution to the majority of the issues users hit with node expansion.

jeremiahsnapp:
Thanks @stevendanna. I agree with the :-1: . I'll close this.

PR #337, opened 2015-06-05
Policy List API

danielsdeleo:
Updates the API for `GET /policies` with a new document format. Parsed into ruby it looks like:

```ruby
        {
          "some_policy_name" => {
            "uri" => api_url("policies/some_policy_name"),
            "revisions" => {
              "909c26701e291510eacdc6c06d626b9fa5350d25" => {}
            }
          }
        }
```

This is the last piece I need to implement the basic `chef show-policy` command.

Note that I reorganized the pedant tests into one file to take advantage of shared fixture data and such. That's all in one commit so it should be easier to review commit-wise.

@chef/lob


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policies.erl
```diff
@@ -67,7 +67,35 @@ malformed_request_message(Any, _Req, _State) ->
...
+    case chef_db:list_all_policy_revisions_by_orgid(DbContext, OrgId) of
+        {error, Why} ->
+            Report = {list_all_policy_revisions_by_orgid, {Why, ReqId}},
+            lager:error("~p", [Report]),
+            error(Report);
+        AllRevisions ->
+            BaseEJSON = build_base_policy_list_ejson(Req, AllRevisions),
+            io:format("BASE EJSON:~p~n", [BaseEJSON]),
+            PolicyEJSON = build_policy_list_ejson(AllRevisions, BaseEJSON),
+            io:format("ALL REVS:~p~n", [PolicyEJSON]),
```
stevendanna: Likely want to use lager here or remove these altogether.

stevendanna:
The code here looks good to me.  I would like if one of the commit messages was more explicit about what the fold format was and how the new format is different, but this is still an experimental feature so I guess that isn't too important.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policies.erl
danielsdeleo: derp, I need to put TODOs next to these since I grep for that before making pull requests.

danielsdeleo:
@stevendanna done

stevendanna:
awesome, thanks :+1:

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policies.erl
```diff
@@ -67,7 +67,33 @@ malformed_request_message(Any, _Req, _State) ->
...
+
+build_nested_base_ejson([{PolicyName, URI}|Rest], EJSON) ->
+    NewEJSON = ej:set_p({PolicyName, "uri"}, EJSON, URI),
+    build_nested_base_ejson(Rest, NewEJSON);
+build_nested_base_ejson([], EJSON) ->
+    EJSON.
+
+build_policy_list_ejson([{PolicyName, RevisionID}|Rest], EJSON) ->
+    NewEJSON = ej:set_p({PolicyName, "revisions", RevisionID}, EJSON, {[]}),
+    build_policy_list_ejson(Rest, NewEJSON);
```
marcparadise: I think I commented on this in an earlier iteration, but won't this only ever capture one list entry? Should we get the current list if it exists and append to it, then use that as the value?

marcparadise:
Looks good outside of the one question I posted.   It might also be good to add some test coverage for multiple revisions.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policies.erl
danielsdeleo: From the pedant tests, this returns data like:

```ruby
        {
          "some_policy_name" => {
            "uri" => api_url("policies/some_policy_name"),
            "revisions" => {
              "909c26701e291510eacdc6c06d626b9fa5350d25" => {}
            }
          },
          "appserver" => {
            "uri" => api_url("policies/appserver"),
            "revisions" => {
              "1111111111111111111111111111111111111111" => {},
              "2222222222222222222222222222222222222222" => {},
              "3333333333333333333333333333333333333333" => {}
            }
          },
          "db" => {
            "uri" => api_url("policies/db"),
            "revisions" => {
              "6666666666666666666666666666666666666666" => {},
              "7777777777777777777777777777777777777777" => {},
              "8888888888888888888888888888888888888888" => {}
            }
          },
          "cache" => {
            "uri" => api_url("policies/cache"),
            "revisions" => {
              "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" => {},
              "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" => {}
            }
          }
        }
```

So each policy is a JSON object with a "revisions" key that is a JSON object, where each revision is a key with a value of an empty JSON object. The reason for the empty JSON object is because I'll probably add an `uri` field when there's a named policy revision endpoint (I could also maybe add other things without breaking compatibility, too).

marcparadise: Okay, cool - thanks!

PR #338, opened 2015-06-05
Notify user during dhparam creation

stevendanna:
Creating a 2048 bit dhparam can take a long time depending on the
state of the machine. This might confuse the user and cause them to
interrupt the process.


stevendanna:
cc @chef/lob This catches me off guard every time and I added the feature, so I imagine it will confuse others as well.

PR #339, opened 2015-06-06
[WIP] custom caching

marcparadise:
Not for merge yet.


scripts/build.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+case $1 in
```
stevendanna: I'm assuming that the main advantage of using a build script like this is doing this conditional environment setup which is harder to do in the makefiles?

marcparadise: Yes I was trying to isolate the environment stuff. The makefiles also don't support it well, in particular capturing and applying bash 'eval' for perl and lua.  Once the caching is behaving I will probably come back and revisit this part again.

marcparadise:
Closing this so that I can convince travis to starting building again, it's ignoring this PR.

PR #340, opened 2015-06-07
allow full path to the package to be specified

### LINKED ISSUE #178: allow full package path when installing add-ons
From @alexpop in chef/opscode-omnibus#746:

> I was following the Chef server [upgrade documentation](https://docs.chef.io/upgrade_server.html). Once upgraded, I moved on and followed the steps to install the latest add-ons(manage, reporting, etc):

> ```
Then run the install command using the --path option to specify the location for the package:
$ chef-server-ctl install NAME_OF_PACKAGE --path /path/to/package
```

> After downloading the required rpm's and reading the above note, I was confident this would:

> ```
chef-server-ctl install opscode-manage --path /media/chef/opscode-manage-1.11.0-1.el5.x86_64.rpm
```

> The command failed with:
```
[2015-04-08T13:01:31+01:00] FATAL: Chef::Exceptions::Package: yum_package[opscode-manage] (private-chef::add_ons_local line 35) had an error: Chef::Exceptions::Package: Package opscode-manage not found:
```

> It looks like the current implementation in the 'private-chef::add_ons_local' recipe is to expect the parent folder for the --path option.

> I opened this issue to allow both the parent folder and the full package path to be specified for the --path option. This would be a more standard install method. I asked a few folks in order to validate this and they all felt for the full package path as I did.

alexpop:
Fixes #178

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_local.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_local.rb
@@ -17,8 +17,13 @@ end
 
 addon_path = node['private_chef']['addons']['path']
 node['private_chef']['addons']['packages'].each do |pkg|
-  # find the newest package of each name
-  pkg_file = Dir["#{addon_path}/#{pkg}*.#{package_suffix}"].sort_by{ |f| File.mtime(f) }.last
+  if ::File.directory?(addon_path)
+    # find the newest package of each name if 'addon_path' is a directory
+    pkg_file = Dir["#{addon_path}/#{pkg}*.#{package_suffix}"].sort_by{ |f| File.mtime(f) }.last
+  else
+    # use the full path to the package
+    pkg_file = addon_path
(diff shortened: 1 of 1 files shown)


stevendanna:
@sdelano Is this something we want to pull into 12.1?  Looks minor and I know it confuses people.

mmzyk:
I'm :+1: We should merge this for 12.1, as this does confuse a lot of people. We'll need to ensure the docs are updated as well to reflect the full path can be used.

mmzyk:
Whoops, I forgot 12.1 was already in RC. This should go with 12.2.

stevendanna:
@alexpop Mind rebasing this against master?  Once rebased I will merge this (or you can if you want :))

Note for the future: You should have write access to this repository so you can also create branches in this repository which will allow whoever decides to merge your PR to do the rebasing themselves if they are in a friendly mode. :)

alexpop:
Sure, thanks

PR #342, opened 2015-06-07
custom cache 2 [wip]

marcparadise:
Trying a new branch to get travis to pick up the build...

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+export COMPONENT_NAME=`basename $1`
+
+case $COMPONENT_NAME in
+  oc_erchef)
+    echo  "Setting env for oc_erchef"
+    # TODO - can we just define this as $COMPONENT/travis-env.sh, and source it?
+    # Perl for local sqitch
+    export PERL5LIB=~/perl5/lib/perl5/x86_64-linux-gnu-thread-multi:~/perl5/lib/perl5:/etc/perl:/usr/local/lib/perl/5.14.2:/usr/local/share/perl/5.14.2:/usr/lib/perl5:/usr/share/perl5:/usr/lib/perl/5.14:/usr/share/perl/5.14:/usr/local/lib/site_perl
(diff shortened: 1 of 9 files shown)


src/oc-id/Makefile
```diff
@@ -1,5 +1,9 @@
+# TODO - looks like we can get built in support for rvm
```
stevendanna: I looked into this but couldn't figure out how.  Their built-in ruby installer only works if you set the `language` to ruby.  If we set the language to ruby then the build-in erlang installer doesn't work.  It looked easier at the time to manage ruby on our own than erlang.

scripts/build.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+export COMPONENT_NAME=`basename $1`
+
+case $COMPONENT_NAME in
+  oc_erchef)
+    # TODO - can we just define this as $COMPONENT/travis-env.sh, and source it?
```
stevendanna: Don't see why that wouldn't work; however, this feels pretty clean as well.

scripts/build.sh
```diff
@@ -0,0 +1,27 @@
...
+    # Setup for gecode build
+    export USE_SYSTEM_GECODE=1
+    export LIBRARY_PATH=$TRAVIS_BUILD_DIR/src/oc_erchef/vendored-gecode/lib
+    export LD_LIBRARY_PATH=$TRAVIS_BUILD_DIR/src/oc_erchef/vendored-gecode/lib
+    export CPLUS_INCLUDE_PATH=$TRAVIS_BUILD_DIR/src/oc_erchef/vendored-gecode/include
+    ;;
+  omnibus)
+    export USE_OMNIBUS_FILES=0
+    export LUALIB=~/.luarocks/lib/lua/5.2
+	  eval $(luarocks path)
```
stevendanna: Weird spacing here.

stevendanna:
@marcparadise I'm gonna close this out.  I think our travis tests aren't too slow to merit the increased complexity here at the moment.  We can keep the branch around and revisit it if travis test speed becomes an issue.

PR #347, opened 2015-06-12
Enabling external rabbitmq for analytics

PrajaktaPurohit:
/cc @sdelano @gregdurham @jamesc

tested with build: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/585/downstreambuildview/

and all looks good.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -117,6 +117,17 @@ default['private_chef']['rabbitmq']['consumer_id'] = 'hotsauce'
 default['private_chef']['rabbitmq']['env_path'] = "/opt/opscode/bin:/opt/opscode/embedded/bin:/usr/bin:/bin"
 
 ####
+# External RabbitMQ
+####
+default['private_chef']['external-rabbitmq']['enable'] = false
+default['private_chef']['external-rabbitmq']['actions_user'] = 'actions'
+default['private_chef']['external-rabbitmq']['actions_password'] = 'changeme'
+default['private_chef']['external-rabbitmq']['actions_vhost'] = '/analytics'
+default['private_chef']['external-rabbitmq']['actions_exchange'] = 'actions'
+default['private_chef']['external-rabbitmq']['node_port'] = '5672'
+default['private_chef']['external-rabbitmq']['vip'] = '127.0.0.1'
+
+####
 # Jetty dummy for logs
 ####
 # Should always be enable = false, we control Jetty+Solr through opscode-solr4
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
@@ -14,6 +14,15 @@ class OmnibusHelper
     {"owner" => owner, "group" => group}
   end
 
+  def rabbitmq_configuration
+    external = node['private_chef']['external-rabbitmq']['enable']
+    if external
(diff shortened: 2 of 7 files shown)


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -117,6 +117,17 @@
 default['private_chef']['rabbitmq']['env_path'] = "/opt/opscode/bin:/opt/opscode/embedded/bin:/usr/bin:/bin"
 
 ####
+# External RabbitMQ
+####
+default['private_chef']['external-rabbitmq']['enable'] = false
```
stevendanna: This is similar to the pattern that has been proposed for the external postgresql, but I'm still feeling like it might be unnecessary to have the user both set down all the config for the external rabbit and flip this switch.  Since we control the the parsing of the configuration file, we should be able to distinguish between these two situations:

1) You've specified a rabbit configuration in chef-server.rb,
2) You have not specified a rabbit configuration in chef-server.rb

which is all we need to know to determine whether to set the action's rabbit config to the local rabbit or whatever the user provided.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -117,6 +117,17 @@
...
+default['private_chef']['external-rabbitmq']['actions_user'] = 'actions'
+default['private_chef']['external-rabbitmq']['actions_password'] = 'changeme'
+default['private_chef']['external-rabbitmq']['actions_vhost'] = '/analytics'
```
stevendanna: Does this fail if this vhost doesn't exist on the external rabbitmq?  If so, what does that failure look like?

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -14,6 +14,15 @@ def ownership
     {"owner" => owner, "group" => group}
   end
 
+  def external_rabbitmq
```
stevendanna: Maybe just name this rabbitmq_configuration since the user doesn't care whether it is external or not?

PrajaktaPurohit: sure

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
PrajaktaPurohit: That is a good point - although I think it might be easier for the users to flip their external rabbitmq on and of with a single config variable rather than having to remove/rewrite all variables into the chef-server.rb

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
PrajaktaPurohit: it does not cause erchef requests to fail, but it fails to send messages to analytics. And that can cause the erchef logs to fillup and / or slow response times.

We will provide the users with a script to create the users, vhosts and grant appropriate permissions - although this might not be the best solution.

stevendanna:
@PrajaktaPurohit and I chatted earlier about my comments.  I think this can go through without them being addressed immediately. :+1: here.

PrajaktaPurohit:
thanks @stevendanna!

jamesc:
+1

sdelano:
:+1: 

* [ ] needs a rebase before merging
* [ ] could use a note in the changelog about preliminary support for external actions queue

sdelano:
:shipit:

PR #348, opened 2015-06-16
Upgrade to redis 2.8.21 for a critical security fix

stevendanna:


omnibus/CHANGELOG.md
```diff
@@ -94,6 +94,12 @@ contains the following components:
 
 ### opscode-omnibus
 
+### redis 8.2.21
```
irvingpop: I think that should be 2.8.21 right? :)

stevendanna: Gracias

sdelano:
:+1: The `omnibus-software` [diff](https://github.com/chef/omnibus-software/compare/0fb66f82eef7b52d866a9fbc153fc4b1d686e8c9...8b0ebc7691d854f9b287107c4b3afca9caf4b9b4) also shows that we're not pulling in new versions of other components. LGTM

PR #350, opened 2015-06-17
Require authentication for redis.

stevendanna:
In standalone mode, redis only listens on 127.0.0.1, meaning that only
local users can connect to the service. However, in HA or Tiered
configurations, redis listens on the backend VIP. In environments
where the VIP may be accessible by untrusted users, the lack of
authentication means that untrusted users have access to redis.

Our previous recommendation was that users should use iptables or
other firewalling to ensure that only trusted users could access the
VIP.

NB: Redis still sends and recieves data in the clear.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/config.lua.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/config.lua.erb
@@ -46,7 +46,6 @@ end
 local function connect_redis()
    local red = redis:new()
    red:set_timeout(<%=@redis_connection_timeout%>)
-   red:set_timeout(1000)
    local ok, err = red:connect("<%= @omnihelper.vip_for_uri('redis_lb') %>", <%=@redis_port%>)
 
    if not ok then
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/config.lua.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/config.lua.erb
@@ -36,15 +36,33 @@ local function array_to_hash(t)
   return h
 end
 
+local function auth_not_required(err)
+  -- conservatively try to determine if auth is not required.
+  -- if auth isn't required, we can safely ignore an error from
+  -- the AUTH command
+  return err == "ERR Client sent AUTH, but no password is set"
+end
+
 local function connect_redis()
-  local red = redis:new()
-  local results
-  red:set_timeout(<%=@redis_connection_timeout%>)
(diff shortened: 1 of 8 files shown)


stevendanna:
cc @chef/lob @manderson26 

This adds authentication to our redis setup, but I'm unsure whether we should merge it.  Namely, I'm concerned about giving a false sense of security since all of the redis traffic happens in the clear.  Redis doesn't natively support TLS and as a result I wouldn't be surprised if the redis clients we use (https://github.com/redis/redis-rb and https://github.com/openresty/lua-resty-redis) don't support TLS either.  To encrypt TLS, I'm guessing we are going to want to do something like:

1) Redis client connects to a TLS stream proxy on localhost
2) TLS steam proxy connects to TLS stream proxy on backend VIP
3) TLS steam proxy connects to Redis on the backend

Nginx 1.9 looks like it can proxy non-HTTP TLS streams (http://nginx.com/blog/tcp-load-balancing-with-nginx-1-9-0-and-nginx-plus-r6/).  Another alternative I've been looking at is spiped (http://www.tarsnap.com/spiped.html).  The consensus in previous discussions seems to be that we'd like to avoid the configuration complexity of stunnel if possible.

stevendanna:
> This adds authentication to our redis setup, but I'm unsure whether we should merge it. Namely, I'm concerned about giving a false sense of security since all of the redis traffic happens in the clear

I think that overall this moves us in the right direction since it lowers who can access redis from "anyone who can get packets routed to the VIP" to "anyone who can listen in on the VIPs traffic from the frontends", which is likely a substantially smaller subset in most environments.

joedevivo:
I have a padlock on the gate that leads to my backyard. If somebody wants to get into my backyard, that padlock is doing very little. If someone want to get in any backyard, they'll probably choose another.

:+1:, although I have to admit I don't know any lua.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/config.lua.erb
```diff
@@ -36,15 +36,33 @@ local function array_to_hash(t)
...
-  local results
-  red:set_timeout(<%=@redis_connection_timeout%>)
-  local ok, err = red:connect("<%= @omnihelper.vip_for_uri('redis_lb') %>", <%=@redis_port%>)
-  if not ok then
-    ngx.log(ngx.ERR, "failed to connect redis: ", err)
-  end
-  return ok, red
+   local red = redis:new()
+   red:set_timeout(<%=@redis_connection_timeout%>)
+   red:set_timeout(1000)
```
marcparadise: I think this probably wasn't meant to be left here?

marcparadise:
Other than the extra `set_timeout` this looks good to me.

joedevivo:
@marcparadise removed the timeout

marcparadise:
:+1: merge away!

PR #352, opened 2015-06-18
Enabled bounded queuing for erchef and bifrost squerl-based queues

stevendanna:
Pooler supported bounded queuing of requests when no connections are
available. Without queuing, many users have (at our suggestion) taken
to increasing the db_pool_size to prevent bursts of requests failing
when all database connections are consumed. In some cases, this can
lead to an level of concurrent connections that the backend can't
handle.

The queuing allows the maximum concurrent database connections to
remain capped at a reasonable level while still providing a means for
users to handle bursts of traffic.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -203,10 +203,11 @@ default['private_chef']['opscode-erchef']['udp_socket_pool_size'] = '20'
 # db_pooler_timeout - the maximum amount of time a request should wait
 # in the queue before timing out.  Request queueing is only effective
 # if db_pooler_timeout > 0
-default['private_chef']['opscode-erchef']['db_pool_size'] = '20'
-default['private_chef']['opscode-erchef']['db_pooler_timeout'] = '0'
+default['private_chef']['opscode-erchef']['db_pool_size'] = 20
+default['private_chef']['opscode-erchef']['db_pool_queue_max'] = 20
+default['private_chef']['opscode-erchef']['db_pooler_timeout'] = 2000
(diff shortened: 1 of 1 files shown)


mmzyk:
Just for my full understanding, the way bounded queuing is achieved is by setting a timeout so that if a request doesn't complete in that time it's dropped and the next request is serviced - thereby not allowing the queue to backup (as much) since no single request will block anything for long. At least, that's how I'm reading the code. 

I'm a bit confused b/c I expected to see a queue size limit applied whereby requests were dropped at the limit. I think what has been done is almost functionally the same and I'm fine with this, just want to make sure I fully understand, since what I expected and what I think I see aren't the same.

stevendanna:
The queue works this way:

1) If a request is made to the pool and the pool has available connections, it is serviced immediately,

2) If a request is made to the pool and the pool has no available connections but has room in the queue (queue is smaller than pool_queue_max), the request is added to the queue.  When a connection is returned to the pool, we immediately check the queue and service the request at the head of the queue.

3) If a request is made to the pool and the pool has no available connections and the queue is full, the request fails immediately.

4) If a queued request is in the pool for too long (user specified number of miliseconds), it is removed from the queue and an error is returned to the caller.

I hope that helps.

mmzyk:
Makes sense. That's what I expected, but I wasn't piecing that together just looking at the options.

I'm +1, although I have no opinion on the defaults that have been chosen here.

stevendanna:
@mmzyk Given the variety of environments we deploy into, the current default was simply my best guess at settings that would (1) make sure it is on by default, (2) not seem absurd in the average environment.  No real science behind it at this point.

cc @chef/lob

joedevivo:
:+1: The defaults here seem better than the current settings. We can always change them in the future if we find something better.

marcparadise:
:+1:

PR #355, opened 2015-06-19
Bump omnibus-software for the latest libxml2

stevendanna:


joedevivo:
:+1: SHAs work for me

PR #370, opened 2015-06-29
Update nginx.rb to use openssl cookbook

charlesjohnson:
- Fix Rubocop violations in recipes/nginx.rb
- Add more comments to recipes/nginx.rb
- Add .gitignore to private-chef cookbook
- Update recipes/nginx.rb to use LWRPs from openssl cookbook for dhparam.pem generation
- Update recipes/nginx.rb to use LWRPs from openssl cookbook for x509 cert generation
- Fix recipes/nginx.rb notification bug
- Change recipes/nginx.rb to use cookbook_file for nginx addons-dir README
- Update nginx addons-dir README.md
- Update private-chef cookbook Berksfile with new source syntax
- Add dependency on openssl cookbook


sdelano:
LGTM. Do we run any rubocop / cookbook tests via travis for this? That would be an amazing addition after putting in all the work here on that front

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -371,12 +371,13 @@
 default['private_chef']['nginx']['ssl_certificate'] = nil
 default['private_chef']['nginx']['ssl_certificate_key'] = nil
 default['private_chef']['nginx']['ssl_dhparam'] = nil
-default['private_chef']['nginx']['ssl_country_name'] = "US"
-default['private_chef']['nginx']['ssl_state_name'] = "WA"
```
stevendanna: I doubt many users were setting these in their configuration, but what will these default to now?

charlesjohnson: The cert only requires 4 fields:
common_name (CN)
Org (O)
Org_unit (OU)
Country (C)

The additional fields can be nil without problem. Also the LWRP doesn't support them, so they became unreferenced by the changeover.

Also also since anything this recipe generates is going to be a self-signed cert, having the additional fields won't make things any better.

PR #372, opened 2015-06-30
Bifrost & Mover build on Olde CI

joedevivo:


stevendanna:
:+1:

marcparadise:
The change looks good, but also take the lines from erchef's build.sh to upload packages for both ubuntu-10.04 and ubuntu-12.04, instead of using the source machine type. You can then delete lines 10,11,12

marcparadise:
Extra note: use caution - when I copy-pasta'd that from erchef to mover, I got tripped up because the deploy of mover used underscores in the file name, while I had used hyphens.  I changed the cookbook in that case, because +1 to consistency...

sdelano:
It seems odd to me that we have to hard-code the version number into this. Is there something we can do consistently across these projects to dynamically generate the release version number?

marcparadise:
Not the version number, just 'ubuntu-10.04' and 'ubuntu-12.04'

sdelano:
:+1:

PR #373, opened 2015-06-30
Fix bug where --expiration-date option was not properly being set in csc key commands.

tylercloke:
Ping @chef/lob 

TODO:
- [x]  Changelog.

[Passing Build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/596/downstreambuildview/)


stevendanna:
:+1: looks reasonable

marcparadise:
:+1:

Issue #154, opened 2015-04-27
packagecloud yum repo metadata can break Chef Server if auto-update is enabled


jeremiahsnapp:
When `chef-server-ctl install` is used to install Chef Server add-on packages it first configures packagecloud yum/apt repository metadata. If the system's package manager is configured to auto-update then it will download and install new packages including Chef Server and add-on packages. Just installing new Chef packages does not complete a proper upgrade since other `ctl reconfigure` commands must be run. This can leave the Chef Server in a broken state if any of the Chef services get restarted. 

The Chef PackageCloud repository metadata should be configured but disabled with respect to the system's package manager auto-updater to prevent this from happening in the first place.

dreamnite:
Suggest adding to the end of the post install script for chef-server-core:
if [ $1 -ge 2 ]; then
 /usr/bin/chef-server-ctl reconfigure

or similar for the addons: 
if [$1 -ge 2]; then 
/usr/bin/$ADDON-ctl reconfigure

$1 in the post install script can be used to determine if the package has been installed or upgraded, see: 
http://docs.fedoraproject.org/en-US/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch09s04s05.html

This should preserve the ability of yum to update the package without causing issues to chef because of a lack of "reconfigure" recipes being run.

jeremiahsnapp:
Hello @dreamnite,

The actual command would need to be `chef-server-ctl upgrade` for chef-server-core but I don't think this is a reliable solution anyway.

Chef Server Tier and HA topologies require more orchestrated steps during the upgrade process. The primary backend needs to be upgraded first and then appropriate files need to be transferred to the other servers in the Chef Server cluster before running the upgrade command on those servers.

I think it is too risky to have a Chef Server (especially a cluster of Chef Servers) upgrade automatically in the first place. The Chef Server is typically a critical part of the infrastructure and should be treated more carefully with scheduled upgrades.

I believe it would be safer to just disable the chef-stable repository metadata with `enabled=0`. When the user is ready to upgrade they just use `yum install chef-server-core --enablerepo=chef-stable`. The `chef-server-install` command can also be configured to use that option when installing other add-ons.

jeremiahsnapp:
I changed the title of this issue because I believe this problem only affects system's using yum.

The default contents of "/etc/apt/apt.conf.d/50unattended-upgrades" on an Ubuntu system would prevent chef-stable packages from being auto upgraded.

```
[...]

PR #374, opened 2015-07-01
Avoid Unintended Updates for RHEL

sean-horn:
ChangeLog-Entry: Several customers on RHEL installs have had unintended Chef Server updates
    with the current enabled=1 setting for the chef-stable repo. We should disable
    the repo by default so RHEL OS updates don't pull in the chef-stable contents as well.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_remote.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_remote.rb
@@ -5,6 +5,10 @@
 # All Rights Reserved
 #
 
+# Only "enabled false" on yum repos, because
+# the default contents of "/etc/apt/apt.conf.d/50unattended-upgrades" 
+# on an Ubuntu system would prevent chef-stable packages from being auto upgraded.
+
 case node['platform_family']
 when 'debian'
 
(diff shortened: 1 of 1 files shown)


sean-horn:
@doubt72 Hmm, maybe this is not complete? What happens during the same run where we have said something like

    chef-server-ctl install opscode-manage

If the repo is disabled, then that should not work at all.

sean-horn:
Ok, updated to override chef-stable repo enable for RHEL as part of `chef-server-ctl install`

stevendanna:
This looks reasonable to me.  What is the situation on apt-based systems, do we need to make any changes there so that the behavior is consistent?

stevendanna:
@jeremiahsnapp answered the ubuntu question here:

https://github.com/chef/chef-server/issues/154#issuecomment-99202090

:+1: from me then.

mmzyk:
We should add a comment to the code that says that Ubuntu doesn't need this guard because of what @jeremiahsnapp said in #154, or else I promise at some point a developer is going to be scratching their head over why only RHEL has an exception.

sean-horn:
@mmzyk Cool, I've added a comment to make things more clear.

mmzyk:
Looks good to merge then.

sean-horn:
@chef/lob Feel free to merge for any upcoming release.

PR #376, opened 2015-07-02
Rename ORGNAME_global_admins groups to ORGNAME_read_access_group

stevendanna:
The ORGNAME_global_admins group has one purpose inside the Chef
Server. To provide READ access to users who join an organization.  Its
name, however, implies great power. This change renames the group to
something more appropriate by:

- changing the organization creation code to create a
  ORGNAME_read_access_group rather than a global_admins group,

- providing a migration for existing global_admins groups,

- changing oc_chef_authz to transparently handle either group


joedevivo:
+1, all the great things I +1'd about #251 without the pesky merge conflicts!

PR #377, opened 2015-07-02
build.sh(s) now grep relx.config for version info

joedevivo:


stevendanna:
Can you also bump the omnibus project config version while we are at it?  It is in omnibus/config/projects/chef-server.rb

joedevivo:
done

marcparadise:
:+1:

stevendanna:
:ship:

PR #378, opened 2015-07-02
dvm omnibus load option

marcparadise:
Make loading of omnibus components before reconfigure an option. …

- adds new customizable vm.omnibus-auto-load configuration option,
  allowing you to specify the omnibus components you want to have loaded
  before the first reconfigure.
- add simple merge of defaults.yml and config.yml.
- remove vm setting natdnshostresolver1, which causes problems on some
  macbook configurations
- add config.yml to gitignore now that it's more-or-less stable, to make local additions
   less messy


dev/config.yml
```diff
@@ -1,12 +1,23 @@
-# Override support for default config entries in defaults.yml.
```
sdelano: Did you mean for this file to be included? It's in the .gitignore

marcparadise: Yeah, I set up the basic version then separately added it to ignore to lock in place. Alternatively I could use config.yml.example.. .

marcparadise: I'll go with .example - it's a fairly standard means of doing things.

PR #379, opened 2015-07-02
ChangeLog-Entry: [omnibus] Make the chef postgres database be owned by the chef postgres user

marcparadise:
Just like it says on the wrapping.


marcparadise:
Note that I'll change the base branch once the parent branch lands.

PR #380, opened 2015-07-02
ChangeLog-Entry: [omnibus] Make the chef postgres database be owned by the chef postgres user

marcparadise:
Just like it says on the package


stevendanna:
:+1: It would be nice if we could add a `not_if` to that ruby block, but my guess is that checking the owner is about as expensive as just unconditionally setting the owner, so the not_if would really be a cosmetic cleanup for the reconfigure output.

PR #382, opened 2015-07-03
Enable running Chefspec tests against private-chef cookbook

charlesjohnson:
- NEW: Add Chefspec & Berkshelf to Gemfile
- NEW: Add spec/spec_helper.rb for Chefspec
- NEW: Chefspec coverage for recipes/nginx.rb
- FIX: Remove unreferenced servers method from libraries/private_chef.rb (Thanks Mark Paradise)
- FIX: (Minor) clean a few comments in recipes/nginx.rb


charlesjohnson:
Notes:
- Passes CI
- Adding Berkshelf to the Gemset may be controversial?
- Servers method was blowing up in Chefspec.
- servers method is unnecessary anyway because it conflicted with the servers attribute in the same module, and the attribute always won.

marcparadise:
:+1:

charlesjohnson:
Need berks there so that we can include it during development. If there's a way to have a dev-specific gemset then I'm down to use that.

stevendanna:
I'm not seeing the chefspec coverager for recipes/nginx.rb that is mentioned in the commit message?  Did that get lost?

charlesjohnson:
D'oh! I'll get it into another commit. Thanks.

PR #386, opened 2015-07-06
WIP: Include missing dependencies in mover build

joedevivo:
I think we're including too many dependencies now, which is better than
not enough, but armed with a good test environment, we should be able to
pull out the dependencies we don't actually need.


stevendanna:
:+1: @chef/lob I think we should consider also forking a 12.1.1 release with just this fix and shipping it ASAP so that people can upgrade to the latest release.

joedevivo:
I'm able to start this release locally now. It fails because it doesn't connect to a database and it can't start chef_reindex. But I'm really not starting it in an environment where I expect it to be able to do these things.

joedevivo:
Merged in #395 and #401. No need for this one.

PR #387, opened 2015-07-06
make external services use an 'external' recipe

marcparadise:
When a service is configured as 'external', this change causes the
service to be disabled locally and instead expects recipe
private_chef::#{service}-external to be available to handle remote
configuration.

ping @chef/lob 

http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/604/downstreambuildview/


marcparadise:
Going to combine this with mp/add-remote-postgres-configuration , no reason not to.

PR #389, opened 2015-07-06
add option for external service config recipe and create one for postgres

marcparadise:
When a service is configured as 'external', this change causes the
service to be disabled locally and instead expects recipe
private_chef::#{service}-external to be available to handle remote
configuration.

This PR also adds a postgresql-external recipe to configure postgres for external use. 

supercedes https://github.com/chef/chef-server/pull/387 to save merge-sequence complexity. 

ping @chef/lob


marcparadise:
CI: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/605/downstreambuildview/

PR #393, opened 2015-07-07
Catch common error cases in global_admins_user_addition migration

stevendanna:


stevendanna:
@chef/lob Running this in Hosted Chef preproduction environment revealed some cases of bad data leading to failed migrations.  Interested in thoughts about the best way to handle some of these cases.

In all but 1 failure, the data is such that the organization cannot be migrated as is.  The failures seen in preproduction were:

- Missing users group
- Missing ORGNAME_global_admins group
- Missing bifrost data for one of the groups
- Cycles in existing bifrost data leading to a 400 from bifrost
- Connection error to bifrost

The following table outlines the rate at which they were seen:

```
| Type                   | %       |
|------------------------+---------|
| no user group          | 2.950 % |
| missing bifrost data   | 0.077 % |
| no global_admins group | 0.035 % |
| cycles in bifrost data | 0.008 % |
| sel_conn_closed        | 0.002 % |
```

(Chef employees, ask me for exact numbers if you are interested).

As can be seen, the "no users group" is the most common case. Unfortunately, this case also isn't fixable since the removal of the users group might have been a policy decision by the user (even though this breaks various org functions).  Additionally, missing bifrost data isn't fixable without assuming the desired policy in the organization and thus allowing access to people who would currently be receiving errors.

Missing global_admins groups is possibly fixable automatically; however, given the low rate we saw this, I decided to lump it in with the others.  We could write a cleanup migration to create the missing global_admin groups that would run before the user_addition migration.

Some requests failed with 400s from bifrost complaining that adding the users group to the global_admins group would create a cycle.  In the cases I checked, the problem was that the bifrost data already had a cycle between the users and admins group (where the users group was added to the admins group and vice-versa).  I've left these as unhandled.

joedevivo:
:+1: code looks good. Does what it says. Somebody (@sdelano?) should weigh in on this being the thing we want to do.

sdelano:
This looks good to me. I think the question to answer is:

> If we leave any of the functionality unfixed, how will behavior differ for said organizations after the migration?

It seems like most of the orgs that aren't migrating are most likely broken already.

stevendanna:
> It seems like most of the orgs that aren't migrating are most likely broken already.

This is my take as well.  If members of those orgs contact us about this feature not working, it is likely that we'll want to do some more comprehensive repair of their data.

stevendanna:
> If we leave any of the functionality unfixed, how will behavior differ for said organizations after the migration?

To answer this directly.  Users of the unmigrated orgs won't get the policy change of being able to read keys between users in that org.  All existing functionality should continue to work and it should be possible to apply that policy change for individual orgs if they do reach out to support.

joedevivo:
There's a separate code path for unmigrated orgs. If we don't successfully migrate them all, do we have to leave the alternate code path in there forever?

stevendanna:
@jd We do not need to leave the path in forever. The alternate code path is for the rename migration (changing the name from ORGNAME_global_admins to ORGNAME_read_access_group).  In preproduction that migration was 100% successful.

The failures we saw were in the migration that attempts to add the user's group to the global_admins/read_access group.

PR #395, opened 2015-07-07
Include missing dependencies in mover build

stevendanna:
I think we're including too many dependencies now, which is better than
not enough, but armed with a good test environment, we should be able to
pull out the dependencies we don't actually need.


stevendanna:
@chef/lob this is the propsed fix for the mover build applied to 12.1.0 with the version bumped to 12.1.1

joedevivo:
:+1:  WE DID IT!

PR #397, opened 2015-07-07
Fix option name so that --background (and -b) work

jcreedcmu:
@marcparadise: is this correct? The `--background` option didn't work for me until I applied this diff.

### DIFF
--- a/dev/dvm/lib/dvm/application.rb
+++ b/dev/dvm/lib/dvm/application.rb
@@ -124,8 +124,8 @@ module DVM
       @projects[project_name].run args
     end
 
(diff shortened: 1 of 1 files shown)


mmzyk:
Looks good. Appears to have been a basic typo (although -b should have still worked)

jcreedcmu:
On the contrary, prior to this, `-b` made `options[:foreground]` true, so that when https://github.com/chef/chef-server/blob/jc/option-name/dev/dvm/lib/dvm/application.rb#L134 looked up  `options[:background]` it still found nothing.

mmzyk:
Ah, makes sense.

marcparadise:
I didn't see this earlier , but yes, good find. We realized we were always specifying foreground and so I changed the default behavior but forgot to update the option.

PR #405, opened 2015-07-08
add remote connection support to postgres helpers

marcparadise:
- updated ec_database and pg_connection to use remote connections when 'external' is set. 
- adds db_superuser and db_superuser_password postgres attributes to avoid having to revisit this code once those updates land 
-


omnibus/files/private-chef-cookbooks/private-chef/providers/pg_database.rb
```diff
@@ -10,33 +10,20 @@ def whyrun_supported?
 use_inline_resources
 
 action :create do
-  execute "create_database_#{new_resource.database}" do
-    command createdb_command
-    user node['private_chef']['postgresql']['username']
-    not_if {database_exist?}
-    retries 30
+  ruby_block "Maybe create database #{new_resource.database}" do
```
stevendanna: Since this is already provider code, I don't see why we need to put this in a ruby_block.

omnibus/files/private-chef-cookbooks/private-chef/providers/pg_database.rb
```diff
@@ -10,33 +10,20 @@ def whyrun_supported?
...
+    block do
+      EcPostgres.with_connection(node) do |connection|
+        result = connection.exec("SELECT datname FROM pg_database WHERE datname='#{new_resource.database}'")
+        if result.ntuples == 0
+          converge_by("Create database #{new_resource.database}") do
+              owner = "WITH OWNER #{new_resource.owner}" if new_resource.owner
+              connection.exec("CREATE DATABASE \"#{new_resource.database}\" #{owner} TEMPLATE #{new_resource.template} ENCODING '#{new_resource.encoding}';")
+              new_resource.updated_by_last_action(true)
```
stevendanna: This updated_by_last_action shouldn't be needed here.  Chef uses the fact that you called a converge_by block to determine that the resource has been updated.

omnibus/files/private-chef-cookbooks/private-chef/providers/pg_database.rb
```diff
@@ -10,33 +10,20 @@ def whyrun_supported?
...
+          end
+        else
+          new_resource.updated_by_last_action(false)
```
stevendanna: Similarly, I think this can be left off.  That would reduce this section of code to:

```ruby
action :create do
  EcPostgres.with_connection(node) do |connection|
    result = connection.exec("SELECT datname FROM pg_database WHERE datname='#{new_resource.database}'")
    if result.ntuples == 0
      converge_by("Create database #{new_resource.database}") do
        owner = "WITH OWNER #{new_resource.owner}" if new_resource.owner
        connection.exec("CREATE DATABASE \"#{new_resource.database}\" #{owner} TEMPLATE #{new_resource.template} ENCODING '#{new_resource.encoding}';")
      end
    end
  end
end
```

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
```diff
@@ -1,14 +1,29 @@
...
-      begin
-        yield connection
-      ensure
-        connection.close
-      end
-    end
+     require 'pg'
+     postgres = node['private_chef']['postgresql']
+     if postgres['external']
+         connection = ::PGconn.open('user' => postgres['db_superuser'],
```
stevendanna: Minor style: We like to use 2 space indents for ruby code.

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
```diff
@@ -1,14 +1,29 @@
...
+       # dedicated postgresql user over pipe.
+       as_user(postgres['username']) do
+         connection = ::PGconn.open('dbname' => database)
+         begin
+           yield connection
+         ensure
+           connection.close
+         end
+       end
+     end
```
stevendanna: Will something like this work to reduce the duplication here? Warn: Untested code written in comment box:

```ruby
def self.with_connection(node, database = 'template1')
  require 'pg'
  postgres = node['private_chef']['postgresql']
  connection = if postgres['external']
                 ::PGconn.open('user' => postgres['db_superuser'],
                               'password' => postgres['db_superuser_password'],
                               'host' => postgres['vip'], 
                               'port' => postgres['port'],
                               'dbname' => database)
               else
                 as_user(postgres['username']) do
                   ::PGconn.open('dbname' => database)
                 end
               end
  begin
    yield connection
  ensure
    connection.close
  end
end
```

omnibus/files/private-chef-cookbooks/private-chef/providers/pg_database.rb
marcparadise: When I did the above without an enclosing ruby_block,  notification did not occur when we created the database which prevent erchef schema deploy from running.  

Explicitly setting updated_by_last_action was my first attempt to get that to behave. I do need to dig into this more, as the ruby_block and updated_by_last_action calls shouldn't be required.

omnibus/files/private-chef-cookbooks/private-chef/providers/pg_database.rb
marcparadise: See above as to why I ended up with a ruby block in there. Hoping to get rid of it.

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
marcparadise: Yep

marcparadise: That is to say - maybe. I'm not sure if moving the usage of connection outside of the as_user block will behave well,but I'll find out.

marcparadise: Because of that user block, that did not work for a local installation (I thought I remembered that from a previous attempt, but double-checked)

stevendanna: @marcparadise Cool, I figured there might be an issue with the `as_user` thing.  thanks for checking!

PR #406, opened 2015-07-08
Adding db superuser to pg_hba.conf

andy-dufour:


marcparadise:
:+1:

omnibus/files/private-chef-cookbooks/private-chef/templates/default/pg_hba.conf.erb
```diff
@@ -64,6 +64,8 @@
 # "local" is for Unix domain socket connections only
 local   all         <%= node['private_chef']['postgresql']['username'] %>                               peer
 
+host    all         <%= node['private_chef']['postgresql']['db_superuser'] %>         127.0.0.1/32           md5
```
stevendanna: Are we able to specify localhost generically vs 127.0.0.1/32 here?  In particular, I'd like to avoid any IPv4 vs IPv6 problems down the road.

andy-dufour: It looks like we could use samehost, or in theory localhost should work.

pg docs --
You can also write all to match any IP address, samehost to match any of the server's own IP addresses, or samenet to match any address in any subnet that the server is directly connected to.

marcparadise:
Let's go with localhost if it works, or samehost. While this should be safe as-is (our ipv6 support requires ipv4 loopback) there's no reason to ask for trouble.

PR #407, opened 2015-07-08
Add PDF docs on the Chef server ACL system

mmzyk:
These are good enough now that we can open this up to help everyone out.


marcparadise:
:+1: this looks great -  MWR

PR #409, opened 2015-07-08
Handle none case in mover_trans_migration_queue:init_queue().

tylercloke:


joedevivo:
:+1:

stevendanna:
:+1: A patch so nice, I :+1: it twice

PR #417, opened 2015-07-10
update private-chef-ctl so that it checks for root before proceeding

marcparadise:
Most private-chef-ctl commands will fail in an unhelpful manner if private-chef-ctl/chef-server-ctl is not run as root.  Add a check for root before invoking omnibus-ctl.

ping @chef/lob


stevendanna:
:+1: This is a major annoyance.

PR #418, opened 2015-07-10
use tcp connections for all possible scenarios when adminstrating local db

marcparadise:
Add a password for the opscode-pgsql superuser when non-external postgres is used, and from that point forward use password auth. 

The goal here is to ensure common connection behavior for all actions that apply to both a remote postgres instance and a local postgres instance, and to make use of other behaviors (connecting via pipe) explicit when needed for a local connection.

This also fixes a bug where the sql_user password for opscode-erchef was regenerated on every reconfigure. This was due to the fact that the password attribute is created under opscode-erchef during the initial run, but for all subsequent runs it can be found under opscode_erchef. 

/cc @chef/lob

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -163,7 +163,9 @@ module PrivateChef
           end
         end
       end
-      # Transitional from erchef's sql_user/password etc lived under 'postgresql' and not under opscode-erchef
+
+      # Transition from erchef's sql_user/password etc living under 'postgresql' in older versions,
+      # to 'opscode_erchef' in newer versions.
       if PrivateChef['postgresql'].has_key? 'sql_password'
         PrivateChef['opscode_erchef']['sql_password'] ||= PrivateChef['postgresql']['sql_password']
         PrivateChef['postgresql'].delete 'sql_password'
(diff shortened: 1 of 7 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -178,15 +180,18 @@ def generate_secrets(node_name)
...
       ha_guard = PrivateChef['topology'] == 'ha' && !me['bootstrap']
 
+      PrivateChef['postgresql']['db_superuser_password'] ||= generate_hex_if_bootstrap(50, ha_guard)
       PrivateChef['redis_lb']['password'] ||= generate_hex_if_bootstrap(50, ha_guard)
       PrivateChef['rabbitmq']['password'] ||= generate_hex_if_bootstrap(50, ha_guard)
       PrivateChef['rabbitmq']['jobs_password'] ||= generate_hex_if_bootstrap(50, ha_guard)
       PrivateChef['rabbitmq']['actions_password'] ||= generate_hex_if_bootstrap(50, ha_guard)
       PrivateChef['drbd']['shared_secret'] ||= generate_hex_if_bootstrap(30, ha_guard)
       PrivateChef['keepalived']['vrrp_instance_password'] ||= generate_hex_if_bootstrap(50, ha_guard)
+      # In a second or later run, opscode_erchef has become opscode-erchef...
```
stevendanna: Can you explain a bit of what you are seeing?  We translate from _ to - as part of a number of services, so my guess is that whatever you are seeing might be broken for other services as well: https://github.com/chef/chef-server/blob/mp/one-connection-type-to-bind-them/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb#L288

stevendanna:
Looks good overall.  Regarding the opscode_erchef vs opscode-erchef issue, it seems like the thing we should do is to be consistent from the start rather than adding this conditional?

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
marcparadise: From what I could see, the only things affected will be those that have secrets in the secrets file.  The only others with underscores are 'oc_' and those are explicitly bypassed in the - > _ conversion. 

That said, on re-read through this code, it looks like it might work if I just ensure that I write out the secrets file key as 'opscode-erchef' and not 'opscode_erchef'.  Taking another look at it now.

stevendanna: > That said, on re-read through this code, it looks like it might work if I just ensure that I write out the secrets file key as 'opscode-erchef' and not 'opscode_erchef'. Taking another look at it now.

I think we might have tooling that reads that file, so we probably don't want to change the format :/

marcparadise:
Result of experiment: nope, that was actually what I did originally. I will keep looking, but in the mean time have updated this to remove that portion of the change - I'll do that under a separate PR.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
marcparadise: The opscode-_erchef entry is new, it previously had no secrets

marcparadise:
Re-running through tests after some squash troubles, but this now contains only the connectivity changes - password regeneration bug fix coming separately. Will merge once CI passes.

PR #420, opened 2015-07-10
Add read-only sql user for oc-id

marcparadise:
This change adds a new read-only database user for the oc_id database. This is in support of an incoming change that adds a new chef-server-ctl psql command, which defaults to connecting to a given application database using the read-only user.


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -536,6 +536,8 @@
 default['private_chef']['oc_id']['sql_database'] = "oc_id"
 default['private_chef']['oc_id']['sql_user'] = "oc_id"
 default['private_chef']['oc_id']['sql_password'] = "snakepliskin"
+default['private_chef']['oc_id']['sql_ro_user'] = "oc_id_ro"
+default['private_chef']['oc_id']['sql_password'] = "look-but-don't-touch"
```
mmzyk: Is this supposed to be sql_ro_password? This is setting sql_password a second time as is.

marcparadise: Yes indeed - it won't break anything, but it's definitely wrong. Thanks!

PR #421, opened 2015-07-10
Migration to decompress node data

jcreedcmu:
:construction: WIP, don't merge yet.

@marcparadise does this look plausible as a starting point for decompressing unnecessarily compressed data in postgres in chef-server? I tested it locally via

``` shell
$ dvm load chef-mover --force && dvm start chef-mover
erlang> mover_manager:migrate(all, 1, mover_uncompress_serialized_objects_callback).
```

and it seems to work fine, and work idempotently at that, since `chef_db_compression:uncompress` checks for the gzip magic number and does nothing if it's not present. Moreover because of that the frontend works fine without any further changes, independently of whether the underlying database items are compressed or not. e.g. I can run this migration, and chef-manage still happily lists node attributes. The questions I want to resolve before expecting substantive review are:
1. Is it important to hunt down all the other database columns that are compressed just like `serialized_object` in table `node`? There aren't that many of them, are there?
2. Is this compression noted in some schema file (formal or informal) that I should change? Or does the schema only say `bytea` and leave it at that?
3. Should I be batching this change org-by-org? I probably should, yeah? I could use some more information about how that ought to be done maximally idiomatically inside `chef-mover`

For the sake of not being redundant, the node `SELECT` might itself want to check for gzip magic number via a `WHERE NOT (get_byte(serialized_object, 0) = 31 AND get_byte(serialized_object, 1) = 139)` in case the migration gets interrupted, but as noted above, I think this is only a matter of performance, not correctness.

Also, it's worth noting that although I am putting up this PR first, it seems to me that the logically prior thing to do is to disable, in `chef_db_compression` itself, any compression of data going forward. I don't know of any downsides to having a database in a mixed state where old rows are gzipped and new rows aren't.

### DIFF
--- a/src/chef-mover/src/mover_uncompress_serialized_objects_callback.erl
+++ b/src/chef-mover/src/mover_uncompress_serialized_objects_callback.erl
@@ -1,0 +1,67 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
+%% ex: ts=4 sw=4 et
+%% @author Jason Reed <jreed@chef.io>
+%% @copyright 2015 Chef, Inc.
+-module(mover_uncompress_serialized_objects_callback).
+
+-export([
+	 migration_init/0,
+	 migration_type/0,
+	 supervisor/0,
+	 migration_start_worker_args/2,
+	 error_halts_migration/0,
+	 reconfigure_object/2,
+	 migration_action/2,
+	 next_object/0
+	]).
+
+-record(node_record, {
+        'id',
+        'serialized_object'
+       }).
+
+-include("mover.hrl").
+
+migration_init() ->
+    mover_transient_migration_queue:initialize_queue(?MODULE,
+                                                     all_unconverted_nodes()),
+    ok.
+
+migration_start_worker_args(Object, AcctInfo) ->
+    [Object, AcctInfo].
+
+migration_action(Row, AcctInfo) ->
+    Id = Row#node_record.id,
+    SerializedObject = Row#node_record.serialized_object,
+    NewObject = chef_db_compression:decompress(SerializedObject),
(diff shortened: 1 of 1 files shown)


stevendanna:
> Is this compression noted in some schema file (formal or informal) that I should change? Or does the schema only say bytea and leave it at that?

I don't think we have any schema to change, but we might want to reach out to CSE and OPS about any maintenance or reporting scripts that might need to be changed either for Hosted or our largest customers. 

> Should I be batching this change org-by-org? I probably should, yeah? I could use some more information about how that ought to be done maximally idiomatically inside chef-mover

If we change erchef to disable inbound compression first, then we can get away with not doing it org by org I think.

If we don't change erchef first, we'll end up in a situation where we have a mix of compressed and uncompressed nodes even after the migration because of in-flight chef-client runs.  Mover theoretically already knows how to do this using the "mover_org_migrator_sup" rather than the transient worker.  However, I think that our per-org tracking is broken since the couchdb->sql migration for orgs, so we might need to fix it up in mover.

jcreedcmu:
@stevendanna does https://github.com/chef/chef-server/pull/427 seem reasonable as a "disable inbound compression" step? Can this PR be merged as-is if I put some kind of "danger danger don't actually run this migration unless you know what you're doing" comment in the header?

marcparadise:
Sorry @jcreedcmu , I didn't see this PR - I will take a look more detailed through on Monday morning.

stevendanna:
@jcreedcmu I did some more research on the schema and left a comment on #427.  Namely, we might need to change the storage types.

stevendanna:
@jcreedcmu Thanks for doing these spikes.  I'm sorry thy never moved foward.  I'm going to close the PRs but keep the branches around in case someone wants to pick them up.  I still think both of these are pretty worthwhile.

PR #424, opened 2015-07-13
Compute version numbers for all Erlang projects

joedevivo:
* Omnibus builds compute version from project.build_version
* local and jenkins builds compute from most recent git tag and commit sha

OMG this was a nightmare. It involved getting two pull requests merged into erlware/erlware_commons and two merged into erlware/relx. It also involved cutting two versions of these (0.14 and 0.15) and (3.3.1 and 3.3.2) respectively.

I am not in love with how it works, but given the limitations (old ci, local, and omnibus builds), I can say that this works.

## The first pass: make rel
`.app.src` files can take `{vsn, git}` to figure out the version number from git. It uses a combination of git tags and sha. Relx did that too, but your tags have to start with 'v'.

erlware/relx#363 and erlware/erlware_commons#83 allows for you to use a custom prefix, in our case `""`; however, the `fgrep` it uses isn't POSIX, so this failed on my omnibus build with ubuntu-10.04. To be nice, erlware/erlware_commons#84 makes it POSIX, but we still have problem.

Even if it didn't fail, the way omnibus builds these things, it doesn't copy over the `./` directory with all of our `.git` info in it. So using git was out of the question unless we copied all of `chef-server` to the vm for all four projects: `bookshelf`, `chef-mover`, `oc_bifrost`, and `oc_erchef`.

It turns out that copying this even one time is a prohibitively expensive operation, and just computing the digest for `chef-server` is... well... long.

## The second pass: VERSION
`rebar` (2.x) allows `{vsn, {cmd, "any command"}}` to set the version string in `.app.src` files, so we could use that to solve the omnibus case of not having a `.git` directory. Oh no! Relx does have that.

Until now! erlware/relx#366 brings the feature to `relx` (3.3.2+). Yay!

But what command will we use?

I introduced two new make targets in each of our 4 erlang projects: `version_clean` and `VERSION`. They work together to create a `VERSION` file whose contents are the version string to use. If you set and environment var `REL_VERSION` it will use that, otherwise it will compute from git.

We can then pass in `REL_VERSION` from omnibus with the `project.build_version` which means every erlang project (and subproject e.g. chef-objects) will be versioned with the same string that the .deb package will ultimately be versioned with. It's prefixed with the one in `omnibus/config/projects/chef-server.rb`, so you still have to change it there.

The way that the `version_clean` and `VERSION` targets work together winds up deleting and recreating the version file alot. this is ok, without `version_clean` it would just use the same version number every time, since make targets love not running if the file exists.

`VERSION` is added to the env `REL_HOOK` so every time you `make rel` you get a new version based on the logic needed from wherever you're building.

## Olde CI support
The jenkins scripts are easier now, since the logic to build the version string used to be in there, but now doesn't need to be. 

## Please test it.

With all that said, I'd like it tested by everyone in @chef/lob to make sure all of our different cases are covered. As far as I know, it should work for everyone, but I only have proof that it works for me.

In order for it to work correctly, you'll need updated relx which means manually removing the following files:

* src/bookshelf/relx
* src/chef-mover/relx
* src/oc_bifrost/relx
* src/oc_erchef/relx

The makefiles should take care of everything else.

## After
I'd like to abstract some of the jenkins scrips and a version.mk out to the root of the `chef-server` project to reduce the duplication of that code, but that can be another PR.


src/bookshelf/custom.mk
```diff
@@ -36,3 +40,13 @@ clean_ct:
 
 ## Pull in devvm.mk for relxy goodness
 include devvm.mk
+
+version_clean:
+	@rm -f VERSION
+
+VERSION: version_clean
+ifeq ($(REL_VERSION),)
+	@echo -n "$$(git log --oneline --decorate | grep -F "tag: " --color=never | head -n 1 | sed  "s/.*tag: \([^,)]*\).*/\1/")-$$(git rev-parse --short HEAD)" > VERSION
```
stevendanna: Why not use `git describe` here?

stevendanna:
The only part of this that makes me nervous given recent events is the big jump in relx versions.  Let's make a note to compare the erlang apps with respect to included dependencies before and after we merge this.

joedevivo:
The thing about this PR is that the build is actually ok: https://travis-ci.org/chef/chef-server/builds/70779384

gh/travis integration seems to be off because I deleted a commit from this branch

src/bookshelf/custom.mk
joedevivo: This is almost exactly what we used for the jenkins build, so I repurposed it here.

joedevivo: `git describe` is trying to tell me that this branch is off of `12.1.0-rc.1`. `git log` is behaving better.

joedevivo:
@stevendanna I'll kick off an Old CI build and see what happens re: deps

tylercloke:
Seems useful and nothing obvious jumps out at me!

A few things:
1) It looks like all (or at least a lot) of our makefiles are similar. Should we document at a high level what all the commands do?
2) Totally with @stevendanna on being worried about jumping two major versions of relx.

LGTM in general though.

joedevivo:
force pushed some whitespace to kick off a new travis. not my proudest moment

joedevivo:
@tylercloke Our Makefile situation has gotten very repeaty. I'll be working on enhancing that experience, but it's the kind of thing I'll have to do in between tasks of higher priority

joedevivo:
Deps check out

screenshot 2015-07-22 11 09 18

stevendanna:
:ship:

PR #427, opened 2015-07-14
Do not gzip compress any chef server database objects

jcreedcmu:
Remove `chef_db_compression:compress` everywhere it is used, and delete it. It should be redundant from a space-use performance standpoint since postgres does its own compression, and should be safe to remove compression going forward since the server can deal uniformly with any existing items that are or aren't compressed; `decompress` already checks for gzip magic number and does the appropriate thing.


stevendanna:
@jcreedcmu 

> It should be redundant from a space-use performance standpoint since postgres does its own compression

Are we sure on this point?  In your demo you mentioned some numbers related to the space change you saw, but I don't remember the specifics. I ask because our schema uses `SET STORAGE EXTERNAL` for most of our serialized_object fields and the documentation (http://www.postgresql.org/docs/9.4/static/storage-toast.html) leads me to believe this may prevent pg's automatic compression:

> EXTERNAL allows out-of-line storage but not compression. Use of EXTERNAL will make substring operations on wide text and bytea columns faster (at the penalty of increased storage space) because these operations are optimized to fetch only the required parts of the out-of-line value when it is not compressed.

Overall I'm :+1: on the idea of removing our compression and doing a sort of "lazy migration" as items are saved back to the table.  But I'm wondering if the above implies that we need a schema change to `SET STORAGE EXTENDED` on some of those columns.

jcreedcmu:
Yeah, you're absolutely right --- unfortunately the perf experiments I did showed that even `SET STORAGE EXTENDED` still yielded about a 2x blow up in the space requirements compared to the status quo of gzip-before-ship-to-db. Using `jsonb` in postgres 9.4 seems to achieve a rough high-variance parity with status-quo-gzip, plus or minus 30% or so on the `fauxhai` samples I tried using. But that _definitely_ requires a schema change.

marcparadise:
I've been thinking about this a bit. I think the PG decompression is done inline on the PG side, as opposed to the receiving side. The reason we initially added this (iirc) was more about moving fewer bits around as opposed to consuming less disk space. 

I also saw that we effectively disable compression for node objects even now but I'm not sure why.  Those are our biggest objects, and storage is set to `EXTERNAL` - so no compression occurs anywhere for those.

I'm thinking our best bet is to enable it chef-server side for all objects if data xfer quantity is still a concern. If anyone has the full history on why enable compresson on the chef-server side, it would be helpful here.

stevendanna:
@jcreedcmu Thanks for doing these spikes.  I'm sorry thy never moved foward.  I'm going to close the PRs but keep the branches around in case someone wants to pick them up.  I still think both of these are pretty worthwhile.

PR #430, opened 2015-07-14
 chef-server-ctl external postgresql sanity checks. 

marcparadise:
This PR adds several sanity checks that occur when we're running with the external postgres flag. It also contains  a rudimentary structure to add additional checks for future externalized components. 

One thing worth noting here is that these checks are run at the start of the chef client run, before attributes are merged in from chef-server.rb. This lets us check missing/incorrect configuration items that we require for external services. 

When an error does occur, it will fail immediately. It will show a numbered error message which contains details about what the problem was and how it can be fixed. 

Still pending are the URLs of the actual help text.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
@@ -1,0 +1,54 @@
+class PreflightValidationFailed < Exception
+
+end
+
+class PreflightValidator
+  attr_reader :node, :previous_run, :helper
+  def initialize(node)
+    @helper = OmnibusHelper.new(node)
+    @node = node
+    @previous_run = node['previous_run']
+    def fail_with error
+      raise PreflightValidationFailed, error
+    end
+  end
+
+  def backend?
+    #
+    # When these preflight checks are run, the chef-server.rb has been ingested into
+    # PrivateChef but has not been merged into the node. This means for accurate results, we'll
+    # need to provide an assembled node object that contains the configuration that
+    # the backend check needs.
+    faux_node = { 'enterprise' => node['enterprise'],
+                  node['enterprise']['name'] =>  PrivateChef }
+    return unless EnterpriseChef::Helpers.backend? faux_node
+
+    return PrivateChef['topology'] == 'standalone' || PrivateChef['role'] == 'backend'
+  end
+end
+
+class PreflightChecks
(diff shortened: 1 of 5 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
```diff
@@ -0,0 +1,36 @@
...
+  # Stop the run immediately if a validation failure occurs.
+  # We will do so in a way that prevents the default at_exit handlers from being invoked -
+  # which means that the error text is immediately visible to the user, without forcing
+  # them to hunt for it in backscroll.
+  #
+  # Validations are expected to be run after chef-server.rb entries are ingested, but before any
+  # defaults are configured via libraries/private_chef.rb
+  def run!
+    begin
+      PostgresqlPreflightValidator.new(node).run!
```
marcparadise: If we're going to extend this mechanism, I'm sure that ruby has ways to discover specific classes so that we don't have to explicitly create and invoke...

marcparadise:
Likely we can do some class shuffling and add a base class for the validators. That would let us discover them (I think) without explicitly invoking each new one we add, as well as define a common helper like `def fail_with` to avoid the repeated 'raises' throughout.

marcparadise:
ping @chef/lob @charlesjohnson @andy-dufour @jeremiahsnapp

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -151,5 +151,9 @@ def ldap_authentication_enabled?
     node['private_chef'].attribute?('ldap') &&
       !(node['private_chef']['ldap'].nil? || node['private_chef']['ldap'].empty?)
   end
+
+  def is_backend?()
```
stevendanna: I believe we have a function like this in enterprise-chef-common?  Could we consider using that one instead?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
stevendanna: Yeah, there are a few examples of factory-style constructors that collect the relevant classes in Chef.  I'm thinking though it might be better to just keep it super explicit in this case.  A nice list of class instantiations that define the set of checks we are gonna do makes it really easy to see what checks are going to run and in what order.

PR #433, opened 2015-07-16
Properly override chef dependency.

tylercloke:


omnibus/config/projects/chef-server.rb
```diff
@@ -91,7 +91,7 @@
 dependency "private-chef-upgrades"
 dependency "private-chef-cookbooks"
 dependency "chef-ha-plugin-config"
-dependency "chef", version: "9a3e6e04f3bb39c2b2f5749719f0c21dd3f3f2ec" # for embedded chef-client -z runs (built from master - build last)
+override "chef", version: "9a3e6e04f3bb39c2b2f5749719f0c21dd3f3f2ec" # for embedded chef-client -z runs (built from master - build last)
```
stevendanna: You still need the dependency line as well.  So keep the original `dependency "chef"` line and move this override up near the other overrides.

PR #438, opened 2015-07-20
Bring Your Own Postgres - phase 1 

marcparadise:
This PR represents the work done in phase 1 of Bring Your Own Postgres.   This is ready for review and is passing CI post-rebase onto master. 

Prior to merge, the following is to be done: 
 * [x] update to knife-ec
 * [x] minor updates to support RDS 
 * [x]  before performing an upgrade, ensure that the 'external' flag value has not changed.  [upgrade runs reconfigure first, which already validates against this.] 
 * [x] ensuring if we use psql in an upgrade, we are using correct userid/password [@marcparadise]
 * [x] run through new upgrade test pipeline to ensure no suprises.

(I've removed a couple of items that now live in the 'trailing tasks' card) 

ping @chef/lob /cc @andy-dufour @charlesjohnson @jeremiahsnapp


omnibus/files/private-chef-ctl-commands/psql.rb
```diff
@@ -0,0 +1,74 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'json'
+
+known_dbs = {
+  "opscode_chef" => {"dbname" => "opscode_chef", "config_key" => "opscode-erchef", "hashseed" => "private_chef"},
```
stevendanna: Perhaps include oc_erchef as an alias for this so it lines up with dvm.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
```diff
@@ -0,0 +1,54 @@
+class PreflightValidationFailed < Exception
```
stevendanna: We should have copywrite headers on these files.

stevendanna: I think it would be better to make these exceptions inherit from StandardError.  We try to avoid inheriting from Exception since it is the base of the exception hierarchy so it is dangerous to rescue it since it includes Interrupts as well.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def backend?
+    #
+    # When these preflight checks are run, the chef-server.rb has been ingested into
+    # PrivateChef but has not been merged into the node. This means for accurate results, we'll
+    # need to provide an assembled node object that contains the configuration that
+    # the backend check needs.
+    faux_node = { 'enterprise' => node['enterprise'],
+                  node['enterprise']['name'] =>  PrivateChef }
+    return unless EnterpriseChef::Helpers.backend? faux_node
```
stevendanna: These two returns are a bit confusing to me.  Also, I would expect this function to return `true` or `false` rather than `nil`.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    return PrivateChef['topology'] == 'standalone' || PrivateChef['role'] == 'backend'
```
stevendanna: The logic here is a bit odd to me.  Can we replace both of these `return` lines with the following single line:

```
EnterpriseChef::Helpers.backend? faux_node
```

or am I missing some nuance here?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
```diff
@@ -0,0 +1,322 @@
+class PostgresqlPreflightValidator < PreflightValidator
+  attr_reader :cs_pg_attr, :node_pg_attr
+  require "pg"
```
stevendanna: Minor style point: we'd typically place these outside of the class definition.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
```diff
@@ -0,0 +1,322 @@
...
+    # ensure it wasn't set to a local host.
+    # Note that this will also prevent people from running external postgres on the same
+    # box as the chef backend. The idea isn't to block 100% of cases, but just to prevent a
+    # possible accidental misconfiguration.
+    #
+    if (cs_pg_attr['vip'].nil? ||
+        cs_pg_attr['vip'] == '::1' ||
+        cs_pg_attr['vip'] == '127.0.0.1'  ||
+        cs_pg_attr['vip'] == 'localhost')
+      fail_with err_missing_external_host
```
stevendanna: Is it possible this check will get in our way later?  What if I just want to run postgresql myself but I still want it on the same machine?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
```diff
@@ -0,0 +1,322 @@
...
+    # see this value change on the front end - so let's not call that out as an error.
+
+    # This test isn't valid until we're bootstrapped (which could
+    # encompass multiple runs in case of error). Before bootstrap,
+    # the default value of 'false' will be in place, and having a
+    # differing value is valid.
+    return unless OmnibusHelper.has_been_bootstrapped?
+    return unless previous_run
+    return unless backend?
+    puts "EXTERNAL #{cs_pg_attr['external']} vs #{previous_run['postgresql']['external']}"
```
stevendanna: Should we use `Chef::Log` here?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
```diff
@@ -0,0 +1,322 @@
...
+  end
+
+  def backend_verify_postgres_version(connection)
+    # Make sure the server is a supported version.
+    r = connection.exec("SHOW server_version;")
+    v = r[0]['server_version']
+    major, minor = v.split(".")
+    # Load up our required major/minor:
+    # NOTE: our current entry in version-manifest.json is 'postgres92',effectively hardcoding the version.
+    # TODO: We will want to ensure that current postgres is always labeled as 'postgres' which will have migration
```
stevendanna: We have no way of turning a TODO here into an action.  Let's put this somewhere else.  Perhaps as an issue against chef-server.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
```diff
@@ -0,0 +1,322 @@
...
+      end
+    end
+  end
+
+  def backend_validation
+    begin
+      connect_as(:superuser) do |connection|
+        backend_verify_database_access(connection)
+        backend_verify_postgres_version(connection)
+        %w{bifrost opscode_chef oc_id}.each {|db| backend_verify_empty_db(connection, db) }
```
stevendanna: Do we actually need a completely empty db?  What if someone is running a large postgresql instance running multiple applications?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
```diff
@@ -0,0 +1,322 @@
+class PostgresqlPreflightValidator < PreflightValidator
```
stevendanna: This class has a lot of conditionals, early returns, special cases, etc.  Could we write some rspec tests for it where we mock out EcPostgresql?

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -84,6 +84,14 @@
   if chef_server_rb_exists
     PrivateChef.from_file(chef_server_path)
   end
+  # Bail out if something is wrong in our configuration.
+  # NOTE: Over time, we can move the validation done in private_chef.rb
+  #       here as well.
+  #
+  # If preflight checks fail, they will abort immediately with a detailed error
+  # message, and without a stacktrace to clutter the screen.
+  PreflightChecks.new(node).run!
```
stevendanna: Is there a reason this happens before the node.consume_attributes? line below?  If so, I might add that to the comment above.

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -230,36 +238,41 @@
   "nginx",
   "keepalived"
 ].each do |service|
-  if node["private_chef"][service]["enable"]
-    include_recipe "private-chef::#{service}"
+  if node["private_chef"][service]["external"]
+    begin
+      # Perform any necessary configuration of the external service:
+      include_recipe "private-chef::#{service}-external"
```
stevendanna: My concern with this organization is that it makes it hard to see what is different in the case of external vs non-external if we aren't very careful about how we write these cookbooks.  I'd like support to be able to ask questions like "What are the differences between internal and external postgresql?" and have the answer being easily attainable by scanning the code.

stevendanna:
>  ensuring if we use psql in an upgrade, we are using correct userid/password

https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-upgrades/001/020_multi_key_schema_migration.rb#L5

That is the only one I can find with grep.

stevendanna:
>  update to knife-ec

Depending on what OSC 11 -> 12 upgrade scenarios we want to support, we may need to update various wrapper scripts as well.  Can you point me in the direction of what you found wasn't working?  All of the commands should support these options:

https://github.com/chef/knife-ec-backup/blob/master/lib/chef/knife/ec_base.rb#L55-L71

but we may need to bubble them up in whatever wrappers we have.

stevendanna:
Overall it looks like there are a lot of good improvements and cleanups here.  If we want to aggressively merge this, we might consider "darklauncing" it by bailing out if we see external set to true for now (perhaps overridable by some environment variable).  That would allow us to merge this, continue testing and iterating on it if needed, but still have master releasable.

dev/sync
```diff
@@ -241,6 +256,7 @@ BEGIN {
         (q) terminate and leave the VM running
         (h) terminate and halt the VM
         (s) terminate and suspend
+        (d) destroy the V
```
thommay: "destroy the VM", presumably

stevendanna:
Another TODO:

- [ ] Teach chef-server-ctl start/stop/restart/status about external postgresql

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
marcparadise: One of them should've been removed.

marcparadise: Also, 'unless' is wrong, this was supposed to just pass through to Helpers.backend?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
marcparadise: I remember talking about this, and deciding that we didn't want to support it as a use case.  I'm open to removing the restriction. I could see an edge case where someone is running this on a beefy box that already has postgres.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
marcparadise: This was lingering debug code, removing it.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
```diff
@@ -0,0 +1,322 @@
...
+    begin
+      connect_as(:invalid_user)
+    rescue ::PG::ConnectionBad => e
+      # A bit messy but PG gem does not expose error codes to us:
+      case e.message
+      when /.*Connection refused.*/
+        fail_with err_postgres_not_available
+      when /.*password authentication failed.*/
+        # This is what we want to see.
+      when /.*no pg_hba.conf entry.*/
```
tylercloke: Checking which error message with regex makes me sad. Can we look at the class of the incoming error? There has to be something a bit more programatic that we can use to separate these...

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
```diff
@@ -0,0 +1,322 @@
...
+      end
+    rescue ::PG::InsufficientPrivilege => e
+      fail_with err_not_superuser
+    rescue ::PG::ConnectionBad => e
+      case e.message
+      when /.*database 'template1', does not exist.*/
+        fail_with err_template1_db_required
+      when /.*password authentication failed.*/
+        fail_with err_invalid_superuser_account
+      when /.*no pg_hba.conf entry.*/
```
tylercloke: Similar comment about the class of the error from above.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
tylercloke: There are a few TODOs in other places. Might be working grepping through if this is the wrong place for them.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
```diff
@@ -0,0 +1,322 @@
...
+                                                    'port' => port}) do |conn|
+       if block_given?
+         yield(conn)
+       end
+    end
+  end
+
+  def err_cannot_change_external_flag
+<<EOM
+CSPG001: The value of postgresql['external'] must be set prior to the initial
```
tylercloke: Are these error codes? What are we using them for?

tylercloke:
Added a few comments to the above. Huge amount of our core configuration and big delta to not have any testing imo (not that we really do for our omnibus definitions as its stands, but maybe this is a good opportunity?). Size of the changes makes me nervous but nothing obvious jumped out at me.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
marcparadise: Specifically it checks to see if our named dbs are present.  I'll rename the function for clarity

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
marcparadise: Unfortunately no. The pg gem gives us two classes of errors, and hides the error codes completely.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
marcparadise: Yes. To make life easier for support. Error codes are easier to grep for than message text

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
marcparadise: The flip side is that wanting to avoid the  'external' conditionals throughout the cookbooks if we  can help it - treating things the same as much as possible makes it easier to wrap your head around it.  Any ideas for  a different approach that still lets this happen?

marcparadise:
@stevendanna re: ctl status - since we don't configure the service, chef-server-ctl status wouldn't known about it at all (my understanding is that it checks run-sv links to determine what is present). I've added the TODO item for follow up though.

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
stevendanna: @marcparadise I think it is probably best to do a few more services and see how it pans out.  In some cases it feels like the only difference between "external" and "internal" ought to be that we need to install the service for "internal".  In that world we could have an "installation" cookbook and a "configure/bootstrap" cookbook for each service.  When external was set, we wouldn't run the installation cookbook.

That said, we can see how it goes, this comment shouldn't block merging this.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb
stevendanna: Ahh cool.

marcparadise:
@tylercloke yeah, we spent time in the first half of the sprint getting server-spec support into this chef-server omnibus --  but that "making sure all new and modified code has tests" conflicted directly with "produce a complete shippable feature in a sprint" - particularly since there's not yet a foundation upon which to build those tests. 

I definitely want to add tests to this, and it's worth doing as the next item on the board. It will also better prepare us for future externalization work - but I also don't want to maintain this separate branch for the time it takes to do that. Thoughts?

marcparadise:
Updated for various code review comments. Also included an update to use error codes in the validator error function names, to make them fully searchable. 
@stevendanna It really already is darklaunched - if you don't set 'external', only a couple of things change (database ownership, and IP-based local access)

marcparadise:
We will also need to update the help links in preflight_postgres_validator.rb as they are currently all placeholders.  We'll do that under a separate PR once docs are wrapped up.

PR #439, opened 2015-07-21
Add features for full-installation reindex to chef-server-ctl reindex

stevendanna:
This adds the following features to `chef-server-ctl reindex`:

- an --all-orgs flag to reindex all organizations on the server,

- a --wait flag to wait for the reindex queue to empty before
  returning,

- a --disable-api flag to put the API into 503 mode at the start of an
  index, and

- a --with-timings flag to report the total time, useful for testing.

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/reindex.rb
+++ b/omnibus/files/private-chef-ctl-commands/reindex.rb
@@ -1,13 +1,143 @@
-# Copyright (c) 2013 Opscode, Inc.
-# All Rights Reserved
+#
+# Copyright 2015 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'optparse'
+require 'chef/config'
+require 'chef/rest'
+require 'chef/org'
+require 'redis'
+
+def all_orgs
+  Chef::Config.from_file("/etc/opscode/pivotal.rb")
+  Chef::Org.list.keys
+end
+
+def expander_queue_size
+  output = `/opt/opscode/embedded/service/rabbitmq/sbin/rabbitmqctl list_queues -p /chef | awk '{sum += $2} END {print sum}'`
+  status = $?
+  if !status.success?
+    $stderr.puts "Failed to get queue size!"
+    exit 1
+  else
+    output.to_i
+  end
+end
+
+def wait_for_empty_queue
+  loop do
+    size = expander_queue_size
+    if size == 0
+      break
+    else
+      puts "\t#{size} objects remaining in queue"
+      sleep 1
+    end
+  end
+end
+
+def enqueue_data_for_org(org)
+  status = Dir.chdir(File.join(base_path, "embedded", "service", "opscode-erchef", "bin")) do
+    run_command("./reindex-opc-organization complete #{org} >/dev/null")
+  end
+  if !status.success?
+    $stderr.puts "Failed to enqueue data for #{org}!"
+  end
+end
+
+def redis
+  @redis ||= begin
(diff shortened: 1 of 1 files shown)


stevendanna:
@chef/lob This is a first stab at a feature to reindex an entire installation.  It works well for small to medium sized installations.  For larger installations or more data we may need to conditional additional modifications (temporarily reconfiguring solr during the reindex, moving to batch updates of solr data, etc).

omnibus/files/private-chef-ctl-commands/reindex.rb
```diff
@@ -1,13 +1,143 @@
...
+  end
+end
+
+def wait_for_empty_queue
+  loop do
+    size = expander_queue_size
+    if size == 0
+      break
+    else
+      puts "\t#{size} objects remaining in queue"
```
tylercloke: Should we only `puts` this if the size has changed?

omnibus/files/private-chef-ctl-commands/reindex.rb
```diff
@@ -1,13 +1,143 @@
...
+end
+
+def disable_api
+  puts "- Disabling the Chef API."
+  redis.hset("dl_default", "503_mode", true)
+end
+
+def enable_api
+  puts "- Re-enabling the Chef API"
+  redis.hdel("dl_default", "503_mode")
```
tylercloke: Do we want to delete this instead of setting it to true? I'm sure it does the same thing, I just didn't know if we had a convention.

omnibus/files/private-chef-ctl-commands/reindex.rb
```diff
@@ -1,13 +1,143 @@
...
+
+  puts "- Enqueueing data for indexing."
+  orgs_to_reindex.each do |org|
+    enqueue_data_for_org(org)
+  end
+
+  if options[:wait]
+    puts "- Waiting for reindexing to complete"
+    wait_for_empty_queue
+  end
```
tylercloke: Might be cool if we organized this in a way that we could output per-org? Not really that important though, just might result in better output.

omnibus/files/private-chef-ctl-commands/reindex.rb
```diff
@@ -1,13 +1,143 @@
...
+  OptionParser.new do |opts|
+    opts.on("-w", "--wait", "Wait for reindex queue to clear before exiting") do |w|
+      options[:wait] = w
+    end
+
+    opts.on("-d", "--disable-api", "Disable writes during reindexing") do |n|
+      options[:disable_api] = n
+    end
+
+    opts.on("-a", "--all-orgs", "Reindex all organizations") do |a|
```
tylercloke: Might be worth mentioning something about `--all-orgs` ignoring argument input.

tylercloke:
A bunch of minor comments, but pretty cool expansion of functionality! If we end up shipping this, we should update the csc docs in docs.chef.io. LGTM.

omnibus/files/private-chef-ctl-commands/reindex.rb
stevendanna: I could see doing that as an option.  The reason we don't do it by default is that the only way to get per-org output with the current setup would be to put an org on the queue, wait, then put the next org on the queue.  This would work but I it would would mean we would reindex more slowly since erchef would be sitting doing nothing while the queue cleared (right now, the queue can be clearing for org 1 while erchef is busy getting the data for org 2).

omnibus/files/private-chef-ctl-commands/reindex.rb
stevendanna: The only time we've used this in my memory, we did a delete to remove the key rather than setting it back to false, so that is what I went with.

tylercloke: Cool, was curious what best practices where around this.

omnibus/files/private-chef-ctl-commands/reindex.rb
tylercloke: Ah speed over slightly better output for sure.

tylercloke:
Still think its worth updating the `--all-orgs` option text to note that arguments will be ignored, but other than that :+1:

omnibus/files/private-chef-ctl-commands/reindex.rb
marcparadise: Can we do this per-org instead of globally?

stevendanna:
Updated the text with @tylercloke's suggestion.  Merging now.

Issue #375, opened 2015-07-01
CS 12.1.0 nginx can't reach OC-ID if localhost resolves to IPv6 address


jeremiahsnapp:
[Chef Server 12.1.0 OC-ID uses rails 4.2.0](https://github.com/chef/chef-server/blob/12.1.0/src/oc-id/Gemfile.lock#L267) which [now listens on localhost instead of 0.0.0.0 by default](http://guides.rubyonrails.org/4_2_release_notes.html#default-host-for-rails-server).

If `/etc/hosts` has localhost resolving to both IPv4 and IPv6 IP addresses then rails will bind to the IPv6 address. This breaks access to OC-ID because `oc_id['vip']` [currently defaults](https://github.com/chef/chef-server/blob/12.1.0/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb#L518) to `127.0.0.1`. This breaks OAuth login for private Supermarket and Analytics servers.

Proposed fix is to change the default to be `oc_id['vip'] = 'localhost'`. This fix works in my test server.

pmoosh:
Yep - happens to me as well. Currently I just remove the localhost entry for IPV6 address. I would assume it should listen on IPV4 as well as IPV6

juliandunn:
@jeremiahsnapp can you submit a PR with that?

mmzyk:
This change was merged to master with this commit: https://github.com/chef/chef-server/commit/1570533ba0fed8c2784dfb585f8bc26c2b43a611

mmzyk:
And this should ship with the next server release.

PR #445, opened 2015-07-22
Set default OC-ID VIP to localhost

jeremiahsnapp:
Rails upgraded to 4.2.0 which listens on localhost instead
of 0.0.0.0 by default. If /etc/hosts has localhost resolving
to both IPv4 and IPv6 IP addresses then rails will bind to
the IPv6 address. Setting OC-ID VIP to localhost allows nginx
to connect properly.


jeremiahsnapp:
This fixes https://github.com/chef/chef-server/issues/375

mmzyk:
:+1: 

Add a changelog line so this gets added to the server change line. (you'll have to look up the syntax). 

You also change the docs to reflect that the default is changing: https://docs.chef.io/config_rb_server_optional_settings.html#oc-id

mmzyk:
This was merged with this commit: https://github.com/chef/chef-server/commit/1570533ba0fed8c2784dfb585f8bc26c2b43a611 

I merged it on @jeremiahsnapp's behalf after a rebase and added the changelog entry.

PR #446, opened 2015-07-22
bookshelf-receiver and keepalived scaffold

stevendanna:
This pull request adds:

- configuration for the bookshelf-receiver rsyncd instance
- scaffolding for performing a failover without using DRBD.  Right now this scaffolding mostly does not actually work, but is there for future development.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/chef-ha-native/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/chef-ha-native/attributes/default.rb
@@ -1,0 +1,11 @@
+#######
+# Bookshelf Receiver
+#######
+default['private_chef']['bookshelf-receiver']['enable'] = true
+default['private_chef']['bookshelf-sender']['ha'] = true
+default['private_chef']['bookshelf-receiver']['dir'] = "/var/opt/opscode/bookshelf-receiver"
+default['private_chef']['bookshelf-receiver']['log_directory'] = "/var/log/opscode/bookshelf-receiver"
+default['private_chef']['bookshelf-receiver']['log_rotation']['file_maxbytes'] = 104857600
+default['private_chef']['bookshelf-receiver']['log_rotation']['num_to_keep'] = 10
+default['private_chef']['bookshelf-receiver']['listen'] = '0.0.0.0'
+default['private_chef']['bookshelf-receiver']['port'] = 11873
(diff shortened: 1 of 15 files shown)


omnibus/files/private-chef-cookbooks/chef-ha-native/attributes/default.rb
```diff
@@ -0,0 +1,25 @@
...
+default['private_chef']['bookshelf-sender']['remote_host'] = nil
+default['private_chef']['bookshelf-sender']['remote_port'] = 11873
+default['private_chef']['bookshelf-sender']['remote_username'] = 'opscode'
+default['private_chef']['bookshelf-sender']['remote_password'] = 'opscode'
+
+#######
+# Bookshelf Receiver
+#######
+default['private_chef']['bookshelf-receiver']['enable'] = true
+default['private_chef']['bookshelf-sender']['ha'] = true
```
stevendanna: These should actually be false.  We'll need to handle these services a bit differently.

stevendanna:
Todo:

- [ ] [bookshelf-receiver] Authorization for rsync (shared secret)
- [ ] [bookshelf-sender] Make it a real ruby project (maybe?)
- [ ] [bookshelf-sender] Rewrite using TDD
- [ ] [bookshelf-sender] Correctly configure bookshelf-sender with peer information
- [ ] [bookshelf-sender] Use inotify to avoid full-syncs
- [ ] Add safety net so that sender and receiver can't both be started tat the same time
- [ ] Multi-node testing in dvm
- [ ] Add chef-ha-native to dvm project load config

Stretch:
- [ ] Encryption for rsync

src/bookshelf-sender/lib/bookshelf/sender/rsync_worker.rb
```diff
@@ -0,0 +1,130 @@
...
+          case an_updates
+          when :full_sync
+            # If our message is full_sync don't even bother, we are going
+            # to clear the queue anyway.
+            return :full_sync
+          when PendingUpdateSet
+            updates += an_updates
+          end
+
+          return updates if queue.empty?
```
stevendanna: I think we might want to remove this attempt at getting multiple items on the queue or add some sort of upper bound.  I'm worried that with a fast enough writer/the right thread scheduling, this will never return.  The UpdateWriter already does work to minimize the items it puts on the queue.

stevendanna:
@chef/lob I've moved the bookshelf-sender stuff inside of bookshelf so that is a different PR now.  This just handles setting up rsyncd on the passive node.

stevendanna:
@chef/lob Moving this over to review since we haven't decided on auth yet.

stevendanna:
Closing this for now based on changing directions.

PR #452, opened 2015-07-27
Removed all ../../includes from -include directives

joedevivo:
Moving the relative path into rebar.config's erl_opts will make it
easier to upgrade to rebar3 when the time comes


stevendanna:
:+1:

PR #454, opened 2015-07-27
Renamed apps/*/test_utils

joedevivo:
We have 3 modules called test_utils. We're probably only getting away
with this because of how rebar2 handles paths.

The following four files have been renamed:

apps/chef_db/test_utils.erl -> chef_db_test_utils.erl
apps/chef_objects/test_utils.erl -> chef_objects_test_utils.erl
apps/oc_chef_authz/test_utils.erl -> oc_chef_authz_test_utils.erl
apps/oc_chef_wm/test_util.erl -> oc_chef_wm_test_utils.erl


tylercloke:
Looks reasonable to me.

PR #455, opened 2015-07-27
Relative locations in test cases

joedevivo:
Making sure EUnit is looking for files in places both rebar2 and 3 can find


tylercloke:
Aw yeee, tests. Looks good to me.

PR #456, opened 2015-07-29
[bookshelf] Add bksw_sync module

stevendanna:
The bksw_sync module handles asyncronous notifications of file uploads
or deletions, using rsync to replicate the files to a remote host.


stevendanna:
cc @chef/lob

src/bookshelf/src/bksw_sync.erl
```diff
@@ -0,0 +1,227 @@
+-module(bksw_sync).
+%%
```
marcparadise: Copyright header

src/bookshelf/src/bksw_sync.erl
```diff
@@ -0,0 +1,227 @@
...
+    {ok, State}.
+
+terminate(_Reason, _State) ->
+    ok.
+
+%%
+%% Internal Functions
+%%
+
+log_rsync_error(Result) ->
```
marcparadise: Is Result always guaranteed to contain at least an empty list?

src/bookshelf/src/bksw_sync.erl
```diff
@@ -0,0 +1,227 @@
...
+%%
+%% Set a timeout in our gen_server response to handle any unsynced
+%% paths if we never hit the max.
+%%
+add_to_unsynced(Element, State = #sync_state{unsynced_paths = UpdateSet,
+                                             base_directory = BaseDir}) ->
+    NewSet = [relative(Element, BaseDir) | UpdateSet],
+    State1 = State#sync_state{unsynced_paths = NewSet},
+    case length(NewSet) of
+        N when N > ?MAX_UNFLUSHED_SIZE ->
```
marcparadise:
```
of
    ?MAX_UNFLUSHED_SIZE ->

Comment (marcparadise) [inline src/bookshelf/src/bksw_sync.erl]: Disregard... I misread.

File: src/bookshelf/src/bksw_sync.erl [hunk 4]
```diff
@@ -0,0 +1,227 @@
...
+to_binary(Path) when is_binary(Path) ->
+    Path.
+
+relative(Path, BaseDir) ->
+    Size = byte_size(BaseDir),
+    case Path of
+        <<BaseDir:Size/binary, Rest/binary>> ->
+            Rest;
+        _Other ->
+            throw(not_child_of_base_dir)
```
marcparadise: Is this a possible path?

src/bookshelf/src/bksw_sync.erl
```diff
@@ -0,0 +1,227 @@
...
+    CmdStatus = exec:run(Cmd, [stdout, stderr, sync]),
+    file:delete(Filename),
+    case CmdStatus of
+        {ok, _Rest} ->
+            lager:info("partial rsync run successful."),
+            {noreply, State#sync_state{unsynced_paths = []}};
+        {error, Result} ->
+            lager:error("partial rsync run failed, retrying in ~p ms", [?MAX_UNFLUSHED_TIME]),
+            log_rsync_error(Result),
+            timer:apply_after(?MAX_UNFLUSHED_TIME, ?MODULE, sync, []),
```
marcparadise: We might want to change this up so that if a partial sync fails N times, we empty the queue and try to force a full sync?

src/bookshelf/src/bksw_sync.erl
```diff
@@ -0,0 +1,227 @@
...
+            {noreply, State}
+    end.
+
+handle_call(status, _From, State) ->
+    % Use a timeout here to prevent status calls from stranding
+    % unsynced paths for longer than expected.
+    {reply, State, State, ?MAX_UNFLUSHED_TIME}.
+
+handle_info(timeout, State = #sync_state{unsynced_paths = UP}) ->
+    lager:info("No updates for ~p ms", [?MAX_UNFLUSHED_TIME]),
```
marcparadise: This will get spammy in the logs in short order, might want to remove it after dev is done.

src/bookshelf/src/bksw_wm_object.erl
```diff
@@ -122,7 +122,14 @@ generate_etag(Rq0, Ctx) ->
 
 delete_resource(Rq0, Ctx) ->
     {ok, Bucket, Path} = bksw_util:get_object_and_bucket(Rq0),
-    {bksw_io:entry_delete(Bucket, Path), Rq0, Ctx}.
+    Res = bksw_io:entry_delete(Bucket, Path),
+    case Res of
+        true ->
+            sync_deletion(bksw_io_names:entry_path(Bucket, Path)),
```
marcparadise: Given that these functions themselves are just one-line pass through calls, maybe we should just call bksw_sync directly?

marcparadise:
This looks good, only a few minor comments above. The biggest change would be forcing a full_sync if partial sync fails more than N times. 

Other than that, would there be an advantage to making the sync intervals and max size tuneables?

marcparadise:
Also, tests!

src/bookshelf/src/bksw_sync.erl
stevendanna: Not from the current call sites no.

src/bookshelf/src/bksw_sync.erl
stevendanna: I was thinking that rather than force a full sync on N failures, force a full sync once the unsync_paths list grows to a certain size.  I think the danger when we are failing a lot of sync attempts in a row is that unsynced_paths keeps growing.

stevendanna:
> Other than that, would there be an advantage to making the sync intervals and max size tuneables?

We should definitely make them tunable.

src/bookshelf/src/bksw_wm_object.erl
stevendanna: :+1:

src/bookshelf/src/bksw_sync.erl
stevendanna: Looks like it:

https://github.com/saleyn/erlexec/blob/master/src/exec.erl#L726-L728

stevendanna: But we can still add a guard and then a fall-through case.

stevendanna:
@chef/lob Updated with most of the review changes:

- copyright header
- configurable wait intervals and max pending
- some logging cleanup
- fall-through case for exec:run error handling
- TESTS!

I also added the following to address a problem I found when playing with it:

- Track whether we've completed the initial sync to ensure that partial sync requests don't pile up while we are in a retry loop for the initial sync but writes are coming in.

@marcparadise I haven't added falling back to full syncs after N partial failures yet.  I'm +1 on special handling for when we've been failing for a long time but want to discuss the right behavior with you in chat today if you have time.

Do we want the staleness-tracking in this first cut of this?  If so I'll get working on that.

marcparadise:
:+1: on this as is, and we can address the excessive failure discussion separately.

src/bookshelf/src/bksw_conf.erl
```diff
@@ -129,6 +131,22 @@ port() ->
             {port, Port}
     end.
 
+rsync_uri() ->
+    case application:get_env(bookshelf, rsync_uri) of
+        undefined ->
+            undefined;
```
tylercloke: Nitpick, should you just catch on
```
WhateverElse ->
   WhateverElse
```

Is that better erlang-ism? Just pass it along? Or do we wanna fail if undef (I'm actually asking lol).

tylercloke:
Looks pretty straightforward! I didn't see any issues. Pretty cool. :+1:

src/bookshelf/src/bksw_conf.erl
stevendanna: To be honest, I was mostly just following the structure of the other methods here, with the only difference being that in this case that my "default" value is also undefined.

PR #462, opened 2015-08-04
[oc_bifrost] Concurrency fix for update_acl

marcparadise:
Added pg_advisory_xact_lock to lock the target id in update_acl

This will block concurrent updates to the target id using
update_acl, which avoids deadlocks seen in cases of multiple
updates to an actor when adding them to multiple groups with concurrent
calls. 

There are some things about this that make it less than ideal:

1. It doesn't address the root cause of why these requests were
   coming in milliseconds apart.
2. This is a blocking lock request. Rather than fail subsequent requests,
   this will allow them through, but it will block them so that they're
   processed first come first served.  In the event that we see a large
   number of concurrent requests for a small number of entities, this could cause 
   a backlog of held connections in the bifrost pool.

   A better long-term solution here might be to make it non-blocking
   (pg_try_advisory_xact_lock) and handle the failure in bifrost, as
   well as handling the specific message in erchef so that we can
   provide meaningful feedback via the API: a second requests to update the
   same ACL is coming in at the same time.

   However, this occurs infrequently  enough that it may not be a
   concern.

3. Because update_acl actually replaces all ACEs this means that when multiple
   requests come in to update the ACL against the same target, the last
   processed will the the final ACL.  There are no guarantees as to
   ordering, so it the last one processed may not be the last request made.

ping @chef/lob - 

ChangeLog-Entry: This corrects a concurrency issue in bifrost where
multiple ACL updates to the same entity could cause a deadlock.


marcparadise:
review note, the actual change is just the one in update_acl.sql.  The others were generated by 'sqitch rework' and represent copies of the previous state for rollback/testing.

marcparadise:
Repro steps captured here: https://gist.github.com/marcparadise/e65234d095ef4e1f4799

stevendanna:
I'm :+1:

> It doesn't address the root cause of why these requests were coming in milliseconds apart.

Do we believe that these concurrent acl updates line up to concurrent user-made API requests or do we think that some operation in erchef generates concurrent acl updates to the same object.  I'd be far more concerned about the latter since it means that we have an erchef operation that is nondeterministic.  I'm also skeptical that it is the latter as I can't think of a request that would need to do that.  In both cases I still think it makes sense to have a lock in this function.

> This is a blocking lock request.

Concurrent updates to *the same* acl should be rare.  Also, because of how our group and acl APIs are exposed to users at the erchef level, I'm pretty sure you can't safely do concurrent ACL or Group updates without a lock somewhere in the user-side code as well.  So any code already doing concurrent updates is probably currently broken, perhaps this will just help the user notice the breakage.

> Because update_acl actually replaces all ACEs this means that when multiple
requests come in to update the ACL against the same target, the last
processed will the the final ACL. There are no guarantees as to
ordering, so it the last one processed may not be the last request made.

I'm pretty sure this doesn't actually change the consistency guarantees of our API.  It has always been a case that "last write wins" and that you can only be sure of request A occurring before request B if request B is started after request A has already returned successfully.

marcparadise:
> Do we believe that these concurrent acl updates line up to concurrent user-made API requests or do we think that some operation in erchef generates concurrent acl updates to the same object. 

I wasn't able to find any.  I remembered that we have an `authz_fanout` parameter, but it appears we're no longer using it.   Tehre is also a path which will lead to oc_httpc:multi_request - but it's not the one that invokes this function.  As far as I could find, we're safe here. 

> So any code already doing concurrent updates is probably currently broken, perhaps this will just help the user notice the breakage
Valid point 

> I'm pretty sure this doesn't actually change the consistency guarantees of our API

True, but I hate knowingly introducing more of the same.

marcparadise:
Alright, I"m going to go ahead and merge this. It will resolve the immediate issue, and worst case lead to some clues as to how the scenario is occurring in the first place.

PR #465, opened 2015-08-06
By default, bind oc_id to IPV4 address

jtimberman:
When using localhost as the vip instead of 127.0.0.1 (as we do
elsewhere), we may end up binding oc_id on port 9090 of ::1 (IPV6).
This can cause nginx to get connection refused, because it will
resolve to 127.0.0.1, but that is not the same.

Setting the vip attribute to 127.0.0.1, AND using the -b option for
rails will ensure that this is set correctly to listen on the IPV4
address.


stevendanna:
I thought this was recently changed to localhost: https://github.com/chef/chef-server/commit/1570533ba0fed8c2784dfb585f8bc26c2b43a611 to avoid this problem?

jtimberman:
on RHEL 7 yesterday, we had the following `/etc/hosts`:

```
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
```

With *either* of the following `rails` startup command in the run script:

```
exec chpst -P -u opscode -U opscode /opt/opscode/embedded/bin/bundle exec rails server -p 9090
exec chpst -P -u opscode -U opscode /opt/opscode/embedded/bin/bundle exec rails server -p 9090 -b localhost
```

The `oc_id` process will listen on `::1:9090`:

```
# w/o -b:
[ec2-user@ip-172-31-9-72 ~]$ sudo netstat -plantu | grep 9090
tcp6       0      0 ::1:9090                :::*                    LISTEN      1421/rails master -
# w/ -b localhost:
[ec2-user@ip-172-31-9-72 ~]$ sudo netstat -plantu | grep 9090
tcp6       0      0 ::1:9090                :::*                    LISTEN      1449/rails master -
```

However, with the following command in the run script (adding `-b 127.0.0.1`, as this PR's change would result):

```
exec chpst -P -u opscode -U opscode /opt/opscode/embedded/bin/bundle exec rails server -p 9090 -b 127.0.0.1
```

It listens on 127.0.0.1:

```
[ec2-user@ip-172-31-9-72 ~]$ sudo netstat -plantu | grep 9090
tcp        0      0 127.0.0.1:9090          0.0.0.0:*               LISTEN      1478/rails master -
```

marcparadise:
We hjave helpers that do the right thing here.  We should set it to 127.0.0.1 and then in https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_id.database.yml.erb use  'OmibusHelper.vip_for_uri('oc_id')
'
That will resolve the correctly based on ipv6 or ipv4 configuration.

marcparadise:
Oops - I linked to the wrong config file, correcting shortly.

marcparadise:
Disregard the above. I've refreshed my memory on ipv6 support (it's been two years) - @jtimberman your change looks good as is.  

The function I referenced is for VIPs that get sent back to the client in URLs or are used over http(s) internally.

Having oc-id on ipv4/127.0.0.1 is the right behavior here - while we allow ipv6 into nginx, local upstreams user 127.0.0.1 even when ipv6 is enabled. 

:+1: on the change.

jtimberman:
And the neat thing about Chef is that vip is an attribute so if someone wants to use an IPV6 address, they can configure that :ice_cream:

stevendanna:
This PR is fine if we want to prefer IPv4, but to be clear, this isn't broken on -current as far as we know.  Using "localhost" ensures nginx talks to oc_id over IPv6 when localhost resolves to IPv6 first.

mmzyk:
@stevendanna is right that we updated this due to bug reports coming in on this not working correctly with IPv6 configured as it was. As far as we knew at the time of the change this was change was fine, although it sounds like perhaps it has consequences we didn't realize. 

That said, we haven't even released the change to default to localhost (so I'm guessing this was found running off of master), but I'm hesitant to change it back and release it when customers are saying this is an issue.

It maybe that the addition of the -b option and changing the attribute setting will fix this, but if that's the case we should also update the docs to make this clear and rope in support.

This PR can show where the docs need to be updated: https://github.com/chef/chef-docs/pull/599

@jeremiahsnapp will be interested in this from support.

jeremiahsnapp:
@jtimberman I originally reported the problem in https://github.com/chef/chef-server/issues/375 and fixed it in https://github.com/chef/chef-server/pull/445 which has already been merged. My PR took a more minimal approach of just using the configuration option that was already available.

This PR adds `-b <%= node['private_chef']['oc_id']['vip'] %> ` and changes `oc_id['vip']` back to `'127.0.0.1'` which will work but it shouldn't be necessary.

However, this PR makes the configuration more explicit so :+1: for that. If rails ever changed its default binding behavior in the future (probably not likely?) this should help us avoid trouble. Plus having more control over configuration is a good thing.

I have no problem seeing this PR merged.

PR #470, opened 2015-08-10
postgres partybus

joedevivo:
This is the partybus change to add remote postgres to migrations. It's working locally, but a build should be run through delivery.


omnibus/partybus/lib/partybus.rb
```diff
@@ -21,6 +21,16 @@ class Config
     attr_accessor :bootstrap_server
     attr_accessor :couchdb_data_dir
 
+    attr_accessor :running_server
+    attr_accessor :postgres
+
+    def initialize
+      if File.exists?("/etc/opscode/chef-server-running.json")
```
stevendanna: Should we be throwing a big error here if chef-server-running.json doesn't exist?

stevendanna: To be a bit more clear about what I mean here, I'm thinking just something like:

```ruby
if File.exists?("/etc/opscode/chef-server-running.json")
  ...
else
  log <<EOF
***
ERROR: Cannot find /etc/opscode/chef-server-running.json
***
Try running `chef-server-ctl reconfigure` before upgrading

EOF
  exit(1)
end
```

stevendanna:
:+1:

PR #475, opened 2015-08-14
Implemented Version 1 of Server Admins

tylercloke:
Implemented Version 1 of [Server Admins](https://gist.github.com/tylercloke/a8d4bc1b915b958ac160).

### Description

Version 1 of Server Admins. Implements flexible user create and read perms.

Now, a global group, called server-admins, will be set up on bootstrap and migration
of existing chef-servers. This global group will have create and read permissions on
the users container, meaning any user placed in the server-admins global group will
have permissions to create and read users system-wide, granting much needed flexability
around our user management for larger orgs. As a side-effect, `knife user` is now
relevant again.

A second migration updates existing users (except pivotal) read, update, and delete permissions to grant server-admins access, as changing container permissions does not update existing objects.

A new set of chef-server-clt tools were added to manage server-admins under Server Admins Commands:
+ grant-server-admin-permissions
Add an existing user to the server-admins global group.
+ list-server-admins
Show all users in the server-admins global group. Should contain pivotal.
+ remove-server-admin-permissions
Removes an existing user from the server-admins global group, removing all special permissions they have.

### Details

Ping @chef/lob 

[Current build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/799/downstreambuildview/)

Questions (please help reviewers):
- [x]  Members of this global group will have READ and CREATE permissions on users. We are not granting UPDATE and DELETE because then a lowly server-admin could mess with pivotal. On the other hand, results in CRUDG behavior where admins could only DELETE / EDIT users they personally created with their own user (vs `chef-server-ctl`) and not migrated users / users they didn't personally create. What do y'all think? I think there are valid arguments for both sides. Its something we could always add in the future with a migration very similar to `mover_server_admins_existing_users_read_permissions_callback.erl`.
- [x] Should we prevent `pivotal` from being removed from the `server-admins` global group via the new `chef-server-ctl` commands?

TODO
- [x] Need one more migration to update the READ permission on existing users, since updating the container does not update the permission for existing objects.
- [x] [Documentation](https://gist.github.com/tylercloke/4f2f55aae84c43349ada)
- [ ] Write basic `chef-server-acceptance` tests and run though upgrade battery.
- [ ] I am doing CRUD, should GRANT be included (seems like no)?

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/server_admins.rb
+++ b/omnibus/files/private-chef-ctl-commands/server_admins.rb
@@ -1,0 +1,193 @@
+#
+# Copyright 2015 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'restclient'
+require 'json'
+require 'pg'
+
+PLACEHOLDER_GLOBAL_ORG_ID = "00000000000000000000000000000000"
+
+add_command_under_category "grant-server-admin-permissions", "server-admins", "Grant a user the ability to create other users by added the user to the server-admins group.", 2 do
+
+  cmd_args = ARGV[3..-1]
+  if cmd_args.length != 1
+    msg = "Username is the only argument to grant-server-admin-permissions.\nPlease pass a single argument."
+    STDERR.puts msg
+    raise SystemExit.new(1, msg)
+  end
+
+  username = cmd_args[0]
+
+  db = setup_erchef_db
+  server_admins_authz_id = get_server_admins_authz_id(db)
+  user_authz_id = get_user_authz_id(db, username)
+
+  bifrost_config = running_service_config('oc_bifrost')
+  superuser_id = bifrost_config['superuser_id']
+  vip = bifrost_config['vip']
+  port = bifrost_config['port']
+
+  # put the user in the server-admins authz group
+  headers = {
+    :content_type => :json,
+    :accept => :json,
+    'X-Ops-Requesting-Actor-Id' => superuser_id
+  }
+
+  base_url = "http://#{vip}:#{port}"
+
+  RestClient.put("#{base_url}/groups/#{server_admins_authz_id}/actors/#{user_authz_id}", "{}", headers)
+
+  puts "User #{username} was added to server-admins. This user can now list, read, create, and delete users (even for orgs they are not members of) for this Chef Server."
+end
+
+add_command_under_category "remove-server-admin-permissions", "server-admins", "Remove all special permission granted to a user from being a server-admin.", 2 do
+
+  cmd_args = ARGV[3..-1]
+  if cmd_args.length != 1
+    msg = "Username is the only argument to remove-server-admin-permissions.\nPlease pass a single argument."
+    STDERR.puts msg
+    raise SystemExit.new(1, msg)
+  end
+
+  username = cmd_args[0]
+
+  if username.downcase == 'pivotal'
+    msg = "You cannot remove the base superuser pivotal from server-admins."
+    STDERR.puts msg
+    raise SystemExit.new(1, msg)
+  end
+
+  bifrost_config = running_service_config('oc_bifrost')
+  superuser_id = bifrost_config['superuser_id']
+  vip = bifrost_config['vip']
+  port = bifrost_config['port']
+
+  db = setup_erchef_db
+  server_admins_authz_id = get_server_admins_authz_id(db)
+  user_authz_id = get_user_authz_id(db, username)
+
+  # put the user in the server-admins authz group
+  headers = {
+    :content_type => :json,
+    :accept => :json,
+    'X-Ops-Requesting-Actor-Id' => superuser_id
+  }
+
+  base_url = "http://#{vip}:#{port}"
+
+  results = JSON.parse(RestClient.get("#{base_url}/groups/#{server_admins_authz_id}", headers))
+
+  users = db.exec_params("SELECT * from USERS WHERE authz_id IN #{create_sql_collection_string(results['actors'])}")
+  user_found = false
+  users.each do |user|
+    if username == user['username']
+      user_found = true
+      break
+    end
+  end
+
+  if !user_found
+    msg = "User #{username} is not a member of server-admins so it cannot be removed."
+    STDERR.puts msg
+    raise SystemExit.new(1, msg)
+  end
+
+  RestClient.delete("#{base_url}/groups/#{server_admins_authz_id}/actors/#{user_authz_id}", headers)
+
+  puts "User #{username} was removed from server-admins. This user can no longer list, read, create, and delete users for this Chef Server except for where they have default permissions (such as within an org)."
+
+end
+
+add_command_under_category "list-server-admins", "server-admins", "List users that have server-admins permissions.", 2 do
+  cmd_args = ARGV[3..-1]
+  if cmd_args.length != 0
+    msg = "This command does not accept arguments."
+    STDERR.puts msg
+    raise SystemExit.new(1, msg)
+  end
+
+  bifrost_config = running_service_config('oc_bifrost')
+  superuser_id = bifrost_config['superuser_id']
+  vip = bifrost_config['vip']
+  port = bifrost_config['port']
+
+  db = setup_erchef_db
+  server_admins_authz_id = get_server_admins_authz_id(db)
+
+  # get all the user authz_ids for all memebers of the server-admins authz group
+  headers = {
+    :content_type => :json,
+    :accept => :json,
+    'X-Ops-Requesting-Actor-Id' => superuser_id
+  }
+
+  base_url = "http://#{vip}:#{port}"
+
+  results = JSON.parse(RestClient.get("#{base_url}/groups/#{server_admins_authz_id}", headers))
+
+  # get the user's authz id
+  users = db.exec_params("SELECT * from USERS WHERE authz_id IN #{create_sql_collection_string(results['actors'])}")
+  users.each do |user|
+    puts user['username']
+  end
+end
+
+# returns string in format '('item1','item2',...,'itemN')'
+def create_sql_collection_string(arr)
+  str = arr.join("','")
+  "('#{str}')"
+end
+
+def setup_erchef_db
+  erchef_config = running_service_config('opscode-erchef')
+  pg_config = running_service_config('postgresql')
+  ::PGconn.open('user' => erchef_config['sql_user'],
+                'host' => pg_config['vip'],
+                'password' => erchef_config['sql_password'],
+                'port' => pg_config['port'],
+                'dbname' => 'opscode_chef')
+end
+
+def get_server_admins_authz_id(db)
+  server_admins_erchef_group = db.exec_params("SELECT authz_id FROM groups WHERE name='server-admins' AND org_id='#{PLACEHOLDER_GLOBAL_ORG_ID}'")
+
+  if server_admins_erchef_group.ntuples == 0
+    msg = "The server-admins global group was not found. Please finish upgrading your Chef Server by following the documentation before using Server Admins related commands."
+    STDERR.puts msg
+    raise SystemExit.new(1, msg)
+  end
+
+  if server_admins_erchef_group.ntuples != 1
+    msg = "More than one server-admins global group was found. Please contact a sysadmin or support (#{server_admins_erchef_group.ntuples} groups found)."
+    STDERR.puts msg
+    raise SystemExit.new(1, msg)
+  end
+
+  server_admins_erchef_group.first['authz_id']
+end
+
+def get_user_authz_id(db, username)
+  user = db.exec_params("SELECT authz_id FROM users WHERE username='#{username}'")
+
+  if user.ntuples != 1
+    msg = "User #{username} was not found."
+    STDERR.puts msg
+    raise SystemExit.new(1, msg)
+  end
+
+  user.first['authz_id']
+end
--- a/src/chef-mover/src/mover_server_admins_existing_users_read_permissions_callback.erl
+++ b/src/chef-mover/src/mover_server_admins_existing_users_read_permissions_callback.erl
@@ -1,0 +1,94 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
+%% ex: ts=4 sw=4 et
+%% @author Tyler Cloke <tyler@chef.io>
+%% @copyright 2015 Chef Software, Inc.
+%%
+%% This migration iterates through all existing users and grants
+%% the server-admins global group READ access on them, as having
+%% access on the container does not update permissions for
+%% existing users.
+
+-module(mover_server_admins_existing_users_read_permissions_callback).
+
+-export([
+         migration_init/0,
+         migration_complete/0,
+         migration_type/0,
+         supervisor/0,
+         migration_start_worker_args/2,
+         migration_action/2,
+         next_object/0,
+         error_halts_migration/0,
+         reconfigure_object/2,
+         needs_account_dets/0
+        ]).
+
+-define(GLOBAL_PLACEHOLDER_ORG_ID, <<"00000000000000000000000000000000">>).
+
+-record(user, {username, authz_id}).
+-record(group, {authz_id}).
+
+migration_init() ->
+    mv_oc_chef_authz_http:create_pool(),
+    mover_transient_migration_queue:initialize_queue(?MODULE, get_users()).
+
+migration_action(UserRecord, _AcctInfo) ->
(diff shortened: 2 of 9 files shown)


omnibus/files/private-chef-ctl-commands/global_admins.rb
```diff
@@ -0,0 +1,139 @@
...
+
+  db = setup_erchef_db
+  global_admins_authz_id = get_global_admins_authz_id(db)
+  user_authz_id = get_user_authz_id(db, username)
+
+  # put the user in the global-admins authz group
+  headers = {
+    :content_type => :json,
+    :accept => :json,
+    'X-Ops-Requesting-Actor-Id' => running_config['private_chef']['oc_bifrost']['superuser_id']
```
marcparadise: Two things - a new function `running_service_config(servicename)` has been added and ready for use once the LOB-10-bugfix PR lands shortly in omnibus-ctl. 

I'd also like to see us just grab it once, `bifrost_config = running_service_config('oc_bifrost')`, etc.

omnibus/config/software/sequel-gem.rb
```diff
@@ -0,0 +1,29 @@
+#
```
marcparadise: Let's stick with using the pg gem, which we already have available. sequel is currently used only for migration from old 11.early sequel-based migrations, and will hopefully be getting yanked soon.

omnibus/files/private-chef-ctl-commands/global_admins.rb
```diff
@@ -0,0 +1,139 @@
...
+
+PLACEHOLDER_GLOBAL_ORG_ID = "00000000000000000000000000000000"
+
+add_command_under_category "grant-global-admin-permissions", "global-admins", "Grant a user the ability to create other users by added the user to the global-admins group.", 2 do
+
+  cmd_args = ARGV[3..-1]
+  if cmd_args.length != 1
+    msg = "Username is the only argument to grant-global-admin-permissions.\nPlease pass a single argument."
+    STDERR.puts msg
+    raise SystemExit.new(1, msg)
```
marcparadise: return exit! 1.  exit! in this case is a Ctl method This also makes sure that post-hooks run or don't run based on the return. Same for the other usage of SystemExit.

marcparadise: Actually skip this comment, I see we're doing this all over the place in existing commands.  I just modified that PR to still do the right thing when SystemExit is raised.

src/chef-mover/src/mover_global_installation_admins_callback.erl
```diff
@@ -0,0 +1,203 @@
...
+        {error, not_found} -> throw_not_found("read");
+        {error, server_error} -> throw_server_error()
+    end,
+
+    CreateUrl = make_url([containers, UserContainerAuthzId, create]),
+    case oc_chef_authz_http:request(CreateUrl, put, [], Body, SuperuserAuthzId) of
+        ok -> true;
+        {error, not_found} -> throw_not_found("create");
+        {error, server_error} -> throw_server_error()
+    end.
```
marcparadise: Let's also add the pivotal user to this group as part of the migration. Maybe one day, we can move entirely to depending on this group instead of the specific pivotal user...

src/chef-server-bootstrap/bin/bootstrap-platform
```diff
@@ -102,13 +113,52 @@ class BootstrapPlatform
...
+    original_create = JSON.parse(RestClient.get(create_url, headers))
+    original_create["groups"] << granted_to_id
+    RestClient.put(create_url, original_create.to_json, headers)
+
+    # grant read perm
+    read_url = "#{base_url}/read"
+    original_read = JSON.parse(RestClient.get(read_url, headers))
+    original_read["groups"] << granted_to_id
+    RestClient.put(read_url, original_read.to_json, headers)
+  end
```
marcparadise: As above, let's add pivotal to this group.

tylercloke:
Interestingly, chef-server-bootstrap does not take a port for the erchef db. Should it?

marcparadise:
Heck yeah - I haven't looked closely yet, but I think that may also resolve an open issue about the server breaking when an alternative port is specified...

danielsdeleo:
I might be tempted to just prevent people from deleting/ruining pivotal in the WM resource(s) and then do whatever is right for this case and ignore pivotal. If there is some reason why a customer _needs_ to delete pivotal then there should be an extra layer of opt-in.

danielsdeleo:
cf. http://lists.opscode.com/sympa/arc/chef/2015-09/msg00128.html this poor guy.

tylercloke:
@danielsdeleo As this branch stands, it is only adding read and create privs to existing users. Since pivotal is created at bootstrap, users won't be able to delete it. They will only be able to delete users they directly created (as they get full perms on objects they made), but not deleting pivotal was part of the motivation for only adding create and read perms.

danielsdeleo:
Right, so you wrote:

> We are not granting UPDATE and DELETE because then a lowly server-admin could mess with pivotal.

If a lowly server-admin could NOT mess with pivotal, then would the best option be to give UPDATE and DELETE to the server admins?

tylercloke:
Sure, we'd just have to update the migration to ignore pivotal. Pivotal is already being created before the global group is added to the container so that would be the only change needed. Its just a matter of how powerful we want server-admins to be.

tylercloke:
@danielsdeleo I think that is probably the sanest approach, as it will be the simplest for the end user. I'll update my branch unless there are objections.

marcparadise:
In general we've been making the move to enforcing policy to prevent user-initiated changes that can break the chef server.  In that context... 

Q1 - I think it would make for a confusing experience if an admin could not delete users that already existed, so it'd be a good move to make the migration that applies to existing users. 

Q2: Yes to that addition - this way we can start to make the move to using only that group permission, and not relying on any one specific user, and eventually remove the next thing I'm going to ask for :) I think there should be two checks there - one to not remove users from  the `superusers` config value present for `oc_chef_wm` in app.config from the global admins group, and one to ensure that the global admins group is never permitted to become empty. 

Because our tooling *does* currently rely on a known superuser account, let's take it one step further and a hard check in  /users/$name endpoint to 403 for any user name present in the `superusers` config value present for `oc_chef_wm` in app.config, with a custom message stating that the default superuser(s) cannot be deleted - that would also prevent the case from the ML above. 

This addition to named user `DELETE` will make the second check (to make sure global admins is not made empty) redundant, but only until our tooling no longer requires the pivotal user. 

These changes now will make it much easier if/when we decide to git rid of the  mostly hard-coded pivotal user  in our tooling. I could see a case for that when we consider what we provide with action log - it could give a way to know who did what as a superuser when using the chef-server-ctl tools. 

==== TL;DR 

For a happy, more break-resistant user experience and to pave the way to a pivotal-less future: 

1. do migrate existing users so that global admins group can do things to them too 
2. do add the check to group update to ensure that any name in the `superusers` config[1] entry cannot be deleted.  (2a. Do we need to ensure nobody can delete the global admins group itself? This becomes relevant in my next comment... )
3. modify `/users/$NAME` `DELETE` to prevent any name in the `superusers` config entry from being deleted.   
4. add a new check to make sure the global admins group is never emptied.   This check will be redundant, but only until we move away from requiring a specific superuser. 

[1] Since the group add/delete handling t in erlang at the moment, might have to fish out where we populate that config entry from to handle it correctly from the ruby side.

danielsdeleo:
So if you really do want to delete a user who's a global admin, you have to remove them from global admins and then it's permitted?

marcparadise:
Couple of trailing thoughts: 

Something to consider - not necessarily for this, but for where we can go next.  I'd like to see us expanding what we have inbound with other changes - let's consider putting global group controls behind the new _admin/ path we're adding for endpoints.  _admin/users or similar.  This'll avoid having this logic split (and sometimes duplicated) in two places - migration rules in erlang, and the new ctl commands.

marcparadise:
@danielsdeleo yes. We already have a policy in place like that for org admins.

marcparadise:
Final thought: There's a DSL-like thing in place for setting permissions instead of having to do each step explicitly.  It's currently geared for use with org setup, but could it be used [with some mods] instead of the explicit setup we have here?  
* https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_org_creator.erl#L68
*  I thought we had a group migration that used this in mover, but I can't seem to find it...

danielsdeleo:
Ok, just wanted to make sure I was understanding everything in your proposal correctly. 

Also, I think I used the acl dsl in the migration that adds the policy containers.

src/chef-mover/src/mover_server_admins_existing_users_read_permissions_callback.erl
```diff
@@ -0,0 +1,87 @@
...
+
+get_users() ->
+    {ok, Users} = sqerl:select(get_users_sql(), [], rows_as_records, [user, record_info(fields, user)]),
+    Users.
+
+get_users_sql() ->
+    <<"SELECT username, authz_id FROM users">>.
+
+get_server_admins_authz_id() ->
+    {ok, [ServerAdmin]} = sqerl:select(get_server_admins_authz_id_sql(), [], rows_as_records, [group, record_info(fields, group)]),
```
stevendanna: I'm wondering if it would be beneficial to use the process dictionary to cache this value so that we don't have to do this select for every user.

tylercloke:
@marcparadise All those checks sound reasonable, but currently the ctl command I wrote is just using SQL since there ins't a global groups ACL endpoint yet. I think we can ship this feature with a minimum viability if we:

1) Grant CRUD on all users for all `server-admins`.
2) Do a simple check in the `chef-server-ctl` remove command added by this PR for pivotal for now.

I like all your ideas, and they'd fall nicely under Version 2 if you read my prep document, but I think we can get this merged down and keep iterating if we do 1 and 2 above.

src/chef-mover/src/mover_server_admins_existing_users_read_permissions_callback.erl
tylercloke: Sound like a good idea.

marcparadise:
Works for me - :+1: with your 1&2, the rest will fit will into when we take it further.

omnibus/files/private-chef-ctl-commands/server_admins.rb
```diff
@@ -0,0 +1,170 @@
...
+                'host' => pg_config['vip'],
+                'password' => erchef_config['sql_password'],
+                'port' => pg_config['port'],
+                'dbname' => 'opscode_chef')
+end
+
+def get_server_admins_authz_id(db)
+  server_admins_erchef_group = db.exec_params("SELECT authz_id FROM groups WHERE name='server-admins' AND org_id='#{PLACEHOLDER_GLOBAL_ORG_ID}'")
+
+  if server_admins_erchef_group.ntuples != 1
```
marcparadise: Add check for zero  and tell them to run chef-server-ctl upgrade before proceeding.

src/chef-mover/src/mover_server_admins_global_group_callback.erl
```diff
@@ -0,0 +1,210 @@
...
+    %% Grant pivotal permissions on server-admins global group
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, actor, ErchefSuperuserAuthzId, group, ServerAdminsAuthzId, create),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, actor, ErchefSuperuserAuthzId, group, ServerAdminsAuthzId, read),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, actor, ErchefSuperuserAuthzId, group, ServerAdminsAuthzId, update),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, actor, ErchefSuperuserAuthzId, group, ServerAdminsAuthzId, delete),
+    ok.
+
+%% Start vendor
+
+%% Vendored from mv_oc_chef_authz since not exported
```
marcparadise: Any reason to not add the exports since we own it anyway? Ideally both in the mv_ version and the original so when we eventually switch away from vendored we don't break on it.

tylercloke:
(Hopefully) final [build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/784/downstreambuildview/).

stevendanna:
cc @chef/lob @chef/server-team This seems like something we want for the next minor version.

marcparadise:
:+1: :shipit:!

tylercloke:
Final build http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/800/

src/chef-mover/src/mover_server_admins_existing_users_read_permissions_callback.erl
```diff
@@ -0,0 +1,94 @@
...
+    mover_transient_migration_queue:initialize_queue(?MODULE, get_users()).
+
+migration_action(UserRecord, _AcctInfo) ->
+    Username = UserRecord#user.username,
+    UserAuthzid = UserRecord#user.authz_id,
+    BifrostSuperuserId = mv_oc_chef_authz:superuser_id(),
+    ServerAdminsAuthzId = get_server_admins_authz_id(),
+    add_permission_to_existing_user_for_server_admins(BifrostSuperuserId, Username, ServerAdminsAuthzId, UserAuthzid, read),
+    add_permission_to_existing_user_for_server_admins(BifrostSuperuserId, Username, ServerAdminsAuthzId, UserAuthzid, update),
+    add_permission_to_existing_user_for_server_admins(BifrostSuperuserId, Username, ServerAdminsAuthzId, UserAuthzid, delete).
```
manderson26: Do we want users to be able to delete the group? I know we don't actually implement a global groups API yet, but when we do it seems like a recipe for breakage.

src/chef-server-bootstrap/bin/bootstrap-platform
```diff
@@ -85,7 +93,26 @@ class BootstrapPlatform
...
+    grant_authz_object_permission("update", "groups", "containers", global_container_authz_ids['users'], server_admins_authz_id, superuser_authz_id, bifrost_host, bifrost_port)
+    grant_authz_object_permission("delete", "groups", "containers", global_container_authz_ids['users'], server_admins_authz_id, superuser_authz_id, bifrost_host, bifrost_port)
+
+    ###########################################
+    # put pivotal in site-admins global group #
+    ###########################################
+    grant_authz_object_permission("create", "actors", "groups", server_admins_authz_id, superuser_authz_id, bifrost_superuser_id, bifrost_host, bifrost_port)
+    grant_authz_object_permission("read",   "actors", "groups", server_admins_authz_id, superuser_authz_id, bifrost_superuser_id, bifrost_host, bifrost_port)
+    grant_authz_object_permission("update", "actors", "groups", server_admins_authz_id, superuser_authz_id, bifrost_superuser_id, bifrost_host, bifrost_port)
+    grant_authz_object_permission("delete", "actors", "groups", server_admins_authz_id, superuser_authz_id, bifrost_superuser_id, bifrost_host, bifrost_port)
```
manderson26: In what circumstances would we want to allow deletion of this group? Is giving pivotal delete a footgun? Not as risky as users above, but still a thing.

omnibus/config/projects/chef-server.rb
```diff
@@ -46,6 +46,7 @@
 dependency "openresty-lpeg"  # lua-based routing
 dependency "runit"
 dependency "chef_backup-gem" # chef-server-ctl backup
+dependency "rest-client-gem" # chef-server-ctl server-admins commands
```
sdelano: This is a larger problem than just this PR, but I want to get the idea out there. What needs the `rest-client` gem, and why doesn't it fetch it for itself? It feels like an anti-pattern to me the more that we specify the gem dependencies of tools here in top level of the project. At the very minimum, that should be `chef-server-ctl.rb` that defines the deps, and eventually have a Gemfile / Gemfile.lock there to handle this.

tylercloke: So...all the _acl endpoints are not written yet, which is why the server-admins commands need `rest-client`. Proposed version 2 of server-admins will add that endpoint in, update the CLI, and remove that dependency (much like the iterative development we did for key rotation). This is the best way to do it iteratively imo. For the whole plan, see here:

https://gist.github.com/tylercloke/a8d4bc1b915b958ac160#version-2

But yes, I can move it into `chef-server-ctl.rb` at the very least.

src/chef-server-bootstrap/bin/bootstrap-platform
tylercloke: Yeah delete on the global-group itself is probably unnecessary / a footgun. Can remove.

src/chef-mover/src/mover_server_admins_global_group_callback.erl
```diff
@@ -0,0 +1,194 @@
...
+            throw(AuthzError)
+    end,
+    
+    case create_server_admins_global_group(ServerAdminsAuthzId, GlobalOrgId) of
+        {chef_sql, {GroupError, _}} ->
+	    lager:error("Could not create new erchef group for server-admins."),
+            throw(GroupError);
+	_ -> true
+    end,
+
```
manderson26: The default permissions are empty on a group created by the superuser. We should add the group to its own read ACE so that members can see who is in it.

manderson26:
Updated permissions look good :+1:

tylercloke:
Once more, with feeling:

http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/801/downstreambuildview/

PR #477, opened 2015-08-17
Add basic testing for the PrivateChef configuration library

stevendanna:
This configuration library is about to undergo a number of changes to
facilitate more modular configuration.  These tests are not complete
but hopefully give us a starting place to help reduce the number of
working configurations we break.


sdelano:
Let's hook this into travis / jenkins

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/private_chef_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+    "superuser_id": "b293108b4a58a346758c091a8ca7462a",
+    "sql_password": "dfaac052d330cb64aa82dd493bd342e93583cb61823fb3ed7792f21f928f9ef0b85d88c995b02ce216d193e9138217f8a5ed",
+    "sql_ro_password": "f3003275a303dccc3e8b93bcde838e2254a82e1c6feb0db034e5e64c263e55643e14f8f75003c7080f7a145328d8a26c8242"
+  },
+  "bookshelf": {
+    "access_key_id": "331fe88b6a86c4801218fd3e831a68b710544069",
+    "secret_access_key": "393dd9330f834102f3650a6ac6938530ccbfbe1c86cb7d732f9893768e4e06eb172cc326da17f435"
+  }}
+EOF
+  }
```
tylercloke: I'm fine with this being in here, but if it gets any longer we should consider loading it from an examples file.

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/private_chef_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    it "disables backend services on the frontend" do
+      expect_existing_secrets
+      rendered_config = config_for("frontend.chef.io")
+      expect(rendered_config["private_chef"]["postgresql"]["enable"]).to eq(false)
+      expect(rendered_config["private_chef"]["bookshelf"]["enable"]).to eq(false)
+      expect(rendered_config["private_chef"]["redis_lb"]["enable"]).to eq(false)
+      expect(rendered_config["private_chef"]["rabbitmq"]["enable"]).to eq(false)
+      expect(rendered_config["private_chef"]["opscode-solr4"]["enable"]).to eq(false)
+    end
```
tylercloke: Its generally considered bad Rspec practice to have more than one assert per `it` block, but I could see how breaking these all out into their own in this instance would be annoying :) Could probably loop or something, but its not a big deal.

tylercloke:
:+1: Two super nitpicky comments that can easily be ignored. Maybe I'll look into getting these running in travis...

stevendanna:
@tylercloke Thanks for hooking this into travis.  I merged as-i.  I agree we should break up some of those it blocks but am hoping to do that as we go down the road of improving this class.

PR #478, opened 2015-08-17
Add production section to newrelic config

jcreedcmu:
Disables new relic in production in on-prem installations.

### DIFF
--- a/omnibus/RELEASE_NOTES.md
+++ b/omnibus/RELEASE_NOTES.md
@@ -1,5 +1,11 @@
 # Chef Server Release Notes
 
+## 12.1.3 (Unreleased)
+
+Take extra steps to ensure that unless newrelic support is expressly
+configured, the service will not call home to newrelic. This fixes a
+problem that affects 12.1.0, 12.1.1, 12.1.2.
+
 ## 12.1.2 (2015-07-16)
 
 Fix issue where chef-server-ctl install could not fetch remote packages via apt.
--- a/src/oc-id/config/newrelic.yml
+++ b/src/oc-id/config/newrelic.yml
@@ -2,3 +2,5 @@ development:
   app_name: oc-id (Development)
   monitor_mode: false
   developer_mode: true
+production:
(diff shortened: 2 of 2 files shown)


jcreedcmu:
See also https://github.com/chef/chef-manage/pull/734

stevendanna:
Please update RELEASE_NOTES.md with a note about this and the versions impacted.  The versions appear to be 12.1.0, 12.1.1, 12.1.2

jcreedcmu:
To the best of my understanding, based on testing in vms, prior to this change, the newrelic gem still attempts to contact the newrelic servers (in a misconfigured state where it has no license key, so no real data is sent, but still) when in "production mode", which will happen on on-premises installs.

I'd really like to have @raskchanky's input on this when he gets back from leave next week.

jcreedcmu:
@smith do you have any information about this? I think the main question I have is "should I be surprised  that it seemed necessary to add a production section at all", i.e. was it just an oversight on @raskchanky's part, or is there something else about this setup that I'm not understading?

smith:
[Newrelic's config docs](https://docs.newrelic.com/docs/agents/ruby-agent/installation-configuration/ruby-agent-configuration#Options) say:

> The Ruby agent's **newrelic.yml** is a standard yaml configuration file. It includes a defaults section at the top, plus sections below for each environment; for example, Development, Testing, and Production.

And the default for monitor mode says it's a Boolean and the default is "(Dynamic)", whatever that means, but in our case it must have meant `true`.

So since there was no production section, it just used the default I guess. Not the behavior I would expect, but I suppose it makes sense.

From looking at https://github.com/chef/chef-server/blob/e26feb3f377a03c5f404845942755b6fd5fb1850/omnibus/files/private-chef-cookbooks/private-chef/recipes/oc_id.rb, it doesn't look like we ever overwrite the newrelic.yml, so it looks like we don't currently support using your own Newrelic accound for on-premises oc-id, but I think we write out the file in the cookbook we use for hosted.

:+1:

PR #479, opened 2015-08-17
Now that luarocks is whitelisted, uncomment full matrix.

tylercloke:
We had a bunch of stuff commented out since luarocks was not whitelisted,
but it was via SSD's ticket:

https://github.com/travis-ci/travis-ci/issues/3716

### DIFF
--- a/src/bookshelf/Makefile
+++ b/src/bookshelf/Makefile
@@ -25,6 +25,9 @@ all: $(REBAR3)
 rel: all
 	@$(REBAR3) release
 
+update: $(REBAR3)
+	@$(REBAR3) update
+
 $(REBAR3):
 	curl -Lo rebar3 $(REBAR3_URL) || wget $(REBAR3_URL)
 	chmod a+x rebar3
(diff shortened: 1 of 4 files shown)


src/bookshelf/Makefile
```diff
@@ -25,6 +25,9 @@ all: $(REBAR3)
 rel: all
 	@$(REBAR3) release
 
+update: $(REBAR3)
```
marcparadise: Were these  pulled in from concrete upstream?   If not, it'll be overwritten on the next concrete update (same for the other Makefile changes)

marcparadise:
Other than confirming the origination of the Makefile changes, LGTM

src/bookshelf/Makefile
tylercloke: Nope, no idea how new tooling works.

joedevivo: CONCRETE IS DEAD! LONG LIVE CONCRETE!

marcparadise:
I've since learned that we have removed concrete from our pipeline, merge when ready :+1:

PR #481, opened 2015-08-18
Add PrivateChef#register_extension method for plugin configuration

stevendanna:
The new register_extension method allows plugins to

- define new top-level configuration entries with defaults

- provide callbacks that will get called at critical points in the
  configuration process

The current callbacks include:

  - gen_secrets
  - gen_backend
  - gen_frontnend
  - gen_api_fqdn

If a callback is not provided, the default implementation is used.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -95,6 +95,8 @@ module PrivateChef
   opscode_org_creator Mash.new
   opscode_certificate Mash.new
 
+  registered_extensions Mash.new
+
   class << self
 
     def from_file(filename)
@@ -115,6 +117,27 @@ module PrivateChef
         if PrivateChef[old_service_key].has_key? configkey
           PrivateChef[new_service_key][configkey] ||= PrivateChef[old_service_key][configkey]
           PrivateChef[old_service_key].delete configkey
+        end
+      end
+    end
+
+    VALID_EXTENSION_CONFIGS= %i(server_config_required config_values gen_backend gen_frontend gen_secrets gen_api_fqdn) unless defined?(VALID_EXTENSION_CONFIGS)
+    def register_extension(name, extension)
+      bad_keys = extension.keys - VALID_EXTENSION_CONFIGS
+
+      bad_keys.each do |key|
+        Chef::Log.warn("Extension #{name} contains unknown configuration option: #{key}")
+        extension.delete(key) # delete mutates extension
+      end
+
+      PrivateChef["registered_extensions"][name] = extension
+      if extension.key?(:config_values)
+        extension[:config_values].each do |k, v|
+          # TODO(ssd): Should we even allow this? Perhaps we should just exit?
+          if PrivateChef.has_key?(k)
+            Chef::Log.warn("Extension #{name} attempted to register configuration default for #{k}, but it already exists!")
+          end
+          PrivateChef.send(k, v)
         end
       end
     end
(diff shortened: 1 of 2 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -152,106 +176,9 @@ def generate_hex_if_bootstrap(chars, ha_guard)
       SecureRandom.hex(chars)
     end
 
-    def generate_secrets(node_name)
```
stevendanna: This is just getting moved to `gen_secrets_default` at line 414 below.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -284,8 +211,9 @@ def generate_hash
 
         # keys for cleanup and back-compat
         "couchdb",
-        "opscode_solr"
-      ].each do |key|
+        "opscode_solr"]
+      keys_from_extensions = PrivateChef["registered_extensions"].map {|name, ext| ext[:config_values].keys.map {|k| k.to_s} if ext[:config_values]}.flatten.compact
```
stevendanna: This craziness should be moved into a function.

marcparadise:
I'll need to take a longer read through this later, but a couple of things that I saw initially: 

* I think we'll want to support mutliple plugins with cumulative effects, rather than just a single plugin callback for a given function? 
*  I'd envisioned optionally allowing the standard configuration to continue - based on the callback return value, so that we don't have to duplicate the things we do want. In many cases we will want a lot of the same things we have now particularly for front ends in a cluster config

stevendanna:
> I think we'll want to support mutliple plugins with cumulative effects, rather than just a single plugin callback for a given function?

Interesting, are you thinking people might mix and match plugins?  My only hesitation here is how we do this without making it debugging hell.  These callbacks basically are mutate a bunch of global state, so it is super helpful to  know what mutations are going to happen before and after you get called.

> I'd envisioned optionally allowing the standard configuration to continue - based on the callback return value, so that we don't have to duplicate the things we do want. In many cases we will want a lot of the same things we have now particularly for front ends in a cluster config

I can see doing this but I kinda want to see what the AWS plugin ends up looking like.  Because the proc's are instance_exec'd they actually have access to any functions in the PrivateChef class, so I was thinking we would break up the monster functions we have now into smaller utility functions and then plugins could call them when they just need to do something standard.  Right now they can call the gen_*_default functions directly if they just want to build on top of them.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -110,6 +112,28 @@ def from_file(filename)
...
+      end
+
+      PrivateChef["registered_extensions"][name] = extension
+      if extension.key?(:config_values)
+        extension[:config_values].each do |k, v|
+          # TODO(ssd): Should we even allow this? Perhaps we should just exit?
+          if PrivateChef.has_key?(k)
+            Chef::Log.warn("Extension #{name} attempted to register configuration default for #{k}, but it already exists!")
+          end
+          PrivateChef.send(k, v)
```
marcparadise: Could we instead use `PrivateChef[k] = v`? I had to think this through for a few seconds and reember what PrivateChef inherited from to realize what this would do. 

We also recently ran into an issue where we had a function sharing a name with a key, and it made serverspec very sad.  Maybe nest it one level under PrivateChef['config_extensions'][k] to avoid trouble?

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -413,12 +416,117 @@ def gen_frontend
...
+
+    def gen_secrets_default(node_name)
+      existing_secrets ||= Hash.new
+      if File.exists?("/etc/opscode/private-chef-secrets.json")
+        existing_secrets = Chef::JSONCompat.from_json(File.read("/etc/opscode/private-chef-secrets.json"))
+      end
+      existing_secrets.each do |k, v|
+        v.each do |pk, p|
+          if not PrivateChef[k]
+            Chef::Log.info("Ignoring unused secret for #{k}.")
```
marcparadise: Maybe split out the existing_secrets load to its own function, or do it before the callback and pass it in since a plugin will likely need to use it as well?

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -495,8 +603,48 @@ def gen_ldap
...
+    def assert_server_config(node_name)
+      unless PrivateChef["servers"].key?(node_name)
+        Chef::Log.fatal <<-EOF
+No server configuration found for "#{node_name}" in /etc/opscode/chef-server.rb.
+Server configuration exists for the following hostnames:
+
+  #{PrivateChef["servers"].keys.sort.join("\n  ")}
+
+EOF
+        exit(1)
```
marcparadise: exit! 1 to prevent a stack trace, at leas until we add our own hooks.  Maybe move this to a preflight check? (Or just note to do so in the future...)

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/private_chef_spec.rb
```diff
@@ -196,4 +196,178 @@ def config_for(hostname)
...
+      PrivateChef.register_extension("different", config_values: {
+                                       new_value: "a different default" })
+    end
+
+    it "warns if the extension contains un unkown configuration option" do
+      expect(Chef::Log).to receive(:warn).with("Extension #{name} contains unknown configuration option: not_here")
+      PrivateChef.register_extension(name, not_here: "a default")
+    end
+
+    it "does not save any unkown keys" do
```
marcparadise: 'unknown'

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -372,7 +306,76 @@ def gen_keepalived(node_name)
...
+      callback = callback_for(:gen_api_fqdn)
+      if ! callback.nil?
+        instance_exec(&callback)
+      else
+        gen_api_fqdn_default
+      end
+    end
+
+    def callback_for(name)
+      extension = PrivateChef["registered_extensions"][PrivateChef["topology"]]
```
marcparadise: Are we sure that extensions will only apply to topologies?

marcparadise:
> Interesting, are you thinking people might mix and match plugins? My only hesitation here is how we do this without making it debugging hell. 

Fair point, I'm okay with this for now. Unless we want to start annotating node attributes with their sources.  Reminds me, isn't there an RFC out there for doing that in chef-client...

marcparadise:
> I can see doing this but I kinda want to see what the AWS plugin ends up looking like. Because the 
>  proc's are instance_exec'd they actually have access to any functions in the PrivateChef class, so I 
> was thinking we would break up the monster functions we have now into smaller utility functions and 
> then plugins could call them when they just need to do something standard. Right now they can call the 
> gen_*_default functions directly if they just want to build on top of them.

Makes sense.  The only one of immediate concern is gen_secrets - I could see plugins that generally want to keep the default stuff, but add or modify a little for their own needs. But the default stuff explicitly rejects the things it does not recognize.

marcparadise:
Some new feedback added, but I don't see need for any major changes.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
stevendanna: I think the answer to this goes hand in hand with whether we end up with plugins for anything other than ha topologies.   I could see us eventually allowing add-ons such as manage to share the chef-server configuration file, but in that case, a lot of this is going to need to change.

tylercloke:
Hate to even bring this up, but the `chef-server.rb` documentation is one of the weaker parts of our docs that James has been wanting help with for forever.

PR #483, opened 2015-08-19
Implement DELETE for Policyfile types

danielsdeleo:
:construction: This only implements some of the required APIs :construction: 

Implements DELETE for policies, policy groups and policy revisions, along with some related GET APIs that will be used client-side for backup/undo functionality.

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policy_group_policy_rev.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policy_group_policy_rev.erl
@@ -4,7 +4,7 @@
 %% @author Jean Rouge <jean@chef.io>
 %% Copyright 2013-2015 Chef Software, Inc. All Rights Reserved.
 
--module(oc_chef_wm_named_policy).
+-module(oc_chef_wm_policy_group_policy_rev).
 
 -include("oc_chef_wm.hrl").
 
--- a/oc-chef-pedant/spec/api/policies/complete_endpoint_spec.rb
+++ b/oc-chef-pedant/spec/api/policies/complete_endpoint_spec.rb
@@ -396,7 +396,166 @@ describe "Policies API endpoint", :policies do
       end
 
     end
-  end
+  end # describe "Listing policies and policy groups"
+
+  describe "Named Policy Groups endpoint /policy_groups/:policy_group_name" do
+
+    let(:named_policy_group_url) { api_url("/policy_groups/example-policy-group") }
+
+    context "when the named group doesn't exist" do
+
+      it "GET /policy_groups/:policy_group_name returns a 404" do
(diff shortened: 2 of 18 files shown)


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policy_group_policy_rev.erl
```diff
@@ -0,0 +1,297 @@
...
+    stash_permissions_objects_authz_ids(Rest, UpdatedBaseState);
+stash_permissions_objects_authz_ids([{policy,AzID}|Rest], #base_state{resource_state = PolicyState} = State) ->
+    UpdatedPolicyState = PolicyState#policy_state{policy_authz_id = AzID},
+    UpdatedBaseState = State#base_state{resource_state = UpdatedPolicyState},
+    stash_permissions_objects_authz_ids(Rest, UpdatedBaseState).
+
+cleanup_orphaned_authz_ids(false, _RequestorId, _AuthzID) ->
+    ok;
+cleanup_orphaned_authz_ids(true, RequestorId, AuthzId) ->
+    oc_chef_authz:delete_resource(RequestorId, object, AuthzId),
```
marcparadise: Should we force the requestor as superuser here? If they ran into a permissions issue creating/updating things, it's possible that they won't be able to delete the IDs. 

Or are these IDs those that would have been generated in this request anyway?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policy_group_policy_rev.erl
```diff
@@ -0,0 +1,297 @@
...
+    {{halt, 500}, Req, State#base_state{log_msg = ERROR}}.
+
+handle_update_result(ok, PolicyName, PolicyData, Req, State) ->
+    LogMsg = {updated, PolicyName},
+    ReqWithBody = chef_wm_util:set_json_body(Req, PolicyData),
+    {true, ReqWithBody, State#base_state{log_msg = LogMsg}};
+handle_update_result(ERROR, _PolicyName, _PolicyData, Req, State) ->
+    {{halt, 500}, Req, State#base_state{log_msg = ERROR}}.
+
+
```
marcparadise: extra blank lines here and below

marcparadise:
I just had the couple of minor comments above.  I'm :+1: on this.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_policy_group_policy_rev.erl
danielsdeleo: Yup, these are generated in this request. This endpoint does create or update against two container/object types, so you might have permission for one and not the other. This gets called in `forbidden` (not using the one from `wm_base`) to clean up before returning the 403.

stevendanna:
:+1:

PR #484, opened 2015-08-19
Purge symlinks before (re)loading erlang projects

danielsdeleo:
Fixes an issue where `dvm load` could corrupt the omnibus-managed
`sys.config` with a bad one created by relx.

@chef/lob


marcparadise:
:+1:

PR #485, opened 2015-08-19
Remove yanked celluloid from Gemfile.lock

danielsdeleo:
I noticed in travis tests for a different branch that the omnibus tests failed to `bundle install`, which was because celluloid 0.16.1 was yanked, though it still appeared in our Gemfile.lock. This updates the bundle to use the previous version instead.

@chef/lob


marcparadise:
:+1: yes please

PR #486, opened 2015-08-19
chef-server-ctl external service support

marcparadise:
The commit messages have full details., but TL;DR: 

Adds external service support in chef-server-ctl.  Adds generalized behavior to support  'status', 'service-list', 'cleanse' for external services.  Disables other runsv commands used specifically against services flagged as external.  Adds external postgresql support for the  commands above. 

gather-logs has been updated to use chef-server-ctl psql for chef server postgres, which already handles remote v local correctly. 

Remaining work: 
* [x] override omnibus-ctl version if we don't change default in omnibus-software
* [x] {Possibly} - move some of the helper functions that determine internal v external services to the base Omibus::Ctl class -- but that only makes sense if we will be standardizing on this method of external service handling. 

ping @chef/lob

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/chef-server-ctl
+++ b/omnibus/files/private-chef-ctl-commands/chef-server-ctl
@@ -1,0 +1,191 @@
+#!/opt/opscode/embedded/bin/ruby
+require 'rubygems'
+gem 'omnibus-ctl'
+require 'omnibus-ctl'
+
+module Omnibus
+  # This implements callbacks for handling commands related to the
+  # external services supported by Chef Server. As additional services
+  # are external-enabled, controls/configuration will need to be added here.
+  class ChefServerCtl < Omnibus::Ctl
+    # Note that as we expand our external service support,
+    # we may want to consider farming external_X functions to
+    # service-specific classes
+    def external_cleanse_postgresql(perform_delete)
+      postgres = external_services['postgresql']
+      exec_list = []
+      # NOTE a better way to handle this particular action may be through a chef_run of a 'cleanse' recipe,
+      # since here we're explicitly reversing the actions we did in-recipe to set the DBs up....
+      superuser = postgres['db_superuser']
+      [%w{oc_bifrost, bifrost},
+       %w{opscode-erchef, opscode_chef},
+       %w{oc_id, oc_id}].each do |service|
+        key, dbname = service
+        exec_list << "DROP DATABASE #{dbname}"
+        exec_list << "REVOKE \"#{running_package_config[key]['sql_user']}\" FROM \"#{superuser}\""
+        exec_list << "REVOKE \"#{running_package_config[key]['sql_ro_user']}\" FROM \"#{superuser}\""
+        exec_list << "DROP ROLE \"#{running_package_config[key]['sql_user']}\""
+        exec_list << "DROP ROLE \"#{running_package_config[key]['sql_ro_user']}\""
+      end
+      if perform_delete
+        delete_external_postgresql_data(exec_list, postgres)
+      else
+        path = File.join(backup_dir, 'postgresql-manual-cleanup.sql')
+        dump_exec_list_to_file(path, exec_list)
+        log warn_CLEANSE002_no_postgres_delete(path, postgres)
+      end
+    end
+
+    def delete_external_postgresql_data(exec_list, postgres)
+      last = nil
+      begin
+        require "pg"
+        connection = postgresql_connection(postgres)
+        while exec_list.length > 0
+          last = exec_list.shift
+          connection.exec(last)
+        end
+        puts "Chef Server databases and roles have been successsfully deleted from #{postgres['vip']}"
+      rescue StandardError => e
+        exec_list.insert(0, last) unless last.nil?
+        if exec_list.empty?
+          path = nil
+        else
+          path = Time.now.strftime('/root/%FT%R-chef-server-manual-postgresql-cleanup.sql')
+          dump_exec_list_to_file(path, exec_list)
+        end
+        # Note - we're just showing the error and not exiting, so that
+        # we don't block any other external cleanup that can happen
+        # independent of postgresql.
+        log err_CLEANSE001_postgres_failed(postgres, path, last, e.message)
+      ensure
+        connection.close if connection
+      end
+    end
+
+    def dump_exec_list_to_file(path, list)
+      File.open(path, 'w') do |file|
+        list.each { |line| file.puts line }
+      end
+    end
+
+    def external_status_postgresql(detail_level)
+      postgres = external_services['postgresql']
+      begin
+        require 'pg'
+        connection =postgresql_connection(postgres)
+        if detail_level == :sparse
+          # We connected, that's all we care about for sparse status
+          # We're going to keep the format similar to the existing output
+          "run: postgresql: connected OK to #{postgres['vip']}:#{postgres['port']}"
+          # to hopefully avoid breaking anyone who parses this.
+        else
+          postgres_verbose_status(postgres, connetion)
+        end
+      rescue StandardError => e
+        if detail_level == :sparse
+          "down: postgresql: failed to connect to #{postgres['vip']}:#{postgres['port']}: #{e.message.split("\n")[0]}"
+        else
+          log err_STAT001_postgres_failed(postgres, e.message)
+          Kernel.exit! 128
+        end
+      ensure
+        connection.close if connection
+      end
+    end
+
+    def postgresql_connection(postgres)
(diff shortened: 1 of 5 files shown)


marcparadise:
Depends on https://github.com/chef/omnibus-ctl/pull/36

marcparadise:
Note: I wanted to get this out there for feedback, but work on correcting tests is still in progress.

omnibus/files/private-chef-scripts/gather-logs
```diff
@@ -24,30 +24,15 @@ fi
...
 
     if [[ ! -e "/opt/$path/bin/$ctl_cmd" ]];
     then
-        echo "ERROR: Enterprise or Chef Server may not be installed."
+        echo "ERROR: Chef Server may not be installed."
         exit 1
     fi
-
-    echo "Gathering logs from Enterprise or Chef Server."
-elif [[ $type == 'OSC' ]];
```
stevendanna: Are we sure the support team doesn't still need this?  I believe they might instruct users to download this file from the repository if they happen to be running a version that of Chef 11 that didn't ship with this.

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+gem 'omnibus-ctl'
+require "omnibus-ctl"
+
+module Omnibus
+
+  # We keep the base behavior of Omnibus::Ctl, but
+  # add our own hooks for before/after command run.
+  # In particular we hook the service calls.
+  class ChefServerCtl < Omnibus::Ctl
+    def command_post_hook(*args)
```
stevendanna: Why isn't this the default implementation of post_host?  It seems like this is the behavior we might want generally?

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+      if respond_to?(method)
+        send(method, *args)
+      end
+    end
+
+    # If a command has a pre-hook defined we will run it.
+    # Otherwise, if it is a run-sv command and the service it refers to
+    # is an external service, we will show an error since we
+    # can't control external services from here.
+    def command_pre_hook(*args)
```
stevendanna: Similarly, it seems like likes 30-34 should be the default implementation of pre_hook.  The implementation could then perhaps just look at the return value from `super`.

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+            log "Deleting data from external service: #{service_name}"
+          end
+          send(to_method_name("external_cleanse_#{service_name}"), perform_delete)
+        end
+    end
+
+    # Add some output headers if we have external services enabled
+    def service_list_pre_hook
+      log_internal_service_header
+      return true
```
stevendanna: Unnecessary `return`

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+          log status
+        end
+      end
+    end
+
+    # Data cleanup requirements for external services aren't met by the standard
+    # 'nuke /var/opt' behavior - this hook allows each service to perform its own
+    # 'cleanse' operations.
+    def cleanse_post_hook(*args)
+        external_services.each_key do |service_name|
```
stevendanna: Extra spaces here.

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+          perform_delete = ARGV.include?("--with-remote")
```
stevendanna: Minor point, but you can move this outside of the loop to avoid doing the look up on every iteration.

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+      File.open(path, "w") do |file|
+        list.each { |line| file.puts line }
+      end
+    end
+
+    def external_status_postgresql(detail_level)
+      postgres = external_services['postgresql']
+      begin
+        require "pg"
+        connection =postgresql_connection(postgres)
```
stevendanna: Space after = here.

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+          # We're going to keep the format similar to the existing output
+          "run: postgresql: connected OK to #{postgres['vip']}:#{postgres['port']}"
+          # to hopefully avoid breaking anyone who parses this.
+        else
+          max_conn = connection.exec("SELECT setting FROM pg_settings WHERE name = 'max_connections'")[0]['setting']
+          total_conn = connection.exec('SELECT sum(numbackends) num FROM pg_stat_database')[0]['num']
+          version = connection.exec('SHOW server_version')[0]['server_version']
+          lock_result =  connection.exec('SELECT pid FROM pg_locks WHERE NOT GRANTED')
+          locks = lock_result.map { |r| r['pid'] }.join(",")
+          if locks.nil? or locks.empty?
```
stevendanna: I think our general ruby style would use || rather than `or` here.  Similarly in a few other conditionals.

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+      ensure
+        connection.close if connection
+      end
+    end
+
+    def postgresql_connection(postgres)
+        PGconn.open('user' => postgres['db_superuser'], 'host' => postgres['vip'],
+                    'password' => postgres['db_superuser_password'], 'port' => postgres['port'], 'dbname' => 'template1')
+    end
+    def external_services
```
stevendanna: newline before this function definition

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+CLEANSE001: While local cleanse of Chef Server succeeded, an error occurred while deleting
+            Chef Server data from the external PostgreSQL server at #{postgres['vip']}.
+
+            The error reported was:
+
+#{format_multiline_message(16, message)}
+
+EOM
+      msg << <<-EOM unless last.nil?
+            This occurred when executing the following SQL:
```
stevendanna: "the following SQL" -> "the following SQL statement"

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+              #{last}
+EOM
+
+      msg << <<-EOM unless path.nil?
+            Please log into that server as superuser and execute the remaining
```
stevendanna: "that server" -> "the PostgreSQL server".

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+
+            #{sql_path}
+
+            See https://docs.chef.io/error_messages.html#cleanse002-postgres-not-purged
+            for more information.
+EOM
+    end
+
+    def format_multiline_message(indent, message)
+      if message.class == String
```
stevendanna: To avoid the explicit type check, perhaps use `if message.respond_to?(:split)`

omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -0,0 +1,328 @@
...
+        message = message.split("\n")
+      end
+      spaces = " "*indent
+      message.map!{|line| "#{spaces}#{line.strip}"}
+      message.join("\n")
+    end
+    def log_internal_service_header
```
stevendanna: newline before this function definition

stevendanna:
@marcparadise I've left a number of comments here and on the linked omnibus-ctl PR.  Most are minor style issues.  The only outstanding issue I think we need to solve before merging is how the hooks work.

omnibus/files/private-chef-ctl-commands/chef-server-ctl
marcparadise: Agreed, I'll factor this and related hook stuff up into omnibus-ctl.

omnibus/files/private-chef-scripts/gather-logs
marcparadise: Ah, alright. I'll add it back in.

PR #488, opened 2015-08-19
Transition Solr memory and JVM settings from OSC11 to Chef 12.

andy-dufour:


andy-dufour:
Still needs to run through CI and I need to do some more upgrade testing, an initial test worked.

Looking for thoughts on if this is the right approach from LOB before I go wild with CI/manual upgrade tests.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -514,6 +515,20 @@ def generate_config(node_name)
         PrivateChef["default_listen_address"] = "::"
       end
 
+      # Transition Solr memory and JVM settings from OSC11 to Chef 12.
```
stevendanna: I am +1 on doing things like this but it should be moved into a function of its own.  Perhaps named "handle_old_solr_config"

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -514,6 +515,20 @@ def generate_config(node_name)
...
+      end
+      if PrivateChef['opscode_solr'].has_key? 'new_size'
+        PrivateChef['opscode_solr4']['new_size'] ||= PrivateChef['opscode_solr']['new_size']
+        PrivateChef['opscode_solr'].delete 'new_size'
+      end
+      if PrivateChef['opscode_solr'].has_key? 'java_opts'
+        PrivateChef['opscode_solr4']['java_opts'] ||= PrivateChef['opscode_solr']['java_opts']
+        PrivateChef['opscode_solr'].delete 'java_opts'
+      end
+
```
marcparadise: Looks good, but could be simplified slightly: 
```
old_config = PrivateChef['opscode_solr'] || {}
[ 'heap_size', 'new_size', 'java_opts' ].each do |key| 
  if old_config.has_key? key
    PrivateChef['opscode_solr4'][key] = old_config[key]
   old_config.delete key 
  end 
end
```

Note that while this will remove the config entries and prevent them from being saved to the node or chef-server-running.json, they will still live on in chef-server.rb / private-chef.rb

marcparadise:
So one other thought here - are we certain that the old config settings from solr1 will be correct for solr4 in the  default case? A lot has changed.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
marcparadise: One step up: `import_legacy_service_config('old_service_key', 'new_service_key', ['config','keys', 'we', 'care', 'about', 'preserving'])` ?

andy-dufour:
I was thinking of a method for importing old keys. I think it's a good idea.

I think heap and new size are relatively safe, I'll do some discovery on what java_opts look like from 1->4.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
stevendanna: @marcparadise :+1: I like that.

andy-dufour:
Added a pretty simple function and used it for the Solr settings, let me know what you think. I need to clean up some comments before merge if it passes CI.

I also looked into java_opts - I think the positives outweigh the negatives of bringing it over. E.g. if a user has overridden java_opts to tune the GC algorithm or something and also set min/max heap settings we could get them in serious trouble after upgrade by ignoring the JVM settings. The settings most people would set (e.g. -Xmx4G -Xms4G -XX:NewSize=128M -XX:+UseConcMarkSweepGC -XX:+UseParNewGC -XX:ParallelCMSThreads=6 kind of stuff) should move without issue.

marcparadise:
:+1: once the commented out stuff is removed.

marcparadise:
Accidental closing.

stevendanna:
:+1:

PR #490, opened 2015-08-20
Allow configuration extensions to load PrivateChef config extensions

stevendanna:
To ensure that config_extensions (an in-development feature) can be
loaded before the configuration, we have to move plugin discovery to
an earlier part of the run. To make the intended ordering more clear,
I've also moved the current configuration logic into its own recipe.


stevendanna:
@chef/lob A first cut at how we can load configuration extensions.  This basically just gives plugins another places to inject some code.  So far we've been operating on the premise that plugins are more or less "trusted".  If we wanted to handle untrusted plugins, this would need a lot more work.

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
```diff
@@ -0,0 +1,82 @@
...
+  # Take care to check that ['previous_run'] exists, though, otherwise
+  # you'll run into trouble doing the initial Chef run.
+  #
+  # TODO: Provide an API for getting this information
+  node.consume_attributes({"previous_run" => old_config['private_chef']})
+end
+
+
+if File.exists?("/etc/opscode/chef-server.json")
+  Chef::Log.warn("Please move to /etc/opscode/chef-server.rb for configuration - /etc/opscode/chef-server.json is deprecated.")
```
marcparadise: I know this is existing, but isn't actually not supported at all anymore? Might be if chef-server.json is present but chef-server.rb/private-chef.rb are not, we should halt?

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
```diff
@@ -0,0 +1,82 @@
+# Copyright:: Copyright (c) 2015 Chef Software, Inc
```
marcparadise: :+1: to having this in its own file , but I'm wondering if we shouldn't follow a convention like `_config.rb` to denote recipes that don't directly change the node?

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
marcparadise: Nevermind, I see it is used.  Still, when did we deprecate it and when can we get rid of it...

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
marcparadise: Same for the plugin_* recipes, to _plugin*.

omnibus/files/private-chef-cookbooks/private-chef/recipes/plugins_discovery.rb
```diff
@@ -0,0 +1,35 @@
...
+
+node.default['available-plugins'] = EnterprisePluginCollection.from_glob("/var/opt/opscode/plugins/*.rb")
+
+#
+# Load plugins from backup locations if we didn't find anything
+# in the plugin directory.
+#
+if node.default['available-plugins'].empty?
+  c = EnterprisePluginCollection.new()
+  BACKUP_PLUGIN_LOCATIONS.each do |path|
```
marcparadise: do we want to prescribe specific plugin dir names? Maybe a `find /opt -type f -name chef-server-plugin.rb`?

marcparadise:
LGTM pending your thoughts on the recipe name changes.

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
stevendanna: @marcparadise Actually, where do you see this being moved?  From what I can see, if you have this file, pretty much nothing will work right since we skip the rest of the configuration.  Perhaps we should exit! here and make the user fix it?

omnibus/files/private-chef-cookbooks/private-chef/recipes/plugins_discovery.rb
stevendanna: My only concern with this is that we assume all the code in those files is trusted so I wanted to limit where we loaded them from.  If we did move to looking for anything with this name, we could likely just another from_glob and avoid some of this.

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
marcparadise: Yeah, if loaded [here] (https://github.com/chef/chef-server/pull/490/files#diff-6e926ec7ad5c51dcb00ecef3c509cf30R37) pretty much everything is gonna break. :+1: to making them fix it.

omnibus/files/private-chef-cookbooks/private-chef/recipes/plugins_discovery.rb
marcparadise: Makes sense re: considering those trusted locations.

PR #493, opened 2015-08-21
Adjust requestor selection to reflect actual query ordering

jcreedcmu:
This is to fix https://getchef.zendesk.com/agent/tickets/5374.

The comments appear to reflect an old state of affairs where clients were returned ahead of users. In fact users are sorted first by the ordering on `type DESC`, which means that if the web-request case wants to prioritize getting a user, it should _not_ reverse the list of requestors it gets after all.

I think a pedant test for this would fit naturally in `oc-chef-pedant/spec/api/keys/keys_spec.rb` in the existing section where a user and client share a name, but I don't yet know how to add `'X-Ops-Request-Source' => 'web'` to the ambient auth headers.

### DIFF
--- a/oc-chef-pedant/spec/api/keys/keys_spec.rb
+++ b/oc-chef-pedant/spec/api/keys/keys_spec.rb
@@ -686,6 +686,40 @@ describe "/keys endpoint", :keys do
       platform.delete_org(other_org_name)
     end
 
+    context "when a client exists with the same name as the user" do
+      let(:colliding_client) do
+        {
+          "name" => org_user_name,
+          "public_key" => keys[:org_client][:public],
+          "private_key" => keys[:org_client][:private]
+        }
+      end
+
+      let(:colliding_client_payload) do
+        { "name" => colliding_client['name'], "public_key" => colliding_client['public_key'],
+          "admin" => "false"
+        }
+      end
+
+      before(:each) do
+        post("#{org_base_url}/clients", superuser, payload: colliding_client_payload)
+        platform.associate_user_with_org($org_name, org_user)
+      end
+
+      after(:each) do
+        delete("#{org_base_url}/clients/#{colliding_client['name']}", superuser)
+        platform.remove_user_from_org($org_name, org_user)
+      end
+
+      it "should allow the user to query /users/:user when user is impersonated via web request'", :focus do
+        # Note that the key used doesn't matter below - by using impersonate we'll be substituting the webui
+        # public key.
+        get("#{org_base_url}/users/#{org_user_name}",
+            impersonate(requestor(org_user_name, colliding_client['private_key'])),
+            headers: {'X-Ops-Request-Source' => 'web'}).should look_like(status: 200)
+      end
+    end
+
     context "posting keys" do
 
       # These behaviors are identical for POSTing to create a client or user key
@@ -1611,4 +1645,3 @@ describe "/keys endpoint", :keys do
     end
   end
 end
-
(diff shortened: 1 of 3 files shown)


jcreedcmu:
cc @marcparadise

marcparadise:
This test should do it, with possible minor tweaking (I haven't run it). Should fail without your changes and pass with them: 
```
    it "should allow user to query /users/:user when that user is impersonated " do
      get("#{platform.server}/users/#{user['name']}", impersonate(user['name']), headers: {'X-Ops-Request-Source' =>  'web'}).should look_like(status: 200)
    end
```

marcparadise:
:+1: on the change post-tests. Ping me here or in slack and I'll get a build kicked off once you're ready.

mmzyk:
Awesome @jcreedcmu - thanks for tracking this all the way through to the cause.

jcreedcmu:
This test still appears to be insufficient. When I run the test *without* the fix in the prior commit, tracing with like
```
dbg:stop_clear().
dbg:start().
dbg:tracer().
dbg:p(all,call).
dbg:tpl(chef_sql, fetch_actors_by_name, 2, [{'_', [], [{return_trace}]}]).
```
shows only one requestor when there ought to be two to exercise the bug.

stevendanna:
@jcreedcmu :+1: on this change overall. The problem with your test is likely the API endpoint you are hitting.  You are hitting the top-level user endpoint so there is no way that the client is going to come back from the query since we aren't passing an org id.  Your test needs to associate the user with the same org as the client and then make a request somewhere in the org's API endpoint.

That was a bit terse, ping me in chat if you need more detail.

jcreedcmu:
@stevendanna yup, came to the same conclusion talking to @marcparadise last friday. I just pushed up a test that does hit the org endpoint, and is properly red before the change and green after, but I think it's making a mess of the conventions in this file ~ is there some better place to situate it to better re-use the test-users and -clients (and associations-of-users-to-orgs) that already exist here? I'm kind of lost among all the afters and befores.

oc-chef-pedant/spec/api/keys/keys_spec.rb
```diff
@@ -686,6 +686,40 @@ def get_client_key(org, client, requestor, key)
...
+        {
+          "name" => org_user_name,
+          "public_key" => keys[:org_client][:public],
+          "private_key" => keys[:org_client][:private]
+        }
+      end
+
+      let(:colliding_client_payload) do
+        { "name" => colliding_client['name'], "public_key" => colliding_client['public_key'],
+          "admin" => "true"
```
stevendanna: The "admin" here here is a bit of a no-op (admin status is now handled via group members), but having it here and set to "true" is a bit confusing. If this client /were/ an admin, your test would pass regardless of which requestor made the call since admins have READ access on users within their orgs.  Which is a long way to say, could we set this to false even though it is a noop?

jcreedcmu:
@stevendanna ok, done. Test still fails/passes before and after server change.

stevendanna:
@jcreedcmu I think the best alternative to what you have here wrt to the test data setup (which I think is fine overall and much clearer than some of the alternatives) would be to add this case (user and client with conflicting name) to the base set of test data set up at the very beginning in our config: 

https://github.com/chef/chef-server/blob/jc/requestor-order/oc-chef-pedant/pedant_config.rb#L139

But, I'd be +1 as is since the current form would be clear to whoever needs to fix a breakage of this test down the road.

jcreedcmu:
Ok, I think I'm going to merge this then, if it's not too objectionable style-wise.

jcreedcmu:
(at least after CI finishes)

PR #496, opened 2015-08-25
Update erlectricity to resolve an encoding issue with multi-byte characters

irvingpop:
Changelog-entry:  Update erlectricity to resolve an encoding issue with multi-byte characters

cc: @sdelano


marcparadise:
:+1:

sdelano:
:+1: - Merge when ready.

PR #498, opened 2015-08-25
Allow standalone-like configurations in plugins

stevendanna:
The previous code assumed that a plugin cares about frontend and backend
nodes. But in cases like AWS where we only have 1 type of node, this is
not the case.


stevendanna:
looking into the test failures

tylercloke:
Looks reasonable to me.

PR #500, opened 2015-08-26
ensure that chef-stable repository is disabled if present to prevent unintended updates

marcparadise:
Testing is locally in flight, but opening this up for review in the mean time. 

ping @chef/lob @mmzyk @sean-horn

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_repository.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_repository.rb
@@ -1,0 +1,48 @@
+#
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+#
+# All Rights Reserved
+#
+
+# Only "enabled false" on yum repos, because
+# the default contents of "/etc/apt/apt.conf.d/50unattended-upgrades"
+# on an Ubuntu system would prevent chef-stable packages from being auto upgraded.
+case node['platform_family']
+when 'debian'
+
+  package "apt-transport-https"
+
+  apt_repository 'chef-stable' do
+    uri "https://packagecloud.io/chef/stable/ubuntu/"
+    key 'https://packagecloud.io/gpg.key'
+    distribution node['private_chef']['addons']['ubuntu_distribution']
+    deb_src true
+    trusted true
+    components %w( main )
+  end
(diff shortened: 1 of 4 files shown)


marcparadise:
force-pushed an amended commit to remove a change we don't need here, and fix a comment

stevendanna:
:+1: Looks good

marcparadise:
Our default `['addons']['packages']` attribute containing a list of all available packages, and the add-ons recipe just loops through and installs everything it sees in `packages`.  This means on reconfigure we'd auto-install all add-ons the first time we did a reconfigure with this recipe included. 

I've modified this PR to split out the addon repo setup into its own recipe and invoke that instead of the base add-on wrapper.

marcparadise:
The failure here appears unrelated. Running it again and will take a look later if it continues to fail.

omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_repository.rb
```diff
@@ -0,0 +1,48 @@
+#
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+#
+# All Rights Reserved
```
stevendanna: This isn't correct anymore, let's fix it.

stevendanna:
:+1: Please fix the copyright notices if you get a chance.

jeremiahsnapp:
I love it @marcparadise . Great fix.

toioiz:
This breaks other items that use chef_ingredient and try to install more than 1 at a time.

Example:
Write chef-stable.repo (enabled)
chef_ingredient "manage"
  -- manage reconfigure
  -- chef-server-ctl reconfigure (disables)
chef_ingredient "reporting" (panic error)
chef_ingredient "push-jobs"

marcparadise:
Well crap.   That seems to rule out any other possible paths to fix this, unless we explicitly enable the resource beforehand , and disable it after.  Ugh.

toioiz:
You can call --enable-repo=chef-stable during the yum command, not sure if the yum resource will allow you to do that inside the chef_ingredient lwrp.

ie. yum --enable-repo=chef-stable install opscode-manage

Note: this may require changes to the yum_package lwrp as well :(

PR #501, opened 2015-08-27
Update jiffy from 0.6.1 to 0.14.1

stevendanna:


stevendanna:
@chef/lob @joedevivo and I noticed that jiffy was pretty far out of date in bifrost.  This brings it up to the same version we ship in erchef and mover.

joedevivo:
:+1: Let's merge this before I start the next thing

PR #506, opened 2015-08-30
not for merge:  chef-server-ctl cluster-join PoC

marcparadise:
It's ugly, but it works.  This shows it's possible without a lot of pain for a node to join an existing cluster without knowing anything about it except its fqdn beforehand. 

Some things that have to be done better; 

* I had to pretend the base chef-server node was a back end in a tiered topology, so that it would auto-configure itself to expose public services on 0.0.0.0   instead of 127.0.0.1.  Once we do the topo and role cleanup (or just support external rabbit..) this won't be needed.  
* The RPC/messaging/tcp stuff is a total hack and is pretty fragile. We'll want to use a state machine that tracks 
* This doesn't address leader handling because I realized it doesn't need to: any FE server can allow any other FE server to come online. There's nothing special that needs doing before that new FE is even configured. 
* nothing is encrypted over the wire at the moment, we're transmitting over tcp secrets, keys, etc to anyone how knows the right way to ask.  I have some ideas here - maybe implement a per-server passphrase to encrypt/decrypt  that the client needs to know about, though that does reduce the simplicity of it slighty.  

ping @chef/lob and  cc/ @manderson26 because I think  you'll find it fun.

### DIFF
--- a/dev/Vagrantfile
+++ b/dev/Vagrantfile
@@ -7,7 +7,7 @@ CS_VM_ADDRESS="192.168.33.100"
 DB_VM_ADDRESS="192.168.33.150"
 DB_SUPERUSER="bofh"
 DB_SUPERPASS="i1uvd3v0ps"
-
+CS_VM_2_ADDRESS="192.168.33.101"
 
 Vagrant.configure("2") do |config|
   attributes = load_settings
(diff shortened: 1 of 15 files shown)


dev/Vagrantfile
```diff
@@ -7,7 +7,7 @@ CS_VM_ADDRESS="192.168.33.100"
 DB_VM_ADDRESS="192.168.33.150"
 DB_SUPERUSER="bofh"
 DB_SUPERPASS="i1uvd3v0ps"
-
+CS_VM_2_ADDRESS="192.168.33.101"
```
marcparadise: By th way, more ugliness ^ :D Simplest way to make dvm work with a second FE, but there are some things we can do much better for dvm if we want to support various topos.

stevendanna:
@marcparadise I think that we should close any PRs that aren't active WIP or that we don't plan on merging.  Given that, thoughts on closing this?  We can keep the branch around if we decide to do something like this.

PR #507, opened 2015-08-31
Alphabetize rebar.lock file

stevendanna:
rebar3 wants to alphabetize the rebar.lock file, who are we to argue?


joedevivo:
Thanks rebar3!

:+1:

PR #509, opened 2015-08-31
Adding suse to package support for local addon installs.

andy-dufour:
We have a couple customers using Suse/SLES even though it's not a supported platform. I'd like them to be able to use chef-server-ctl install addon <package> --path <path> to install their addons.

This is a pretty simple two line change for the local add-on wrapper to not fail when the platform is suse based, and instead use Zypper/RPM to install the packages.

Passed Wilson CI testing: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/701/

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_local.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_local.rb
@@ -8,7 +8,7 @@
 case node['platform_family']
 when 'debian'
   package_suffix = 'deb'
-when 'rhel'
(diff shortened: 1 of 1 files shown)


irvingpop:
:+1: LGTM

marcparadise:
I see this is only for local add ons which makes sense -- the change itself looks fine.  However, I  do have some concerns. Since we don't run any testing for the suse platform we can't make any guarantees of compatibility.  By enabling add-ons for install via chef-server-ctl on those platforms aren't we implicitly saying this is a thing we'll support - both in generally for chef server, and specifically for add ons?

PR #514, opened 2015-09-04
Node policy integration

danielsdeleo:
Integrate node objects and policyfile attributes by updating node create/update to accept `policy_group` and `policy_name`

These things are dependent but will be done later to keep patch size small:
* add endpoint `GET /policies/:policy_name/nodes`
* add endpoint `GET /policy_groups/:policy_group/nodes`

### DIFF
--- a/src/oc_erchef/apps/chef_objects/src/chef_node.erl
+++ b/src/oc_erchef/apps/chef_objects/src/chef_node.erl
@@ -29,6 +29,7 @@
          ejson_for_indexing/2,
          extract_recipes/1,
          extract_roles/1,
+         fields_for_insert/1,
          fields_for_fetch/1,
          fields_for_update/1,
          id/1,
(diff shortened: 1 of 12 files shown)


danielsdeleo:
@chef/lob not complete yet, but could use review on the schema stuff especially.

danielsdeleo:
BTW, Adding the two columns to the nodes table didn't seem to make much difference in testing against preprod: https://gist.github.com/danielsdeleo/53154586743b4dbbe217

src/oc_erchef/apps/chef_objects/src/chef_node.erl
```diff
@@ -76,12 +77,16 @@
...
          ]}).
 
 -define(VALID_KEYS,
         [<<"name">>, <<"chef_environment">>, <<"json_class">>, <<"chef_type">>,
-         <<"normal">>, <<"default">>, <<"override">>, <<"automatic">>, <<"run_list">>]).
+         <<"normal">>, <<"default">>, <<"override">>, <<"automatic">>, <<"run_list">>,
+         <<"policy_name">>, <<"policy_group">> ]).
 
 -define(create_if_missing_fields,
         [{<<"chef_environment">>, <<"_default">>},
```
marcparadise: Do we want to provide defaults for the new fields?

danielsdeleo: I don't think there is a sane default. I was planning on having `null` for these fields be an indication to `chef-client` that it's running in "not policyfiles mode". The only thing I think I could put as a default is something that would be invalid for the user to upload, like `!nothing` but that leads to a lot of ugly elsewhere, I think.

tylercloke:
Looks good!

marcparadise:
:+1:

PR #516, opened 2015-09-08
Set forward_agent = true in Vagrant ssh config

stevendanna:
Occasionally we want to clone private repositories inside the dev-vm.
The forwarded ssh-agent makes this easier since the user can add their
github ssh key to their agent and it will be available inside the VM.


marcparadise:
:+1:  This used to be in there

PR #518, opened 2015-09-08
Lock port compiler to 0.3.0 

stevendanna:
An update to the port compiler on hex caused an incompatibility with the
version of rebar we had vendored. Here we lock the pc plugin to avoid
the problem.

We should try to move to a newer rebar3 and update the port compiler
once rebar3 does a release.

### DIFF
--- a/src/bookshelf/rebar.config
+++ b/src/bookshelf/rebar.config
@@ -54,6 +54,8 @@
     ]}
 ]}.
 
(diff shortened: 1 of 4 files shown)


danielsdeleo:
LGTM, hope travis likes it. :+1:

marcparadise:
:+1:

PR #520, opened 2015-09-09
Support directly posting index updates to Solr

stevendanna:
:warning: See reviewer notes, much of this code is vendored and does not need review. :warning: 

Previously, all index updates were placed on rabbitmq and
processed by opscode-expander. This changeset allows the user to
configurably directly send updates from erchef to solr. The
object expansion code was imported from chef_reindex which is
used in the solr to solr4 migration.

[Running Build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/711/downstreambuildview/)

TODO:
- [x] New build and test csc reindex.
- [x] https://github.com/chef/chef-server/pull/520#discussion_r39109962


stevendanna:
Reviewer Notes:

The C code for fast_xs and related erlang module were directly imported from chef_reindex and likely do not need much review.

Similarly, the object expansion code in chef_index_expand was imported from chef_reindex.  However, the send_item, send_delete, doc_for_index, and doc_for_delete functions are new.

The chef_index_batch module is a gen_server responsible for batching together updates when the queue_mode is set to batch.  The goal of this gen_server is to batch requests when possible but still regularly flush to solr on a none interval regardless of batch size.

The chef_index module is a top level entry for for the chef_index code which handles dispatch into the various indexing modes: rabbitmq (old path), inline (expansion and post done in the webmachine process directly), batch (expansion done in the webmachince process, posting done via the batch gen_server).

omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
```diff
@@ -155,7 +155,19 @@
                 {rabbitmq_password, <<"<%= node['private_chef']['rabbitmq']['password'] %>">>},
                 {rabbitmq_vhost, <<"<%= node['private_chef']['rabbitmq']['vhost'] %>">>},
                 {rabbitmq_exchange, <<"">>},
-                {solr_url, "http://<%= @helper.vip_for_uri('opscode-solr4') %>:<%= node['private_chef']['opscode-solr4']['port'] %>/solr"}
+		{search_provider, <%= node['private_chef']['opscode-erchef']['search_provider'] %>},
```
tylercloke: Tabbing is hard. Not worth re-pushing for unless we find an actual issue though.

src/oc_erchef/apps/chef_index/src/chef_index_sup.erl
```diff
@@ -71,7 +73,10 @@ amqp_child_spec() ->
             error_logger:info_msg("Connecting to Rabbit at ~p:~p ~p~n",
                                   [Host, Port, {VHost, ExchangeName}]),
 
-            [bunnyc_spec(VHost, Host, Port, User, Password, ExchangeName)]
+            [bunnyc_spec(VHost, Host, Port, User, Password, ExchangeName)];
+        _  -> %% TODO should we not start up batch if in inline mode?
```
sdelano: Is there an inline mode of this this comment out of date?

src/oc_erchef/apps/chef_index/src/chef_solr.erl
```diff
@@ -65,19 +64,10 @@ add_org_guid_to_query(Query = #chef_solr_query{filter_query = FilterQuery},
...
-                    {ok, non_neg_integer(), non_neg_integer(), [binary()]} |
-                    {error, {solr_400, string()}} |
-                    {error, {solr_500, string()}}.
-search(#chef_solr_query{} = Query, SolrUrl) ->
+search(#chef_solr_query{} = Query) ->
     %% FIXME: error handling
-    Url = SolrUrl ++ make_solr_query_url(Query),
-    {ok, Code, _Head, Body} = ibrowse:send_req(Url, [], get),
+    Url = make_solr_query_url(Query),
+    {ok, Code, _Head, Body} = chef_index_http:request(Url, get, []),
```
sdelano: :+1: on factoring this out

src/oc_erchef/apps/chef_index/src/chef_index_batch.erl
```diff
@@ -0,0 +1,230 @@
...
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
+
+-record(chef_idx_batch_state, {
+          wrapper_size :: non_neg_integer(),
+          current_size :: non_neg_integer(),
+          max_size :: non_neg_integer(),
+          item_queue = [], %% [{Pid, Timestamp, Doc}]
```
tylercloke: Did we wanna update this?

src/oc_erchef/apps/chef_index/src/chef_index_sup.erl
tylercloke: Hrm I don't think we need to be starting chef_index_batch if inline mode is requested. I think we just need to change `_` to `batch`. `inline` mode doesn't need to start any gen_servers.

src/oc_erchef/apps/chef_index/src/chef_index.erl
```diff
@@ -0,0 +1,77 @@
...
+
+-module(chef_index).
+
+-export([delete/4,
+         add/5,
+         add_async/4,
+         add_async/5
+        ]).
+
+add(TypeName, Id, DbName, IndexEjson, ReqId) ->
```
sdelano: Can we get comments on the difference between `chef_index:add` and `chef_index:add_async`? For one, if we're configured to use RabbitMQ, that's also technically async. What's the difference between the two?

tylercloke:
Things generally LGTM other than the few noted comments (nothing too serious). Should be easy to clean up.

src/oc_erchef/apps/chef_index/src/chef_index.erl
stevendanna: :+1: Maybe we should also have a different name?  When in a mode that isn't rabbitmq, chef_index:add() will block until we have a response from solr.  chef_index:add_async() will not block and returns ok, regardless of the queue_mode configurable.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -234,6 +234,20 @@
 default['private_chef']['opscode-erchef']['couchdb_max_conn'] = '100'
 default['private_chef']['opscode-erchef']['ibrowse_max_sessions'] = 256
 default['private_chef']['opscode-erchef']['ibrowse_max_pipeline_size'] = 1
+# general search settings used to set up chef_index
+default['private_chef']['opscode-erchef']['search_provider'] = "solr" # solr, TODO cloudsearch
+default['private_chef']['opscode-erchef']['search_queue_mode'] = "rabbitmq" # rabbitmq, batch, or inline
```
sdelano: We've talked about this a few time, but having 3 options makes me sad about the confusion that it's going to add. We're going to need to clearly document the differences between the three, and also the behavior of `batch`. A batch size of 5 MILLION (:drevilface:) is pretty insane as a default. What does that mean? What is the `10` for wait? 10 seconds or 10 milliseconds?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -234,6 +234,20 @@
...
+default['private_chef']['opscode-erchef']['search_batch_max_size'] = "5000000"
+default['private_chef']['opscode-erchef']['search_batch_max_wait'] = "10"
+# solr_service configuration for erchef. These are used to configure an opscoderl_httpc pool
+# of HTTP connecton workers.
+default['private_chef']['opscode-erchef']['solr_timeout'] = 30000
+default['private_chef']['opscode-erchef']['solr_http_init_count'] = 25
+default['private_chef']['opscode-erchef']['solr_http_max_count'] = 100
```
sdelano: Is there any where else that we set the init count to be something different than the max count? I know for db connections we always make them the same. What about for HTTP connections?

src/oc_erchef/apps/chef_index/c_src/fast_xs.c
```diff
@@ -0,0 +1,210 @@
+#include <assert.h>
```
sdelano: I'll have to look at the individual commits to find this, but is there anything different in this code than in the old `chef_reindex` module? Also, why vendor that model over here? Does mover now use `chef_index` for the reindex migration or are we further digging ourselves into a mover hole?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
stevendanna: Definitely need some documentation here.  the max_size is actually the byte_size() of the document we are going to post.  This is a forward-looking feature since cloudsearch has a hard limit on the size of its updates.

The max_wait time is in milliseconds.  We should also document that it is the "max_wait" (1) for how long something will sit in the queue before initializing the post and (2) it is a "max" in the same sense that that the bus schedule tells you when the bus is supposed to come.

src/oc_erchef/apps/chef_index/src/chef_index.erl
tylercloke: Afaik `add_async` is only called during reindex. It follows the same code that `chef_index:add()` does, it executes the add in a `spawn` block since there isn't a specific erchef request waiting on it (`chef-server-ctl reindex` just calls the [reindex module directly](https://github.com/chef/chef-server/blob/fce335d0f84cf042ba291e2cfe7aa9204e1188d6/src/oc_erchef/priv/reindex-opc-organization#L101). We don't care about the exact time it finishes reindexing a specific item, just that it eventually does.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -1035,11 +1036,11 @@ create_from_json(#wm_reqdata{} = Req,
     %% a 500 and client can retry. If we succeed and the db call fails or conflicts, we can
     %% safely send a delete to solr since this is a new object with a unique ID unknown to
     %% the world.
-    ok = oc_chef_object_db:add_to_solr(ObjectRec, ObjectEjson),
+    ok = oc_chef_object_db:add_to_solr(ObjectRec, ObjectEjson, ReqId),
```
sdelano: What's the necessity of passing around the request id with all of these?

src/oc_erchef/apps/chef_index/c_src/fast_xs.c
stevendanna: The fast_xs code is nearly all unchanged (one change for dialzyer in the erlang side fast_xs module). As for the rest here is a summary:

- chef_index_http: unchaged
- chef_index_expand: The core expansion code was not touched, but the gen_server was removed and a small public facing API was exposed.  The gen_server in chef_index_expand was focused on batches of a known size, which we generally won't have in the case of indexing inflight requests, thus the chef_index_batch gen_server.
-

src/oc_erchef/apps/chef_index/src/chef_index_sup.erl
sdelano: @tylercloke that sounds reasonable to me.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
stevendanna: This is pretty much verbatim the pool config we use for our authz http connections.   Both are definitely worth looking at skeptically given what we've seen in recently chef-server issues with a well-known online retailer.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
tylercloke: Looks like it was added so that stats_hero could be used to time solr requests:

https://github.com/chef/chef-server/pull/520/files#diff-164d5c5b7e47072b0967d26bee658451R57

src/oc_erchef/apps/chef_index/c_src/fast_xs.c
sdelano: So `chef_index` and `chef_index_batch` are new. Is the batching in `chef_index_batch` largely from the previous `chef_index_expand`, or is that also new?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
sdelano: :+1:

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
sdelano: Thanks for the clarification

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
sdelano: :+1: - thanks @tylercloke

src/oc_erchef/apps/chef_objects/rebar.lock
```diff
@@ -0,0 +1 @@
+[].
```
stevendanna: Remove this file.

src/oc_erchef/apps/chef_index/c_src/fast_xs.c
stevendanna: chef_index_batch is new (but obviously similar since both are batching things for solr).  chef_index_batch only handles adds, tracks the byte_size of the docs being posted rather than the number of items, and is designed to handle numerous unrelated processes adding items to the same batch.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
marcparadise: Can we change the unit, or accept a unit indicator to make it more clear? "5M" or "5MB" vs "5000000"; or `search_batch_max_mb`?

src/oc_erchef/apps/chef_index/c_src/fast_xs.c
marcparadise: Is it an option to move the chef_reindex application into oc_erchef, or at least under chef-server/src/chef_reindex?

joedevivo:
:+1: Let's get it in there!

marcparadise:
:+1:

PR #523, opened 2015-09-10
Dt/user by email

doubt72:
This is mainly intended for the authenticate user endpoint, i.e., we want to be able to login via email address, not just username.

This has the side effect of making the /users/:name endpoint (where :name = email address) work as well.  Pro: this makes password resets possible in manage using either username or email address.  Possible con: this makes any user lookup with email address work, including (probably) a knife lookup.

The question is: do we care about that, or does it open any holes we might be worried about?  If so, we should do this another way (i.e., a distinct lookup for the authenticate_user endpoint, and another way of looking up users by email for password resets).

### DIFF
--- a/oc-chef-pedant/spec/api/authenticate_user_spec.rb
+++ b/oc-chef-pedant/spec/api/authenticate_user_spec.rb
@@ -153,6 +153,37 @@ describe 'authenticate_user', :users do
       end
     end
 
+    context 'with correct (email) credentials', :smoke do
+      let(:body) { { 'username' => username + '@opscode.com', 'password' => password } }
+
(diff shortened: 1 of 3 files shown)


marcparadise:
I am :+1: on this once we're through CI, but would like it if others from @chef/lob could take a look and make sure there are no edge cases with email-as-username that we haven't considered.

sdelano:
I'm :-1: on this. I'll continue to explain in a bit.

marcparadise:
After discussion in Slack, we're going to take another approach here - adding a new query and new `chef_user:fetch` fun head to use it when username is not provided, but email is.

PR #524, opened 2015-09-11
cleanup static nginx files

marcparadise:
This removes unreferenced js files from the nginx html directory, and updates the default API index.html page to include instructions on installing manage. This includes an outdated jqueery version -t he only component that uses jquery appears to be oc-id, and they pull it in at build time via the jquery-rails gem 

ping @chef/lob /cc  @raskchanky @jcreedcmu


marcparadise:
Note that the CI failure above looks transient, and occurred in post-test cleanup. Re-running it.

marcparadise:
This is what the api index page looks like with these updates. 

snapshot1

marcparadise:
snapshot2
Updated image after editing the intro paragraph.

marcparadise:
snapshot1

stevendanna:
:+1:

stevendanna:
Is this good to merge?  Would love someone from VI to confirm.

omnibus/files/private-chef-cookbooks/private-chef/files/default/html/index.html
```diff
@@ -10,20 +10,38 @@
...
+          </p>
+
+          <p>
+            If you do want to use your browser to manage your server, then
+            install the Management Console. It's free for up to 25 nodes.
+          </p>
+
+          <p>
+            One way to install the Management Console is to log in as root and type this command:
+            <pre> # chef-server-ctl install opscode-manage </pre>
```
smith: Note that this will be changed to `chef-manage` soon, but we'll make sure that happens when we make that migration.

smith:
:+1:

marcparadise:
Some text updates from marketing inbound. I thought I pushed them already.
Will do so when I'm at my desk

PR #528, opened 2015-09-13
Avoid spurious HTTP 400 after GET to /_status

stevendanna:
After making a GET to status using a persistent HTTP/1.1 connection, the
client might see a spurious HTTP 400 response (after already receiving a
response to their status request). Since most clients that make requests
to /_status either use HTTP 1.0 or close their connection after reading
the response, they never see the 400.

The 400 seems to be sent whenever the process started with spawn_monitor
exits. gather_health_workers is the only function with a receive clause
for the DOWN messages, but that function likely has already returned
when it is recieved. Calling demonitor/2 with the flush option ensures
those message never hit our mailbox (and are removed if they are already
in our mailbox).

It is unclear to me whether this represents an upstream bug in
mochiweb/webmachine.


joedevivo:
This works. We can merge it, or not monitor the processes in the first place.

PR #533, opened 2015-09-16
stabilize chef_index_batch tests

stevendanna:
A number of timeouts we were seeing seems to be a race condition in
shutting down the gen_server. To avoid this, we monitor the gen_server
pid and wait to get notified of its exit.


tylercloke:
:+1:

Issue #361, opened 2015-06-22
chef-server-ctl ha-status incorretly reports opscode-chef-mover as down in HA setup on primary


vinyar:
Chef-server 12.0.8 on RHEL 6.6 

If I understand correctly, the opscode-chef-mover service is only used during an upgrade, and does not need to be running all of the time. Reporting an error on ha-status is scary and causes cold sweat. 

Expected behavior:
No sweat should be shed.
Should not report an error if it's not an actual error.

```ruby
[someuser@someserver]$ sudo chef-server-ctl ha-status
[OK] keepalived HA services enabled.
[OK] cluster status = master
[OK] found VIP IP address and I am master
[OK] found VRRP communications interface eth2
[OK] bookshelf is running correctly, and I am master.
[OK] keepalived is running.
[OK] nginx is running correctly, and I am master.
[OK] oc_bifrost is running correctly, and I am master.
[OK] oc_id is running correctly, and I am master.
[ERROR] opscode-chef-mover is not running.
[OK] opscode-erchef is running correctly, and I am master.
[OK] opscode-expander is running correctly, and I am master.
[OK] opscode-expander-reindexer is running correctly, and I am master.
[OK] opscode-solr4 is running correctly, and I am master.
[OK] postgresql is running correctly, and I am master.
[OK] rabbitmq is running correctly, and I am master.
[OK] redis_lb is running correctly, and I am master.

[ERROR] ERRORS WERE DETECTED.
```

vinyar:
@glasschef for additional input

poliva83:
@vinyar Did you get anywhere with this? I am am seeing the same output.

```

root@chef-server-be-1a:~# chef-server-ctl ha-status
[OK] keepalived HA services enabled.
[OK] cluster status = master
[OK] found VIP IP address and I am master
[OK] found VRRP communications interface eth0
[OK] bookshelf is running correctly, and I am master.
[OK] keepalived is running.
[OK] nginx is running correctly, and I am master.
[OK] oc_bifrost is running correctly, and I am master.
[OK] oc_id is running correctly, and I am master.
[ERROR] opscode-chef-mover is not running.
[OK] opscode-erchef is running correctly, and I am master.
[OK] opscode-expander is running correctly, and I am master.
[OK] opscode-expander-reindexer is running correctly, and I am master.
[OK] opscode-solr4 is running correctly, and I am master.
[OK] postgresql is running correctly, and I am master.
[OK] rabbitmq is running correctly, and I am master.
[...]

PR #536, opened 2015-09-17
Don't consider opscode-chef-mover service status when checking ha-status

poliva83:
Obvious fix for https://github.com/chef/chef-server/issues/361

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/ha.rb
+++ b/omnibus/files/private-chef-ctl-commands/ha.rb
@@ -94,6 +94,10 @@ add_command_under_category "ha-status", "high-availability", "Show the status of
   ha_services = Dir.chdir(running_config['runit']['sv_dir']){Dir.glob('**/keepalive_me').map{|f| File.dirname(f)}}.sort
 
   get_all_services.sort.each do |service_name|
+    # opscode-chef-mover service is only used during an upgrade, and does not need to be running all of the time
+    # do not consider any hidden services when checking ha status
+    next if hidden_services.include?(service_name)
+    
     if service_enabled?(service_name)
       status = run_command("/opt/opscode/init/#{service_name} status >/dev/null 2>&1")
       if ha_services.include?(service_name)


omnibus/files/private-chef-ctl-commands/ha.rb
```diff
@@ -94,6 +94,9 @@
   ha_services = Dir.chdir(running_config['runit']['sv_dir']){Dir.glob('**/keepalive_me').map{|f| File.dirname(f)}}.sort
 
   get_all_services.sort.each do |service_name|
+    # opscode-chef-mover service is only used during an upgrade, and does not need to be running all of the time
+    next if service_name.eql? "opscode-chef-mover"
```
stevendanna: We may want to use `omnibus-ctl`'s concept of hidden services here.  In that case, this check would become:

```
next if hidden_services.include?("opscode-chef-mover")
```

stevendanna:
Thanks! This is a good improvement, I left a minor comment about using the hiddent_services array that omnibus-ctl already provides.

poliva83:
@stevendanna I can change the PR to use hidden_services array. But can you please explain to me how this array becomes available to the ha.rb script?

The only reference to it I found is here:
https://github.com/chef/chef-server/blob/003067dfe6d42c9adfedd77a0dc8d69fb0f169de/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb#L24

But to be perfectly honest I didn't really understand how ha.rb script was able to leverage the get_all_services array either.

stevendanna:
@poliva83 Happy to try to explain.

TL;DR: You have access to any instance method in the Omnibus::Ctl class.

When we load a command definition like ha.rb, an instance of
Omnibus::Ctl eval's the relevant file:

https://github.com/chef/omnibus-ctl/blob/master/lib/omnibus-ctl.rb#L172-L174

The add_command_under_category function used to define new
commands, is actually a methods on the Omnibus::Ctl class. It
uses the block provided to add_command_under_category to define a
new instance method the Omnibus::ctl class:

https://github.com/chef/omnibus-ctl/blob/master/lib/omnibus-ctl.rb#L190

That new instance method, like any other instance method in that
class, has access to all other the other methods already defined
in Omnibus::Ctl.

Both hidden_services and get_all_services are instance methods in
the Omnibus::Ctl class

  https://github.com/chef/omnibus-ctl/blob/master/lib/omnibus-ctl.rb#L320
  https://github.com/chef/omnibus-ctl/blob/master/lib/omnibus-ctl.rb#L397

which is why you have access to them here.

poliva83:
@stevendanna Thanks for explanation, things are more clear now. I tested and updated my pull request. Should be good now.

stevendanna:
:+1 This looks good to merge.

Issue #534, opened 2015-09-17
chef-server-ctl backup and return code


tiago-cruz-movile:
Using Chef 12.2.0, *chef-server-ctl backup* always return exit code <> 0:

```
[root@mordor2 ~]# chef-server-ctl backup --yes
Starting Chef Server backup
Bringing down the Chef Server
ok: down: bookshelf: 0s, normally up
ok: down: nginx: 1s, normally up
ok: down: oc_bifrost: 0s, normally up
ok: down: oc_id: 0s, normally up
ok: down: opscode-chef-mover: 0s, normally up
ok: down: opscode-erchef: 0s, normally up
ok: down: opscode-expander: 0s, normally up
ok: down: opscode-expander-reindexer: 0s, normally up
ok: down: opscode-solr4: 0s, normally up
ok: down: rabbitmq: 1s, normally up
ok: down: redis_lb: 0s, normally up
Dumping Postgresql database to /tmp/chef_backup20150916-2948-zhsm7b/chef_backup-2015-09-16-19-49-59.sql
could not change directory to "/root"
ok: down: postgresql: 0s, normally up
Writing backup manifest
Creating backup tarball
tar: Removing leading `/' from member names
Bringing up the Chef Server
ok: run: bookshelf: (pid 3121) 1s
ok: run: nginx: (pid 3168) 0s
ok: run: oc_bifrost: (pid 3183) 0s
ok: run: oc_id: (pid 3231) 1s
ok: run: opscode-chef-mover: (pid 3247) 0s
ok: run: opscode-erchef: (pid 3289) 0s
ok: run: opscode-expander: (pid 3312) 0s
ok: run: opscode-expander-reindexer: (pid 3334) 1s
ok: run: opscode-solr4: (pid 3360) 1s
ok: run: postgresql: (pid 3392) 0s
ok: run: rabbitmq: (pid 3429) 1s
ok: run: redis_lb: (pid 3655) 0s
Exporting tarball to /var/opt/chef-backup
Cleaning up /tmp/chef_backup20150916-2948-zhsm7b
Backup Complete!
[root@mordor2 ~]# echo $?
1
```

Confirmed by @irvingpop :)

PR #537, opened 2015-09-17
Fix chef-server-ctl backup always returning 1

ryancragun:
Fixes https://github.com/chef/chef-server/issues/534
http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/721/downstreambuildview/

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/backup.rb
+++ b/omnibus/files/private-chef-ctl-commands/backup.rb
@@ -5,8 +5,8 @@ require 'ostruct'
 
 add_command_under_category 'backup', 'general', 'Backup the Chef Server', 2 do
   unless running_config
-    puts '[ERROR] cannot backup if you haven\'t completed a reconfigure'
-    exit 1
+    log('You must reconfigure the Chef Server before a backup can be performed', :error)
+    exit(1)
   end
 
   options = OpenStruct.new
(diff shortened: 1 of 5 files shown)


stevendanna:
I'm :+1: on just merging; however, if you have a few minutes, here are some things I had questions about:

1) Error output should probably go to standard error rather than standard out.
2) Given our purpose, I think the bare rescue makes sense, but I'd prefer to see it just around foreign code (ChefBackup::Runner).  The other code we are wrapping looks to be related to transforming user-input into configuration.  If something is wrong with those cases, I think we are going to want something more descriptive than the ruby exception message since it probably means the user has input they need to fix.

omnibus/files/private-chef-ctl-commands/backup.rb
```diff
@@ -25,10 +25,15 @@
...
+    Chef::Mixin::DeepMerge.deep_merge!(stringify_keys(options.to_h), running_config['private_chef']['backup'])
+    ChefBackup::Runner.new(running_config).backup
+  rescue => e
+    puts "[ERROR] #{e.message}"
+    exit(1)
+  end
 
-  status = ChefBackup::Runner.new(running_config).backup
-  exit(status ? 0 : 1)
+  exit(0)
```
marcparadise: the way omnibus-ctl works you can just 'return 0' , and above 'return 1'. That will be used as the exit code.

marcparadise:
@stevendanna to the best of my recollection, all of the omnibus-ctl error output (except wrapped chef runs) go to stdout.

ryancragun:
@stevendanna All error output now goes through STDERR and we only rescue on the foreign code.
@marcparadise I've seen early returns in omnibus-ctl cause jump exceptions because omnibus-ctl will try and yield a non-existent block, hence the specific exits.

stevendanna:
Thanks for the updates, :+1: from me.  Will merge at EOD if there are no objections

PR #543, opened 2015-09-23
Add favicons for oc-id and static files.

smith:
Uses the favicon from https://www.chef.io/favicon.ico.

oc-id had a blank file in that place, while the static files did not
have one. Adding the files and the configuration to let nginx serve it.

ChangeLog-Entry: Add a favicon.ico


stevendanna:
Why do we need the favicon.ico in two places?

smith:
@stevendanna I suppose you're right and we can change the link in the oc-id template. Will change.

smith:
Actually, it will just attempt to load /favicon.ico if you're loading /id/anything so even better.

smith:
@stevendanna updated.

stevendanna:
:+1: here.

PR #545, opened 2015-09-24
Use configured port everywhere we talk to postgres

### LINKED ISSUE #459: I can't change postgresql port
I need to change the port where Chef's PostgreSQL is listening but it fails. I've added the options "postgresql['port'] = 6432" to the chef-server.rb. chef-server-ctl reconfigure fails, because there are some resources in the recipes that use the psql command without getting the port information from somewhere.

marcparadise:
Fixes #459  and allows a custom port to be specified in chef-server postgresql settings.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
@@ -198,7 +198,7 @@
            {ip_mode, [ <%=PrivateChef['use_ipv6'] ? "ipv6,ipv4" : "ipv4" %> ] },
            %% Database connection parameters
            {db_host, "<%= node['private_chef']['postgresql']['vip'] %>"},
-           {db_port, 5432},
+           {db_port, <%= node['private_chef']['postgresql']['port'] %>},
            {db_user, "<%= node['private_chef']['opscode-erchef']['sql_user'] %>"},
            {db_pass, "<%= node['private_chef']['opscode-erchef']['sql_password'] %>"},
            {db_name, "opscode_chef" },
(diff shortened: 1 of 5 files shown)


omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
```diff
@@ -197,7 +197,7 @@
            {ip_mode, [ <%=PrivateChef['use_ipv6'] ? "ipv6,ipv4" : "ipv4" %> ] },
            %% Database connection parameters
            {db_host, "<%= node['private_chef']['postgresql']['vip'] %>"},
-           {db_port, 5432},
+           {db_port, <%= node['private_chef']['opscode-erchef']['sql_user'] %>},
```
stevendanna: I imagine this should be `<%= node['private_chef']['postgresql']['port'] %>` rather than sql_user?

An aside: We should probably (eventually) always have 2 attributes 1 that controls what port the service listens on when we set it up, and another to control what port applications should talk to.  By default they would be the same, but it would give us the flexibility to add connection pooling/tpc proxies such that erchef would actually talk to a different port.

marcparadise: D'oh. Corrected.   As we move forward with different permitted  postgres backends per-service, we'll actually push the postgres host/vip/superuser  config down into the application itself, probably with a default value that is used if it's not specified continuing to live in `postgresql`

sdelano:
:+1: - lgtm. Also :+1: to @stevendanna's suggestion of listen and connect port differences, someday.

stevendanna:
:+1: Looks good, MWR

PR #550, opened 2015-09-30
Override bundler from omnibus-software default of 1.5.3 to 1.10.6.

tylercloke:
Ping @chef/lob 

This will fix the build.

http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/731/downstreambuildview/


stevendanna:
:+1: Thanks for tracking this down.

TODO: fix omnibus-software as well, the default is megaold.

PR #551, opened 2015-10-01
[oc_chef_wm] Return the correct "total" in search responses

stevendanna:
Solr may return IDs that are no longer in our database. We filter such
IDs from the response to account for this. Unfortunately, the previous
code was returning the count of responses from Solr, which will be
incorrect if filtering has occurred. This patch updates the code to use
the count of responses from Postgresql (i.e the filtered count).

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
@@ -129,8 +129,7 @@ to_json(Req, #base_state{chef_db_context = DbContext,
                                  filter_permitted_results(ReqId, RequestorId, OrgId, DbContext, IndexType, Ids);
                              false -> Ids
                          end,
-            {DbNumFound, Ans} = make_search_results(BulkGetFun, FilteredIds, BatchSize,
-                                                    Start, SolrNumFound),
+            {DbNumFound, Ans} = make_search_results(BulkGetFun, FilteredIds, BatchSize, Start),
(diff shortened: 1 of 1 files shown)


stevendanna:
cc @chef/lob

stevendanna:
Hrm, is this going to cause a problem for paginated searches?

marcparadise:
I see you already commented on paginated searches.  This looks good for searches that fit in a single set, but looks like it will still report inaccurately for multi-page.  With that said - I don't know that there's a good solution to that since we can't reasonably validate the full result set. 

In addition, with increased delete latency, we're probably going to face issues where the result sets differ on subsequent attempts - so fetching the second set  may return results that overlap with the first.

stevendanna:
@mp Right, even for paginated sources, returning the solr total is a bit odd because you then have to write your client to assume that it might never actually see the number of results it was seen in total.  We could alternatively simply remove all test assertions that look at the total field and that would solve the problem on the cloudsearch branch.

stevendanna:
We definitely can't merge this as-is since the total needs to indicate to the caller that there is more pagination to be done.  After some thinking, I believe that the best bet for now is to adjust our tests to never make assertions about the "total" field in the response.  I don't see a way to fix this without changing the API response. 

@marcparadise  Interestingly, I believe the chef-client code is using the *current* API incorrectly (partially my fault :(). Fix here:https://github.com/chef/chef/pull/4029

stevendanna:
Closing this.  The cloudsearch-support branch now fixes the tests to remove assertions based on the total field of the response.

PR #552, opened 2015-10-01
[oc-chef-pedant] Fix search tests incorrectly using with_search_polling

stevendanna:
with_search_polling works by retrying when an exception is raised. An
empty response from search will not raise an exception, rather the
assertions on the results should also be inside the with_search_polling
block.


stevendanna:
cc @chef/lob

marcparadise:
:+1:

PR #555, opened 2015-10-03
Gather Logs Updates

sean-horn:
* Lets gather the Manage config too, if found

* Use path variables, fix a related bug

    Cleaned up more static specification of config files.
    This helped fix an issue with mistakenly using chef-server.rb as the file to
    grep on EC 11.3.2 when it should have been private-chef.rb.

* Collect config and log info for chef-sync master replica

* Fix major usage of private-chef/chef-server-ctl
  We should prefer the private-chef-ctl symlink for now, so that EC 11.3.2
  still works properly

Tested on EC 11.3.2 and CS 12.0.5


stevendanna:
This looks reasonable. @sean-horn If this is still relevant to merge, mind rebasing the branch against master?  I'll merge after that is done.

sean-horn:
@stevendanna Thanks for taking a look. I just rebased.

PR #556, opened 2015-10-05
Warns rather than fail if external project isn't linked

stevendanna:
The user should not need oc_reporting to do  a
`dvm load oc_erchef`.


marcparadise:
:+1:

PR #562, opened 2015-10-07
Make the oc_chef_authz app use the chef_objects app.

billstclair:
This ensures that chef_objects gets compiled before oc_chef_authz, so that the chef_object behaviour is available at compile time for oc_chef_authz/src/oc_chef_policy.erl.

Fixes broken build.

### DIFF
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.app.src
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.app.src
@@ -22,6 +22,7 @@
   {registered, []},
   {applications, [kernel,
                   stdlib,
+		  chef_objects,
                   stats_hero,
                   sqerl]},
   {mod, {oc_chef_authz_app, []}}


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.chef.io/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## GitHub Users Who Are Not Authorized To Contribute

The following GitHub users do not appear to have signed a CLA:

* @billstclair

[Please sign the CLA here.](https://supermarket.chef.io/become-a-contributor)

chef-supermarket:
Hi. Your friendly Curry bot here. Just letting you know that all commit authors have become authorized to contribute. I have added the "Signed CLA" label to this issue so it can easily be found in the future.

stevendanna:
@billstclair Thanks for digging into this.  We might take a little time to merge it as we want to understand why this doesn't seem to be required for those of us who are working with chef-server day to day. However, I'll make sure this (or a similar fix) gets in soon.  If you don't mind signing the CLA, that would help alot.

billstclair:
CLA signed, as you can see above.

I talked with Oliver about this. It appears from oc_errchef.app.src, that chef_objects should be built before oc_chef_authz, but doesn't happen in my world without this patch.

I tried your recommendation for "./rebar3 update", and I built a rebar3 from the latest master branch sources. Both compile oc_chef_authz before chef_objects without my patch.

But my pull request doesn't pass travis tests, and it fixes a problem that nobody but me has, which obviously makes it a low priority.

billstclair:
I've been investigating this. It compiles fine on my Mac. It seems that only my Digital Ocean Ubuntu VM has the problem. I've spent a bunch of time looking inside rebar3, but so far without any luck at finding what makes it compile the apps/ in a different order.

I pulled chef_object.erl out of the chef_objects app into its own chef_object app, and that make the oc_erchef travis tests succeed, but the chef-mover Travis tests are failing. Which is weird, since chef-mover has its own separate copies of some of the oc_erchef apps/.

Anyway, closing this pull request, since it's not a problem that anybody working on Chef is encountering.

marcparadise:
Thanks @billstclair , I'm going to be spinning up a DO 14.04 vm sometime this week to try it out.  I don't currently run into it with local builds on my 15.04 desktop system.

billstclair:
Another way of fixing this, should it become necessary, would be to move each of the apps/ sub-directories into its own git repository, and include them as rebar dependencies, not sub-applications.

stevendanna:
@billstclair That is the structure we used to have but we moved away from it because unfortunately these modules are a bit more tightly coupled than we would like them to be.  It became the case that implementing a feature in Chef Server required PRs across 4 or 5 different projects.  Hopefully, however, as we continue to improve some of the interface boundaries, we can split out truly reusable pieces into libraries again.

I haven't had much time to track down what you saw yet, but I did find this commit, which seems like it might have been addressing a similar issue:

https://github.com/rebar/rebar3/commit/f82874de5802dbf9e2029a9aa07c7cbae12a1324

billstclair:
That rebar3 patch is in the master branch, with an update to all_apps_deps/1 to make it handle two forms for each of the rebar_app_info:deps(App), and to merge the sorts of Applications & Deps rather than sorting the concatenation.

https://github.com/rebar/rebar3/blob/master/src/rebar_digraph.erl#L93

Using the master rebar, freshly built, still fails in my DO VM compiling oc_chef_authz before chef_objects.

```
$ rebar3 version
rebar 3.0.0-beta.3+build.3081.ref6e4ab1e on Erlang/OTP 17 Erts 6.4
```

elliott-davis:
I just bumped into this on a Debian Jessie VM. This PR solved the problem.

PR #563, opened 2015-10-08
Remove "hosted" from page copy

smith:
Just call it "Chef account" and "Chef username" and put it into the i18n
config.

ChangeLog-Entry: Remove "hosted" from some text in oc-id


stevendanna:
:+1: On the idea of the change.  It looks superficially request but needs a :+1: from someone who knows our web stack.

mmzyk:
:+1:  This is just using Rails i18n setup and then using the shortcut syntax for views. This all looks good.

stevendanna:
@smith Rebase and merge when ready.

PR #566, opened 2015-10-12
dvm powerdown ssh fail and more

marcparadise:
A few dvm enhancements: 
- we no longer load dvm itself as a gem, instead just running it in place.  Because we no longer mount off of /vagrant, this elimiates the problem where mount fails during reboot, causing VM to never complete the process of coming online. 
- remove deep_merge hack in dvm loading/config  because we already have deep_merge available via the chef-gem -  so we'll include that one instead. 
- added AUTOPACKAGE env var for `vagrant up` for the truly lazy who don't like typing the extra 
  key when presented with a menu of packages. 
- use absolute paths in dvm itself, so that can works without installation as a gem. Possibly we can just add /vagrant/dvm  and /vagrant/dvm/lib to PATH  manage better not being a gem.  Next pass I will probably just use `RUBYLIB` in profile.d/omnibus-embedded.sh and get away from fixed paths. 
 - omnibus-embedded.sh now has the right permissions so that path is loaded correctly no matter if you're vagrant or root (which means sudo -s now works as well as sudo -i) 
- fixed project loading so that we don't check for the existence a path until we try to load it. 

ping @chef/lob


dev/Vagrantfile
```diff
@@ -58,7 +58,6 @@ def define_chef_server(config, attributes)
              "opscode_erchef['db_pool_size']" => 10,
              "oc_id['db_pool_size']" => 10,
              "oc_bifrost['db_pool_size']" => 10 }
-      # TODO merge with other options nicely.
       json = simple_deep_merge(json, { "provisioning" => { "chef-server-config" => pg } })
```
tylercloke: Did you mean to remove this line? I just see the TODO being removed without an actual change...

tylercloke:
:+1: Other than my comment, all sound like decent changes.

dev/Vagrantfile
marcparadise: nice spot.   The actual change was adding that `simple_deep_merge` and the to-do was a leftover for when that wasn't done

PR #568, opened 2015-10-13
Cloudsearch Support

stevendanna:
This patchset adds support for using AWS CloudSearch as the search
index used by the Chef Server. CloudSearch provides a Solr-like API that
can provide a search service similar to that provided by the Chef Server's 
managed version of Solr 4. 

When using CloudSearch as a backend, the following limitations apply:

- All searches are case-insensitive
- Search results are limited to the first 10000 results.
- The rabbitmq search indexing queue cannot be used.  Only 
  direct-from-erchef (inline or batch) modes are supported.
- CloudSearch must be correctly configured by the user with the correct
  index fields and access policies.

We hope to remove the 10000 result max in an future version of chef-client 
and chef-server.

To support CloudSearch this PR makes the following changes:

- The chef_solr module now passes through many function calls to 
   provider-specific modules that understand the particular requirements
   of the relevant API.  

- The chef_index_expand module has been altered to know how to 
  construct post bodies for both solr and cloudsearch.  The difference in 
  the post bodies include different XML payload signatures and different key-value 
  separators in the expansion.

- chef_index_expand and chef_lucene now call chef_solr:transform_* functions when
  processing data. These functions are responsible for altering the user-inputed data
  to work around CloudSearch's tokenization which is not configurable.  Specifically,
  most ASCII punctuation results in CloudSearch breaking the data into multiple terms.
  This is problematic for most Chef Node data where special characters such as '-' can
  be expected to appear in what the user considers a single searchable term.  To avoid this 
  problem,  nearly all ASCII punctuation is replaced with markers both when indexing and making
  search requests.

- A cs_escape NIF was added to speed up the data transformations mentioned above.

- A minor refactor was done to remove unnecessary code from chef_index_expand

[Latest build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/738/downstreambuildview/)

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
@@ -29,14 +29,13 @@ class SolrPreflightValidator < PreflightValidator
   end
 
   def warn_changed_external_flag
-    if OmnibusHelper.has_been_bootstrapped? && backend?  && previous_run
+    if OmnibusHelper.has_been_bootstrapped? && backend? && previous_run
       if external_flag_changed?
         Chef::Log.warn <<-EOM
 
-The value of opscode_solr4['external'] has been changed. Search
-results against the new external search index may be incorrect. Please
-run `chef-server-ctl reindex --all` to ensure correct results
-
+CSSOLR001: The value of opscode_solr4['external'] has been changed. Search
+           results against the new external search index may be incorrect. Please
+           run `chef-server-ctl reindex --all` to ensure correct results
 EOM
       end
     end
@@ -51,10 +50,9 @@ EOM
     if cs_solr_attr['external'] & (! cs_solr_attr['external_url'])
       fail_with <<EOM
 
-No external url specified for Solr depsite opscode_solr4['external']
-being set to true. To use an external solr instance, please set
-opscode_solr4['external_url'] to the external solr endpoint.
-
+CSSOLR002: No external url specified for Solr depsite opscode_solr4['external']
+           being set to true. To use an external solr instance, please set
+           opscode_solr4['external_url'] to the external solr endpoint.
 EOM
     end
   end
@@ -67,25 +65,24 @@ EOM
       if ! ['batch', 'inline'].include?(@cs_erchef_attr['search_queue_mode'])
         fail_with <<-EOM
 
-The cloudsearch search provider is only supported by the batch or
-inline queue modes. To use the cloudsearch search provider, please
-also set:
+CSSOLR003: The cloudsearch search provider is only supported by the batch or
+           inline queue modes. To use the cloudsearch search provider, please
+           also set:
 
-opscode_erchef['search_queue_mode'] = 'batch'
+           opscode_erchef['search_queue_mode'] = 'batch'
 
-in /etc/opscode/chef-server.rb
-
+           in /etc/opscode/chef-server.rb
 EOM
       end
     when 'solr'
     else
       fail_with <<-EOM
 
-The specified search provider (#{provider}) is not currently supported.
-Please choose from one of the following search providers:
+CSSOLR004: The specified search provider (#{provider}) is not currently supported.
+           Please choose from one of the following search providers:
 
-solr
-cloudsearch
+           solr
+           cloudsearch
 EOM
     end
   end
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
@@ -23,25 +23,28 @@ class SolrPreflightValidator < PreflightValidator
   end
 
   def run!
-    warn_unchanged_external_flag
+    warn_changed_external_flag
     verify_external_url
     verify_erchef_config
   end
 
-  def warn_unchanged_external_flag
+  def warn_changed_external_flag
     if OmnibusHelper.has_been_bootstrapped? && backend?  && previous_run
-      if cs_solr_attr.has_key?('external') && (cs_solr_attr['external'] != previous_run['opscode-solr4']['external'])
+      if external_flag_changed?
         Chef::Log.warn <<-EOM
 
-The value of opscode_solr4['external'] has been changed.  Search
+The value of opscode_solr4['external'] has been changed. Search
 results against the new external search index may be incorrect. Please
 run `chef-server-ctl reindex --all` to ensure correct results
 
 EOM
       end
-    else
-      return true
     end
+  end
+
+  def external_flag_changed?
+    cs_solr_attr.has_key?('external') &&
+      (cs_solr_attr['external'] != previous_run['opscode-solr4']['external'])
   end
 
   def verify_external_url
@@ -77,6 +80,7 @@ EOM
     when 'solr'
     else
       fail_with <<-EOM
+
 The specified search provider (#{provider}) is not currently supported.
 Please choose from one of the following search providers:
 
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
@@ -1,0 +1,88 @@
+#
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+class SolrPreflightValidator < PreflightValidator
+  attr_reader :cs_solr_attr, :node_solr_attr, :cs_erchef_attr
+
+  def initialize(node)
+    super
+    @cs_solr_attr = PrivateChef['opscode_solr4']
+    @node_solr_attr = node['private_chef']['opscode-solr4']
+    @cs_erchef_attr = PrivateChef['opscode_erchef']
+  end
+
+  def run!
+    warn_unchanged_external_flag
+    verify_external_url
+    verify_erchef_config
+  end
+
+  def warn_unchanged_external_flag
+    if OmnibusHelper.has_been_bootstrapped? && backend?  && previous_run
+      if cs_solr_attr.has_key?('external') && (cs_solr_attr['external'] != previous_run['opscode-solr4']['external'])
+        Chef::Log.warn <<-EOM
+
+The value of opscode_solr4['external'] has been changed.  Search
+results against the new external search index may be incorrect. Please
+run `chef-server-ctl reindex --all` to ensure correct results
+
+EOM
+      end
+    else
+      return true
+    end
+  end
+
+  def verify_external_url
+    if cs_solr_attr['external'] & (! cs_solr_attr['external_url'])
+      fail_with <<EOM
+
+No external url specified for Solr depsite opscode_solr4['external']
+being set to true. To use an external solr instance, please set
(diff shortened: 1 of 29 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
```diff
@@ -0,0 +1,88 @@
...
+    @cs_erchef_attr = PrivateChef['opscode_erchef']
+  end
+
+  def run!
+    warn_unchanged_external_flag
+    verify_external_url
+    verify_erchef_config
+  end
+
+  def warn_unchanged_external_flag
```
marcparadise: `warn_changed_external_flag` may be more accurate

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
```diff
@@ -0,0 +1,88 @@
...
+      if cs_solr_attr.has_key?('external') && (cs_solr_attr['external'] != previous_run['opscode-solr4']['external'])
+        Chef::Log.warn <<-EOM
+
+The value of opscode_solr4['external'] has been changed.  Search
+results against the new external search index may be incorrect. Please
+run `chef-server-ctl reindex --all` to ensure correct results
+
+EOM
+      end
+    else
```
marcparadise: Since we're not using the return, can we just omit from here down?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
```diff
@@ -0,0 +1,88 @@
...
+      return true
+    end
+  end
+
+  def verify_external_url
+    if cs_solr_attr['external'] & (! cs_solr_attr['external_url'])
+      fail_with <<EOM
+
```
marcparadise: Remember your kindly support personnel and add error numbers

omnibus/files/private-chef-cookbooks/private-chef/templates/default/pedant_config.rb.erb
```diff
@@ -61,7 +61,11 @@ default_orgname <%= @default_orgname ? "\"#{@default_orgname}\"" : "nil" %>
 # testing location, you should not specify a value for this parameter.
 # The tests will still run, albeit slower, as they will now need to
 # poll for a period to ensure they are querying committed results.
+<% if node['private_chef']['search_provider'] == "solr" %>
 search_server "<%= @solr_url %>"
+<% else %>
+# search_server "<%= @solr_url %>"
```
marcparadise: I didn't see anything in the test changs to handle no server url - will its absence cause trouble for pedant?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
stevendanna: Yes, yes indeed.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/pedant_config.rb.erb
stevendanna: Nope!  This is a bit convoluted but is explained a bit by the comment above.  When search_server isn't present, what changes is how we poll for search changes.  When it is present we send an explicit commit and then assume that the next search will have up to date result.  When it isn't present we repeat the search over and over until the test passes or we time out.  CloudSearch doesn't use the explicit commit so we have to do the latter, which is much much slower.

src/oc_erchef/apps/chef_index/src/chef_index_batch.erl
```diff
@@ -144,11 +143,11 @@ flush(State = #chef_idx_batch_state{item_queue = Queue,
     {PidsToReply, Timestamps, DocsToAdd} = lists:unzip3(Queue),
     Doc = wrap(DocsToAdd, State),
     Self = self(),
-    spawn_link(
+    spawn(
```
marcparadise: It looks like we were  never doing anything with the termination notification, that's why the change to `spawn`?

src/oc_erchef/apps/chef_index/src/chef_solr.erl
```diff
@@ -24,17 +24,80 @@
...
+         delete_search_db/1,
+         delete_search_db_by_type/2,
+         %% Helpers used by providers
+         db_from_orgid/1
         ]).
 
 -include("chef_solr.hrl").
+-define(PROVIDER_FUNC(FuncName),
+        begin
+            Mod = list_to_atom(atom_to_list(search_provider()) ++ "_provider"),
```
marcparadise: I think we talked about having the config render the actual provider name? That would let these look like: 
```
-define(PROVIDER_FUNC(FuncName) (search_provider()):FuncName() ).
```
and 
```
-define(PROVIDER_FUNC(FuncName,Args) (search_provider()):FuncName(Args) ).
```

src/oc_erchef/apps/chef_index/src/chef_index_batch.erl
stevendanna: Yes.  We could alternatively add a handle_info for the EXIT messages.  However, I think the best we can do is log a message since we no longer have the list of Pids waiting for a response.  Those  requests will all fail with a gen_server timeout.

src/oc_erchef/apps/chef_index/src/chef_solr.erl
```diff
@@ -44,22 +107,32 @@ make_query_from_params(ObjType, QueryString, Start, Rows) ->
...
--spec add_org_guid_to_query(#chef_solr_query{}, binary()) ->
-                                   #chef_solr_query{}.
+-spec add_org_guid_to_query(#chef_solr_query{}, binary())
+                           -> #chef_solr_query{}.
 add_org_guid_to_query(Query = #chef_solr_query{filter_query = FilterQuery},
                       OrgGuid) ->
-    Query#chef_solr_query{filter_query = "+" ++
-                              search_db_from_orgid(OrgGuid) ++
-                              " " ++ FilterQuery}.
+    Query#chef_solr_query{filter_query = ?PROVIDER_FUNC(add_org_guid_to_fq,
```
marcparadise: Since we  have   `search_provider` in the  `chef_solr_query`,  can we  directly  Provider:add_org_guid_to_fq(...)? 

Same in other places where we're operating off of a `chef_solr_query` record.

src/oc_erchef/apps/chef_index/src/chef_solr.erl
```diff
@@ -44,22 +107,32 @@ make_query_from_params(ObjType, QueryString, Start, Rows) ->
     FilterQuery = make_fq_type(ObjType),
     %% 'sort' param is ignored and hardcoded because indexing
     %% scheme doesn't support sorting since there is only one field.
-    Sort = "X_CHEF_id_CHEF_X asc",
+    Sort = binary_to_list(id_field()) ++ " asc",
```
marcparadise: Given the comment about the parameter being ignored - is it not ignored in cloudsearch? And is it really ignored in solr? 

I could see enforcing a sort as a means to get consistent ordering when paging, but am just curious if it's respected by one, both or neither provider(s).

stevendanna: @marcparadise I believe that comment refers to the user-supplied sort parameter.  The Sort parameter we send to Solr and Cloudsearch are definitely not ignored and ensure that we can page results consistently.

src/oc_erchef/apps/chef_index/src/chef_solr.erl
```diff
@@ -70,13 +143,7 @@ search(#chef_solr_query{} = Query) ->
...
     case Code of
         "200" ->
-            SolrData = jiffy:decode(Body),
-            Response = ej:get({<<"response">>}, SolrData),
-            Start = ej:get({<<"start">>}, Response),
-            NumFound = ej:get({<<"numFound">>}, Response),
-            DocList = ej:get({<<"docs">>}, Response),
-            Ids = [ ej:get({<<"X_CHEF_id_CHEF_X">>}, Doc) || Doc <- DocList ],
-            {ok, Start, NumFound, Ids};
+            handle_successful_search(Body);
```
marcparadise: I think this is another case where we can get the provider from the record, and pass it into handle_successful_search instead of needing to look it up in the `handle_successful_search` function.

src/oc_erchef/apps/chef_index/src/chef_solr.erl
stevendanna: Yup, apologies, I forgot about that early review comment.

src/oc_erchef/apps/chef_index/c_src/cs_escape.c
```diff
@@ -0,0 +1,224 @@
+/* Copyright 2015 Chef Software, Inc.
```
marcparadise: I combed through this pretty thoroughly earlier in the week - any changes since then?

marcparadise:
Two general comments: 

Module naming - prefixing the module name with the app name (or at least with `chef_`) helps avoid conflicts with future libraries. That'd probably mean a change to the nif name as well. 

Macros  - when a macro does more than a very simple thing it makes me nervous, only because it hides complexity behind something that people assume to be a simple call.  In this case, we end up doing  envy:get (application:gen_env, + envy validation) in addition to the envy list function for each call into the search provider. 

I'd rather see us thread  the provider through the function calls to the extent possible - (maybe even in the exposed API as well, so that the caller needs to tell us which to use? ) . If that's not practical, maybe next-best would be a combination of threading whereit makes sense (eg using the query `search_provider` field we've already captured and passing it into internal functions) ,  capturing the provider if we need to make multiple calls instead of using the macro multiple times, and making the macro into a function - which is less likely to cause assumptions about its behavior 

Outside of that - the changes themselves look good. I would like to get more pedant coverage in for actual search behaviour, but even that is tricky - then we're getting into testing the behaviors of an external system in our own tests.

src/oc_erchef/apps/chef_index/c_src/cs_escape.c
```diff
@@ -0,0 +1,224 @@
...
+  "__QS__", // ?
+  "__AT__", // @
+  "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K",
+  "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V",
+  "W", "X", "Y", "Z", // No replacements
+  "__OB__", // [
+  "__BS__", // Back Slash
+  "__CB__", // ]
+  "__CA__", // ^
+  "_", // _
```
sdelano: Does this mean that we're replacing underscores with underscores? How does replacing other characters with `__XX__` (including the underscores) affect underscores still being there?

sdelano: I would have guessed that we might replace `_` with `__US__` or something.

sdelano: I see it's handled in the tests as well. Perhaps there's some backstory on it.

sdelano: My only concern here would be that actual data might have something like `__EQ__` in it, and then we hit that unintentionally because we search for `=`.

stevendanna: > Does this mean that we're replacing underscores with underscores? How does replacing other characters with __XX__ (including the underscores) affect underscores still being there?

Yes, although in most cases replacement is guarded against a check that the characters is one we need to replace.  We could certainly replace literal underscores as well as I agree it could cause problems in places where people have data that is,s ay "foo-bar" and then search for "foo_*" which will return "foo-bar".  It is handled in the tests mostly because it is one of the few special characters that we weren't replacing.

stevendanna: > My only concern here would be that actual data might have something like __EQ__ in it, and then we hit that unintentionally because we search for =.

Yup. This is indeed a problem.  One alternative that we talked about was changing this to do the  replacements with unicode characters that are currently impossible to search for.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
stevendanna: DOne!

stevendanna:
Fairwell for now.

PR #569, opened 2015-10-13
Comment vendored depsolver test

stevendanna:
This test is noisy and fails at random on Travis.  It is part of
chef-mover's vendored copy of depsolver.


marcparadise:
:+1:

PR #570, opened 2015-10-14
rabbitmq queue monitor + queue stability enhancements

metadave:
This PR adds several enhancements for RabbitMQ stability.

1. a `max-length` policy is set on the `/analytics` exchange
2. The RabbitMQ Management console is enabled by default with SSL on port `15672`. A `rabbitmgmt` user is created to access this console.
3. A Queue Length monitoring gen\_server (`chef_wm_actions_queue_monitoring.erl`). This gen\_server periodically connects to the management endpoint to check max-length and current length of the analytics queue. If the queue is at > 80% capacity, a warning is issued via lager. If the queue reaches 100% capacity, toggle a "queue at capacity" flag until messages have been read/purged from the queue. Code that publishes messages to RabbitMQ can check the state of the "queue at capacity" flag before enqueuing. 
4. Queue length statistics are reporting via the `_status` endpoint.

More details are available in `chef_wm_actions_queue_monitoring.erl`.

---
### Notes

* This PR adds the PropER testing tool, which is ***GPL V3***. HOWEVER, this is a ***test-only dependency***.

* SSL for the RabbitMQ Management has been configured to only accept `tlsv1.2` and `tlsv1.1` (POODLE, BEAST) via:

		{ssl, [{versions, ['tlsv1.2', 'tlsv1.1']}]}

---

### TODO: 

- [x] tests in `chef_wm_status_tests.erl`
- [x] integration tests/local rabbitmq setup
- [x] vhost/queue hardcoded, remove!


omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
```diff
@@ -74,13 +75,37 @@
...
+
 component_runit_service "rabbitmq"
 
 if is_data_master?
   rmq_ctl = "/opt/opscode/embedded/bin/rabbitmqctl"
+  rmq_plugins = "/opt/opscode/embedded/bin/rabbitmq-plugins"
   opc_ctl = "/opt/opscode/bin/private-chef-ctl"
   opc_username = OmnibusHelper.new(node).ownership['owner']
   rmq_ctl_chpost = "/opt/opscode/embedded/bin/chpst -u #{opc_username} -U #{opc_username} #{rmq_ctl}"
+  rmq_plugins_chpost = "/opt/opscode/embedded/bin/chpst -u #{opc_username} -U #{opc_username} #{rmq_plugins}"
```
stevendanna: I see that the `_chpost` extension is used above, but part of me feels that this is a typo for `_chpst` that has lived for a while.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,472 @@
...
+         start_timer/0,
+         stop_timer/0
+
+        ]).
+
+-behaviour(gen_server).
+-define(SERVER, ?MODULE).
+
+
+-record(state, {
```
stevendanna: It might be useful to give this state a descriptive prefix so that it can be recognized at a glance if it ever shows up in a stacktrace.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,472 @@
...
+    case MaxResult of
+        {ok, "200", _, MaxLengthJson} ->
+            parse_max_length_response(MaxLengthJson);
+        {error, {conn_failed,_}} ->
+            lager:info("Can't connect to RabbitMQ management console"),
+            undefined;
+        {ok, "404", _, _} ->
+            lager:info("RabbitMQ max-length policy not set"),
+            undefined;
+        Resp -> lager:error("Unknown response from RabbitMQ management console: ~p", [Resp]),
```
stevendanna: indentation here doesn't match the rest of the cases.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,472 @@
...
+    Interval = envy:get(oc_chef_wm, rabbitmq_queue_length_monitor_millis, pos_integer),
+    {ok, TRef} = timer:send_interval(Interval, status_ping),
+    TRef.
+
+% this function just returns ok, as a return value is communicated back to
+% the gen_server via check_current_queue_length
+-spec check_current_queue_state(pid(), integer()) -> ok.
+check_current_queue_state(ParentPid, DroppedSinceLastCheck) ->
+    case get_max_length() of
+        undefined -> ok;
```
stevendanna: In this case is the only way the ParentPid gets notified via the EXIT message for the spawn_link this is called inside of?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,472 @@
...
+    {noreply,State#state{dropped_since_last_check = 0}};
+handle_info({'EXIT', From, Reason}, #state{worker_process = WorkerPid,
+                                           sync_response_process = SyncPid} = State) ->
+    % Check to see if the EXIT came from our worker
+    case From == WorkerPid of
+        true ->
+              % check if we need to reply to a sync_check_current_state() call
+              case SyncPid of
+                  undefined -> ok;
+                  Pid -> gen_server:reply(Pid, ok)
```
stevendanna: `ok` seems like an odd response in this case, should we return an error tuple with the reason?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,472 @@
...
+            lager:info("Can't connect to RabbitMQ management console"),
+            undefined;
+        {ok, "200", _, CurrentStatusJson} ->
+            parse_current_length_response(CurrentStatusJson);
+        {ok, "404", _, _} ->
+            lager:info("Queue not bound in /analytics exchange"),
+            undefined;
+        Resp ->
+            lager:error("Unknown response from RabbitMQ management console: ~p", [Resp]),
+                 undefined
```
stevendanna: indentation inconsistency here

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
metadave: the ParentPid doesn't need to be notified in this case, there are no results to add to the state

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
metadave: maybe, but this is mostly just to signal the end of the synchronous op (via sync_check_current_state) which is used for testing only. Any downstream function calls that fail should log an error message.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
stevendanna: Yeah, cool, figured that out as I took in more of the module.  I thought I had deleted that particular comment but I guess I wasn't quick enough :

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
stevendanna: :+1: Seems reasonable.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_status.erl
```diff
@@ -61,10 +61,20 @@ check_health() ->
...
+    StatList =
+      case envy:get(oc_chef_wm, rabbitmq_queue_length_monitor_enabled, false, boolean) of
+        false -> % chef_wm_actions_queue_monitoring isn't running, skip it
+                 StatList0;
+        true -> AnalyticsQ = chef_wm_actions_queue_monitoring:status(),
+                StatList0 ++ [{<<"analytics_queue">>, {AnalyticsQ}}]
+    end,
+    {Status, chef_json:encode({StatList})}.
 
 overall_status(Pings) ->
```
stevendanna: I'm wondering if we should set overall_status to `fail` if the queue is at capacity and the drop messages configurable is true. My concern is that we really want to surface the fact that messages are being dropped.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
metadave: :+1:

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_status.erl
metadave: would we want another state, say `overload`, to signal that the server is still functioning in a degraded state? As the whole point of this PR is to improve stability of Chef Server + Rabbit, I would be sad if someone restarted Chef Server upon receiving `fail`.

metadave: I'm easy either way, I think changing the overall_status is a good idea.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -116,6 +116,45 @@
...
+
+# don't send messages to rabbitmq if it has reached it's configured max_length
+default['private_chef']['rabbitmq']['drop_on_full_capacity'] = true
+
+# rabbit_mgmt_service configuration for erchef. These are used to configure an opscoderl_httpc pool
+# of HTTP connecton workers.
+default['private_chef']['rabbitmq']['rabbit_mgmt_timeout'] = 30000
+default['private_chef']['rabbitmq']['rabbit_mgmt_http_init_count'] = 25
+default['private_chef']['rabbitmq']['rabbit_mgmt_http_max_count'] = 100
+default['private_chef']['rabbitmq']['rabbit_mgmt_http_cull_interval'] = "{1, min}"
```
marcparadise: In order to keep this friendlier if anyone who has to configure it, can we just use seconds, and do the conversion to correct form in the config template? (For the two  below as well).

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -116,6 +116,45 @@
...
+default['private_chef']['rabbitmq']['rabbit_mgmt_http_max_age'] = "{70, sec}"
+default['private_chef']['rabbitmq']['rabbit_mgmt_http_max_connection_duration'] = "{70,sec}"
+
+# comma sep list of tuples, without surrounding []'s
+# rendered as a list in oc_erchef.config.erb, including basic_auth info
+default['private_chef']['rabbitmq']['rabbit_mgmt_ibrowse_options'] =  "{connect_timeout, 10000}"
+
+# rabbitmq management console SSL settings
```
marcparadise: Would it make sense to use the existing cert we create elsewhere, instead of having to generate a separate one?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -116,6 +116,45 @@
 default['private_chef']['rabbitmq']['consumer_id'] = 'hotsauce'
 default['private_chef']['rabbitmq']['env_path'] = "/opt/opscode/bin:/opt/opscode/embedded/bin:/usr/bin:/bin"
 
+default['private_chef']['rabbitmq']['max_length'] = 100000
```
marcparadise: Given that this is only applied to the /analytics vhost, let's call it something like `analytics_max_length`.  Another thought here is that instead of using a separate flag for `enabled` we could just check to see if the length is > 0.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
```diff
@@ -103,7 +103,28 @@
               {actions_vhost, <<"<%= @actions_vhost %>">>},
               {actions_exchange, <<"<%= @actions_exchange %>">>},
               {actions_fqdn, <<"<%= node['private_chef']['lb']['api_fqdn'] %>">>},
-
```
marcparadise: Can we move these into {rabbitmq, [{management, [ ...proplist ...]}]}? It's an extra step to extract it, but a bit cleaner. (I've been wanting to do the same with the actions_ prefixed stuff too, but that'll require a change to reporting since I think it uses this config block.)

omnibus/files/private-chef-cookbooks/private-chef/templates/default/rabbitmq.config.erb
```diff
@@ -0,0 +1,13 @@
+[
+ {ssl, [{versions, ['tlsv1.2', 'tlsv1.1']}]},
+ {rabbitmq_management,
+  [{listener, [{port,     <%= node['private_chef']['rabbitmq']['management_port'] %> },
```
marcparadise: Will this by default bind to the same address as specified in ['rabbitmq']['vip']?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
...
+        parse_integer(proplists:get_value(<<"max-length">>, Defs, undefined))
+    catch
+        Error:Rsn->
+            lager:error("Invalid RabbitMQ response while getting queue max length ~p ~p",
+                        [Error, Rsn]),
+            undefined
+    end.
+
+
+-spec parse_integer(any()) -> integer | undefined.
```
marcparadise: I think we have one or two other safe-list-to-int-converters kicking around here somewhere.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
metadave: I tired that first, but didn't have any luck with the nginx certs for some reason or another. I'll recheck now that it's working.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
...
+status() ->
+    gen_server:call(?SERVER, status).
+
+% is current queue length >= max queue length?
+% Note, this simply checks the gen_server state, and does NOT
+% ping the RabbitMQ management console.
+% Pinging the management console is handled by an internal timer
+% so we never block waiting for an HTTP response.
+is_queue_at_capacity() ->
+    Timeout = envy:get(oc_chef_wm, rabbitmq_queue_length_timeout_millis, 5000, non_neg_integer),
```
marcparadise: A couple of concerns with the timeout, the conditions that could lead up to it, and the effects it'll  have if we start to hit it. I'll write them up shortly in a standalone comment.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
...
+    ok.
+
+delete_pool() ->
+    Pools = get_pool_configs(),
+    [ok = oc_httpc:delete_pool(PoolNameAtom) || {PoolNameAtom, _Config} <- Pools],
+    ok.
+
+get_pool_configs() ->
+    Config = envy:get(oc_chef_wm, ?POOLNAME, [], any),
+    [{?POOLNAME, Config}].
```
marcparadise: I'm probably missing something really obvious, but where is `POOLNAME` defined?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
metadave: Sure, but note that the default timeout of `gen_server:call()` is 5000 ms anyways. This is *just in case* you want to tweak that.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
metadave: https://github.com/chef/chef-server/pull/570/files#diff-63ce4ca65846a36f23d55cce7620703eR126

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
...
+                ok
+    end.
+
+
+-spec calc_ratio_and_percent(integer(), integer()) -> {float(), float()}.
+calc_ratio_and_percent(0, _MaxLength) ->
+    {0.0, 0.0};
+%% the max_length policy should ensure that CurrentLength <= MaxLength,
+%% but return 100% full if this ever happens
+calc_ratio_and_percent(CurrentLength, MaxLength) when CurrentLength > MaxLength ->
```
marcparadise: Could be `>=`?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
+% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
+%% ex: ts=4 sw=4 et
+%%% @author Dave Parfitt <dparfitt@chef.io>
+%%% @doc
+%%% monitor RabbitMQ length of analytics queues
+%%% @end
+%% Copyright 2011-2015 Chef Software, Inc. All Rights Reserved.
```
marcparadise: Just 2015 and no `All Rights Reserved`

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
...
+    lager:info("Queue monitor check still running, skipping next check"),
+    {noreply, State};
+handle_info(Info, State) ->
+    lager:debug("Unknown info: ~p", [Info]),
+    {noreply, State}.
+
+terminate(_Reason, #queue_monitor_state{timer=undefined}) ->
+    ok;
+terminate(_Reason, #queue_monitor_state{timer=Timer}) ->
+    {ok, cancel} = timer:cancel(Timer),
```
marcparadise: Do we want to crash it if timer cancel failed? We can probably safely ignore the result and save some potentially misleading crash log spew.

metadave: :+1:

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
...
+    Config = envy:get(oc_chef_wm, ?POOLNAME, [], any),
+    [{?POOLNAME, Config}].
+
+%%-------------------------------------------------------------
+
+init([]) ->
+    % used to catch worker msgs
+    process_flag(trap_exit, true),
+    TRef = start_update_timer(),
+    {ok, #queue_monitor_state{timer=TRef, worker_process = undefined}}.
```
marcparadise: If I'm processing correctly, this means we won't have status available for 60 seconds after startup, though we will start servicing webmachine requests much sooner than that. If someone has restarted because they saw the queue filled up and panicked, this will mean lost requests for that first 60 seconds. Can we get an initial status in  the supervisor and pass it in?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
...
+    {reply, ok, State#queue_monitor_state{queue_at_capacity = AtCapacity}};
+handle_call(status, _From, State) ->
+    % return a term to be converted to JSON
+    Stats = [{queue_at_capacity,State#queue_monitor_state.queue_at_capacity},
+             {dropped_since_last_check, State#queue_monitor_state.dropped_since_last_check},
+             {max_length, State#queue_monitor_state.max_length},
+             {last_recorded_length, State#queue_monitor_state.last_recorded_length},
+             {total_dropped, State#queue_monitor_state.total_dropped}],
+    {reply, Stats, State};
+handle_call(sync_check_current_state, From, #queue_monitor_state{sync_response_process = undefined} = State) ->
```
marcparadise: If a status_ping is already in flight, this will cause a second one to be created. In that case could we update the state to capture it and not issue the message, so that the in-flight request will service the `sync_check_current_state` request?

metadave: `status_ping` only kicks off if `worker_process = undefined`.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
...
+     self() ! status_ping,
+     {noreply, State#queue_monitor_state{sync_response_process = From}};
```
marcparadise: If a secondary process does a sync_check_current_state while another process is waiting, we'll fall through and reply `ignored` and log it as unknown.  Maybe handle the case where `sync_response_process` exists?

metadave: Yes, but it's only used for testing.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
metadave: +1, but why pass it in from the supervisor? All the HTTP requests to get status are async, so doing it in the supervisor would be messy.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_status.erl
marcparadise: :+1: on changing overall_status.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
```diff
@@ -0,0 +1,473 @@
...
+%% but return 100% full if this ever happens
+calc_ratio_and_percent(CurrentLength, MaxLength) when CurrentLength > MaxLength ->
+    {1.0, 100.0};
+calc_ratio_and_percent(CurrentLength, MaxLength) ->
+    Ratio = CurrentLength / MaxLength,
+    Pcnt = round(Ratio * 100.0),
+    {Ratio, Pcnt}.
+
+-spec rabbit_mgmt_server_request(string()) -> oc_httpc:response().
+rabbit_mgmt_server_request(Path) ->
```
marcparadise: I think I'd like to see from here on down moved to its own `chef_wm_rabbitmq_management` module, with vhost and queue name being passed in where needed.  Looks like `create_pool` and `delete_pool`  should be in it too. 

This will separate out the stuff that actually talks to the external service into its own interface, which will smooth the way if/when we want to support multiple queues.(And makes it a little cleaner if we need to do an initial direct request from the supervisor before the queue monitor is started.)

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
marcparadise: Hah - I must have been searching with the ? in place

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
marcparadise: As a sync request (see other comment  about splitting out some functions)  Basically it's a way to make sure erchef itself can't start handling requests until monitoring is in place - since the application complete its start until the supervisor reports success.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
marcparadise: :+1:

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_actions_queue_monitoring.erl
metadave: ah, cool, I like it.

marcparadise:
This looks good, with some relatively minor comments above. 

My only large area of concern is the volume of requests this will be handling.  The timeout option made me think of it, but realistically if we ever hit that timeout we're already going to be in a really bad state.   

This is very lightweight as implemented, and I'm probably being overly paranoid - I'd feel better if we had decent integrated load suite that didn't require spinning up a ton of AWS nodes.

metadave:
thanks for the review @marcparadise @stevendanna, I'll clean things up on Monday.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
irvingpop: +1

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
metadave: there are 3 other pools that use similar settings. Should all pools be updated?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
metadave: fixed

omnibus/files/private-chef-cookbooks/private-chef/templates/default/rabbitmq.config.erb
metadave: I'm unsure, but I believe I can specify the IP via

```      
{ip, "<%= node['private_chef']['rabbitmq']['vip'] %>"},
```
if that works for you

omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
metadave: should this include the queue monitor settings as well? `rabbitmq_queue_length_monitor_enabled` etc

metadave:
ping @chef/lob

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -116,6 +116,54 @@
 default['private_chef']['rabbitmq']['consumer_id'] = 'hotsauce'
 default['private_chef']['rabbitmq']['env_path'] = "/opt/opscode/bin:/opt/opscode/embedded/bin:/usr/bin:/bin"
 
+default['private_chef']['rabbitmq']['ssl_versions'] = "'tlsv1.2', 'tlsv1.1'"
```
irvingpop: super minor nitpick, would this be better expressed as an `Array`?

metadave: sure, do you mean `"['tlsv1.2', 'tlsv1.1']"` or `['tlsv1.2', 'tlsv1.1']`?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -116,6 +116,54 @@
...
+# cull interval specified in seconds
+default['private_chef']['rabbitmq']['rabbit_mgmt_http_cull_interval'] = 60
+# max age specified in seconds
+default['private_chef']['rabbitmq']['rabbit_mgmt_http_max_age'] = 70
+#max connection duration specified in seconds
+default['private_chef']['rabbitmq']['rabbit_mgmt_http_max_connection_duration'] = 70
+
+# comma sep list of tuples, without surrounding []'s
+# rendered as a list in oc_erchef.config.erb, including basic_auth info
+default['private_chef']['rabbitmq']['rabbit_mgmt_ibrowse_options'] =  "{connect_timeout, 10000}"
```
irvingpop: just so I understand, if I wanted to add more options here I would change it to:   `"{option_a, 100}, {option_b, 200}"`

metadave: that's correct, because of the way oc_httpc reads basic auth info, I have to combine these into a larger options array.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -116,6 +116,54 @@
...
+# if the queue monitor is busy and this timeout has been exceeded,
+# assume that rabbit is in a bad state and don't send messages to it
+# 5000 is the default of gen_server:call()
+default['private_chef']['rabbitmq']['queue_length_monitor_timeout_millis'] = 5000
+
+# don't send messages to rabbitmq if it has reached it's configured max_length
+default['private_chef']['rabbitmq']['drop_on_full_capacity'] = true
+
+# prevent erchef from starting if queue is at capacity
+default['private_chef']['rabbitmq']['prevent_startup_on_full_capacity'] = true
```
irvingpop: This is concerning to me for two reasons:
1)  it is in the rabbitmq namespace, but it actually impacts erchef without the word `erchef` anywhere in the variable
2)  Is this the default behavior we want?   I think this comes down to a discussion on "fail open vs fail closed" -  and I believe our users would prefer we default to "fail open"  (prefer API availability)

metadave: I can change the name to `prevent_erchef_startup_on_full_capacity`. 

Regarding a default value, I'll defer to @marcparadise @stevendanna

metadave: defaulted to `false`.

marcparadise:
:+1:

PR #573, opened 2015-10-16
[dvm] List installer packages starting at 1 not 0

stevendanna:
Other parts of the installer selection code assume that the user
gave us a number starting from 1.


marcparadise:
:+1:

PR #577, opened 2015-10-21
Replace _'s in data and queries in cloudsearch mode

stevendanna:
If we don't replace `_`, then it is possible that search terms like
`foo_*` would turn up data where the value was `foo-bar` because
the search "foo-bar" will be converted to `foo__DS__bar` which will
match `foo_*`.

This commit converts `_`'s to `__US__`.  In the above example, the search
query `foo_*` would be converted to `foo__US__*` which would not match
`foo__DS__bar`.

This required special handling during query parsing since we want to
ensure we only ever replace `_`'s that were in the original data.


stevendanna:
We are going a different direction.

PR #579, opened 2015-10-23
Avoid proxies when talking to local chef-server

stevendanna:
This configuration file is used by chef-server-ctl to talk to the API
locally. Proxy configs in the users environment often cause problems
because the LB VIP is almost always 127.0.0.1, which causes the proxy to
try to connect to itself rather than back to the chef-server.


danielsdeleo:
LGTM :+1:

When @jkeiser is back I'm going to chat with him about whether we should just have this in core chef.

PR #581, opened 2015-10-23
Mp/escape to unicode

marcparadise:
Add work break tests, and take a stab at using unicode replacement characters.

WIP.


stevendanna:
Closing for now.  The new tests were pulled into the elasticsearch branch.  Thanks mark!

PR #585, opened 2015-10-27
Update mochiweb to the latest from mochi/mochiweb

irvingpop:
The primary motivation for updating mochiweb is to pull in this fix: https://github.com/mochi/mochiweb/issues/149

In local testing mochiweb appears as version `2.13.0` and passes a full pedant test (`chef-server-ctl test --all`)


irvingpop:
@stevendanna per your feedback,  I traced the mochiweb dependency:

irvingpop:
One afterthought/question:   Should I update the rebar config and lock for bookshelf and bifrost in this same PR, a different PR, or leave them alone?

irvingpop:
wilson build:  http://wilson.ci.chef.co/job/chef-server-12-build/1509/

stevendanna:
> One afterthought/question: Should I update the rebar config and lock for bookshelf and bifrost in this same PR, a different PR, or leave them alone?

We've been trying to keep core deps like this consistent across projects, so I would definitely be +1.

marcparadise:
This means we're on the wrong webmachine too.  We should be using https://github.com/webmachine/webmachine

irvingpop:
hey @marcparadise @stevendanna since the scope of this PR is creeping bigger and bigger, I propose that we keep the current PR as-is in an effort to be GEFN  (fix the issue that’s affecting customers)  and then work on the `opscoderl_wm` changes in another PR?

irvingpop:
New wilson build containing latest mochiweb for erchef, bifrost and bookshelf:  http://wilson.ci.chef.co/job/chef-server-12-build/1517/

irvingpop:
Update:   Validated the new mochiweb in erchef, bifrost and bookshelf (build 1517) in ec-metal's loadtest environment -  performed as expected.

cc: @sdelano

marcparadise:
Replaced by the changes in #592 , which pull in the new deps via updated https://github.com/chef/opscoderl_wm.

PR #587, opened 2015-10-28
Update gather-logs Reporting Info Collection

sean-horn:
* Get sqitch tags for Reporting

* Check postgres socket to make sure we are on the primary before doing queries

* Bugfix: opscode-reporting-ctl status subcommand does not exist

ChangeLog-Entry: [omnibus] gather-logs update reporting gathering

### DIFF
--- a/omnibus/files/private-chef-scripts/gather-logs
+++ b/omnibus/files/private-chef-scripts/gather-logs
@@ -138,6 +138,7 @@ then
     echo "SELECT CURRENT_TIMESTAMP; SELECT * FROM pg_stat_activity;" | chef-server-ctl psql oc_erchef --as-admin --options -tA > "$tmpdir/pg_stat_activity.txt" 2>&1
     echo "SELECT * FROM sqitch.tags;" |  chef-server-ctl psql oc_erchef --as-admin --options -tA  > "$tmpdir/opscode_chef_sqitch_tags.txt" 2>&1
     echo "SELECT * FROM sqitch.tags;" |  chef-server-ctl psql bifrost --as-admin --options -tA > "$tmpdir/bifrost_sqitch_tags.txt" 2>&1
+    echo "SELECT * FROM sqitch.tags;" |  chef-server-ctl psql reporting --as-admin --options -tA > "$tmpdir/reporting_sqitch_tags.txt" 2>&1
     if [[ -S "/tmp/.s.PGSQL.5432" ]]; then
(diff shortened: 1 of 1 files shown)


omnibus/files/private-chef-scripts/gather-logs
```diff
@@ -176,9 +177,9 @@ then
     found_manage=true
 fi
 
-if [[ -e /opt/opscode-reporting/bin/opscode-reporting-ctl ]];
+#Use socket to make sure we're on the primary in an HA system
+if [[ -e /opt/opscode-reporting/bin/opscode-reporting-ctl && -S "/tmp/.s.PGSQL.5432" ]];
```
marcparadise: Unfortunately this isn't a reliable indicator if reporting (or reporting via chef server) is  configured with external postgres.

sean-horn: @marcparadise That's right. Forgot that was a thing. Any thoughts on a way to proceed?

It's not critical that the information be gathered, but it's nice to have it.

What about basing the decision on something like this?

    /opt/opscode/embedded/bin/ruby -rjson -e "puts JSON.parse(STDIN.read)['private_chef']['postgresql']['external']" < /etc/opscode/chef-server-running.json

I noticed that you did this earlier to give our best effort and fail gracefully if it doesn't work.

```
    # postgres server may be external - we'll go ahead and try to gather what we can in any case, and just capture errors to the output
    # without showing them to the user
    echo "SELECT CURRENT_TIMESTAMP; SELECT * FROM pg_stat_activity;" | chef-server-ctl psql oc_erchef --as-admin --options -tA > "$tmpdir/pg_stat_activity.txt" 2>&1
```

marcparadise: So with the support we have in place, chef-server-ctl psql opscode- reporting (or maybe opscode_reporting) --as-admin should  connect if it's available, and otherwise not.  The drawback is that it won't work if reporting is configured with its own external DB, independent of chef server. That's not a concern now as the version that supports this was only just released yesterday, and it's not explicitly called out as a thing that you can do in any case - but at some point down the road, it may become one.

marcparadise: The ruby bit above looks good too, but would have the same limitation re: separate external reporting database.

sean-horn:
I'll go ahead and close for now and reopen when I have time to get as many of the possible cases supported as I can.

PR #588, opened 2015-10-28
dvm fixes to fix unhelpful error messages and enable successful loading of bifrost

irvingpop:
Thanks to @marcparadise for help with these.   3 specific fixes that got me unstuck:

1. `dev/dvm/bin/dvm`:   The rescue was masking the underlying failure -  loosen up what we rescue in order to always show the underlying failure
2. `dev/dvm/lib/dvm.rb`:  a class definition went missing which was masking the `raise` messages with an unhelpful message
3.  `src/oc_bifrost/Makefile`:  remove an errant `bundle` target from `oc_bifrost` so that it can be successfully loaded

### DIFF
--- a/dev/dvm/bin/dvm
+++ b/dev/dvm/bin/dvm
@@ -11,6 +11,6 @@ require "dvm"
 
 begin
   DVM::Application.start(ARGV)
-rescue DVM::DVMArgumentError => e
+rescue Exception => e
   say HighLine.color(e.message, :red)
 end
--- a/dev/dvm/bin/dvm
+++ b/dev/dvm/bin/dvm
@@ -11,6 +11,6 @@ require "dvm"
 
 begin
   DVM::Application.start(ARGV)
-rescue Exception => e
+rescue DVM::DVMArgumentError => e
   say HighLine.color(e.message, :red)
 end
--- a/dev/dvm/lib/dvm.rb
+++ b/dev/dvm/lib/dvm.rb
@@ -1,3 +1,7 @@
+module DVM
+  class DVMArgumentError < RuntimeError; end
+end
+
 require_relative "dvm/tools"
 require_relative "dvm/project"
 require_relative "dvm/application"
--- a/src/oc_bifrost/Makefile
+++ b/src/oc_bifrost/Makefile
@@ -53,7 +53,6 @@ omnibus: $(REBAR3) distclean
 	$(REBAR3) do compile, release
 
 ## For dvm
-DVM_HOOK = bundle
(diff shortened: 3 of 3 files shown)


dev/dvm/bin/dvm
```diff
@@ -11,6 +11,6 @@ require "dvm"
 
 begin
   DVM::Application.start(ARGV)
-rescue DVM::DVMArgumentError => e
```
sdelano: If it wasn't getting rescued before, where did it end up?

dev/dvm/lib/dvm.rb
```diff
@@ -1,3 +1,7 @@
+module DVM
+  class DVMArgumentError < RuntimeError; end
+end
```
sdelano: How did that block of code ever work before? Now I'm uber-confused.

dev/dvm/bin/dvm
sdelano: It seems like this would have ended raising the whole exception anyways.

marcparadise: It would/should have dumped a stack if it wasn't an error specifically thrown by dvm internals. (I overused the crap out of DVMArgumentError ).  

@irvingpop is this one still needed after the re-addition of the missing DVMArgumentError class?  Generally dvmargumenterrors are things that will have a hint as to how to fix (now that they actually exist again), while the general Exception case we'd want to see the stack trace to figure out where it broke.   

Or am I misremembering the default behavior of the thor app when it hits an unhandled exception?

dev/dvm/lib/dvm.rb
marcparadise: @sdelano my best guess is that when I changed over to avoid having to package and install a dvm gem into the chef-server runtime, I somehow nuked that definition - because since then, we've been getting singularly unhelpful undefined class errors instead of something useful.

marcparadise:
:+1: after confirmation that we do still want to consume all exceptions without printing a stack for those that didn't originate directly from explicit dvm `raise`es.

marcparadise:
Forgot to add - thanks for making this PR @irvingpop , some of this had been written down on my 'things to follow up' list for about two weeks, and kept getting bumped.

dev/dvm/bin/dvm
irvingpop: Thanks @marcparadise @sdelano now that the underlying raise is working in `dvm/lib/dvm.rb` this isn't needed, so I put it back.

stevendanna:
:+1:

sdelano:
LGTM

PR #591, opened 2015-10-30
ChangeLog-Entry: [oc_erchef] queue monitor doesn't affect overall_sta…

metadave:
…tus by default

add queue_at_capacity_affects_overall_status boolean attribute

cc @chef/lob @irvingpop


irvingpop:
:+1: from a cookbook/configuration perspective

marcparadise:
:+1:

PR #592, opened 2015-10-30
update rebar locks for bookshelf, bifrost, and erchef

marcparadise:
Pulls in mochiweb from the "mochi" org via updated `opscoderl_wm`.  Also updating other deps for latest fixes and enhancements, since it's been a while.  

Replaces https://github.com/chef/chef-server/pull/585

ChangeLog-Entry: [oc_erchef] [bookshelf] [oc_bifrost] pull in the latest
webmachine and mochiweb dependencies to resolve an issue which could
lead to requests being rejected under sudden load.

ping @chef/lob


stevendanna:
:+1: iff pedant passes.  It is a lot of dependency change but the majority of these have been safe-to-upgrade in the past.

marcparadise:
Yeah, we did have a clean pedant run.  For future releases, I'd like to make updating rebar locks the first thing we do *after* we cut a new release - sot hat we have more time to run with them -  instead of the last thing we do *before* we cut a release.

stevendanna:
@marcparadise :+1: on that idea.  We should add it to the release process.  Integrate as early as possible with a set up update deps.

PR #608, opened 2015-11-05
allow dvm to create an external reporting db vm

metadave:


marcparadise:
:+1: one day fairly soon I'm going rework external db so that it's just an lxc container in the same vm, but until then ....

metadave:
thanks!

PR #609, opened 2015-11-05
missing dev template from PR 608

metadave:
forgot to `git add` this file for https://github.com/chef/chef-server/pull/608


marcparadise:
:+1:

PR #611, opened 2015-11-06
Don't create unused SERVICE/etc dirs for erlang services

stevendanna:
Nothing lives in the etc/ directories since the move to rebar3.


marcparadise:
:+1: once it has a changelog-entry.

PR #612, opened 2015-11-06
Follow filename convention for migration 26

stevendanna:


marcparadise:
:+1:

marcparadise:
I'm starting to feel like it's time to disable the ct /eunit for the mover stuff pulled in via dependency.

PR #613, opened 2015-11-06
Don't run eunit on mover's vendored code.

marcparadise:
We don't modify this, and it's currently preserved from a fixed
version to ensure mover's back-compatibility.  Because it's a regular
source of intermittent CI failures that we can't address(because we
can't change the vendored code), this change explicitly tests
only the parts of mover that can change.

ChangeLog-Entry: [chef-mover] do not run intermittently failing tests on dependencies
that we can't change for backward-compatibility reasons.

### DIFF
--- a/src/chef-mover/Makefile
+++ b/src/chef-mover/Makefile
@@ -30,7 +30,9 @@ distclean:
 install: $(REBAR3) distclean
 	$(REBAR3) update
 
-travis: all
+travis: $(REBAR3)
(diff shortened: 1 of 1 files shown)


marcparadise:
ping @chef/lob

stevendanna:
I'm :+1: We'll have to remember to run these tests manually, however, if we do ever need to fix something in that vendored code.  However, these tests are currently failing regularly since we haven't backported stability fixes, so I think it is working

PR #614, opened 2015-11-09
Add option to set rspec seed value

manderson26:
Rspec by default runs tests in a random order, which normally is
good. However sometimes bugs manifest themselves as state leftover from
prior tests, and it's hard to sort those out when the order changes
every time.

Add a --seed flag to pedant to set the rspec seed value.


marcparadise:
:+1:

PR #622, opened 2015-11-16
Pedant changes to support `chef-zero`

randomcamel:


danielsdeleo:
Agree w/ @stevendanna here, it's unclear to me what was wrong before.

jkeiser:
@stevendanna @danielsdeleo I've updated the commit message, and added in the "run pedant against chef-zero on every checkin" checkin to boot. I'll send through wilson this morning.

.gitignore
```diff
@@ -1,3 +1,6 @@
 .DS_Store
 erln8.config
 config.yml
+.bundle
+vendor
+Gemfile.lock
```
sdelano: Unless you're willing to fix everything that breaks every time something gets checked into chef/chef-zero (or any other dependent gem), please do not remove the Gemfile.lock from a piece of software that we use to run integration tests (both in CI and on-site with customers).

What benefit do we get from unlocking the gems? There's a very good reason there's a Gemfile.lock in this repo.

jkeiser:
Wilson has passed. Merging.

PR #624, opened 2015-11-17
Relax dependencies so that chef and chef-pedant no longer conflict

jkeiser:
Chef pins rspec to ~> 3.3, we pin to ~> 3.2. This lets us float further.


manderson26:
Looks pretty reasonable to me.

marcparadise:
Looks good, but I do have one question.Will we not use our version of rest-client with this change? I seem to recall upstream requires tlsv3 and breaks some stuff . maybe also be some ipv6 stuff in there iirc?

jkeiser:
@marcparadise this won't change where it gets rest-client from, just relaxes the versions it can use (which seems useless for rest-client, at least, but ultimately harmless.)

sdelano:
@marcparadise - In the Gemfile.lock it also doesn't appear to change the version of `rest-client` that we pull in. It's still the same.

sdelano:
@jkeiser - @marcparadise makes a good point though, we need at least `>= 1.6.7` for `rest-client`. You've actually opened up the versions in the wrong  direction by allowing `>= 1.6`. I understand the methodology behind the other gems that use the spermy operator, but for GTE, it doesn't make sense.

jkeiser:
@sdelano yep! I'll change it.

marcparadise:
:+1: thanks

tylercloke:
:+1:

lamont-granquist:
chef no longer depends on rest-client, ripped that out in 12.x and the dep pin in 11.x is unlikely to ever change.

stevendanna:
I believe this PR has already been merged as part of another branch. Closing this to avoid confusion and wasted time.  Please reopen if I'm confused.

lamont-granquist:
Yeah the failure on the chef/chef repo magically disappeared...

PR #627, opened 2015-11-19
update minimal cookbook versions query to force query plan

sdelano:
This query uses two CTEs (inputs and min_cve) to force the query planner
to first join cookbook_versions (min, maj, patch) with cookbooks (id, name)
before filtering with inputs.

Without the min_cv CTE, the query planner would choose to join inputs with
cookbook_versions before filtering by name. In the worst case scenario (all
cookbooks have the same version number), the inputs and cookbook_versions
join behaved like a cartesian product of all the inputs and the cookbook versions.

/cc @chef/lob

### DIFF
--- a/src/oc_erchef/apps/chef_db/priv/pgsql_statements.config
+++ b/src/oc_erchef/apps/chef_db/priv/pgsql_statements.config
@@ -363,19 +363,31 @@
 %% Used in implementation of environemnts/ENVNAME/cookbook_versions endpoint.
 %%
 %% Depends on the cbv type existing in the schema
+%%
+%% This query uses two CTEs (inputs and min_cve) to force the query planner
+%% to first join cookbook_versions (min, maj, patch) with cookbooks (id, name)
+%% before filtering with inputs.
+%%
+%% Without the min_cv CTE, the query planner would choose to join inputs with
+%% cookbook_versions before filtering by name. In the worst case scenario (all
+%% cookbooks have the same version number), the inputs and cookbook_versions
+%% join behaved like a cartesian product of all the inputs and the cookbook versions.
 {bulk_fetch_minimal_cookbook_versions,
- <<"WITH inputs AS ( select * from unnest($2::cbv[]) )
-  SELECT v.frozen, v.meta_deps, v.metadata, v.serialized_object,
-         c.authz_id, c.org_id, c.name
+ <<"WITH inputs AS ( SELECT * FROM unnest($2::cbv[]) ),
+         min_cv AS ( SELECT v.frozen, v.meta_deps, v.metadata, v.serialized_object,
+                            v.major, v.minor, v.patch, c.authz_id, c.org_id, c.name
+                       FROM cookbooks as c,
+                            cookbook_versions as v
+                      WHERE c.org_id = $1
+                        AND c.id = v.cookbook_id )
+  SELECT min_cv.frozen, min_cv.meta_deps, min_cv.metadata, min_cv.serialized_object,
+         min_cv.authz_id, min_cv.org_id, min_cv.name
     FROM inputs,
-         cookbooks as c,
-         cookbook_versions as v
-   WHERE c.org_id = $1
-     AND inputs.name = c.name
-     AND v.cookbook_id = c.id
-     AND v.major = inputs.major
-     AND v.minor = inputs.minor
-     AND v.patch = inputs.patch">>}.
+         min_cv
+   WHERE inputs.name = min_cv.name
+     AND inputs.major = min_cv.major
+     AND inputs.minor = min_cv.minor
+     AND inputs.patch = min_cv.patch">>}.
 
 %% Used in implementation of environments/ENVIRONMENT/recipes endpoint
 %%


sdelano:
Here are some details on the worst case scenario mentioned in the PR:

### Worst Case

* 201 cookbooks, `cookbook_(1..200)` and `master`
* `master` includes all other cookbooks
* no other dependencies
* all cookbooks are at version `1.1.1`

Before this PR, this is what `explain analyze` on the minimal cookbook versions query returns for this data set:

```
                                                              QUERY PLAN                                                               
---------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=4.75..28.42 rows=1 width=578) (actual time=3.825..418.399 rows=201 loops=1)
   Join Filter: (inputs.name = c.name)
   Rows Removed by Join Filter: 40200
   CTE inputs
     ->  Function Scan on unnest  (cost=0.00..1.00 rows=100 width=56) (actual time=0.062..0.350 rows=201 loops=1)
   ->  Hash Join  (cost=3.75..27.03 rows=1 width=536) (actual time=1.541..68.215 rows=40401 loops=1)
         Hash Cond: ((v.major = inputs.major) AND (v.minor = inputs.minor) AND (v.patch = inputs.patch))
         ->  Seq Scan on cookbook_versions v  (cost=0.00..21.01 rows=201 width=528) (actual time=0.005..0.636 rows=201 loops=1)
         ->  Hash  (cost=2.00..2.00 rows=100 width=56) (actual time=1.479..1.479 rows=201 loops=1)
               Buckets: 1024  Batches: 1  Memory Usage: 15kB
               ->  CTE Scan on inputs  (cost=0.00..2.00 rows=100 width=56) (actual time=0.067..1.056 rows=201 loops=1)
   ->  Index Scan using cookbooks_pkey on cookbooks c  (cost=0.00..0.37 rows=1 width=82) (actual time=0.003..0.005 rows=1 loops=40401)
         Index Cond: (id = v.cookbook_id)
         Filter: (org_id = 'bb996f532e8358779d236750526af7c8'::bpchar)
 Total runtime: 418.822 ms
(15 rows)
```

After this patch is applied, the query runs as such:

```
                                                                            QUERY PLAN                                                                            
------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Hash Join  (cost=36.80..43.85 rows=1 width=393) (actual time=2.462..6.652 rows=201 loops=1)
   Hash Cond: ((minimal_cv.name = inputs.name) AND (minimal_cv.major = inputs.major) AND (minimal_cv.minor = inputs.minor) AND (minimal_cv.patch = inputs.patch))
   CTE inputs
     ->  Function Scan on unnest  (cost=0.00..1.00 rows=100 width=56) (actual time=0.048..0.349 rows=201 loops=1)
   CTE minimal_cv
     ->  Hash Join  (cost=8.03..31.80 rows=201 width=602) (actual time=0.905..3.174 rows=201 loops=1)
           Hash Cond: (v.cookbook_id = c.id)
           ->  Seq Scan on cookbook_versions v  (cost=0.00..21.01 rows=201 width=528) (actual time=0.005..0.334 rows=201 loops=1)
           ->  Hash  (cost=5.51..5.51 rows=201 width=82) (actual time=0.851..0.851 rows=201 loops=1)
                 Buckets: 1024  Batches: 1  Memory Usage: 23kB
                 ->  Seq Scan on cookbooks c  (cost=0.00..5.51 rows=201 width=82) (actual time=0.014..0.440 rows=201 loops=1)
                       Filter: (org_id = 'bb996f532e8358779d236750526af7c8'::bpchar)
   ->  CTE Scan on minimal_cv  (cost=0.00..4.02 rows=201 width=417) (actual time=0.913..4.233 rows=201 loops=1)
   ->  Hash  (cost=2.00..2.00 rows=100 width=56) (actual time=1.501..1.501 rows=201 loops=1)
         Buckets: 1024  Batches: 1  Memory Usage: 15kB
         ->  CTE Scan on inputs  (cost=0.00..2.00 rows=100 width=56) (actual time=0.052..1.026 rows=201 loops=1)
 Total runtime: 7.085 ms
(17 rows)
```

### Best Case

On the best case scenario, where each cookbook has a unique version, both queries return within 5-7ms on my laptop in the development VM.

### Normal Case

This case isn't tested. The best case above includes one cookbook version for each cookbook and the version numbers are unique across the space. The case that isn't tested is the "normal" case, where each cookbook has multiple cookbook versions and they could be overlapping with other cookbooks. Before this patch, there would be the potential to "cross polinate" (my term) the join of `inputs` and `cookbook_versions`, where we want the result for cookbook `foo` at `1.2.3` but cookbooks `bar`,`baz`,and `bat` are also available at that version. We're still going to join their data before filtering it by name in the next step. Given the above, I'm strongly confident that this patch improves the normal case for this query as well.

sdelano:
Credit goes to @jamesc for the multi-CTE magic.

manderson26:
Still parsing the change; but one thought is that your breakdown of the change times (in fact your whole second post above) would be useful to have in the actual commit message.

marcparadise:
:+1: , agree with including the rational in the commit message. 

There is an alternative that may also be worth checking to see if how it performs in comparison: Keep the query the same, but prefix the statement with `SET LOCAL join_collapse_limit=1;`.  This enforces applying the joins in the order we specify, for the duration of the current transaction.

manderson26:
:+1: for merge.

stevendanna:
:+1:

sdelano:
@marcparadise I definitely tried `SET LOCAL join_collapse_limit=1` to force the query planner but it didn't help. I even added explicit joins in. Although thinking about how to apply the `SET LOCAL` made me think that we'd have to wire it all the way through `epgsql` as well (like timeouts). I can drop you a `chef-server-ctl backup` and the query I was running if you'd like to play around yourself.

marcparadise:
Yeah, turns out we couldn't include it directly in the prepared statement itself.  It wouldn't have to be wired all the way through pgsql, but sqerl would definitely have to be taught about it. 

Bright side, now I have another creator-day project - allow sqerl to understand that prepared statements may include  transaction-local config vars as an additional parameter ... 

That aside, seems like it won't benefit in this case anyway. 

:+1: on the change

PR #628, opened 2015-11-19
Update to the latest rebar3

stevendanna:


stevendanna:
Still a WIP.  Working through various dependency/plugin issues

PR #629, opened 2015-11-19
Pin rebar3_neotoma_plugin until we can upgrade to rebar3-beta4

stevendanna:


stevendanna:
This is really suboptimal.  We should try to get on the latest rebar3-beta4 ASAP.  The goal of this is to unblock master in the near term.

sdelano:
:+1: :shipit:

Issue #50, opened 2014-12-23
Impossible to upload cookbooks if non-standard port is used


oleg-z:
My chef-server.rb:

```
nginx['non_ssl_port'] = 8081
nginx['ssl_port'] = 4000
```

After running chef-server-ctl reconfigure I can't upload cookbook: 

```
INFO: Uploading /var/default/chef-data/cookbooks/squeeze64-1.13.0/yum/templates/default/main.erb (checksum hex = e5ab84fe45a83c038ff442722be03dbd) to https://127.0.0.1:4000/bookshelf/organization-55800a42d41ca4067b8d9cc3f9d1ab51/checksum-e5ab84fe45a83c038ff442722be03dbd?AWSAccessKeyId=e737796ac367dbe4a94c96ad3ed439d9a3099d17&Expires=1419376039&Signature=clYmIZ0ViZA3dXgosSXDTiH9LfA%3D
INFO: HTTP Request Returned 500 Internal Server Error: internal service error
ERROR: Server returned error for https://127.0.0.1:4000/organizations/default/sandboxes/9cc3f9d1ab51f248e5ce1ccd3913b6da, retrying 1/5 in 3s
INFO: HTTP Request Returned 500 Internal Server Error: internal service error
ERROR: Server returned error for https://127.0.0.1:4000/organizations/default/sandboxes/9cc3f9d1ab51f248e5ce1ccd3913b6da, retrying 2/5 in 8s                                              
INFO: HTTP Request Returned 500 Internal Server Error: internal service error                                                                                                                
ERROR: Server returned error for https://127.0.0.1:4000/organizations/default/sandboxes/9cc3f9d1ab51f248e5ce1ccd3913b6da, retrying 3/5 in 10s
```

Reason:
incorrect erchef template which assumes that default protocol port is used. Attempt to specify vip parameter with port (e.g. 1.1.1.1:4000) causes issue because normalize_host method parses specified string as IPv6

```
{chef_objects, [
                  {s3_access_key_id, "<%= node['private_chef']['bookshelf']['access_key_id'] %>"},
                  {s3_secret_key_id, "<%= node['private_chef']['bookshelf']['secret_access_key'] %>"},
                  {s3_url, "<%= node['private_chef']['nginx']['x_forwarded_proto'] %>://<%= @helper.vip_for_uri('bookshelf') %>"},
```

stevendanna:
[...]

PR #637, opened 2015-11-20
Use the correct ssl port for bookshelf and pedant when it's not default

marcparadise:
1. Update erchef's `s3_url` to properly reflect the ssl port in use
2. Similarly update pedant's `api_url`. 

Addresses #50

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
@@ -1,6 +1,6 @@
 %% -*- mode: erlang -*-
 %% -*- tab-width: 4;erlang-indent-level: 4;indent-tabs-mode: nil -*-
-%% ex: ts=4 sw=4 ft=erlang et
+%% ex: ts=4 sw=4 ft=eruby.erlang et
 [
  %% SASL config
  {sasl, [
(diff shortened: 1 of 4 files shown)


omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
```diff
@@ -206,7 +206,7 @@
   {chef_objects, [
                   {s3_access_key_id, "<%= node['private_chef']['bookshelf']['access_key_id'] %>"},
                   {s3_secret_key_id, "<%= node['private_chef']['bookshelf']['secret_access_key'] %>"},
-                  {s3_url, "<%= node['private_chef']['nginx']['x_forwarded_proto'] %>://<%= @helper.vip_for_uri('bookshelf') %>"},
+                  {s3_url, "<%= "#{@s3_proto}://#{@helper.vip_for_uri('bookshelf')}:#{NginxErb.new(node).port_for_proto(@s3_proto)}"%>"},
```
stevendanna: I think this will be incorrect for people who have configured their Chef Server to point at AWS?

stevendanna: Also, but less likely, it would break in cases where the frontends have been reconfigured to listen on a non-standard port, but the backend is still in the default configuration.

marcparadise: I think we use `s3_external_url` for thecase but I will confirm.

I'll look into both though , to be sure.

stevendanna: I think s3_external_url is the url we use to construct s3 links while s3_url is the url used to talk to s3 (for instance, HEAD requests to confirm files are there).

PR #638, opened 2015-11-21
Restrict 'other' permissions for chef-server.rb

itmustbejj:
This PR removes all permissions for 'other' on /etc/opscode/chef-server.rb. This file can hold sensitive data such as the LDAP bind_password, so it should not be accessible by anyone other than owner/group (both root in this case).


stevendanna:
@itmustbejj Looks good, mind updating your commit message to include a line at the end with:

```
ChangeLog-Entry: Restrict permissions on /etc/opscode/chef-server.rb
```

PR #639, opened 2015-11-21
Restrict 'other' permissions for chef-server.rb.

itmustbejj:
This PR removes all permissions for 'other' on /etc/opscode/chef-server.rb. This file can hold sensitive data such as the LDAP bind_password, so it should not be accessible by anyone other than owner/group (both root in this case).

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
@@ -72,6 +72,11 @@ else
   end
 
   if chef_server_rb_exists
+    # Restrict 'other' permissions on chef-server.rb as it may contain sensitive info,
+    # such as the LDAP bind password. Using FileUtils to fix, as a file resource only
+    # sets permissions in absolute mode. This addresses config files created before the
+    # fix was implemented in the postinst script.
+    FileUtils.chmod('o-rwx', chef_server_path)
     PrivateChef.from_file(chef_server_path)
   end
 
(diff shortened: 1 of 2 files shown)


stevendanna:
:+1: looks good

tylercloke:
:+1: Let's ge this in before our next release.

stevendanna:
@itmustbejj Do you mind rebasing against master?  Once that is done I'll merge this.

stevendanna:
@itmustbejj I went to rebase this myself and had a thought.  Do we want to fix this for existing installs as well?  We can potentially do this in this recipe: https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb

Thoughts @chef/chef-server-maintainers

itmustbejj:
@stevendanna Rebased and added a commit in the config.rb you mentioned above for existing config files.

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
```diff
@@ -51,6 +51,7 @@
 
   if chef_server_rb_exists
     chef_server_rb_not_empty = !File.zero?(chef_server_path)
+    FileUtils.chmod('o-rwx', chef_server_path)
```
stevendanna: I think we probably want to use a `file` resource.  It would probably also be a bit better to do the permissions fixup furhter down after all the symlinking and special handling is over.

itmustbejj: Using File.chmod requires permissions to be set in absolute mode. I figured it was more pragmatic to remove 'other' permissions than to explicitly set all permission for owner, group, and other.

I agree on moving it down after symlinks and whatnot.

stevendanna: That explanation makes sense. Maybe add a comment that mentions it

PR #641, opened 2015-11-24
don't monitor rabbit queue length w/ actions disabled

sdelano:
When setting `dark_launch['actions'] = false` in
`/etc/opscode/chef-server.rb`, the actions queue is never created and
the rabbitmq actions queue monitoring code returns a 500 on `/_status`
when it's unable to connect to the queue. This PR disabled the
monitoring behavior when the actions feature is disabled.


irvingpop:
:+1:

metadave:
+1

marcparadise:
:+1:

PR #642, opened 2015-11-25
Remove bksw_sync module

stevendanna:
We aren't using this module and its exec dependency is a constant source
of development main.


stevendanna:
cc @chef/platform-engineering We implemented this a while ago but I think that we are gonna end up going a different direction and the `exec` module this depends on is a constant source of pain.

marcparadise:
:+1:

PR #644, opened 2015-11-25
Organize RAML and add all endpoints

jkeiser:
This does not include all endpoint methods or schemas, but it does include a lot more.  Had to reorganize a bit for sharing of schemas.


src/oc_erchef/raml-docs/Makefile
```diff
@@ -12,7 +12,7 @@ else
 		echo "You don't have raml2html, but we'll assume you want it" ; \
```
stevendanna: I think the test in this if is wrong. I never got this error message and rather got a command not found error on line 15.

The problem can be solved either by using the `-z` test rather than `-f` or by putting quotes around the variable before passing it to the -f test.

stevendanna:
My only concern with this is causing confusion about what the "official" API doc is.  We also have: https://docs.chef.io/api_chef_server.html.  We should add some works to one of our README-style docs that describe how we manage this documentation and how to update the main docs site based on changes here.

jkeiser:
@stevendanna I'm working with @jamescott to get the RAML-generated docs onto docs.chef.io via a Makefile that pulls chef-server and generates the HTML from it.

PR #645, opened 2015-11-30
Implement RFC 14 - Add universe endpoint

thommay:
Implementation based on https://docs.chef.io/api_cookbooks_site.html#universe and https://github.com/chef/chef-rfc/blob/master/rfc014-universe-endpoint.md

Questions:
 - Should we think about caching the compiled JSON? Supermarket does cache, but it has a much more complex and larger set of cookbooks, and is returning the same json consistently to all users.
 - How can I write this better?
 - Does it need eunit tests as well as pedant?

cc @lamont-granquist


thommay:
@chef/lob

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_universe.erl
```diff
@@ -0,0 +1,100 @@
...
+                              {wm_req(), chef_wm:base_state()}.
+validate_request('GET', Req, #base_state{organization_guid = OrgId, server_api_version = ApiVersion} = State) ->
+    {Req, State#base_state{resource_state = #oc_chef_organization{id = OrgId, server_api_version = ApiVersion}}}.
+
+auth_info(Req, State) ->
+    {{container, cookbook}, Req, State}.
+
+to_json(Req, #base_state{chef_db_context = DbContext,
+                         organization_guid = OrgId} = State) ->
+  DependencyList = chef_db:fetch_all_cookbook_version_dependencies(DbContext, OrgId),
```
stevendanna: It's unfortunate IMHO that chef_db transforms `{ok, Result}` -> `Result`, since it means we can't match on the `ok` here.  Rather, we will end up passing the error tuple along meaning that a database problem is going result in a stack trace whose issues is `bad generator` down at line 97.  Given the small size of the module, it isn't a huge problem now.    One possibility is to use a case statement here to handle the error.  I'll see if we have a good example of that somewhere.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_universe.erl
```diff
@@ -0,0 +1,100 @@
...
+  Universe = make_universe(Req, DependencyList),
+  {chef_json:encode({Universe}), Req, State}.
+
+make_dependencies(Dependencies) ->
```
stevendanna: These sort of small functions are often straightforward to write a dialyzer spec for and they help point out issues like the one I mentioned above.  Thoughts on adding type specs to make_dependencies, make_version_list, and make_universe?

stevendanna:
> Does it need eunit tests as well as pedant?

I'm OK with no unit tests for our top-level webmachine modules provided we've covered all the relevant behavior in pedant.  Double so if we have some dialyzer specs on some of the trickier functions.

stevendanna:
> Should we think about caching the compiled JSON? Supermarket does cache, but it has a much more complex and larger set of cookbooks, and is returning the same json consistently to all users.

I assume supermarket is using a timeout-based expiration for the cache?  How often is that cache refreshed?  My worry is that the cache won't buy us much but will add complexity to a rather simple endpoint.  As you mentioned, supermarket serves the same response to everyone.  For Chef Server, a single org would have to hit the endpoint often to take advantage of the cache unless the timeout is very long.

If this endpoint becomes a performance problem, part of me would rather explore improving the postgresql queries we use such that we don't have to do so much json munging in erlang and can just directly deliver the response of the database.  This could perhaps be via a materialized view, which would be pretty similar to caching the results.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_universe.erl
```diff
@@ -0,0 +1,100 @@
...
+                           malformed_request/2,
+                           ping/2,
+                           forbidden/2,
+                           is_authorized/2,
+                           service_available/2]}]).
+
+-export([allowed_methods/2,
+        to_json/2 ]).
+
+
```
stevendanna: extra newline here

oc-chef-pedant/spec/api/universe_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    it "is successful" do
+      expect(universe.code).to eq(200)
+    end
+
+    it "returns the expected results" do
+      response_obj = parse(universe.body)
+      expect(response_obj).to be_a_kind_of(Hash)
+      expect(response_obj["foo"]["1.2.3"]).to be_a_kind_of(Hash)
```
stevendanna: Rather than digging into the response object and making assertions, can we define what the entire response should look like with something like `expect(response_obj).to eq(expected_resp)`.  This will make it easier to look at the spec and understand what we expect to see on the wire.

thommay:
> I assume supermarket is using a timeout-based expiration for the cache? How often is that cache refreshed?

Supermarket expires the cache every time a new cookbook is uploaded, rather than doing a time based expiration. 

> If this endpoint becomes a performance problem, part of me would rather explore improving the postgresql queries we use such that we don't have to do so much json munging in erlang and can just directly deliver the response of the database.

I definitely agree that I'm not sure the endpoint will ever get hit enough to make it worthwhile, and I was also kind of assuming that the depsolver queries would be reasonably efficient given they're hit regularly.

thommay:
@ssd fixed your concerns, I think.

tylercloke:
Hey, you are also gonna wanna rebase on master and update the new API RAML documentation with your new endpoint [here](https://github.com/chef/chef-server/tree/master/src/oc_erchef/raml-docs).

thommay:
@tylercloke @stevendanna rebased, added yaml docs. How does changelog work now?

src/oc_erchef/raml-docs/universe.yml
```diff
@@ -0,0 +1,36 @@
+#%RAML 0.8
+---
+description: |
+  Universe API
+
+  The universe is the known collection of cookbooks that have been uploaded
+  to the organisation.
```
stevendanna: organization :D

stevendanna:
@thommay For the changelog, add a line in your commit message like:

```
ChangeLog-Entry: Your desired message here
```

This is a big enough feature (in terms of user impact) that you might want also want to add some proposed text to RELEASE_NOTES.md under a new heading for the unreleased version.

src/oc_erchef/raml-docs/universe.yml
thommay: GAH GAH GAH. Organization gets me EVERY SINGLE TIME.

stevendanna:
:+1: from me

.travis.yml
```diff
@@ -73,7 +73,7 @@ matrix:
       before_install: rm -f oc-chef-pedant/Gemfile.lock
       before_script: cd oc-chef-pedant
       script: bundle exec rake chef_zero_spec
-      env: "GEMFILE_MOD=\"gem 'rake'; gem 'chef-zero', github: 'chef/chef-zero'\""
+      env: "GEMFILE_MOD=\"gem 'rake'; gem 'chef-zero', github: 'chef/chef-zero', branch: 'tm/ignore-pedant-universe'\""
```
tylercloke: Do we want to merge with this, or do we need to merge down a chef-zero branch before merging?

tylercloke:
:+1: besides single comment on gemfile.

.travis.yml
thommay: that's in a separate commit so we can just nuke it and then merge. The relevant chef-zero PR is https://github.com/chef/chef-zero/pull/176 .

stevendanna:
This has 2 :+1: and is ready to merge.  @thommay Would you mind rebasing against master and then merging when ready.

joerg:
Hi,

Now that this version of the chef server has been released I was playing around with the universe endpoint and found that it is authenticated (knife raw /universe works while curl localhost/universe does not). Usually, universe endpoints are public and the berkshelf-api-client [1] also has no means of authenticating this request. I did not want to open up another issue, that is why I am asking this here: Bug or feature?
If it is a feature, then it explicitly does not work with Berkshelf, which should be documented somewhere.

[1] https://github.com/berkshelf/berkshelf-api-client/blob/master/lib/berkshelf/api_client/connection.rb

thommay:
Feature. See https://discourse.chef.io/t/chef-server-12-4-0-and-12-4-1-have-been-released/7742/8 for further details

PR #655, opened 2015-12-05
Compare parsed json, not exact body-with-whitespace

jkeiser:


tylercloke:
Looks reasonable :+1:

PR #657, opened 2015-12-09
Add nodes endpoint methods

jkeiser:


stevendanna:
:+1: Looks reasonable

PR #658, opened 2015-12-09
Ma/bookshelf on pgsql

manderson26:
WIP for review only at this time.


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -598,6 +598,12 @@
 default['private_chef']['bookshelf']['secret_access_key'] = "generated-by-default"
 # Default: set to Host: header. Override to hardcode a url, "http://..."
 default['private_chef']['bookshelf']['external_url'] = :host_header
+default['private_chef']['bookshelf']['db_pool_size'] = 20
+default['private_chef']['bookshelf']['db_pool_queue_max'] = 20
+default['private_chef']['bookshelf']['db_pooler_timeout'] = 2000
+default['private_chef']['bookshelf']['sql_db_timeout'] = 5000
```
metadave: can a `_millis` suffix be added to `db_pooler_timeout` and `sql_db_timeout` so the unit is explicit?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/bookshelf.config.erb
```diff
@@ -82,7 +82,35 @@
...
+ {sqerl, [
+           {db_driver_mod, sqerl_pgsql_client},
+           {ip_mode, [ <%=PrivateChef['use_ipv6'] ? "ipv6,ipv4" : "ipv4" %> ] },
+           %% Database connection parameters
+           {db_host, "<%= @postgresql['vip'] %>"},
+           {db_port, <%=  @postgresql['port'] %>},
+           {db_user, "<%= @postgresql['sql_user'] %>"},
+           {db_pass, "<%= @postgresql['sql_password'] %>"},
+           {db_name, "bookshelf" },
+           {idle_check, 10000},
```
metadave: should this be configurable?

src/bookshelf/src/bksw_sql.erl
```diff
@@ -0,0 +1,274 @@
...
+        Error ->
+            Error
+    end.
+
+mark_file_done(DataId, Size, Chunks, SumMD5, SumSha256, SumSha512) ->
+    case sqerl:statement(update_file_data_done, [DataId, Size, Chunks, SumMD5, SumSha256, SumSha512], count) of
+        {ok, 1} ->
+            ok;
+        Error ->
+            Error
```
metadave: what happens to the file data if `update_file_data_done` fails? Should there be a way to manage orphaned file data?

metadave:
is it possible for users to access/upload files in orgs they don't have access to?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
stevendanna: If we add _millis here we should add them to all of the sql_db_timeout configurables and make sure existing config files still work.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/bookshelf.config.erb
```diff
@@ -82,7 +82,35 @@
...
+           {column_transforms,
+                              [{<<"created_at">>,
+                                {sqerl_transformers, convert_YMDHMS_tuple_to_datetime}},
+                               {<<"updated_at">>,
+                                {sqerl_transformers, convert_YMDHMS_tuple_to_datetime}}]
+           }
+          ]},
+ {pooler, [
+           {pools, [[{name, sqerl},
+                     {max_count, 1 },
```
stevendanna: Why are we setting this to 1 vs db_pool_size?

src/bookshelf/src/bksw_sql.erl
```diff
@@ -0,0 +1,274 @@
...
+    end.
+
+delete_bucket(BucketName) ->
+    case sqerl:statement(delete_bucket, [BucketName], count) of
+        {ok, 1} ->
+            ok;
+        Error ->
+            Error
```
stevendanna: There is some inconsistency around whether functions in the module return `{error, Error}` or `Error`.  Similarly, for `{ok Result}` vs `Result`.  My preference would be that we *always* return tagged tuples so that callers can use a pattern match to ensure errors don't unexpectedly propagate unhandled too far from their source.

stevendanna:
What is the upgrade strategy here?  Since we have pulled out all of the filesystem handling code I'm assuming we are aiming at a one-time partybus upgrade rather than some sort of lazy upgrade?

marcparadise:
I apologize in advance for the comment I'm about to leave :D

marcparadise:
I'd like to see consistency with our other schemas and their naming, mostly where the table name doesn't repeat in the primary key field name - eg "bucket.id" vs "bucket.bucket_id",etc.

src/bookshelf/src/bksw_sql.erl
```diff
@@ -0,0 +1,277 @@
+%% @copyright 2015 Chef, Inc. All Rights Reserved
```
stevendanna: I think we usually go `Chef Software, Inc` in copyright notices.

src/bookshelf/src/bksw_sql_io.erl
```diff
@@ -0,0 +1,80 @@
+%% Copyright 2012 Opscode, Inc. All Rights Reserved.
```
stevendanna: A file from the past! :D

src/bookshelf/src/bksw_wm_sql_bucket.erl
```diff
@@ -0,0 +1,88 @@
+%% @copyright 2012-2013 Opscode, Inc. All Rights Reserved
+%% @author Tim Dysinger <dysinger@opscode.com>
```
stevendanna: Seems unlikely ;)

stevendanna:
An even more glorious pull request is now tracking this.

PR #660, opened 2015-12-10
groups, principals, search and org members endpoints

jkeiser:


src/oc_erchef/raml-docs/organizations/search.yaml
```diff
@@ -1,8 +1,108 @@
+description: |
+  A list of the indexes that can be searched. The indices include
```
marcparadise: I think this might be missing context? What endpoint provides the list of indices?

src/oc_erchef/raml-docs/organizations/search.yaml
```diff
@@ -1,8 +1,108 @@
...
+        "a": {
+          "b": {
+            "c": "hi"
+          }
+        }
+      }
+    }
+    ```
+
+    Will be returned if you search for d:hi, *or* a_b_c:hi.
```
marcparadise: should be `c:hi`?

marcparadise:
A couple of minor comments, but :+1: this is awesome!

src/oc_erchef/raml-docs/organizations/search.yaml
jkeiser: @marcparadise the endpoint declaration is in index.raml (that's just the way it has to be broken up, AFAICT).

PR #664, opened 2015-12-11
Update rspec-rerun to latest to get rid of backtrace issues

jkeiser:


oc-chef-pedant/lib/pedant/config.rb
```diff
@@ -80,19 +78,20 @@ def self.rspec_args
...
         # even if they all pass!
-        FileUtils.rm(::RSpec::Rerun::Formatters::FailuresFormatter::FILENAME, :force => true)
-      end
+        FileUtils.rm(::RSpec::Rerun::Formatter::FILENAME, :force => true)
 
-      _test_dirs = test_directories
-      puts "Running tests from the following directories:", _test_dirs
-      args.concat _test_dirs
+        # This is the set of tests we're running.
+        _test_dirs = test_directories
```
tylercloke: Why is this an underscore variable?

manderson26:
Looks pretty reasonable to me. I notice that it removes the ability to rerun with a list of directories. Not sure that it matters, (or if that even works the way you'd expect).

stevendanna:
@jkeiser This looks fine.  It would be nice if "backtrace issues" was described in a second paragraph of your commit message. :+1: here.

PR #666, opened 2015-12-11
Don't treat host:port as an IPv6 address

nvwls:
Addresses chef-server/665

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
@@ -27,10 +27,10 @@ class OmnibusHelper
   # Normalizes hosts. If the host part is an ipv6 literal, then it
   # needs to be quoted with []
   def self.normalize_host(host_part)
-    # Make this simple: if ':' is detected at all, it is assumed
+    # Make this simple: if more than one ':' is detected, it is assumed
     # to be a valid ipv6 address. We don't do data validation at this
     # point, and ':' is only valid in an URL if it is quoted by brackets.
-    if host_part =~ /:/
+    if host_part.count(':') > 1
       "[#{host_part}]"
     else
       host_part


stevendanna:
Thanks for the contribution.  However, I think this would exclude a lot of valid IPv6 addresses since a single `:` is used to separate groups in an ipv6 address.

stevendanna:
D'oh I misread this.  I think this will work apologies for any confusion.

stevendanna:
:+1:

manderson26:
:+1: Makes sense.

marcparadise:
This change has the potential to break other configuration - `OmnibusHelper.vip_for_url` uses this function, and everywhere that we use `vip_for_uri`, we're appending or otherwise specifying port,. here's one example: 

https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb#L160

In that case, if someone specified a vip that included a port for bifrost, erchef would not be able to communicate with bifrost. It's the same for other usages. 

We have an existing issue #50, was the underlying issue you're trying to resolve with this?

stevendanna:
@marcparadise In thinking about this, while I agree with you at a high level, I don't see this change opening up new issues.  Nothing /prevented/ users from specifying a port.  In the past, such specifications would fail with whatever weird error the relevant component would raise when configured with an invalid IPv6 address. 

As far as I know, there isn't a valid IPv6 address that contains a single `:`, this does allow for an easy work around for the issue with bookshelf when nginx is listening on non-443, and it doesn't open the user up to particularly more errors that they are open to now.

The only downside to merging this would the cost of continuing to support `hostname:port` specifications in these config items in the future even once we have a more complete fix.

nvwls:
Based on the comment:

        The only downside to merging this would the cost of continuing to support hostname:port specifications

Is this being deprecated? Or will this be addressed in one form or another?

marcparadise:
I believe the underlying need for this has been addressed by https://github.com/chef/chef-server/pull/833, which adds a bookshelf_vip attribute that accepts port.  @nvwis could you confirm that the reason for this change was the inability to set a custom port for bookshelf?

marcparadise:
Closing this as resolved by #833.   Please re-open this issue if the addition of `bookshelf[vip_port]` doesn't address your concern.

PR #667, opened 2015-12-11
FLOW-67: ACL endpoints for policies and policy groups

doubt72:
This PR adds:

FLOW-155: ACL access for policies and policy groups (both read and write) to oc_erchef
FLOW-156: pedant test coverage for the endpoints


marcparadise:
:+1: MWR

stevendanna:
:+1: Looks reasonable

PR #668, opened 2015-12-14
Add a simple package-list script to easily download builds

stevendanna:
This is a simple script to download builds from the current channel, making it easy to get any build built off of master for testing.

We should eventually rewrite this using mixlib-install:

https://github.com/chef/mixlib-install

but currently mixlib install doesn't really support any of the server-side packages.

### DIFF
--- a/dev/scripts/package-list
+++ b/dev/scripts/package-list
@@ -1,0 +1,35 @@
+#!/usr/bin/env ruby
+require 'artifactory'
+
+product= ARGV[0] || "chef-server"
+package_name = if product == "chef-server"
+                 "chef-server-core"
+               else
+                 product
+               end
+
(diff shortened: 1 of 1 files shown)


stevendanna:
One problem with this is that promoted builds are removed from current, so you need to s/current/stable/ in the url for those builds.  We could make a HEAD request and make sure the build is there when buidling the URL, but I kinda want to keep this simple and put support for server-side projects into mixlib-install

tylercloke:
@stevendanna maybe just pass `current` vs `stable` in as an arg?

PrajaktaPurohit:
+1

marcparadise:
allowing optional arg to specify 'current' would be good, otherwise  :+1:

PR #671, opened 2015-12-15
Automated reporting matrix testing

tylercloke:


dev/Vagrantfile
```diff
@@ -172,7 +172,17 @@ end
...
+                  if ENV['FORCE_INSTALLER'] =~ /^.*chef-server-core.*deb$/ and File.file?(ENV['FORCE_INSTALLER'])
+                    puts "Valid $FORCE_INSTALLER override detected. Using #{ENV['FORCE_INSTALLER']} as Chef Server 12 build."
+                    ENV['FORCE_INSTALLER']
+                  else
+                    puts "$FORCE_INSTALLER set to #{ENV['FORCE_INSTALLER']} but is not a valid Chef Server 12 build."
+                    exit 1
+                  end
+                else
+                  prompt_installer
+                end
```
marcparadise: Didn't we already have $INSTALLER for this?

dev/scripts/reporting/chef-server-reporting-latest-independant-external-dbs.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+#
+# Copyright 2012-2014 Chef Software, Inc.
```
marcparadise: 2015 for this and the others.

dev/Vagrantfile
marcparadise: https://github.com/chef/chef-server/blob/tc/modified-ssd/dev/Vagrantfile#L199

tylercloke: `$INSTALLER` still requires human interaction, it just adds `$INSTALLER` to the list to select from. Since my scripts are spinning DVM up and down automatedly, it needs a way to bypass that menu in general.

dev/scripts/reporting/chef-server-reporting-latest-independant-external-dbs.sh
tylercloke: :derp: thanks!

dev/scripts/reporting-functions.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
stevendanna: This file seems to share a lot of content with `scripts/reporting/reporting-functions.sh`. Do we need both?

stevendanna:
Left a question re a file that looks like a duplicate.  Otherwise :+1:, this is a good first step.

dev/scripts/reporting-functions.sh
tylercloke: Ah no, that was left by mistake from earlier iterations. Deleting.

dev/Vagrantfile
marcparadise: INSTALLER=x + AUTOPACKAGE=1 

Not as straightforward as this variable, but it does have the same effect without another layer atop selection.

tylercloke: Ah wasn't aware I could do that. Cool.

PR #673, opened 2015-12-16
Fix up some pedant spec descriptions that lagged behind API developments

randomcamel:


stevendanna:
:+1: This looks reasonable to me.

oc-chef-pedant/spec/api/cookbook_artifacts/delete_spec.rb
```diff
@@ -21,7 +21,7 @@
 
   include Pedant::RSpec::CookbookUtil
 
-  context "DELETE /cookbooks/<name>/<version>" do
+  context "DELETE /cookbook_artifacts/<name>/<version>" do
```
stevendanna: Minor quibble but below we call "version" a "cookbook_identifier".  Perhaps that is the better name?

randomcamel: It is better, and I'll fix it. Due to some conceptual and code copy-paste, the code and tests often play fast and loose with terminology.

stevendanna: @randomcamel I'm fine leaving that call to you or @danielsdeleo who know this code better than I most likely.

PR #677, opened 2015-12-19
Simplify bootstrap, move pivotal key gen,  pg retries

marcparadise:
* This moves the existing bootstrap logic into a cookbook library,
  reusing the EcPostgres library and node attributes in place of rendered
  config templates. The existing src/chef-server-bootstrap componet has
  been removed, as well as associated templates.
* This also moves pivotal key creation into the bootstrap recipe itself -
  it turns out we didn't actually use it before that point [1]
* Bootstrap will accept and use an existing pivotal key if one is present.
* Retries have been added all ec_postgres connection attempts over tcp.
  This should take care of the sporadic reports we receive of errors when
  attempting to create users or schema because the database is still coming online.
* Retries have also been added to bifrost requests within bootstrap. We
  have several reports of bootstrap failures due to bifrost being unavailable
  for a short period at startup, and this should fix all cases of that.
* Finally, in light of recent conversations about the consistency
  risks inherent to our two-phase object creation process, it separates
  out the object creation steps as follows:

     1. Create authz objects
     2. Assign authz membership and permissions
     3. create corresponding erchef objects

This helps improve the idempotency of a failed or interrupted
bootstrap[2].  An authz failure may result in orphaned authz objects,
but it will not prevent the bootstrap from succeeding on
a subsequent run because of partially created objects.

Additional notes and tasks:

   1. [x] I need to verify that erchef won't be sad on startup if it can't
      add the key to its keychain, but bootstrap already forces an
      erchef restart after it completes - so risk should be small if
      any at all.
   2. [x] This is still pending some work comining in from LOB-95,
      specifically the check to the state of bootstrap & pivotal user

All pedant tests pass, but more manual validation of bootstrap and
pivotal + admin group ACLs is required.  It may also be a good
opportunity to add post-bootstrap pedant tests that specifically pivotal and
admin user group ACEs, and execute after the initial reconfigure is 

TODO:
* [x] apache license fixup
* [x] remove include_recipe for opscode-erchef
* [x] cleanup recipe for old bootstrap dir (@sdelano)

From #704, which is now merged into this branch: 
* [x] track down some pedant test failures: sporadic errors (due to insufficent FE node memory) 

* [x] clean up the git history
* [x] couple of debug statements to remove
* [x]  add silent mode to pgsql retries and use in preflight checks that expect failure - 
* [x] fix preflight file name
* [x] add remediation for bootstrap errors 6 and 7
* [x] figure out why bookshelf tests are failing now. -*Edit: removing this.*  Bookshelf is going off the backend very soon and tracking this down does not seem worth the time.


marcparadise:
in particular, /cc @tylercloke for bootstrap changes

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
```diff
@@ -3,11 +3,26 @@ class EcPostgres
...
+      connection = ::PGconn.open('user' => postgres['db_superuser'],
+                                 'host' => postgres['vip'],
+                                 'password' => postgres['db_superuser_password'],
+                                 'port' => postgres['port'],
+                                 'dbname' => database)
+    rescue e
+      if retries > 0
+        retries -= 1
+        Chef::Log.warn "Error connecting to postgresql: #{e.message}, retrying after 1s sleep. #{retries} retries remaining."
+        sleep 1
```
stevendanna: Do we want something like exponential backoff here instead?

marcparadise: While I'm not sure it'll be neededin the failure cases we're looking at, it can't hurt either. I'll change that.

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
```diff
@@ -3,11 +3,28 @@ class EcPostgres
...
-                               'dbname' => database)
+    connection = nil
+    retries = 5
+    begin
+      connection = ::PGconn.open('user' => postgres['db_superuser'],
+                                 'host' => postgres['vip'],
+                                 'password' => postgres['db_superuser_password'],
+                                 'port' => postgres['port'],
+                                 'dbname' => database)
+    rescue e
```
stevendanna: Is there a more specific class of error we could rescue here?

sdelano: A quick skim of the PGconn [docs](http://phrogz.net/rubylibs/rdoc/classes/PGconn.html#M000207) didn't turn up anything incredibly useful:

```
PGconn.connect( pghost, pgport, pgoptions, pgtty, dbname, login, passwd ) => conn
PGconn.new( pghost, pgport, pgoptions, pgtty, dbname, login, passwd ) => conn
PGconn.setdb( pghost, pgport, pgoptions, pgtty, dbname, login, passwd ) => conn
PGconn.setdblogin( pghost, pgport, pgoptions, pgtty, dbname, login, passwd ) => conn
pghost:	server hostname (String)
pgport:	server port number (Integer)
pgoptions:	backend options (String)
pgtty:	tty to print backend debug message (ignored in newer versions of PostgreSQL) (String)
dbname:	connecting database name (String)
login:	login user name (String)
passwd:	login password (String)

On failure, it raises a PGError exception.
```

Clicking the link on PGError takes you to a blank page.

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
```diff
@@ -1,70 +1,59 @@
 #
-# Author:: Adam Jacob (<adam@opscode.com>)
-# Copyright:: Copyright (c) 2011 Opscode, Inc.
+# Author:: Adam Jacob <adam@chef.io>
+# Copyright:: Copyright (c) 2011-2015 Chef Software, Inc.
```
sdelano: For future reference, if we're updating the license headers we should make them Apache Licensed. Although we do have language all over [the source](https://github.com/chef/chef-server/tree/master/src/oc_bifrost#license) saying that everything is Apache 2.

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
```diff
@@ -1,70 +1,59 @@
...
 #
 # All Rights Reserved
 #
 
-require 'securerandom'
-
 # Enterprise Addon Install
```
sdelano: While we removed the meat of the comment, this one is still out of date. We do zero addon install here.

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
```diff
@@ -1,70 +1,59 @@
...
-# On a new install, download and install all of the Chef
-# addons from package repositories. Only perform this
-# action during a fresh install, and not during an upgrade.
+include_recipe "private-chef::opscode-erchef"
```
sdelano: Why do we have to require the `opscode-erchef` recipe here? IIRC, this is now the only place where we include recipes for individial services. There's a loop in the base recipe the includes all the services that are required. My concern here is that we'll start up and configure erchef even if the config has explicitly turned it off.

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
```diff
@@ -1,70 +1,59 @@
...
-template File.join(opscode_test_config_dir, "config.rb") do
-  source "bootstrap-config.rb.erb"
-  owner "root"
-  group "root"
-  mode "0600"
+pivotal_key_path = "/etc/opscode/pivotal.pem"
+if File.exists?(pivotal_key_path)
+  pivotal_key = OpenSSL::PKey::RSA.new(pivotal_key_path)
+else
+  pivotal_key = OpenSSL::PKey::RSA.generate(2048)
```
sdelano: An interesting thought just popped into my head. What are the implications / feasibility of creating a pivotal key for each frontend and using the key rotation feature?

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
```diff
@@ -1,70 +1,59 @@
...
-  source "bootstrap-script.rb.erb"
-  owner "root"
-  group "root"
-  mode "0600"
-  not_if { OmnibusHelper.has_been_bootstrapped? }
+# Clean up original bootstrap component in existing installations.
+directory "/opt/opscode/embedded/service/chef-server-bootstrap" do
+  action :delete
+  recursive true
+  ignore_failure true
```
sdelano: Should this be handled in the clenaup recipe?

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
marcparadise: Yeah, pg_conn shoves everything into PGError. In postgres preflight validator I had to match on postgres error strings :/

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
marcparadise: Because this fails if we don't: 

https://github.com/chef/chef-server/pull/677/files#diff-1f3de27173d5e39c86242170696a4f19R48

Maybe I could just add the service declaration instead? Will that cause trouble if I do?

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
marcparadise: I'd been thinking the same as I went through bootstrap.  There is no reason not to add them as multiple pivotal keys, and it simplifies things for the config picture.

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
marcparadise: That works for me.

Issue #646, opened 2015-12-01
Instructions are incorrect after installing a package during upgrade


andy-dufour:
After installing an upgrade package, the post-install instructions reference performing 'chef-server-ctl upgrade' followed by 'chef-server-ctl cleanup'.

After a chef-server-ctl upgrade, services remain down. We should either have chef-server-ctl upgrade start services or explicitly tell users to perform a chef-server-ctl start afterwards.

/cc @jaymzh

jaymzh:
I don't think the instructions are correct - I think it's a bug. `chef-server-ctl` should ideally leave the services in whatever state it found them in, or barring that, leave them up.

tylercloke:
The documentation does say to run `chef-server-ctl start` [here](https://docs.chef.io/upgrade_server.html#standalone) so we should be consistent between the two at the very least. 

Interestingly, it also says to run a `chef-server-ctl stop` before the upgrade which means everything should be down going into an upgrade, so it seems like the sane behavior would actually be to leave everything down by default.

Long-term, we should probably detect what is up, handle bringing everything down in the upgrade command, and then bring back up what was up before upgrade was run.

jaymzh:
The RPM explicitly states to run upgrade and then cleanup when you install it. the instructions the RPM tell me superceed everythign else :)

tylercloke:
Agreed that the two should be consistent. Easy enough to fix.

MathildeBarr:
@tylercloke @andy-dufour Any update here?

sdelano:
Closed by #689

PR #689, opened 2016-01-14
Fixing upgrade instructions in package postinstall script

andy-dufour:
Fixing #646

Simply adding some information via echo to tell the user they need to run chef-server-ctl start after the upgrade command completes.

Currently if you follow our post-install packages instructions explicitly you end up with a chef-server with all of its services down. Adding an explicit chef-server-ctl start and a link to our verbose docs upgrade instructions corrects the situation.

### DIFF
--- a/omnibus/package-scripts/chef-server/postinst
+++ b/omnibus/package-scripts/chef-server/postinst
@@ -48,7 +48,7 @@ if [ -e /etc/opscode/chef-server-running.json ]; then
   echo "(Add the '--no-op' option to see what would be removed by"
   echo "this command)"
   echo
-  echo "For verbose upgrade instructions please see:"
+  echo "For detailed upgrade instructions please see:"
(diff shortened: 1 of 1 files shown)


omnibus/package-scripts/chef-server/postinst
```diff
@@ -40,6 +45,9 @@ if [ -e /etc/opscode/chef-server-running.json ]; then
   echo
   echo "(Add the '--no-op' option to see what would be removed by"
   echo "this command)"
+  echo
+  echo "For verbose upgrade instructions please see:"
```
metadave: verbose -> detailed?

stevendanna:
:+1: This looks reasonable to me

andy-dufour:
This also passed Wilson: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/48/

irvingpop:
Love it!  :+1:

PR #691, opened 2016-01-14
Modify postgres preflight checks to have correct assumptions

### LINKED ISSUE #599: backend postgresql preflight validation runs on front-end nodes.
Some of these  (specifically if DB and roles already exist)  should onyl be running on backend nodes, and possibly only for the  backend bootstrap node so that we don't see the same issue when bringing up a replacement back end. 

Because bringing front-ends online typically involves copying files from a backend, the existing check for `chef-server-running.json` being present prevents this from coming up in most cases. 

However, if chef-server-running is omitted from the files being copied (and there's no reason for it to be copied, so this should be supported), the front-end will fail when external postgresql is configured

rmoshier:
- We now look for combination of previous run and secrets when
  deciding whether or not to check for existing databases and roles.
  Previously, we looked only for the presence of a previous run which
  was wrong because front end nodes won't always have a
  chef-server-running.json copied over, but they should always have
  private-chef-secrets.json in place prior to the initial reconfigure.

- Added first_run? and secrets_exist? to PreflightValidator to make
  logic and testing more clear.

- Added preflight_postgres_validator_spec.rb
  - Added test for existing errors CSPG016 and CSPG017
  - Added tests of backend_validation for previous_run and secrets

- Fixes #599

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
@@ -28,6 +28,14 @@ class PreflightValidator
     end
   end
 
+  def first_run?
+    previous_run.nil?
+  end
+
+  def secrets_exists?
+    PrivateChef.existing_secrets.length > 0
+  end
+
   def backend?
     #
     # When these preflight checks are run, the chef-server.rb has been ingested into
--- a/omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/preflight_postgres_validator_spec.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/preflight_postgres_validator_spec.rb
@@ -1,0 +1,110 @@
+require_relative '../../libraries/helper.rb'
+require_relative '../../libraries/preflight_checks.rb'
+require_relative '../../libraries/preflight_postgres_validator.rb'
+
+describe PostgresqlPreflightValidator do
+  let(:node) { { 'private_chef' => {'opscode-erchef' => { 'sql_user' => 'blah' }, 'oc_bifrost' => { 'sql_user' => 'blah' }, 'oc_id' => { 'sql_user' => 'blah' }} } }
+  let(:first_run_response) { false }
+  let(:secrets_exists_response) { false }
+  let(:postgres_validator) {
+    validator = PostgresqlPreflightValidator.new(node)
+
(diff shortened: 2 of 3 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
```diff
@@ -28,6 +28,17 @@ def fail_with error
     end
   end
 
+  def first_run?
```
stevendanna: Minor style point.  This method could be:

```
def first_run?
  previous_run.nil?
end
```

which would be more in line with the ruby style we lean towards.

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/preflight_postgres_validator_spec.rb
```diff
@@ -0,0 +1,110 @@
+require_relative '../../libraries/helper.rb'
+require_relative '../../libraries/preflight_checks.rb'
+require_relative '../../libraries/preflight_postgres_validator.rb'
+
+describe PostgresqlPreflightValidator do
+  let(:node) { { 'private_chef' => {'opscode-erchef' => { 'sql_user' => 'blah' }, 'oc_bifrost' => { 'sql_user' => 'blah' }, 'oc_id' => { 'sql_user' => 'blah' }} } }
+  let (:first_run_response) { false }
```
stevendanna: a bit of inconsistent spacing here

stevendanna:
Two minor comments.  Otherwise this looks reasonable to me. :+1: MWR

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
rmoshier: What do you think, @marcparadise? The tests still pass if we make this change. @stevendanna, FYI, Marc and I messed with this a lot trying to figure out what was most clear.

sdelano: I'm :+1: with @stevendanna's recommendation.

marcparadise: I'm :+1: with the suggested change.  I think we made it super-explicit to keep things clear as we talked through it, and didn't revisit it before finishing.

PR #692, opened 2016-01-15
Add a flag to ignore quirks specific to chef-zero (like name size = 255)

jkeiser:


oc-chef-pedant/spec/api/policies/complete_endpoint_spec.rb
```diff
@@ -804,7 +804,7 @@ def rm(assoc)
             end
           end
 
-          context "when the name is close to the maximum size" do
+          context "when the name is close to the maximum size", :chef_zero_quirks do
 
             # On Chef Zero, some backends will append `.json` to a file name,
             # which can exceed the common limit of 255 characters.
```
danielsdeleo: Probably worth cleaning up this comment and changing the line below back to 255 characters, since I only made it 250 in order to accommodate zero in the old implementation.

danielsdeleo:
:+1: on the overall thing, would be nice to effectively revert the old change I made to accommodate zero.

oc-chef-pedant/spec/api/policies/complete_endpoint_spec.rb
jkeiser: FYI, you added both tests (250 and 255), so we're still testing 255 on server--this lets us keep a 250-char test on zero.

danielsdeleo: I thought the intention here is that Chef Zero will filter out `:chef_zero_quirks` in its pedant runs, right? So it seems to me that either we can filter this test out, in which case we might as well change the size to erchef's maximum of 255, or not filter it out and use the zero-friendly value of 250, yeah? FWIW, the other maximum size tests below are for different properties (policy_name vs. revision_id vs. cookbook identifier).

PR #693, opened 2016-01-15
Add SQL storage support to bookshelf

stevendanna:
The majority of this work was completed by @manderson26 as the commits show.  I've written up an overview of how this works to facilitate review.

# Storage System Configuration

A new configuration item, `storage_type`, can now be set to sql or filesystem.  Rather than branching at specific points in the code, we load two complete separate webmachine route configurations which point at complete separated modules.

https://github.com/chef/chef-server/compare/ssd+ma/bookshelf-pgsql#diff-21c31fcff1b6745b56681df1341bb492R127

This results in some duplicated logic at the webmachine layer; however it ensure that the filesystem code is left untouched by any of this work.

# SQL Storage Schema

The schema is rather straightforward.  Buckets are rows in the bucket_names table.  A file consists of an entry in the file_names table which references a bucket_id and a data_id.  The data_id points to an entry in the file_data table which contains metadata about the content stored.  All actual content is stored in the file_chunks table.  A given file may span multiple file_chunks.  The chunking allows for streaming upload without holding the entire file in memory at once.

# File deletion

When a file is deleted, the file_data and file_chunk entries for that file are not immediately removed from the database as this may cause a race condition if another client is actively trying to read that file.  Rather, a trigger marks the relevant file_data entry for deletion in the future.  A worker processes in bookshelf cleans up entries marked for deletion periodically, reducing the chance of the race condition.

# User-facing configuration and tooling

The filesystem mode is still the default and there is no upgrade path currently.  To use SQL based storage, the user must enable it at installation time with:

```
bookshelf['storage_type'] = :sql
```

A preflight verification ensure that the setting is not changed after the initial installation. the `chef-server-ctl psql` command has been updated with knowledge of the new database.

# Testing 

Bookshelf uses ct as its main set of tests.  These tests have been expanded to test both the filesystem and sql storage_types.


stevendanna:
TODO:

- [ ] Run stest before merging

src/bookshelf/src/bksw_sql_io.erl
```diff
@@ -0,0 +1,87 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
```
stevendanna: Can we maybe nuke this and merge it into bksw_sql.io?

src/bookshelf/schema/deploy/base_schema.sql
```diff
@@ -0,0 +1,134 @@
...
+-- good pattern if we use blobs with OIDs, and it also allows
+-- deduplication.
+--
+
+-- Encoding choices for hashes.
+
+CREATE TABLE IF NOT EXISTS file_data(
+    data_id           bigserial PRIMARY KEY,
+    upload_started_at timestamp without time zone default (now() at time zone 'utc'),
+    tombstoned_at timestamp without time zone default NULL,
```
stevendanna: alignment problem

src/bookshelf/schema/deploy/smart_delete.sql
```diff
@@ -0,0 +1,70 @@
...
+   UPDATE file_data SET tombstoned_at = my_tombstone WHERE data_id = id AND tombstoned_at IS NULL;
+   RETURN my_tombstone;
+END;
+$$
+LANGUAGE plpgsql VOLATILE;
+
+CREATE OR REPLACE FUNCTION delete_file_last_reference() RETURNS TRIGGER as $delete_file_last_reference$
+   BEGIN
+       IF NOT EXISTS (SELECT 1 FROM file_names WHERE data_id = OLD.data_id) THEN
+         raise WARNING 'marking % deleted', OLD.data_id;
```
stevendanna: Remove raise debug stuff

src/bookshelf/schema/deploy/utility_functions.sql
```diff
@@ -0,0 +1,71 @@
+-- Deploy bookshelf:utility_functions to pg
```
stevendanna: Do we need this?  Most aren't called in the code or are called by functions which aren't called according to xref.

src/bookshelf/priv/pgsql_statements.config
```diff
@@ -0,0 +1,85 @@
+%% -*- mode:erlang, erlang-indent-level: 4;indent-tabs-mode: nil -*-
```
stevendanna: Review for dead statements

src/bookshelf/src/bksw_sql.erl
```diff
@@ -0,0 +1,280 @@
...
+    end.
+
+%% TODO IMPROVE
+delete_file(Bucket, Name) ->
+    case find_file(Bucket, Name) of
+        {ok, not_found} ->
+            {error, file_not_found};
+        {ok, #db_file{file_id = FileId}} ->
+            delete_file(FileId);
+        Error ->
```
stevendanna: Should we used a tagged tuple here

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,315 @@
...
+    %% Uploads create a new data segment, regardless of file existence
+    maybe_upload(Rq, Ctx, bksw_sql:insert_file_data()).
+
+%%===================================================================
+%% Internal Functions
+%%===================================================================
+
+send_streamed_body(#context{entry_md = #db_file{chunk_count = ChunkCount},
+                            transfer_state = #file_transfer_state{next_chunk = ChunkCount} = _TransferState0}) ->
+                          %% TODO CHECK hash checksums/length before exiting successfully
```
manderson26: Look into checking the checksum

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,315 @@
...
+            io:format("Empty chunk sent Chunk ~p ~p~n", [ChunkId, DbFile]),
+            {[], fun() -> send_streamed_body(Ctx) end};
+        {ok, Data} ->
+            TransferState1 = bksw_sql:update_transfer_state(TransferState0, Data, 1),
+            {Data, fun() -> send_streamed_body(Ctx#context{transfer_state = TransferState1}) end};
+        {error, no_connections} ->
+            %% Sleep a bit then retry.
+            %% not pretty, there should be a nicer way of redoing this. At minimum we probably should push this down
+            %% into bksw_sql or sqerl somehow.
+            timer:sleep(?PGSQL_RETRY_INTERVAL),
```
manderson26: Remove this and up the pooler queue size. Requires stress testing.

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,315 @@
...
+
+maybe_upload(Rq, #context{entry_md = #db_file{} = Obj0}= Ctx0, {ok, DataId}) ->
+    Obj1 = Obj0#db_file{data_id = DataId},
+    Ctx1 = Ctx0#context{entry_md = Obj1, transfer_state = bksw_sql:init_transfer_state()},
+    Resp = write_streamed_body(wrq:stream_req_body(Rq, ?BLOCK_SIZE), Rq, Ctx1),
+    Resp;
+maybe_upload(Rq, Ctx, {error, _Error}) -> %% Clean up and return useful error
+    {halt, Rq, Ctx};
+maybe_upload(Rq, Ctx, Error) -> %% Clean up and return useful error
+    io:format("maybe_upload ~p ~p ~p~n", [Rq, Ctx, Error]),
```
manderson26: Make this logger call

src/bookshelf/src/bksw_cleanup_task.erl
```diff
@@ -0,0 +1,114 @@
...
+    case sqerl:select(cleanup_abandoned_uploads, [Interval], first_as_scalar, [cleanup_abandoned_uploads]) of
+        {ok, Count} ->
+            lager:debug("Cleanup task: cleaned up %p expired file_data elements", [Count]),
+            Count;
+        _Error ->
+            lager:debug("Cleanup task: error cleaning up expired file_data elements", []),
+            0
+    end.
+
+
```
stevendanna: extra space

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,315 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
```
stevendanna: This file has a lot of TODO comments.  We should do them or excise the comments.

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,315 @@
...
+                    Rq1 = bksw_req:with_etag(RawRequestMd5, Rq0),
+                    {true, wrq:set_response_code(202, Rq1), Ctx1};
+                _ ->
+                    {true, wrq:set_response_code(406, Rq0), Ctx1}
+            end
+    end;
+write_streamed_body({Data, Next}, Rq0,
+                    #context{entry_md = #db_file{data_id = DataId},
+                             transfer_state = #file_transfer_state{next_chunk = ChunkId} = TransferState0} = Ctx0 ) ->
+%%    io:format("Context 3 ~p~n", [Ctx0]),
```
manderson26: Remove this

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,315 @@
...
+    end.
+
+%%
+%% Takes a chunk from webmachine, and streams it out to the db
+%%
+%% A future version of this may want to re-chunk things;
+write_streamed_body({Data, done}, Rq0,
+                    #context{entry_md = #db_file{data_id = DataId} = File0,
+                             transfer_state = #file_transfer_state{next_chunk = ChunkId} = TransferState0} = Ctx0 ) ->
+%%    io:format("Context 2 ~p~n", [Ctx0]),
```
manderson26: Remove this

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,315 @@
...
+
+    #file_transfer_state{next_chunk = ChunkCount} = TransferState1 =
+        case Data of
+            <<"">> ->
+                TransferState0;
+            Data ->
+                %% TODO: This can fail; we need some method for managing failure here.
```
manderson26: This is a real issue, if there's no pooler connection we drop data on the floor; we should at least detect and return an error code so that the user can retry.

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,315 @@
...
+finalize_maybe_create_file(_Rq, _Ctx,
+                           #db_file{file_id = undefined,
+                                    bucket_name = BucketName,
+                                    name = Name,
+                                    data_id = DataId}) ->
+    case bksw_sql:create_file_link_data(BucketName, Name, DataId) of
+        {ok, _} ->
+            ok;
+        {error, _Reason} ->
+            %% TODO ERROR HANDLING
```
manderson26: Add error handling

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,315 @@
...
+            ok
+    end;
+%% If we have a file_id, we're replacing an old file
+finalize_maybe_create_file(_Rq, _Ctx,
+                           #db_file{file_id = FileId,
+                                    data_id = DataId}) ->
+    case bksw_sql:link_file_data(FileId, DataId) of
+        _ ->
+            ok
+            %% TODO ERROR HANDLING
```
manderson26: Add error handling

src/bookshelf/src/bksw_wm_sql_object.erl
stevendanna: These are the items left:

```
src/bksw_wm_sql_object.erl
213:                          %% TODO CHECK hash checksums/length before exiting successfully
288:                %% TODO: This can fail; we need some method for managing failure here.
323:    %% TODO: This can fail; we need some method for managing failure here.
```

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -596,8 +596,18 @@
 default['private_chef']['bookshelf']['stream_download'] = true
 default['private_chef']['bookshelf']['access_key_id'] = "generated-by-default"
 default['private_chef']['bookshelf']['secret_access_key'] = "generated-by-default"
+default['private_chef']['bookshelf']['auth_check_disabled'] = false
 # Default: set to Host: header. Override to hardcode a url, "http://..."
 default['private_chef']['bookshelf']['external_url'] = :host_header
+default['private_chef']['bookshelf']['storage_type'] = :filesystem
+default['private_chef']['bookshelf']['db_pool_size'] = 20
+default['private_chef']['bookshelf']['db_pool_queue_max'] = 20
```
stevendanna: I think we should consider increasing this substantially if we want to support a lot of concurrency out of the box.

stevendanna: We'll decide on this after looking at the results of stest.

src/bookshelf/README_SQL.md
```diff
@@ -0,0 +1,79 @@
+
```
stevendanna: Should we nuke this before merging? I think that we've covered most of the TODOs in here that are critical.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bookshelf_validator.rb
```diff
@@ -0,0 +1,64 @@
...
+Bookshelf's storage_type was previously '#{previous_value}'; however
+the current configuration would result in a value of '#{current_value}'.
+At this time it is not possible to change the bookshelf storage_type post-installation.
+
+Please set
+
+bookshelf['storage_type'] = :#{previous_value}
+
+in /etc/opscode/chef-server.rb
+EOM
```
sdelano: A couple thoughts here:

Would it be useful to be explicit in the case logic about each of the 5 cases? As the code changes, the implicit "failures" may be hard to debug.

On this line:

```
Bookshelf's storage_type was previously '#{previous_value}'; however
```

if the previous value is `nil`, that appears that it will cause a confusing error message, especially with

```
bookshelf['storage_type'] = :#{previous_value}
```

Do we have tests for this helper? I can't find them.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
```diff
@@ -73,6 +73,7 @@ def run!
     begin
       PostgresqlPreflightValidator.new(node).run!
       SolrPreflightValidator.new(node).run!
+      BookshelfPreflightValidator.new(node).run!
```
sdelano: I definitely love that we're continuing to add preflight checks for things that we know are either unsupported or just won't work. Nice job. :100:

omnibus/files/private-chef-cookbooks/private-chef/recipes/bookshelf.rb
```diff
@@ -48,12 +48,18 @@
 
 bookshelf_config = File.join(bookshelf_dir, "sys.config")
 
+bookshelf_params = node['private_chef']['bookshelf'].to_hash.dup # is dup implicit in #to_hash?
+bookshelf_params['postgresql'] = node['private_chef']['postgresql'].dup
+['sql_user', 'sql_password', 'sql_db_timeout'].each do |v|
+  bookshelf_params['postgresql'][v] = node['private_chef']['opscode-erchef'][v]
+end
```
sdelano: We've configured explicit users and passwords for the `bookshelf` database and users [above](https://github.com/chef/chef-server/pull/693/files#diff-3aaae794e6628ce1d91334c7dd493e7aR609). Why are we overriding those to use the `opscode-erchef` ones here?

omnibus/files/private-chef-cookbooks/private-chef/recipes/postgresql-external.rb
```diff
@@ -20,4 +20,5 @@
   include_recipe "private-chef::erchef_database"
   include_recipe "private-chef::bifrost_database"
   include_recipe "private-chef::oc_id_database"
+  include_recipe "private-chef::bookshelf_database"
```
sdelano: Are we going to create this regardless of whether we've configured the storage to be `:sql`?

src/bookshelf/README_SQL.md
sdelano: :+1: Is this file still useful?

src/bookshelf/schema/sqitch.conf
```diff
@@ -0,0 +1,8 @@
+[core]
+	engine = pg
+	# plan_file = sqitch.plan
+	# top_dir = .
+# [engine "pg"]
+	# target = db:pg:
+	# registry = sqitch
+	# client = psql
```
sdelano: Are these lines all commented out? It seems like we actually do that across most of our `sqitch.conf` files. Any specific reason why?

omnibus/files/private-chef-cookbooks/private-chef/recipes/bookshelf.rb
stevendanna: This was left over from when we didn't have a separate db user for bookshelf.  Will remove.

omnibus/files/private-chef-cookbooks/private-chef/recipes/postgresql-external.rb
stevendanna: Yes.  Currently the bookshelf code starts sqerl and connects to the database regardless of the config, so we always bootstrap the database.

omnibus/files/private-chef-cookbooks/private-chef/recipes/bookshelf_database.rb
```diff
@@ -0,0 +1,37 @@
...
+  access_profile :read
+end
+
+private_chef_pg_sqitch "/opt/opscode/embedded/service/bookshelf/schema" do
+  hostname postgres_attrs['vip']
+  port     postgres_attrs['port']
+  username  postgres_attrs['db_superuser']
+  password  postgres_attrs['db_superuser_password']
+  database "bookshelf"
+end
```
sdelano: FYI - this is going to deploy schema changes **ALWAYS**. This deviates from the way in which we've been deploying schemas for the Chef Server itself by way of `partybus` migrations. Given that we don't have any additional schema changes yet, this is okay, but it's worth keeping in mind now and potentially worth working into the code now if we'd like to control the schema changes in the same way in the future, in case we forget about it.

src/bookshelf/rebar.config
```diff
@@ -24,6 +24,10 @@
           {branch, "master"}}},
         {meck, ".*",
          {git, "https://github.com/eproxus/meck.git", {tag, "0.8.3"}}},
+        {sqerl, ".*",
+         {git, "git://github.com/chef/sqerl.git", {branch, "master"}}},
```
sdelano: Why is `envy` fetched over HTTP but `sqerl` over git://?

src/bookshelf/rebar.lock
```diff
@@ -1,11 +1,15 @@
 [{<<"envy">>,
-  {git,"git://github.com/manderson26/envy.git",
-       {ref,"954c87adcfea892db462f56059eb8bc2857ab50b"}},
-  1},
+  {git,"https://github.com/manderson26/envy",
+       {ref,"8ba233bae37d87c319f3b4c5ae6973576986b36c"}},
+  0},
```
sdelano: Just a curiosity question here, is the `0` (last tuple element) here a reference to the depth of the node in the dependency tree?

src/bookshelf/schema/deploy/smart_delete.sql
```diff
@@ -0,0 +1,69 @@
...
+RETURNS integer
+AS $$
+DECLARE
+   expire_time file_data.tombstoned_at%TYPE;
+   deleted integer;
+BEGIN
+   SELECT (now() at time zone 'utc') - (delay * '1 millisecond'::interval) INTO expire_time;
+   WITH a AS (DELETE FROM file_data WHERE tombstoned_at < expire_time RETURNING 1)
+   SELECT count(*) FROM a INTO deleted;
+   RETURN deleted;
```
sdelano: What's the difference here between just returning the results of `SELECT` instead of doing a `SELECT INTO` followed by a `RETURN`?

src/bookshelf/schema/sqitch.conf
stevendanna: I think most of our sqitch.conf's are mostly autogenerated by sqitch.

src/bookshelf/rebar.lock
stevendanna: Looks like it: https://github.com/rebar/rebar3/blob/master/src/rebar_prv_lock.erl#L63-L65

src/bookshelf/src/bksw_conf.erl
```diff
@@ -114,46 +123,45 @@ reset_dispatch() ->
 dispatch() ->
     {keys, {AccessKeyId, SecretAccessKey}} = keys(),
     Config = [{stream_download, stream_download()},
+              {auth_check_disabled, auth_check_disabled()},
               {access_key_id, AccessKeyId},
-              {secret_access_key, SecretAccessKey}],
+              {secret_access_key, SecretAccessKey}|sql_retry()],
```
sdelano: It appears that we're setting the sql portion of the config in 3 different ways in this file:

1. explicitly on lines 56/57
2. via `lists:flatten` on line 70
3. and here on line 128 via head / tail construction

src/bookshelf/src/bksw_conf.erl
```diff
@@ -97,14 +109,11 @@ ends_with(Char, String) ->
...
+    envy:get(bookshelf, auth_check_disabled, false, boolean).
 
 ip() ->
-    case application:get_env(bookshelf, ip) of
-        undefined ->
-            [{ip, "127.0.0.1"}];
-        {ok, Ip} ->
-            [{ip, Ip}]
-    end.
+    {ip, envy:get(bookshelf, ip, "127.0.0.1", string)}.
```
sdelano: We used to return `[{ip, Ip}]` and now we're returning `{ip, Ip}`.

1. Is this a bug
2. If not, should we make a `spec()` for this?

sdelano: After a quick search, looks like we were saving ourselves previously with `lists:flatten`

omnibus/files/private-chef-cookbooks/private-chef/recipes/bookshelf_database.rb
marcparadise: True for erchef, but we do deploy bifrost changes in the same way as this.

src/bookshelf/rebar.config
marcparadise: all s/b https to prevent builder issues, iirc.

src/bookshelf/schema/deploy/smart_delete.sql
marcparadise: Is the intent here just to indicate if something was deleted or not? If so, can we just use a normal delete query from sqerl, and check if the returned number of affected rows is > 1? 

Same for `cleanup_abandoned_uploads`, can it be justa  straight up delete instead?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -596,8 +596,18 @@
 default['private_chef']['bookshelf']['stream_download'] = true
 default['private_chef']['bookshelf']['access_key_id'] = "generated-by-default"
 default['private_chef']['bookshelf']['secret_access_key'] = "generated-by-default"
+default['private_chef']['bookshelf']['auth_check_disabled'] = false
```
sdelano: TODO: we should remove this from the attributes file

src/bookshelf/src/bksw_sql.erl
```diff
@@ -0,0 +1,253 @@
...
+    case sqerl:statement(insert_bucket, [BucketName], count) of
+        {ok, 1} ->
+            ok;
+        {conflict, _} ->
+            error("Never should happen");
+        Error ->
+            Error
+    end.
+
+-spec create_bucket_b(binary()) -> boolean().
```
marcparadise: From the fun heads, I'm not sure what the significance of '`bucket_b` is compared to `delete_bucket` - just a rename for clarity around `b` would help.

src/bookshelf/src/bksw_sql.erl
```diff
@@ -0,0 +1,253 @@
...
+    case sqerl:select(link_file_data, [FileId, DataId], first_as_scalar, [success]) of
+        {ok, _} ->
+            ok;
+        {error, Reason} ->
+            {error, Reason}
+    end.
+
+find_file(Bucket, Name) ->
+    case sqerl:select(find_file, [Bucket, Name], {first_as_record, ?DB_FILE_TX_FM }) of
+        %% Awkward sanity check that we got back the expected record type here.
```
marcparadise: I think we can nuke these. THere's no value in them, the record must match what we give it based on the way sqerl assembles the response.

src/bookshelf/stest/rebar.config
```diff
@@ -20,6 +21,6 @@
 %% markdown can be checked in and will be browsable on github. The
```
marcparadise: Can we remove this block entirely, since it's for stest and I edown isn't a thing in our tests?

src/bookshelf/stest/src/bookshelf_bench.erl
```diff
@@ -80,6 +81,7 @@ init([]) ->
     %% basho_bench sets cwd to ./tests/current so we need to
     %% traverse up an extra directory to get to deps
     code:add_paths(filelib:wildcard("../../../deps/*/ebin")),
+    code:add_paths(filelib:wildcard("../../deps/*/ebin")),
```
marcparadise: I think we can do a single add_paths with results of both?

src/bookshelf/schema/deploy/base_schema.sql
```diff
@@ -0,0 +1,133 @@
...
+
+ALTER TABLE file_chunks ADD CONSTRAINT file_chunks_id_fk FOREIGN KEY (data_id) REFERENCES file_data(data_id) ON DELETE CASCADE ;
+
+CREATE OR REPLACE VIEW expanded_files AS SELECT bucket_name, fd.* FROM bucket_names b INNER JOIN
+       (SELECT bucket_id, name, created_at, file_id, d.* FROM
+               file_names f INNER JOIN file_data d ON f.data_id = d.data_id) fd
+	ON b.bucket_id = fd.bucket_id;
+
+--
+-- This atomically links a file data item with
```
marcparadise: This comment is

src/bookshelf/src/bksw_sql.erl
```diff
@@ -0,0 +1,253 @@
...
+    Path = filename:join([code:priv_dir(bookshelf), "pgsql_statements.config"]),
+    {ok, Statements} = file:consult(Path),
+    Statements.
+
+create_bucket(BucketName) ->
+    case sqerl:statement(insert_bucket, [BucketName], count) of
+        {ok, 1} ->
+            ok;
+        {conflict, _} ->
+            error("Never should happen");
```
sdelano: We're returning `error("Never should happen")` here. Why? Are we guarding this call with a check to the database to determine if a bucket exists? What about simultaneous calls to the bucket creation API interleaving these database requests?

1. You're right, it "shouldn't" happen, but it definitely can
2. Given that the Chef Server (for which bookshelf is the only provider) uses one bucket ever, it's not a big deal

stevendanna: >  What about simultaneous calls to the bucket creation API interleaving these database requests?

I think the only caller of this is create_bucket_b which checks if the bucket exists first.  `create_bucket_b` is what is called in the controller, so concurrent bucket creates should still never hit this.  That said, this could probably be refactored a bit for sure.

src/bookshelf/stest/rebar.config
stevendanna: :+1: let's nuke it.  It would be could if we used edown, but right now we don't.

src/bookshelf/src/bksw_sql.erl
```diff
@@ -0,0 +1,253 @@
...
+        Error ->
+            Error
+    end.
+
+-spec create_bucket_b(binary()) -> boolean().
+create_bucket_b(Bucket) ->
+    case bksw_sql:find_bucket(Bucket) of
+        {ok, none} ->
+            case bksw_sql:create_bucket(Bucket) of
```
sdelano: Why all the calls to `bksw_sql`? This is the same module. Is this just leftover from a refactor?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bookshelf_validator.rb
stevendanna: I've left the "default to failure", but I've more explicitly listed the missing case.

src/bookshelf/schema/deploy/base_schema.sql
sdelano: LOL

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/preflight_bookshelf_validator_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        s = BookshelfPreflightValidator.new({'private_chef' => {}})
+        allow(s).to receive(:previous_run).and_return({'bookshelf' => {}})
+        allow(s).to receive(:fail_with).and_return(:i_failed)
+        s
+      }
+
+      it "succeeds if set to filesystem" do
+        expect(PrivateChef).to receive(:[]).with('bookshelf').and_return({'storage_type' => 'filesystem'})
+        expect(subject.verify_storage_type_unchanged).to eq(true)
+      end
```
sdelano: Should we additionally test setting the values to symbols like we suggest?

src/bookshelf/src/bksw_sql.erl
sdelano: `bksw_sup:ensure_default_bucket` calls this and is guarded by another database call. I can see multiple versions of bookshelf starting up for the first time and calling this. While contrived, it's definitely possible.

Additionally, `create_bucket_b` makes a call to check if it exists and then an additional call to create it and fails (badly) on a conflict. Two simultaneous calls to this function from different webmachine workers (let alone two separate servers) can easily trigger a race condition and conflict mode here.

sdelano:
```
so concurrent bucket creates should still never hit this
```

How so? If I have a bucket that does not exist and I make two concurrent calls to create, it's not too far fetched for the following to happen:

```
* process A checks if bucket exists: NO
* process B checks if bucket exists: NO
* process A creates bucket: SUCCESS
* process B creates bucket: CONFLICT
```

src/bookshelf/src/bksw_wm_sql_index.erl
```diff
@@ -0,0 +1,55 @@
...
+    {['GET'], Rq, Ctx}.
+
+content_types_provided(Rq, Ctx) ->
+    {[{"text/xml", to_xml}], Rq, Ctx}.
+
+resource_exists(Rq, Ctx) ->
+    {erlang:is_list(bksw_sql:list_buckets()), Rq, Ctx}.
+
+to_xml(Rq, Ctx) ->
+    Buckets = bksw_sql:list_buckets(),
```
sdelano: It looks like we're making the same database call with `bksw_sql:list_buckets` twice here. Can we save it in the state from the first call?

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,336 @@
...
+    {[{MT, upload}], Rq, Ctx}.
+
+resource_exists(Rq0, Ctx) ->
+    case wrq:method(Rq0) of
+        %% Buckets always exist for writes since we create them on the fly
+        'PUT' ->
+            case fetch_entry_md(Rq0, Ctx) of
+                {error, _Error} ->
+                    {false, Rq0, Ctx};
+                {not_found, #context{} = Ctx1 } ->
```
sdelano: Why do we explicitly match on a `#context{}` record here but not in the `#db_file{}` case below?

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,336 @@
...
+            {halt, Rq0, Ctx}
+    end.
+
+maybe_retry(#context{sql_retry_count = 0}, _RetryFun, FailFun) ->
+    FailFun();
+maybe_retry(#context{sql_retry_delay = Delay, sql_retry_count = Count} = Ctx, RetryFun, _FailFun) ->
+    timer:sleep(Delay),
+    RetryFun(Ctx#context{sql_retry_count = Count - 1}).
+
+%% Return `{Obj, CtxNew}' where `Obj' is the entry meta data `#object{}' record or the atom
```
sdelano: Is this comment out of date? We're returning a `#db_file{}` record not an `#object{}` record. Should we add a dialyzer spec for this?

src/bookshelf/src/bksw_wm_sql_object.erl
sdelano: I'm also confused why we're returning `true` for `resource_exists` when the resource doesn't actually exist. Perhaps we should talk through this because reading the [webmachine state machine](https://raw.githubusercontent.com/webmachine/webmachine/master/docs/http-headers-status-v3.png) isn't really helpful to me right now.

sdelano: Reading more of the above-linked state machine, I'd imagine that we'd actually return false here and continue down the path of: `H7, I7, I4, P3, P11, CREATED`

src/bookshelf/test/bkswt_api_SUITE.erl
```diff
@@ -0,0 +1,687 @@
...
+                 end,
+    load_config(ConfigPath).
+
+load_inline_config() ->
+    StaticBookshelfConfig = [{ip, "127.0.0.1"},
+                             {port, 4321},
+                             {keys, {"e1efc99729beb175",
+                                     "fc683cd9ed1990ca"}},
+                             {disk_store, "/tmp/bukkits"},
+                             {sql_retry_count, 5},
```
sdelano: In the rest of the code, and as configured by default w/ no way to override the functionality, we set the `sql_retry_count` to 0, but it's set to 5 here in testing. Should we revisit that?

src/bookshelf/src/bksw_wm_sql_object.erl
sdelano: After talking with @manderson26 - this follows the [original code](https://github.com/chef/chef-server/blob/42b27d417bd4502aaa811d9fbc26295a5c6aec74/src/bookshelf/src/bksw_wm_object.erl#L93), which is sufficient for me.

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,336 @@
...
+%% upload/2 flow.
+validate_content_checksum(Rq, Ctx) ->
+    {true, Rq, Ctx}.
+
+allowed_methods(Rq, Ctx) ->
+    {['HEAD', 'GET', 'PUT', 'DELETE'], Rq, Ctx}.
+
+content_types_provided(Rq, Ctx) ->
+    CType =
+    case wrq:get_req_header("accept", Rq) of
```
sdelano: The whitespace is off here. Perhaps we should make this method consistent with `content_types_accepted`

src/bookshelf/src/bksw_sql.erl
```diff
@@ -0,0 +1,253 @@
...
+init_transfer_state() ->
+    #file_transfer_state{
+       size = 0,
+       next_chunk = 0,
+       hash_context_md5 = crypto:hash_init(md5),
+       hash_context_sha256 = crypto:hash_init(sha256),
+       hash_context_sha512 = crypto:hash_init(sha512)
+      }.
+
+update_transfer_state(#file_transfer_state{next_chunk = NextChunk} = UploadState, Data, ChunkIncr) when size(Data) == 0 ->
```
sdelano: We should comment on this function that it's used for both streaming uploads and downloads, and perhaps change the variable name `UploadState` to something less confusing.

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,336 @@
...
+            Rq1 = bksw_req:with_etag(base64:encode(HashMd5), Rq0),
+            {true, wrq:set_response_code(202, Rq1), Ctx1};
+        RawRequestMd5 ->
+            RequestMd5 = base64:decode(RawRequestMd5),
+            case RequestMd5 of
+                HashMd5 ->
+                    Rq1 = bksw_req:with_etag(RawRequestMd5, Rq0),
+                    {true, wrq:set_response_code(202, Rq1), Ctx1};
+                _ ->
+                    {true, wrq:set_response_code(406, Rq0), Ctx1}
```
sdelano: Is it true that at this point, especially when overwriting a file, we've actually replaced the contents of the file (as far as bookshelf is concerned) and now we're like "oh shit that wasn't the right data". Is this a "too bad" moment? Can we do better?

src/bookshelf/src/bksw_sql.erl
stevendanna: `_b` was intended to convey that it returns a bool

src/bookshelf/src/bksw_sql.erl
stevendanna: :+1:

src/bookshelf/src/bksw_sql.erl
stevendanna: D'oh. Yes.  I think we should just change this to return an atom like {error, conflict}.  Things that don't care abotu conflict can match appropriately.

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/preflight_bookshelf_validator_spec.rb
stevendanna: Added below

src/bookshelf/schema/deploy/base_schema.sql
```diff
@@ -0,0 +1,199 @@
...
+RETURNS integer
+AS $$
+DECLARE
+   expire_time file_data.tombstoned_at%TYPE;
+   deleted integer;
+BEGIN
+   SELECT (now() at time zone 'utc') - (delay * '1 millisecond'::interval) INTO expire_time;
+   WITH a AS (DELETE FROM file_data WHERE tombstoned_at < expire_time RETURNING 1)
+   SELECT count(*) FROM a INTO deleted;
+   RETURN deleted;
```
stevendanna: Some comments lost their place on rebase.  Previously @sdelano asked:

> What's the difference here between just returning the results of SELECT instead of doing a SELECT INTO followed by a RETURN?

And @marcparadise followed up with:

> Is the intent here just to indicate if something was deleted or not? If so, can we just use a normal delete query from sqerl, and check if the returned number of affected rows is > 1?

Same for cleanup_abandoned_uploads, can it be justa straight up delete instead?

omnibus/files/private-chef-cookbooks/private-chef/recipes/bookshelf_database.rb
stevendanna: I think we decided that we want to make all oc_erchef, bifrost, and bookshelf all look the same?  For clarity, it might be best to do that in a follow up commit shortly after this is merged.

src/bookshelf/src/bksw_sql.erl
stevendanna: @marcparadise Any thoughts on a name that would convey "create bucket, returning true as long as it exists afterwards" and "delete bucket, returning true as log as it doesn't exist afterwards"?  I was thinking "ensure_create" and "ensure_delete" if we don't like the "_b" postfix.  However "ensure" to me would imply we actually try to do something about errors and such, which we don't.

src/bookshelf/test/bkswt_api_SUITE.erl
stevendanna: I think we should definitely test with whatever the default is for retries (currently 0 as you mention).  Do we want to make this configurable for end users?  A concern I have is that it isn't applied universally to all sql interactions. cc @manderson26

src/bookshelf/src/bksw_wm_sql_object.erl
stevendanna: > Is it true that at this point, especially when overwriting a file, we've actually replaced the contents of the file (as far as bookshelf is concerned) and now we're like "oh shit that wasn't the right data".

Yes, that is my reading too.

>  Is this a "too bad" moment? Can we do better? 

We could move line 310 into the "success" branch of this case statement.

> How does the current code behave?

My reading is that the old code would behave more or less the same way.

stevendanna: Holding off on this one until we get @manderson26's thoughts.

src/bookshelf/stest/src/bookshelf_bench.erl
stevendanna: Anything using rebar3 doesn't actually use the `deps` folder anymore so it won't be there in bookshelf in a clean checkout, so I'm thinking '../../..' isn't applicable anymore.  A possible alternative is `../../../_build/default/lib/*/ebin`.

omnibus/files/private-chef-cookbooks/private-chef/recipes/bookshelf_database.rb
sdelano: :+1:

src/bookshelf/stest/rebar.config
sdelano: Let's delete this and see what happens.

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/preflight_bookshelf_validator_spec.rb
sdelano: :+1:

src/bookshelf/test/bkswt_api_SUITE.erl
sdelano: TODO:
* [ ] set the test default to 0
* [ ] expose the config value to attributes.rb and the config templates

src/bookshelf/src/bksw_wm_sql_object.erl
sdelano: TODO:
* [ ] move the behaviour of `finalize_maybe_create_file` to occur after checking the checksum

src/bookshelf/src/bksw_wm_sql_object.erl
```diff
@@ -0,0 +1,341 @@
...
+    Ctx1 = Ctx0#context{entry_md = File1},
+    bksw_sql:update_metadata(DataId, Size, ChunkCount, HashMd5, HashSha256, HashSha512),
+
+    case get_header('Content-MD5', Rq0) of
+        undefined ->
+            Rq1 = bksw_req:with_etag(base64:encode(HashMd5), Rq0),
+            %% Failures here turn into 500s and stack traces; we might want to do better if this
+            %% happens much.
+            ok = finalize_maybe_create_file(Rq1, Ctx0, File1),
+            {true, wrq:set_response_code(202, Rq1), Ctx1};
```
stevendanna: As far as I can tell, this, and the error cases below, will all return HTTP 204 because we aren't sending a response body and webmachine doesn't seem to care what we've set in the second argument. If we want to set a response code I think we need to use `{{halt, 202}, Rq1, Ctx1}` as our return value here.  However, I'll note that this same pattern is used in a number of places so I'm guessing most of our client code doesn't care too much about 202 vs 204; however, we definitely want to make sure we return non-2XX at line 328.

PR #695, opened 2016-01-15
svlogd can't write with /var/log at 0700

sean-horn:
Some customers set /var/log 's mode to 0700, which breaks Chef Server 12.3.1+ initial reconfigure.

/var/log/opscode is either not present or unwritable even if present and the reconfigure fails.


marcparadise:
The change itself looks fine, but I think customers who have set it to 0700 may be upset to find that installing Chef Server has changed that without telling them?

marcparadise:
To clarify further: I think this belongs in preflight checks, and we can fail gracefully with an error message telling them to how to create the directory manually and set required permissions.

marcparadise:
Closing in favor of issue #847 after discussion with @sean-horn

PR #697, opened 2016-01-15
WIP - Correct handling of no rabbitmq in controls endpoint

rmoshier:
:skull: :construction: WIP :construction: :skull: 

TODO:
- [x] 410 response to return JSON
- [x] modify test to make sure it returns JSON

Note: the bulk of the diff in the control spec is a whitespace change :exclamation:

### DIFF
--- a/oc-chef-pedant/spec/api/controls_spec.rb
+++ b/oc-chef-pedant/spec/api/controls_spec.rb
@@ -12,419 +12,438 @@ require 'pedant/rspec/node_util'
 
 describe "Controls API Endpoint", :controls do
 
-  describe "v0.1.0" do
-    include Pedant::RSpec::NodeUtil
+  if Pedant::Config.actions_enabled
 
-    # The reason for this is that we need to verify
-    # that the principal endpoint functions properly in OPC, as
-    # currently the only consumer of this endpoint is chef-client-audit which
-    # absolutely requires the OPC control endpoint to work.
-    let(:input) {
-      {
-        :message_type => "control_groups",
-        :message_version => "0.1.0",
-        :organization_name => "ponyville",
-        :chef_server_fqdn => "api.opscode.piab",
-        :recorded_at => "2014-11-07T14:53:58Z",
-        :remote_hostname => "33.33.33.1",
-        :request_id => "g3IAA2QAEGVyY2hlZkAxMjcuMC4wLjEDAABM1wAAAAAAAAAA",
-        :node_name => "client.opscode.piab",
-        :id => "a4247480-bb92-42a5-8116-3729dcc486eb",
-        :run_id => "76296628-7de0-4509-adaa-9a525409142d",
-        :control_groups => [
-          {
-            :name => "mysql audit",
-            :status => "failure",
-            :number_success => 1,
-            :number_failed => 1,
-            :id => "59186fdc-2d0a-4653-af83-bbe79302fcaf",
-            :controls => [
-              {
-                :name => "should be installed",
-                :status => "success",
-                :details => "null",
-                :resource_type => "Package",
-                :resource_name => "mysql",
-                :context => [],
-                :sequence_number => 1
-              },
-              {
-                :name => "should have mysql service",
-                :status => "failure",
-                :details => "expected Service \"mysql\" to be enabled",
-                :resource_type => "null",
-                :resource_name => "null",
-                :context => [],
-                :sequence_number => 2
-              }
-            ]
-          }
-        ]
+    describe "v0.1.0" do
+      include Pedant::RSpec::NodeUtil
+
+      # The reason for this is that we need to verify
+      # that the principal endpoint functions properly in OPC, as
+      # currently the only consumer of this endpoint is chef-client-audit which
+      # absolutely requires the OPC control endpoint to work.
+      let(:input) {
+        {
+          :message_type => "control_groups",
+          :message_version => "0.1.0",
+          :organization_name => "ponyville",
+          :chef_server_fqdn => "api.opscode.piab",
+          :recorded_at => "2014-11-07T14:53:58Z",
+          :remote_hostname => "33.33.33.1",
+          :request_id => "g3IAA2QAEGVyY2hlZkAxMjcuMC4wLjEDAABM1wAAAAAAAAAA",
+          :node_name => "client.opscode.piab",
+          :id => "a4247480-bb92-42a5-8116-3729dcc486eb",
+          :run_id => "76296628-7de0-4509-adaa-9a525409142d",
+          :control_groups => [
+            {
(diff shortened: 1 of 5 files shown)


stevendanna:
Github trick to avoid looking at whitespace changes: https://github.com/chef/chef-server/pull/697/files?w=1

oc-chef-pedant/spec/api/controls_spec.rb
```diff
@@ -12,419 +12,432 @@
...
-      it 'returns a 405 ("Method Not Allowed") for normal_user' do
-        get(api_url("/controls"),
-            normaluser,
-            :payload => input ) do |response|
-              response.should look_like({
-                :status => 405
-              })
-            end
-      end
+        it "returns a 201 ('Created') for client" do
```
sdelano: What's the reasoning behind changing the string formatting here? It's generally accepted good Ruby style to use single quotes `'foo'` for strings that don't need in interpolation and double quotes `"#{foo}"` for strings that do need interpolation. That appeared to be generally true here, and I just wanted to see if there was a reason for moving away from it.

Having said that, there's no clear performance advantage to either and it's more for consistency than anything else. I'll also admit that there's not much consistency in this file to begin with (see following line), so I'm really just raising a point that probably doesn't need much follow up, unless you wanted to enforce consistency on the whole file :)

oc-chef-pedant/spec/api/controls_spec.rb
```diff
@@ -12,419 +12,432 @@
...
-                :details => "expected Service \"mysql\" to be enabled",
-                :resource_type => "null",
-                :resource_name => "null",
-                :context => [],
-                :sequence_number => 2
-              }
-            ]
-          }
-        ]
+  if Pedant::Config.actions_enabled
```
sdelano: This is an interesting approach and I'd like to get @marcparadise's feedback here as well. Is it standard practice for us to conditionally test based on configuration variables, and if so, how do we ensure in all of our integration testing that this case is actually covered?

oc-chef-pedant/spec/api/controls_spec.rb
stevendanna: FWIW, in core chef we have moved to using double-quotes everywhere.

sdelano: Right on! Like I said, consistency :)

marcparadise: I think this was an incidental change when we undid an earlier version of the updates.  Still, we'll keep it consistent with what's in the rest of the file.

oc-chef-pedant/spec/api/controls_spec.rb
marcparadise: The closest we have is LDAP testing, where config variable drives enabling various LDAP auth-related tests. Other than that, we have done this for darklaunch-flagged features in the past (though all of those have been cleaned up). 

As with LDAP and previous darklaunch, there's really no way that our current test setup can handle it. 


Alternative: I'd  given thought to having the test toggle the config flag via rpc- that'd let us  just test both.  I decided against it because time but we can revisit it.  Having that ability built into pedant would actually open up some interesting additional testing options.

marcparadise: Further note: that said, the time-consuming aspect that prevented me from doing this was more related to ensuring that we didn't break when we weren't running on a local instance  (such as running pedant against a dns-load-balanced cluster) that didn't have access to talk to the erlang node.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_controls.erl
```diff
@@ -48,6 +48,22 @@ init(Config) ->
 init_resource_state(_Config) ->
     {ok, #control_state{}}.
 
+service_available(Req, State) ->
```
metadave: nice :-)

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_controls.erl
```diff
@@ -48,6 +48,22 @@ init(Config) ->
...
+    case oc_chef_wm_base:service_available(Req, State) of
+        {true, Req1, State1} ->
+            ActionsEnabled = envy:get(oc_chef_wm, enable_actions, false, boolean),
+                case ActionsEnabled of
```
stevendanna: Slightly odd spacing here, `case` should start at the same column as `ActionsEnabled` in the line above.

stevendanna:
:+1: overall on this, nice job.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_controls.erl
rmoshier: Thanks for pointing this out. I'm still getting used to the wonky Erlang alignment!

oc-chef-pedant/spec/api/controls_spec.rb
```diff
@@ -12,419 +12,438 @@
...
-                :status => 405
-              })
-            end
-      end
+          delete(api_url("clients/clientwithnonode.opscode.piab"), platform.superuser)
+          delete("#{platform.server}/clients/clientwithnonode.opscode.piab", platform.superuser)
+        end
 
-    end
+        it 'returns a 201 ('Created') for client' do
```
sdelano: @rmoshier - This is the line that's causing your test failures, you have multiple instances of single quotes.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_controls.erl
```diff
@@ -48,6 +48,22 @@ init(Config) ->
 init_resource_state(_Config) ->
     {ok, #control_state{}}.
 
+service_available(Req, State) ->
+    case oc_chef_wm_base:service_available(Req, State) of
+        {true, Req1, State1} ->
+            ActionsEnabled = envy:get(oc_chef_wm, enable_actions, false, boolean),
+            case ActionsEnabled of
```
sdelano: You can squash

```erlang
            ActionsEnabled = envy:get(oc_chef_wm, enable_actions, false, boolean),
            case ActionsEnabled of
```

into 

```erlang
            case envy:get(oc_chef_wm, enable_actions, false, boolean) of
```

oc-chef-pedant/spec/api/controls_spec.rb
sdelano: I'm fine with this, thanks for the clarification.

rmoshier:
@sdelano Good suggestion! I am going to keep it like this for now and keep that sort of refactor in mind for the future.

PR #699, opened 2016-01-19
Add ACLs to policy groups

jkeiser:


marcparadise:
:+1: after changing the commit summary msg to include an [api-docs] prefix, makes it easier when building changelogs.

PR #700, opened 2016-01-20
Add test for knife-opc org creation

jaym:
There was a breakage in knife-opc caused by changes to chef/chef.
Creating an org with a user in it raises the following:
```
Exception: NoMethodError: undefined method `get_rest' for #<Chef::ServerAPI:0x00000002347a18>`
```

The test will pass when https://github.com/chef/chef/pull/4432 is merged


stevendanna:
@jaym Looks good to me....but why are the tests passing already?

jaym:
@stevendanna they are not getting run against chef-zero: https://github.com/chef/chef-zero/blob/master/spec/run_oc_pedant.rb#L99

tylercloke:
:+1: sanity tests!

PR #702, opened 2016-01-20
Allow opt-out of RVM/bundler busting in knife pedant tests

danielsdeleo:
Travis runs ruby using rvm, but the environment variable munging we're doing to run knife in the pedant tests is breaking rvm. In order to run these tests against Chef Zero, we need a way to opt-out. Since Chef Zero includes Chef in its bundle, it's also not necessary to "bundle bust" the command, so I made the opt-out disable that as well. Since my goal is to make this work on travis, the easiest path to take is to switch on an environment variable, but I could make this part of pedant's config if there's a desire to do that instead.

This seems to have fixed https://github.com/chef/chef-zero/pull/188 (there was an error on the travis side, so travis is in the process of re-running those things right now).

@chef/chef-server-maintainers

### DIFF
--- a/oc-chef-pedant/lib/pedant/rspec/knife_util.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/knife_util.rb
@@ -73,17 +73,30 @@ module Pedant
         # Convenience method for creating a Mixlib::ShellOut representation
         # of a knife command in our test repository
         def shell_out(command_line)
-          # All the environment variable munging is so Bundler doesn't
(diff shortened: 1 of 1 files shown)


stevendanna:
This works for now. If we need an option later, we can add it then. :+1:

danielsdeleo:
I think the travis failures are failures in travis. Status page says something about their logging infrastructure being busted, which maybe is resulting in builds looking like they failed.

marcparadise:
:+1:

PR #703, opened 2016-01-21
Add validation tag to header validation test

danielsdeleo:
...so Chef Zero can skip them because it doesn't implement checking for things that are 400s.

@chef/chef-server-maintainers


sdelano:
:+1:

PR #704, opened 2016-01-21
HA-5 - standalone front-end bootstrapping 

marcparadise:
I will re-open this PR against master once #677 merges, but this shows the diff more effectively without pulling in extra stuff. This PR: 

- adds a bootstrap bypass when we determine we shouldn't be running bootstrap 
- adds bootstrap preflight check to ensure : 
 * pivotal.pem and private-chef-secrets.json are in a consistent state based on whether we have a previous run or not 
 * Ensures if we're bypassing data bootstrap on a node, pivotal user is present in opscode_chef db. 
 * Ensures that if we're not bypassing data  bootstrap, pivotal user is NOT present 
- blocks out further additions we'll want to consider for bootstrap preflight when we're in the 'polishing' phase of this work. 
- factors up postgres-related stuff from PreflightPostgresValidator into the base PreflightValidator class, since it's now needed in two places. 

Pending: 
* [x] track down some pedant test failures: sporadic errors (due to insufficent FE node memory) 
* [ ] figure out why bookshelf tests are failing now. 
* [ ] clean up the git history
* [x] couple of debug statements to remove
* [ ] add silent mode to pgsql retries and use in preflight checks that expect failure
* [ ] fix preflight file name
* [ ] add remediation for bootstrap errors 6 and 7 (@sdelano)
* [ ] add database exists checks before checking pivotal user


omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap.validator.rb
```diff
@@ -0,0 +1,238 @@
+#
```
stevendanna: This file is named a bit oddly.  Should it be `preflight_bootstrap_validator.rb`?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap.validator.rb
```diff
@@ -0,0 +1,238 @@
...
+
+class BootstrapPreflightValidator < PreflightValidator
+  def initialize(node)
+    super
+  end
+
+  def run!
+    # Sanity checks: for any configuration we should expect that
+    # neither pivotal.pem nor private-chef-secrets.json exists, or that both exist.
+  puts "Checking bootstrap!"
```
stevendanna: indent level off here

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
```diff
@@ -4,23 +4,24 @@ def self.with_connection(node, database = 'template1', opts = {})
...
         retries -= 1
         sleep_time = 5 - retries
         sleep_time *= sleep_time
-        Chef::Log.warn "Error connecting to postgresql: #{e.message}, retrying after #{sleep_time}s sleep. #{retries} retries remaining."
+        Chef::Log.debug "Error connecting to postgresql: #{e.message}, retrying after #{sleep_time}s sleep. #{retries} retries remaining."
         sleep sleep_time
         retry
       else
-        Chef::Log.warn "Error from postgresql: #{e.message}, retries have been exhausted."
+        Chef::Log.info "Error from postgresql: #{e.message}, retries have been exhausted."
```
sdelano: should this be `error` or `fatal` instead of `info`?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap.validator.rb
sdelano: or even `bootstrap_preflight_validator.rb` to match the class (I know there's some oddity with the alphabet sorting).

@marcparadise perhaps the `.` vs `_` is why you were seeing odd errors this morning?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap.validator.rb
```diff
@@ -0,0 +1,238 @@
...
+      puts e.message
+      # DB not present, or other more critical failure that will be captured
+      # down the line via  postgres preflight.
+      false
+    end
+  end
+
+
+  #TODO:
+  #These error messages and condition will need further refinement, details, and validation.
```
sdelano: Yes.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap.validator.rb
marcparadise: Ugh. In an earlier iteration I had it this way, I then renamed it with _s , but then kept editing the one with .s.  I'll fix it. 

I don't think it's related @sdelano  - the sort would remain the same, but specific 'require_relative' fixed it.

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
marcparadise: It was, but then it start spamming the prostgresvalidator preflight checks with alarming fatal messages when errors were actually expected.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
```diff
@@ -47,6 +58,42 @@ def backend?
...
+                                                'port' => port,
+                                                # By default, pass exceptions up without retrying
+                                                # We have a few scenarios in which we
+                                                # expect errors, and don't want to delay the bootstrap
+                                                # for unnecessary retries.
+                                                'retries' => 0 }) do |conn|
+       if block_given?
+         yield(conn)
+       end
+    end
```
sdelano: Shouldn't all of these postgres-specific functions exist in the postgres validator, or are they shared by the bootstrap validator?

sdelano: I don't see them used anywhere except the postgres validator, which is where they've been moved from.

marcparadise: connect_as is shared between them.  The other two were done preemptively - when it's time to refine and finalize the bootstrap validator, they will also need to be there. 

I can actually use one of them to get rid of a nested begin/rescue/end block, now that I think about it. I'll make that change, and move the roles checking function back into pg validator until we need it elsewhere.

dev/Vagrantfile
```diff
@@ -98,6 +98,8 @@ def define_chef_server(config, attributes)
     # We're also going to do a share of the slower vboxsf style, allowing us to auto-checkout dependencies
     # and have them be properly synced to a place that we can use them.
     config.vm.synced_folder installer_path, "/installers"
+    chef_backend_config_path = File.absolute_path(File.join(Dir.pwd, "../../chef-backend/dev"))
```
marcparadise: This will be removed

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
```diff
@@ -4,23 +4,24 @@ def self.with_connection(node, database = 'template1', opts = {})
...
+    retries = opts['retries'] || 5
     begin
       connection = ::PGconn.open('user' => postgres['db_superuser'],
                                  'host' => postgres['vip'],
                                  'password' => postgres['db_superuser_password'],
                                  'port' => postgres['port'],
                                  'dbname' => database)
-    rescue e
+    rescue => e
+      raise if opts['caller_wants_exception']
```
marcparadise: Please to delete

omnibus/files/private-chef-cookbooks/private-chef/libraries/ec_postgres.rb
sdelano: I think it makes sense to default to `:warn` and `:error` here and have the preflight validators pass in a `:silent` flag to turn off the output (they expect failure sometimes and will rescue accordingly).

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap.validator.rb
```diff
@@ -0,0 +1,238 @@
...
+BOOT007: The secrets file (/etc/opscode/private-chef-secrets.json) is present
+         but the file /etc/opscode/pivotal.pem is missing.
+
+         Pending: remediation walkthrough.
+EOM
+  end
+
+  def err_BOOT008_pivotal_public_key_mismatch
+<<EOM
+BOOT007: The pivotal key file /etc/opscode/pivotal.pem exists, but its public key
```
sdelano: `BOOT007`? (more of these below as well)

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap.validator.rb
```diff
@@ -0,0 +1,238 @@
...
+EOM
+
+  end
+
+  def err_BOOT007_secrets_with_no_pivotal
+<<EOM
+BOOT007: The secrets file (/etc/opscode/private-chef-secrets.json) is present
+         but the file /etc/opscode/pivotal.pem is missing.
+
+         Pending: remediation walkthrough.
```
sdelano: Let's add the remediation for 6 and 7.

PR #706, opened 2016-01-25
Fix rspec failures for BookshelfPreflightValidator

stevendanna:
The base class access the postgresql configuration in its initialize
methods, forcing us to mock that out in our tests.


sdelano:
:+1:

manderson26:
:+1:

PR #710, opened 2016-01-26
[bookshelf/omnibus] Convert storage_type to a string consistently

stevendanna:
A new function, transform_to_consistent_types, now converts the
bookshelf storage_type config value to a string when parsing the
user-provided configuration. Internally, we now assume that it is always
a string.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bookshelf_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bookshelf_validator.rb
@@ -38,11 +38,11 @@ class BookshelfPreflightValidator < PreflightValidator
       true
     else
       previous_value = previous_run['bookshelf']['storage_type']
-      current_value = user_attrs['storage_type'] || :filesystem
+      current_value = user_attrs['storage_type'] || 'filesystem'
 
-      if previous_value.nil? && current_value.to_s == 'filesystem' # case (2)
+      if previous_value.nil? && current_value == 'filesystem' # case (2)
         true
-      elsif previous_value.nil? && current_value.to_s != "filesystem" # case (3)
+      elsif previous_value.nil? && current_value != "filesystem" # case (3)
         fail_with <<EOM
 
 Bookshelf's storage_type was previously the default of 'filesystem';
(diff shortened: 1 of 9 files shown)


stevendanna:
@sdelano I believe this should address the issue you pointed out yesterday about all the calls to to_s for the storage_type code.

manderson26:
Looks reasonable to me.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -121,6 +121,14 @@ def import_legacy_service_config(old_service_key, new_service_key, keys)
       end
     end
 
+    # Mutate PrivateChef to account for common cases of user-provided
+    # types not being what we want
+    def transform_to_consistent_types
+      if PrivateChef['bookshelf']['storage_type'] && PrivateChef['bookshelf']['storage_type'].is_a?(Symbol)
```
marcparadise: Can we just `to_s` here without the `is_a?` conditional?

marcparadise:
:+1: , just the one comment.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bookshelf_validator.rb
```diff
@@ -67,7 +67,7 @@ def verify_storage_type_unchanged
 
 Please set
 
-bookshelf['storage_type'] = :#{previous_value}
+bookshelf['storage_type'] = '#{previous_value}'
```
tylercloke: I'm confused about how this previously worked. IRB doesn't seem to like `:#{previous_value}`, so I'm unsure, but it seems like we want double quote here? We probably want to eval `previous_value` instead of passing `#{previous_value}` to storage_type?

tylercloke: Or actually, if previous value is a string now, do we just want?:

`bookshelf['storage_type'] = previous_value`

stevendanna: Note that this is all inside a a HEREDOC.  So this is basically changing:

    ":#{previous_value}"

to 

    "'#{previous_value}'"

tylercloke:
:+1:

marcparadise:
Are we good to go ahead with this? Or should we close it out?

manderson26:
I'm still +1 for merging.

stevendanna:
@marcparadise I've updated this to address your feedback.  IIRC this was just some cleanup and wasn't to address anything in particular, but it still looks reasonable to me.

PR #711, opened 2016-01-26
[bookshelf] Return 204 on cookbook upload

stevendanna:
Despite the old code that was attempting to set the response code to
202, we have tests that depend on these endpoints returning 204.


sdelano:
:+1:

manderson26:
:+1:

PR #714, opened 2016-01-27
Create cookbook artifacts with all fields filled in

danielsdeleo:
Uploads from exising client tooling will include empty arrays for file
types that aren't present in the cookbook. Changing the test to better
reflect what happens in real use.

Really, erchef should be normalizing this data on the way in or out so that it's always serving data with these fields but for now this change will allow us to implement the endpoint in Chef Zero in a way that works correctly with Chef Client.

@chef/chef-server-maintainers


stevendanna:
Do we think we want a new set of tests that test the "minimal" payload?  That is, do we *require* clients to send all the fields?  If not, do we need to keep some test coverage for clients that don't send all the fields?

danielsdeleo:
It's a _de facto_ requirement for clients to send all fields because chef-client won't work if you don't, but it's not enforced anywhere. I think the optimal solution would be for erchef to normalize all the different segments to empty arrays if not present but I don't have time to do that just now.

stevendanna:
I made a quick issue as a placeholder.  :+1: on this change to move chef-zero forward.

PR #715, opened 2016-01-27
bifrost and bookshelf schemas explicit upgrade

marcparadise:
Instead of upgrading bifrost and bookshelf schemas to the
latest with every chef-server-ctl reconfigure, we're taking the more
conservative approach used for the opscode_chef schema:

1. Install the schema to the latest version on a new install
2. Changes to schemas during upgrades must be explicitly performed
   via a partybus migration.

### DIFF
--- a/omnibus/files/private-chef-upgrades/001/022_cbv_type_addition.rb
+++ b/omnibus/files/private-chef-upgrades/001/022_cbv_type_addition.rb
@@ -1,6 +1,7 @@
 define_upgrade do
   if Partybus.config.bootstrap_server
     must_be_data_master
-    run_sqitch("@cbv-type", "@1.0.4")
+    # Performance improvements for cookbook fetching.
+    run_sqitch('@cbv-type', 'oc_erchef')
   end
 end
--- a/omnibus/partybus/lib/partybus/migration_api/v1.rb
+++ b/omnibus/partybus/lib/partybus/migration_api/v1.rb
@@ -97,19 +97,39 @@ EOF
         runner.run_command(command, options)
       end
 
-      def run_sqitch(ec_target, osc_target="@1.0.4")
-        args =  " --db-user #{Partybus.config.postgres['db_superuser']}"
-        args << " --db-name opscode_chef"
-        args << " --db-host #{Partybus.config.postgres['vip']}"
-        args << " --db-port #{Partybus.config.postgres['port']}"
-        args << " --engine pg"
+      # Caller may use 'opts' to override :username :password,
+      # :database, and/or :path.
+      def run_sqitch(target, service, opts = {})
+        options = default_opts_for_service(service).merge(opts)
+        command = <<-EOM.gsub(/\s+/," ").strip!
+          sqitch --engine pg
+            --db-name #{options[:database]}
+            --db-host #{Partybus.config.postgres['vip']}
+            --db-port #{Partybus.config.postgres['port']}
+            --db-user #{options[:username]}
+            --top-dir /opt/opscode/embedded/service/#{options[:path]}
+            deploy #{target} --verify
+        EOM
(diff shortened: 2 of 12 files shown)


marcparadise:
Testing in progress, but opening for review now. Note that depending on ongoing discussion, we may expand this to include the necessary partybus changes to support the requirement. The current state of partybus means that we can't perform a schema migration against anything other than the `opscode_chef` database.

stevendanna:
> we may expand this to include the necessary partybus changes to support the requirement

I think that is a good idea.  I'd like the support to be there for the next person who has to do a schema update.

omnibus/partybus/lib/partybus/migration_api/v1.rb
```diff
@@ -97,19 +97,26 @@ def run_command(command, options={})
...
-        cmd << " && cd baseline && sqitch #{args} deploy --to-target #{osc_target} --verify"
-        cmd << " && cd .. && sqitch #{args} deploy --to-target #{ec_target} --verify;"
-        run_command(cmd,
-                    :cwd => "/opt/opscode/embedded/service/opscode-erchef/schema",
-                    :env => {"PGPASSWORD" => Partybus.config.postgres['db_superuser_password']})
+      # credentials is a required hash containing
+      # EITHER: :service - name of key into 'secrets' that should contain
+      #                    sql_user and sql_password entries,
+      #    OR:  :username, :password for sql auth
+      def run_sqitch(target, path, database, credentials)
```
stevendanna: Perhaps this might be a bit nicer from the callers perspective if we just had to pass in a single 'service' argument that would know how to look up the correct path, username, password, and database.  We could leave the final param as an optional opts hash for when you need to set things manually:

```
def run_sqitch(target, service, opts={})
   username = opts[:username] || username_for_service(service)
   password =  opts[:password] || password_for_service(service)
   database = opts[:database] || database_for_service(service)
   path = opts[:path] || path_for_service(service)
   ...
```

stevendanna:
Overall, looks good.  Made a minor suggestion regarding the API of run_sqitch, let me know what you think.

omnibus/partybus/lib/partybus/migration_api/v1.rb
marcparadise: I started down that path, but currently can't recall why I didn't stick with it.  In any case - that sounds good to me.

marcparadise:
Updated with a variant on the comments above.

omnibus/files/private-chef-upgrades/001/022_cbv_type_addition.rb
```diff
@@ -1,6 +1,7 @@
 define_upgrade do
   if Partybus.config.bootstrap_server
     must_be_data_master
-    run_sqitch("@cbv-type", "@1.0.4")
+    # Performance improvements for cookbook fetching.
+    run_sqitch('@cbv-type', 'oc_erchef/schema')
```
stevendanna: This should be 'oc_erchef' I think?

sdelano:
:+1:

omnibus/partybus/lib/partybus/migration_api/v1.rb
```diff
@@ -97,19 +97,40 @@ def run_command(command, options={})
...
+            --db-port #{Partybus.config.postgres['port']}
+            --db-user #{options[:username]}
+            --top-dir /opt/opscode/embedded/service/#{options[:path]}
+            deploy #{target} --verify
+        EOM
+        run_command(command, env: {"PGPASSWORD" => options[:password]})
+      end
+
+      def default_opts_for_service(service)
+        username = Partybus.config.secrets[service]['sql_user']
```
sdelano: The `sql_user` isn't stored in the `secrets` hash, which is why this is failing tests.

PR #716, opened 2016-01-27
Tag tests with :authorization, remove degenerate test

jkeiser:


jkeiser:
@chef/chef-server-maintainers @marcparadise

oc-chef-pedant/spec/api/account/account_acl_spec.rb
```diff
@@ -14,7 +14,7 @@
 
     context "GET /users/<user>/_acl"  do
 
-      let(:actors) { ["pivotal", username] }
+      let(:actors) { ["pivotal", username].uniq }
```
stevendanna: I'm confused by this.  It must be a no-op yes?

oc-chef-pedant/spec/api/account/account_acl_spec.rb
```diff
@@ -29,7 +29,7 @@
         it "can get user acl" do
           get(request_url, platform.superuser).should look_like({
              :status => 200,
-             :body_exact => acl_body
+             :body => acl_body
```
stevendanna: @jkeiser Can you provide a more detailed commit message that explains why this is correct since it is presumably making the test less strict and allowing some cases that are currently prohibited?

oc-chef-pedant/spec/api/account/account_acl_spec.rb
jkeiser: @stevendanna not if username is `pivotal` :) Which is the issue we were encountering, over and over.

oc-chef-pedant/spec/api/account/account_acl_spec.rb
jkeiser: @stevendanna will do. The issue is that the server returns things like `[ 'pivotal', 'pivotal' ]` and cares about the order of the members of actors and groups, and the tests shouldn't care about duplicates or order. `body` lets us handle that.

oc-chef-pedant/spec/api/account/account_acl_spec.rb
stevendanna: nvm, missed that username wasn't a string.

stevendanna: D'oh. MIssed that wasn't a string. Thanks!

stevendanna:
:+1:

PR #720, opened 2016-01-28
[dvm] change dvm install package attr name

marcparadise:
Ohai recently added a 'packages' top-level attribute, which
doesn't place nicely with dvm's usage of node['packages'] to
capture which additional packages to install at startup.

This change just renames the attribute to install_packages to avoid
conflict.


stevendanna:
:+1:

PR #721, opened 2016-01-29
remove unsupported upgrades

marcparadise:
1.      remove upgrades that exist under EC 11.
2.      remove  Sequel db migration support

More details are in the two commit messages, but this removes migrations that are not reachable in an EC11.1+ upgrade to 12.  It also removes Sequel support from partybus, since it was only used in the removed migrations.


marcparadise:
For reference, here is the state of migrations as of 11.1.0: 

 https://github.com/chef/chef-server/tree/11.1.0/files/private-chef-upgrades/001

marcparadise:
Note that I'll be running this through the upgrade test pipeline prior to merging, and will post here once that's done.

marcparadise:
This has passed upgrade testing pipeline as well.

tylercloke:
This generally looks good! A couple of quick questions:

1) I'm not seeing in the docs that we only support 11.3.0 or later. If this is actually our policy, we should update the docs before / right after we merge this.
2) Can we add a check that checks the version being upgraded from? I'd hate for a migration to just crash out because someone was on EC 11.2.0. Be nice if it crashed with an error message explaining what needed to happen.

stevendanna:
> 2) Can we add a check that checks the version being upgraded from to our upgrade command somewhere? I'd hate for a migration to just crash out because someone was on EC Less Than 11.3.0. Be nice if it crashed with an error message explaining what needed to happen.

Given how much work we've been putting into providing better errors, I think this is a must-have for this PR.

marcparadise:
The next steps for this item: 

- set up Conflicts directive for deb package 
- set up Conflicts directive or preinst script for rpm package that prevents installing from versions we can't upgrade from[
- update official upgrade documentation to reflect  the minimum version requirement (EC 11.2 or later) for upgrades.

Additional note: this will not impact upgrades from Open Source CS 11, as that follows a different path.

stevendanna:
I think we should consider moving some version of this forward.  As we've seen in some recent support situation, in most cases the presence of these migrations only serve to confuse.  In addition to Marc's list, I would like to see:

- `chef-server-ctl upgrade` Checks the migration-level and bails out if it doesn't match the minimum migration level we support (also bail out if it doesn't exist)

stevendanna:
Closing because of the age.

PR #725, opened 2016-02-01
Only define LINE_SEP on first load

stevendanna:
Since we call require_relative on this file multiple times, the
constant definition produces warnings during the `chef-server-ctl
reconfigure`. Guarding the defintion with defined? prevents the
warning.


marcparadise:
:+1:

PR #729, opened 2016-02-03
[chef-server-ctl] Cleanse bookshelf database when storage_type is sql

stevendanna:
The key is storage_type, not storage_mode.


sdelano:
:+1:

PR #734, opened 2016-02-10
Updating gatling-rsync configuration

dmccown:
This is to fix a change in `0.9.0` version of
`vagrant-gatling-rsync` where rsync will run by default. Here's the
changelog from the plugin:
https://github.com/smerrill/vagrant-gatling-rsync/blob/v0.9.0/CHANGELOG.md#090-june-28-2015.


stevendanna:
Looks reasonable to me :+1:

marcparadise:
:+1:

tylercloke:
Make it all faster! :+1:

PR #735, opened 2016-02-11
Upgrade Posgtresql to 9.2.15.

rhass:
This mitigates CVE-2016-0773.

As a side note, is there any reason not to update this in omnibus-software and then just reference the correct version? I didn't see anything in the files which would compile them differently.

### DIFF
--- a/omnibus/config/software/postgresql92.rb
+++ b/omnibus/config/software/postgresql92.rb
@@ -1,5 +1,5 @@
-#
-# Copyright 2012-2014 Chef Software, Inc.
+#1
+# Copyright 2012-2016 Chef Software, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


marcparadise:
@rhass yes, complications around upgrading major version, and data directory locations.

rhass:
I realize upgrading between major versions is a much more involved issue, but I what mean is to just define the `default_version` value and then keep the other bits like the hash and whatnot in the omnibus-software repo.

rhass:
Nevermind, I missed this line: https://github.com/rhass/chef-server/blob/CVE-2016-0773/omnibus/config/software/postgresql92.rb#L35

omnibus/config/software/postgresql92.rb
```diff
@@ -15,10 +15,10 @@
 #
 
 name "postgresql92"
-default_version "9.2.14"
+default_version "9.2.15"
 
-source url: "http://ftp.postgresql.org/pub/source/v9.2.14/postgresql-9.2.14.tar.bz2",
-       md5: "ce2e50565983a14995f5dbcd3c35b627"
+source url: "http://ftp.postgresql.org/pub/source/v9.2.15/postgresql-9.2.15.tar.bz2",
```
stevendanna: Mind changing this to https while you're at it?

stevendanna:
Other than my minor comment, :+1:.

omnibus/config/software/postgresql92.rb
rhass: Ah! Good catch, Done!

marcparadise:
:+1:

stevendanna:
@rhass Apologies for the delay, mind rebasing this? This is ready to merge after rebase.

rhass:
@stevendanna Should be good to go.

PR #740, opened 2016-02-16
Split keys_spec.rb into user_, client_keys_spec.rb

jrunning:
Supports work to bring chef-zero closer to parity with chef-server. See chef/chef-zero#199.


danielsdeleo:
:+1: For some background, user multi-key is troublesome in some modes of Chef Zero operation that don't support multi-org (like when using ChefFS). The way the tests were organized before, you could skip the actual tests by tagging them appropriately, but many of the `before` blocks would still run tasks relevant just to users.

tylercloke:
Running it through wilson http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/53/downstreambuildview/

jrunning:
Fixed issues with erchef (hopefully), running a new job on Wilson now: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/54/downstreambuildview/

tylercloke:
:+1: Looks reasonable.

Issue #741, opened 2016-02-17
chef-server-ctl install fails on Amazon Linux


jamesc:
The yum. repos.d  is written as:

`* template[/etc/yum.repos.d/chef-stable.repo] action create
       [chef-stable]
       name=Chef Stable Repo
      baseurl=https://packagecloud.io/chef/stable/el/2015/$basearch
       enabled=0
       gpgcheck=1
       gpgkey=file:///opt/opscode/embedded/keys/packages-chef-io-public.key`

this is due to platform_family being e.g `2016.09` instead of `7.X`

Responsible code is https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/recipes/add_ons_repository.rb#L29

This needs to have special casing for amazon linux (always use 7 ?)

juliandunn:
I'm going to assume this is no longer an issue given that we are shipping it as part of the AIO on Opsworks which is on Amazon Linux.

PR #742, opened 2016-02-17
Add logic to support configure yum repos for Amazon Linux (#741)

jamesc:
We need to hard-code AL to use EL7 repositories

ChangeLog-Entry: [omnibus] [chef-server/741] Support chef-server-ctl install on Amazon Linux


marcparadise:
:+1:

mmzyk:
:+1:

Issue #743, opened 2016-02-17
Using public EC2 name for manage + API exceeds nginx hash bucket size


jamesc:
If you use a public EC2 name e.g. `ec2-52-36-198-136.us-west-2.compute.amazonaws.com` for the `fqdn_api` in `chef-server.rb` and the `fqdn` in `manage.rb` then nginx won't start and shows in the logs:

```2016-02-17_13:29:17.97029 nginx: [emerg] could not build the server_names_hash, you should increase server_names_hash_bucket_size: 64```

This already was patched in chef-analytics with an insightful comment added at https://github.com/chef/chef-analytics/blob/master/omnibus/files/cookbooks/opscode-analytics/templates/default/nginx/analytics.conf.erb#L24-L27.

We should do the needful.

PR #745, opened 2016-02-17
Increase default nginx server_names_hash_bucket_size to 128 from 64

jamesc:
This is needed on standalone/combined installs on EC2 where the
combination of several long DNS names can take you over the default
hash_bucket_size.

ChangeLog-Entry: [omnibus] [chef-server/743] Increase default nginx
server_names_hash_bucket_size to 128 from 64

Fixes https://github.com/chef/chef-server/issues/743

How to repro on EC2:
1. Set the api_fqdn for chef-server.rb to the external FQDN
2. Set the fqdn in manage.rb to another name, e.g. a CNAME to that FQDN

Do reconfigure on manage and chef-server.  nginx will start throwing the errors if bucket_size is 64.


marcparadise:
:+1:

PR #747, opened 2016-02-18
Tag tests that expect 400 return w/ `validation`

danielsdeleo:
I'm investigating adding API v1 support for Zero, and found these tests that expect HTTP 400 return codes but do not have validation tags.

@chef/chef-server-maintainers


marcparadise:
:+1:

stevendanna:
:+1: MWR

PR #749, opened 2016-02-22
Update how we pin ruby version in omnibus

ksubrama:
This needs to go in after https://github.com/chef/omnibus-software/pull/595 has been merged and our Gemfile.lock is correctly bumped.


stevendanna:
@ksubrama Mind writing a commit message that explains what is going on here.  From the looks of it this isn't changing what version of ruby we will get or how it is getting built, it is just using some new omnibus-software changes that are there to better support other platforms?

ksubrama:
I'm withdrawing this because we're holding off on that refactor - so there's no impact here.

PR #753, opened 2016-02-25
Make org creation optional in APIv1 spec

danielsdeleo:
This makes pedant fail when running against servers that don't have multi-org (Chef Zero in ChefFS mode).


stevendanna:
Seems reasonable. :+1:

PR #754, opened 2016-02-26
[omnibus] Remove old access_by_lua nginx config and allow custom acce…

ryancragun:
This removes the old `access_by_lua` entry that probably could have been removed in https://github.com/chef/chef-server/commit/f22c1f111e94751f5b1d02227390fa10fe964478

It will also allow us to set an `access_by_lua_file` location in `chef-server.rb` which is super useful for some work that we have to do for cloud marketplaces.


stevendanna:
:+1: Seems reasonable to me.

tylercloke:
:+1: A comment explaining what this lua file does (and explaining that it isn't the lua routing code) would be nice. If pedant passes, I'm good with this.

ryancragun:
I updated the commit message, left a comment in the nginx template, and will test pedant with a build soon.  http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/81/downstreambuildview/

tylercloke:
:+1: if pedant is passing!

Issue #688, opened 2016-01-12
rabbitmq['management_enabled'] setting doesn't actually do anything


jeremiahsnapp:
The `rabbitmq['management_enabled']` setting doesn't actually do anything.

https://github.com/chef/chef-server/blob/12.3.1/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb#L127

The ability to easily disable rabbitmq management would be really nice especially when trying to avoid other issues such as https://github.com/chef/chef-server/issues/686

stevendanna:
Looks like this was solved in https://github.com/chef/chef-server/commit/83ba83552a34d41f55f40f56c5b969174c3fdd8c  Thanks for reporting it!

PR #756, opened 2016-02-29
['private_chef']['rabbitmq']['management_enabled'] should be respected

jmink:
This is to solve https://github.com/chef/chef-server/issues/688
Previously management_enabled was being completely ignored

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
@@ -183,13 +183,24 @@ if is_data_master?
     retries 10
   end
 
-  execute "#{rmq_plugins} enable rabbitmq_management" do
-    environment (rabbitmq_env)
-    user opc_username
-    not_if "#{rmq_plugins} list | grep rabbitmq_management | grep -v rabbit_management_agent | grep E"
-    # management plugin needs a rabbit restart
-    notifies :restart, 'runit_service[rabbitmq]', :delayed
-    retries 10
+  rabbitmq_management_is_up = "#{rmq_plugins} list | grep rabbitmq_management  | grep -v rabbitmq_management_agent | grep -v rabbitmq_management_visualiser | grep E"
+  if rabbitmq['management_enabled']
+    execute "#{rmq_plugins} enable rabbitmq_management" do
+      environment (rabbitmq_env)
+      user opc_username
+      not_if rabbitmq_management_is_up
+      # management plugin needs a rabbit restart
+      notifies :restart, 'runit_service[rabbitmq]', :delayed
+      retries 10
+    end
+  else
+    execute "#{rmq_plugins} disable rabbitmq_management" do
+      environment (rabbitmq_env)
+      user opc_username
+      notifies :restart, 'runit_service[rabbitmq]', :delayed
+      if rabbitmq_management_status_is_up
+      retries 10
+    end
   end
 
   execute "#{rmq_ctl} set_user_tags #{rabbitmq['management_user']} administrator" do


stevendanna:
I'm wondering if we should add a pre-flight check to ensure that management is enable if https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb#L135 is enabled. Or to auto-tune one based on the other.

omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
```diff
@@ -186,12 +186,20 @@
   execute "#{rmq_plugins} enable rabbitmq_management" do
     environment (rabbitmq_env)
     user opc_username
-    not_if "#{rmq_plugins} list | grep rabbitmq_management | grep -v rabbit_management_agent | grep E"
+    not_if { !rabbitmq['management_enabled'] || shell_out("#{rmq_plugins} list | grep rabbitmq_management | grep -v rabbit_management_agent | grep E").status.success? }
```
stevendanna: I think it is possible to put two not_if attributes on the resource, might be a bit cleaner:

```
not_if { !rabbitmq['management_enabled'] }
not_if "#{rmq_plugins} list | grep rabbitmq_management | grep -v rabbit_management_agent | grep E"
```

A further cleanup would be to to put the command that checks whether rabbitmq is enabled in a library so it could be:

```
not_if  { !rabbitmq['management_enabled'] || RabbitHelper.plugin_enabled?("rabbitmq_management") }
```

but that is probably a bit much for this small fix.

sdelano: If `rabbitmq['management_enabled']` is a node attribute, why not just put all of that in an `if` block and leave the `not_if` the same?

stevendanna: That works too, in fact, I think I prefer it.  Let's do that instead.

sdelano: e.g.:

```
if rabbitmq['management_enabled']
  execute "rmq enable" do
    # things
  end
else
  exectue "rmq disable" do
    # things
  end
end
```

jmink: Happy to make that change.

jmink:
Updated!

ksubrama:
:+1: on the changes.

omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
```diff
@@ -183,13 +183,22 @@
...
+    execute "#{rmq_plugins} enable rabbitmq_management" do
+      environment (rabbitmq_env)
+      user opc_username
+      not_if "#{rmq_plugins} list | grep rabbitmq_management | grep -v rabbit_management_agent | grep E"
+      # management plugin needs a rabbit restart
+      notifies :restart, 'runit_service[rabbitmq]', :delayed
+      retries 10
+    end
+  else
+    execute "#{rmq_plugins} disable rabbitmq_management" do
```
sdelano: Should we additionally have a `not_if` block here on this service? What happens if we run `disable rabbitmq_management` multiple times w/o it enabled? We'll also restart rabbit on every chef-client run, which isn't ideal.

jmink: Good call.  There is one now.

jmink:
Updated

tylercloke:
:+1:

stevendanna:
I'm OK with this merging without the preflight check, we can always add one later. :+1:

sdelano:
:+1:

omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
```diff
@@ -183,13 +183,24 @@
...
+      environment (rabbitmq_env)
+      user opc_username
+      notifies :restart, 'runit_service[rabbitmq]', :delayed
+      not_if "#{rabbitmq_management_status} | grep -v E"
```
ksubrama: Don't you mean `'if`

jmink: No, I mean not_if.  What that block is doing is ensuring we only try to change the state of the addons when they aren't already in the desired state.

I could have used if without the -v, but -v looks for lines without the pattern.

ksubrama: D'oh!

ksubrama:
:+1:

PR #757, opened 2016-03-01
Bump omnibus-software to pick up latest server-jre

scottopherson:
omnibus-software now defaults the server-jre version to 8u74
so I've updated the Gemfile.lock to pick up this new version.
However, this newer version of omnibus-software no longer
contains the cacerts version we were overriding with (2014.08.20)
so I had to update that as well and chose to use the default.


stevendanna:
:+1: This looks OK to me.  These are the changes I get when running the changelog generator:

```
* cacerts (2014.08. -> 2016.01.)
* openssl (1.0.1q -> 1.0.1r)
* server-jre (8u31 -> 8u74)
* libxml2 (2.9.2 -> 2.9.3)
* nokogiri (1.6.3.1 -> gem install latest, probably 1.6.7.2)
* ohai (237129a0 -> a15a010c)
* appbundler (0.6.0 -> 959b9037)
* chef (09227432 -> 55753254)
```

ksubrama:
:+1:

manderson26:
:+1:

PR #760, opened 2016-03-01
Added event reporting to a file for insights

elliott-davis:
* This is	the first step in adding Insights eventing to chef-server for action events.
* Initially, Insights will hook into the same action eventing system as Analytics to consume the same events.
* Insights will need chef-server to persist its events to a file from which Insights will consume.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -93,6 +93,11 @@ default['private_chef']['opscode-solr']['data_dir'] = "/var/opt/opscode/opscode-
 ####
 default['private_chef']['server-api-version'] = 0
 
+####
+# Inisghts
+####
+default['private_chef']['insights']['enable'] = false
+default['private_chef']['insights']['log_dir'] = '/var/log/insights'
 
 ####
 # RabbitMQ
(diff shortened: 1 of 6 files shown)


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -93,6 +93,11 @@
 ####
 default['private_chef']['server-api-version'] = 0
 
+####
+# Inisghts
+####
+default['private_chef']['insights']['enable'] = false
+default['private_chef']['insights']['log_dir'] = '/var/log/insights'
```
stevendanna: We may want to make this`/var/log/opscode/insights` to match our other services.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_action_insights.erl
```diff
@@ -0,0 +1,27 @@
...
+
+-module(oc_chef_action_insights).
+
+-export([ingest/2]).
+
+-spec ingest(ShouldSend :: boolean(), Data :: binary()) -> ok.
+ingest(true, Data) ->
+    InsightsLogDir = envy:get(oc_chef_wm, insights_log, string),
+    InsightsLog = filename:join(InsightsLogDir, "insights"),
+    ok = file:write_file(InsightsLog, Data, [append]);
```
sdelano: My only concern with this is that `file:write_file` is most likely going to serialize access to this file through a single gen_server or process mailbox. We've seen issues with this in bookshelf doing synchronous request logging that caused the Erlang VM to hang, balloon in memory, and eventually crash.

elliott-davis:
Closing until we have a more baked idea of handling concurrent file writing

PR #764, opened 2016-03-02
[omnibus] Don't build rebar2, we don't use it

stevendanna:
All of the erlang projects vendor a version of rebar3 that they use to
build.


stevendanna:
Wilson passes: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/82//downstreambuildview

tylercloke:
Good catch :+1:

sdelano:
:+1:

ksubrama:
:+1:

PR #766, opened 2016-03-03
Added GET /groups/:group/transitive_member/actors/:actor endpoint for checking recursive membership.

tylercloke:
This endpoint will tell you if an actor is a member of a passed group or if that actor is a member of a group is recursively a member of a passed group.

For example, if brian is a member of the group clients, and clients is a member of the group superusers,
/groups/superusers/transitive_member/actors/brian would return true.

It returns JSON like:
  {"is_member" => true or false}

The framework is laid for adding GET /groups/:group/transitive_member/groups/:group as well, but this code currently depends on a db function called groups_for_actor,
an a similar groups_for_group does not exist yet. Notes have been left in the code to this effect.

Changelog-Entry: [bifrost] Added GET /groups/:group/transitive_member/actors/:actor endpoint for checking recursively checking group membership for a given actor on a given group.

[Build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/85/downstreambuildview/)

### DIFF
--- a/src/oc_bifrost/apps/bifrost/priv/pgsql_statements.config
+++ b/src/oc_bifrost/apps/bifrost/priv/pgsql_statements.config
@@ -53,6 +53,14 @@
       WHERE authz_id=$1
     ) AS exists">>}.
 
+{is_actor_recursive_member_of_group,
+<<"SELECT EXISTS(
+     SELECT auth_group.id FROM auth_actor, auth_group
+     WHERE auth_actor.authz_id=$1
+     AND auth_group.authz_id=$2 AND auth_group.id IN (
+       SELECT groups_for_actor(auth_actor.id))
+     ) AS exists">>}.
+
 %% Updating ACLs
 {update_acl,
  <<"SELECT update_acl AS success
--- a/src/oc_bifrost/apps/bifrost/src/bifrost_db.erl
+++ b/src/oc_bifrost/apps/bifrost/src/bifrost_db.erl
@@ -17,7 +17,8 @@
          has_permission/4,
          remove_from_group/3,
          statements/0,
-         update_acl/5]).
+         update_acl/5,
+         is_recursive_member_of_group/3]).
 
 ping() ->
     case select(ping, [], first_as_scalar, [ping]) of
@@ -184,6 +185,21 @@ group_membership(TargetType, GroupId) ->
             L;
         {ok, none} ->
             [];
+        {error, Error} ->
+            {error, Error}
+    end.
+
+%% TODO currently only supports actor since db query depends on db function
+%% groups_for_actor and a similar db function for groups does not exist.
+is_recursive_member_of_group_query(actor) -> is_actor_recursive_member_of_group.
+
+%% Returns all groups this actor is a member of,
+%% including parents of those groups, recursively.
+-spec is_recursive_member_of_group(auth_type(), auth_id(), auth_id()) -> boolean() | {error, term()}.
+is_recursive_member_of_group(Type, MemberAuthzId, ParentGroupAuthzId) ->
+    case select(is_recursive_member_of_group_query(Type), [MemberAuthzId, ParentGroupAuthzId], rows_as_scalars, [exists]) of
+        {ok, [Answer]} ->
+            Answer;
         {error, Error} ->
             {error, Error}
     end.
--- a/src/oc_bifrost/apps/bifrost/src/bifrost_wm_transitive_member_resource.erl
+++ b/src/oc_bifrost/apps/bifrost/src/bifrost_wm_transitive_member_resource.erl
@@ -1,0 +1,59 @@
+-module(bifrost_wm_transitive_member_resource).
+
+-include("bifrost_wm_rest_endpoint.hrl").
+-include_lib("stats_hero/include/stats_hero.hrl").
+
+-export([to_json/2]).
+
+-spec init([any()]) -> {ok, base_state()}.
+init(Config) ->
+    bifrost_wm_base:init(?MODULE, Config).
+
+-spec allowed_methods(wm_req(), base_state()) ->  {['GET'], wm_req(), base_state()}.
+allowed_methods(Req, State) ->
+    {['GET'], Req, State}.
+
+-spec auth_info('GET') -> read.
+auth_info('GET') ->
+    read.
+
+-spec validate_request(wm_req(), base_state()) ->
+			      {{halt, 401 | 403 | 404}, wm_req(), base_state()} |
+			      {boolean(), wm_req(), base_state()}.
+validate_request(Req, State) ->
+    bifrost_wm_base:validate_requestor(Req, State).
+
+%% TODO currently only supports actor since db query depends on db function
+%% groups_for_actor and a similar db function for groups does not exist, but
+%% will leave code in for groups in case we ever want to implement.
+-spec to_json(wm_req(), base_state()) ->
+		     {{halt, 404 | 400}, wm_req(), base_state()} |
+		     {binary(), wm_req(), base_state()}.
+to_json(Req, #base_state{reqid = ReqId,
+			 authz_id = GroupAuthzId,
+			 member_id = MemberAuthzId,
+			 member_type = MemberType} = State) ->
+    try
+	%% check if member_id exists since it doesn't get checked like base_state(authz_id)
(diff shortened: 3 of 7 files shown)


src/oc_bifrost/apps/bifrost/priv/pgsql_statements.config
```diff
@@ -53,6 +53,24 @@
       WHERE authz_id=$1
     ) AS exists">>}.
 
+{get_actor_id,
+<<"SELECT id FROM auth_actor WHERE authz_id=$1">>}.
```
ksubrama: These don't feel right.  id is a database internal identifier.  It's a bigserial, not a char(32).  It isn't supposed to be exposed outside the db if my intuition about how these schemas are designed is right.

src/oc_bifrost/apps/bifrost/src/bifrost_db.erl
```diff
@@ -88,6 +90,18 @@ exists(Type, AuthId) ->
     {ok, Answer} = select(StatementName, [AuthId], first_as_scalar, [exists]),
     Answer.
 
+get_id_query(actor)     -> get_actor_id;
+get_id_query(container) -> get_container_id;
+get_id_query(group)     -> get_group_id;
+get_id_query(object)    -> get_object_id.
+
+-spec get_id(auth_type(), auth_id()) -> not_found | boolean().
```
ksubrama: This type is definitely weird.  get_id_query returns a bigserial, not an auth_id() (which is a string).

src/oc_bifrost/apps/bifrost/priv/pgsql_statements.config
```diff
@@ -53,6 +53,24 @@
...
+<<"SELECT id FROM auth_container WHERE authz_id=$1">>}.
+
+{get_object_id,
+<<"SELECT id FROM auth_object WHERE authz_id=$1">>}.
+
+%% Returns every group an actor is a member of recursively,
+%% including parent groups.
+{get_all_recursive_groups_for_actor,
+<<"SELECT authz_id FROM auth_group 
+   WHERE id IN (SELECT groups_for_actor($1))">>}.
```
ksubrama: groups_for_actor has a comment above it saying that it takes internal database ids and is not designed to be called from outside.  But we pass $1 directly to it.  Wouldn't it be better to take an authz_id and then use `actor_id($1)`

src/oc_bifrost/apps/bifrost/src/bifrost_wm_transitive_member_resource.erl
```diff
@@ -0,0 +1,58 @@
...
+-spec to_json(wm_req(), base_state()) ->
+		     {{halt, 404 | 400}, wm_req(), base_state()} |
+		     {binary(), wm_req(), base_state()}.
+to_json(Req, #base_state{reqid = ReqId,
+			 authz_id = GroupAuthzId,
+			 member_id = MemberAuthzId,
+			 member_type = MemberType} = State) ->
+    try
+	%% check if member_id exists and get it's id (authz_id already validated)
+	case ?SH_TIME(ReqId, bifrost_db, get_id, (MemberType, MemberAuthzId)) of
```
ksubrama: If you only need the internal database id for this check, you can use the method exist/2 to perform the check.

src/oc_bifrost/apps/bifrost/src/bifrost_wm_transitive_member_resource.erl
```diff
@@ -0,0 +1,58 @@
...
+    catch
+	throw:{db_error, Error} ->
+            bifrost_wm_error:set_db_exception(Req, State, Error)
+    end.
+
+%% @private
+-spec is_member(request_id(), auth_type(), auth_id(), auth_id()) -> boolean().
+is_member(ReqId, MemberType, MemberId, GroupAuthzId) ->
+    %% Recursively get all groups that contain ActorAuthzId and see if GroupAuthzId is in the list.
+    case ?SH_TIME(ReqId, bifrost_db, get_all_parent_groups, (MemberType, MemberId)) of
```
ksubrama: A potential race exists where MemberId can be deleted (maybe?) between your looking and your use here.  Would you consider writing this entire membership check in SQL?

ksubrama:
I'm new to the world of erlang - so if some of my feedback makes no sense, it's because it probably doesn't.

src/oc_bifrost/oc-bifrost-pedant/spec/bifrost/groups_spec.rb
```diff
@@ -891,4 +891,95 @@ def self.creates_group_as(requestor, headers = {})
       end # for groups
     end # DELETE
   end # /groups/<group_id>/<member_type>/<member_id>
+
+  context '/groups/<group_id>/transitive_member/<member_type>/<member_id>' do
+    should_not_allow :PUT,
+      "/groups/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee/transitive_member/actors/ffffffffffffffffffffffffffffffff"
```
kmacgugan: Can you add a comment about it needing to be 32 characters in length? Otherwise I think someone might try to "refactor" the repeating letters.

kmacgugan:
+1, My comment is nitpicky but could be helpful for others trying to understand the test. Also looks like a squash is still needed but after that LGTM

src/oc_bifrost/oc-bifrost-pedant/spec/bifrost/groups_spec.rb
tylercloke: Sure, I can, but the rest of test code uses it.

kmacgugan: oh i don't mean refactor it, just add a note about why it is that long name? It was just confusing to me.

src/oc_bifrost/apps/bifrost/src/bifrost_db.erl
```diff
@@ -188,6 +189,21 @@ group_membership(TargetType, GroupId) ->
             {error, Error}
     end.
 
+%% TODO currently only supports actor since db query depends on db function
+%% groups_for_actor and a similar db function for groups does not exist.
+is_recursive_member_of_group_query(actor) -> is_actor_recursive_member_of_group.
```
Sgtpluck: should we match other potential types and leave a log message?

Sgtpluck:
+1, though i left a comment

src/oc_bifrost/apps/bifrost/src/bifrost_wm_transitive_member_resource.erl
```diff
@@ -0,0 +1,59 @@
...
+			 member_id = MemberAuthzId,
+			 member_type = MemberType} = State) ->
+    try
+	%% check if member_id exists since it doesn't get checked like base_state(authz_id)
+	case ?SH_TIME(ReqId, bifrost_db, exists, (MemberType, MemberAuthzId)) of
+	    false ->
+		{{halt, 404}, Req, State};
+	    true ->
+		IsMember = is_member(ReqId, MemberType, MemberAuthzId, GroupAuthzId),
+		{bifrost_wm_util:encode({[{<<"is_member">>, IsMember}]}), Req, State}
```
stevendanna: I think that the majority of the bifrost API communicates success/failure of a check via return codes rather than the response body.  Might we want to do that here as well?

stevendanna:
Overall this looks good to me.  As we discussed in chat, I'm unsure on the speed/style of the SQL query, but I think it will yield the correct result which is probably good enough for now.

src/oc_bifrost/apps/bifrost/src/bifrost_wm_transitive_member_resource.erl
tylercloke: Oh really? I was looking for an example but I couldn't find one.

tylercloke: Interesting, I'm not sure what HTTP code I'd use if it wasn't a member. I don't really wanna use 404 since then it'd be hard to tell if the actor wasn't a member of the group from the actor or group don't exist.

ksubrama:
:+1: on the SQL query - it looks better than before!

PR #768, opened 2016-03-07
DO NOT MERGE PLZ; Added event reporting to a file for insights

elliott-davis:
This will be the first of two PR's to our initial pass at adding insights to chef-server. This PR essentially taps the actions events and dumps them in a file.  The second PR will have a new service that attaches to this file for rotation and shipping.


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_action_insights.erl
```diff
@@ -0,0 +1,67 @@
+%% Copyright 2012-2014 Chef Software, Inc. All Rights Reserved.
```
metadave: old copyright

elliott-davis: guess that means I copied-wrong Caruso

stevendanna:
What are we thinking for rotation of this file?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_action_insights.erl
```diff
@@ -0,0 +1,67 @@
...
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+init([]) ->
+    file:open(envy:get(oc_chef_wm, insights_log, string), [append]).
+
+handle_call(_, _, State) ->
+    {reply, ok , State}.
+
+handle_cast({ingest, Data}, IODevice) ->
+    file:write(IODevice, [Data, "\n"]),
```
stevendanna: We might want to check the response and at least print a warning in the case of an error writing to this file.

elliott-davis:
@stevendanna Re: file rotation - I updated the description.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_action_insights.erl
stevendanna: More of a note to self: I'm assuming that we are writing the newline since the ingester is going to be line-oriented?  I believe that our json encoding library ensures that any newlines are transformed escaped so I think we are safe from there being any newlines in the data.  If that isn't the case, an alternative might be to write a fix-length field that would store the length of the content and then write the message.  The ingester could then read the length and then read the specified number of bytes.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_action.erl
```diff
@@ -93,15 +93,12 @@ log_action(Req, #base_state{resource_state = ResourceState} = State) ->
...
+    ShouldSendInsights = envy:get(oc_chef_wm, enable_insights, boolean),
     {FullActionPayload, EntityType, EntitySpecificPayload} = extract_entity_info(Req, ResourceState),
-    Payload = case ShouldSendBody of
-        true -> FullActionPayload;
-        false -> []
-    end,
     Task = task(Req, State),
     MsgType = routing_key(EntityType, Task),
-    Msg = construct_payload(Payload, Task, Req, State, EntitySpecificPayload),
+    Msg = construct_payload(FullActionPayload, Task, Req, State, EntitySpecificPayload),
```
stevendanna: Do we want to avoid constructing a payload for a message we aren't going to send?

stevendanna:
@elliott-davis Thanks for the update.  Overall this looks reasonable to me.  I think that in the long run we are going to want a more sophisticated gen_server that might allow us either to load-shed or provide backpressure if we can't write to this file fast enough to keep up with requests.   However, I don't think that is required at this time as it is still an in-development feature.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_action_insights.erl
```diff
@@ -0,0 +1,67 @@
...
+ingest(true, Data) ->
+    gen_server:cast(?MODULE, {ingest, Data});
+ingest(false, _) -> ok.
+
+-spec start_link() -> {ok, pid()}.
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+init([]) ->
+    file:open(envy:get(oc_chef_wm, insights_log, string), [append]).
```
stevendanna: I believe `file:open` returns {ok, File}, so down below in handle_cast you'll probably get a badarg error since you aren't passing an IODevice but {ok, IODevice}.  Additionally, even though this gen_server only has one bit of state now, you still might want to store it in a record.

src/oc_erchef/rebar.config
```diff
@@ -52,7 +52,9 @@
         {uuid, ".*",
          {git, "https://github.com/okeuday/uuid", {tag, "v1.3.2"}}},
         {sync, ".*",
-         {git, "https://github.com/rustyio/sync.git", {branch, "master"}}}
+         {git, "https://github.com/rustyio/sync.git", {branch, "master"}}},
+        {hoax, ".*",
```
stevendanna: This should probably go in the test profile down below rather than up here.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_action.erl
tduffield: We are still putting the Msg into the actions queue if actions is enabled

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_action_insights.erl
tduffield: We'll take a look at this when we add the ingester that will read from the file.

tduffield:
@stevendanna would love to chat about any gen_server optimizations you had in mind.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_action_insights.erl
elliott-davis: I think the reason this works is because init is expected to return {ok, State}. Since file:open returns {ok, IODevice} gen_server is okay with it. We could probably be more explicit about what is being returned, or as you suggested, return a record.

stevendanna: D'oh, you are correct, sorry about that.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -93,6 +93,11 @@
 ####
 default['private_chef']['server-api-version'] = 0
 
+####
+# Inisghts
+####
+default['private_chef']['insights']['enable'] = false
+default['private_chef']['insights']['log_file'] = '/var/log/insights/insights.log'
```
tduffield: Should this be nested under `/var/log/chef-server`?

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
```diff
@@ -7,10 +7,13 @@
...
 opscode_erchef_log_dir = node['private_chef']['opscode-erchef']['log_directory']
 opscode_erchef_sasl_log_dir = File.join(opscode_erchef_log_dir, "sasl")
+
+include_recipe 'private-chef::insights' if node['private_chef']['insights']['enable']
+
 [
   opscode_erchef_dir,
   opscode_erchef_log_dir,
-  opscode_erchef_sasl_log_dir
+  opscode_erchef_sasl_log_dir,
```
tduffield: Remove trailing comma?

stevendanna:
:+1: I think this is good to merge.

tduffield:
:+1:

mmzyk:
A few thoughts: the description on this PR is great. I'd love to start seeing more descriptions like that of why make it into commit messages too. The description from the PR will make it down into the merge commit, but as a general musing. If I only saw the commit message it would take me a lot longer to decipher what was going on.

Some actions tests were removed. Why was that? It appeared care was taken to make sure Actions still functioned (and it should) so I'm confused why the tests were removed.

Overall, from a brief read through, code structure seems fine and I'm :+1:. This seems like a great candidate to do a code walk through with someone from the chef server team on, though. It would help spread knowledge and I think would spur discussion around design that a PR like this tends to miss.

stevendanna:
> The description from the PR will make it down into the merge commit, but as a general musing. 

I don't think that is true.  Only the summary line ends up in the merge commit.

mmzyk:
@stevendanna Huh. Be good to confirm that for future reference, but I wouldn't hold up this PR over it. We can adjust as we go.

elliott-davis:
Re: removal of the tests - There were two pairs of tests that were essentially duplicates of each other.  The difference between the tests in the pairs was a set_env that was never added in sys.config. As a part of this PR, the env var check was removed (since it wasn't used) and consequently the env toggle in the tests wasn't needed. 

TL;DR defunct test cruft was removed due to tested code path removal.

mmzyk:
@elliott-davis :+1: Appreciate the updates and explanations.

elliott-davis:
Approaching this a different way from insights. Closing.

PR #769, opened 2016-03-07
Added /orgs/org/users/user/keys(/key) endpoints and changed default perms on org scoped key GETs.

tylercloke:
Merge https://github.com/chef/chef-zero/pull/205 and re-run Travis once two :+1: acquired.

Now, all org-scoped key GETs will allow access to any requestor within the same org (and the superuser).
The endpoints affected are
```
+ /orgs/org/clients/client/keys
+ /orgs/org/clients/client/keys/key
+ /orgs/org/users/user/keys
+ /orgs/org/users/user/keys/key
```
of which the latter two are new endpoints.

The new permissions where implemented by overriding default permissions for those GETs and checking if the requestor of actor_keys_access_group instead (which is a global group that exists per org that contains the users and clients groups, similar to the existing read_access_group).

The existing user key endpoints
```
+ /users/user/keys
+ /users/user/keys/key
```
were not modified and contain the same permissions as beforehand.

A large part of the diff is just tests being reorganized.

[Passing Build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/107/downstreambuildview/)


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_routes.erl
```diff
@@ -116,6 +116,12 @@ org_route(client_key, Req, Args) ->
     {BaseURI, Org} = extract_from_req(Req),
     Template = template_for_type(client_key),
     render_template(Template, BaseURI, [Org, ParentName, Name]);
+org_route(org_user_key, Req, Args) ->
+    {object_name, ParentName} = lists:keyfind(object_name, 1, Args),
+    {name, Name} = lists:keyfind(name, 1, Args),
+    {BaseURI, Org} = extract_from_req(Req),
+    Template = template_for_type(client_key),
```
stevendanna: Should this be `template_for_type(org_user_key)`?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_keys.erl
```diff
@@ -114,30 +159,91 @@ auth_info(Req, #base_state{resource_args = TargetType, resource_state = #key_sta
...
+    GroupId = Group#oc_chef_group.authz_id,
+    case oc_chef_authz:is_actor_transitive_member_of_group(oc_chef_authz:superuser_id(), ActorId, GroupId) of
+        {error, not_found} ->
+            %% Let oc_chef_wm_base:forbidden sort things out, it handles 404s.
+            {{actor, ActorId}, Req, State};
+        true ->
+            {authorized, Req, State};
+        false ->
+            {{halt, 403}, Req, State}
+    end.
```
stevendanna: I'm wondering if it makes more sense to implement this in oc_chef_wm_base so that future endpoints can use the same mechanism.  `auth_info` in this module could return `{{member_of, GroupName}, Req, State}` and this logic could be moved into the oc_chef_wm_base:forbidden (or some function it calls).

oc-chef-pedant/lib/pedant/platform.rb
```diff
@@ -195,18 +195,18 @@ def create_client(name, org = self.test_org)
       payload = { "name" => clientname }
 
       if server_api_version == 0
-        r = post(api_url('/clients'), org.validator, :payload => payload)
+        r = post(api_url('/clients', org), org.validator, :payload => payload)
```
jmink: Is it safe to change the requests for old api versions?  Does that mean we're changing the contract on people?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_keys.erl
tylercloke: Long-term that probably makes the most sense, but this PR is already getting pretty big. What do you think?

tylercloke: Open to doing it, but makes sense to break up the work to me.

oc-chef-pedant/lib/pedant/platform.rb
tylercloke: This is just for testing, so no, but also I'm fixing a bug. This code worked in the default case (where you didn't pass in an org) that was mostly being used but was broken if you passed an org in. Just fixing a bug.

stevendanna: Interesting, what API requests were these being made to without passing an org? You commit message mentions the default_org.  Do you mean the default org as configured for the chef-server under test?  Does that imply these tests previously polluted a user's organization if run on a live cluster?

stevendanna:
I'm :+1: on the existing changes; however, per our chat discussion this needs a mover migration for existing orgs/users.

oc-chef-pedant/lib/pedant/platform.rb
tylercloke: If you wanna do a test with multiple orgs, you have to make one :) This code tried to supported that, but was wrong.

tylercloke:
Right...the org user keys endpoints will need lua route updates too...

oc-chef-pedant/spec/api/keys/client_keys_spec.rb
```diff
@@ -674,35 +676,34 @@ def rand_id
...
+        end
 
+        shared_examples_for 'successful client key get' do
           context "when GET /organizations/org/clients/client/keys is called (list keys)" do
             it "all keys should be listed with correct expiry indicators" do
-              list_client_keys(org_name, org_client_name, superuser).should look_like(
-                status: 200,
-                body_exact: [
+              list_client_keys(org_name, org_client_name, current_requestor).should look_like({
+                :status => 200,
```
marcparadise: Why `:` to `=>`?

oc-chef-pedant/spec/shared_context/keys_context.rb
```diff
@@ -0,0 +1,73 @@
...
+def rand_id
+  rand(10**7...10**8).to_s
+end
+
+RSpec.shared_context 'when the current_requestor is the superuser' do
+  let(:current_requestor) { superuser }
+end
+
+RSpec.shared_context 'when the current_requestor is a user in the main org' do
+  let(:test_user_name) { "pedant_test_user_#{rand_id}" }
```
marcparadise: Let's have the name indicate the context `"pedant_test_user_keys_#{rand_id}"`. Makes it easier to identify the source when we sometimes  find that we're not cleaning up properly.

src/chef-mover/src/mover_actor_keys_access_group_callback.erl
```diff
@@ -0,0 +1,158 @@
...
+         migration_action/2,
+         next_object/0,
+         error_halts_migration/0,
+         reconfigure_object/2,
+         needs_account_dets/0
+        ]).
+
+-record(org, {name, id, authz_id}).
+-record(group, {name, id, authz_id}).
+-define(GLOBAL_PLACEHOLDER_ORG_ID, <<"00000000000000000000000000000000">>).
```
marcparadise: I am pretty sure we have this available in an hrl we can pull in from somewhere?

src/chef-mover/src/mover_actor_keys_access_group_callback.erl
```diff
@@ -0,0 +1,158 @@
...
+	ok ->
+	    process_group(ClientGroup, AccessGroupAuthzId, OrgName)
+    end.
+
+process_group(Group, AccessGroupAuthzId, OrgName) ->
+    case add_group_to_group(Group, AccessGroupAuthzId) of
+        ok ->
+            ok;
+        {error, failure_creating_global_group} ->
+            lager:info("Failed to create keys access group for Organization ~p so cannot be migrated", [OrgName]),
```
marcparadise: warn instead of info, perhaps, for the cases where we can't migrate anything.

src/oc_erchef/apps/chef_objects/src/chef_key_base.erl
```diff
@@ -230,4 +230,7 @@ key_owner_type(Owner) when Owner =:= client;
     client_key;
 key_owner_type(Owner) when Owner =:= user;
                            Owner =:= chef_user ->
-    user_key.
+    user_key;
+key_owner_type(Owner) when Owner =:= org_user;
+                           Owner =:= chef_org_user ->
```
marcparadise: Where does `chef_org_user` come from?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_db.erl
```diff
@@ -158,10 +159,16 @@ fetch_global_group(#oc_chef_authz_context{} = Ctx, GroupName) ->
     fetch_group(Ctx, ?GLOBAL_PLACEHOLDER_ORG_ID, GroupName).
 
 make_read_access_group_name(OrgName) ->
-  lists:flatten(io_lib:format("~s_read_access_group", [OrgName])).
+    make_org_prefixed_group_name(OrgName, "read_access_group").
 
 make_global_admins_group_name(OrgName) ->
```
marcparadise: Should this still be a thing?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_organization.erl
```diff
@@ -127,10 +128,17 @@ from_json(Req, #base_state{resource_state = #organization_state{
 %%
 delete_resource(Req, #base_state{chef_db_context = DbContext,
                                  requestor_id = RequestorId,
+                                 organization_name = OrgName,
                                  resource_state = #organization_state{
                                                      oc_chef_organization = Organization}
                                 } = State) ->
+    %% TODO: Convert to delete_global_group.
```
marcparadise: TODO to do.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_organization.erl
```diff
@@ -163,5 +171,20 @@ delete_read_access_group(DbContext, AuthzContext, OrgName, RequestorId) ->
...
 
+%% TODO: Factor similar methods into each other.
+delete_global_group(GroupName, OrgName, #base_state{chef_db_context = DbContext,
+                                                    chef_authz_context = AuthzContext,
+                                                    requestor_id = RequestorId} = _State) ->
+    delete_global_group(GroupName, DbContext, AuthzContext, OrgName, RequestorId).
+
+delete_global_group(GroupName, DbContext, AuthzContext, OrgName, RequestorId) ->
+    case oc_chef_authz_db:fetch_global_group(AuthzContext, GroupName) of
+        #oc_chef_group{} = ReadAccessGroup ->
```
marcparadise: s/b `GlobalGroup` and not `ReadAccessGroup` in keeping with the function.

marcparadise:
On the whole :+1: with minor comments. 

The only outstanding question (currently under discussion) is whether this would be better as an org-owned group instead of a global group, since the constraints that led us to creating global_read (global_admins) don't necessarily apply here. I don't think it would be a major change to the PR to switch it over, but we're in process of talking this through.

marcparadise:
After further discussion, we’ve decided to make the group a normal org group, with permissions to prevent deletion.  This allows membership to be edited as needed in situations where the customer has concerns about exposing public key and expiration data to clients.

src/oc_erchef/apps/chef_objects/src/chef_key_base.erl
tylercloke: Probably nowhere. Just being consistent with the above.

src/chef-mover/src/mover_actor_keys_access_group_callback.erl
tylercloke: Following a pattern from past migrations. I'm fine with warn though.

oc-chef-pedant/spec/shared_context/keys_context.rb
tylercloke: Which name?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_db.erl
tylercloke: Probably not but not in the scope of this PR to clean up imo.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_org_creator.erl
```diff
@@ -87,6 +88,7 @@
            {add_acl,
             [mk_tl(container, ?CONTAINERS), mk_tl(group, [admins, clients, users]), {organization}],
             ?ALL_PERMS, [{group, admins}]},
+           {add_acl, [{group, public_key_read_access}], [read, update], [{group, admins}]},
```
manderson26: These changes look good to me. I prefer read, update over adding read, grant. There probably would be no harm in adding others to the read list, but I'm fine with where it is.

tylercloke: Sweet, I think read, update is the right thing too.

manderson26:
:+1: with changes to make it org owned.

PR #770, opened 2016-03-07
[chef-server-ctl] Fix several bugs in chef-server-ctl backup

ryancragun:
* Bump chef_backup dependency
* Allow passing pg_options to backup/restore database operations
* Add early exit if rsync is not installed
* Add early exit if commands not run as root
* Use a temporary working directory when executing commands

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/backup.rb
+++ b/omnibus/files/private-chef-ctl-commands/backup.rb
@@ -22,8 +22,12 @@ add_command_under_category 'backup', 'general', 'Backup the Chef Server', 2 do
       options.pg_options = pg_options
     end
 
-    opts.on('-c', '--config-only', 'Backup only the config on the frontends. No data will be backedup.') do
+    opts.on('-c', '--config-only', 'Only backup configuration, no data') do
       options.config_only = true
+    end
+
+    opts.on('-t', '--timeout [string]', 'Set the Maximum amount of time to wait for shell commands') do |shell_out_timeout|
+      options.shell_out_timeout = shell_out_timeout
     end
 
     opts.on('-h', '--help', 'Show this message') do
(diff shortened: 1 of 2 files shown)


stevendanna:
:+1: This looks reasonable.  Mind pulling in the change from https://github.com/chef/chef-server/pull/736 as well?

coderanger:
Should probably hold off until https://github.com/chef/chef_backup/issues/19 is actually fixed correctly.

stevendanna:
@coderanger Why would that bug block these improvements?

coderanger:
Because the supposed fix for that ticket both didn't fix that issue and introduced a new timeout incorrectly.

coderanger:
If you're going to do roll-up PRs, live by the sword die by the sword ;-)

stevendanna:
I understand your point that chef/chef_backup#19 isn't fix by ryan's changes.  But I don't see how moving to this version makes us worse off than before.  Even in the case of the statement_timeout, it would appear from your comment that the limiting reagent will still be the shellout timeout anyways.

coderanger:
Fair, it's not actually worse, will still time out at the 10m mark :)

stevendanna:
:+1: here

omnibus/files/private-chef-ctl-commands/backup.rb
```diff
@@ -19,6 +18,14 @@
       options.agree_to_go_offline = true
     end
 
+    opts.on('--pg-options [string]', 'Additional options to pass to postgress during backups') do |pg_options|
```
tylercloke: Nitpick, but should probably end with a `.` consistently.

tylercloke:
:+1: One nitpick but it looks good!

PR #774, opened 2016-03-10
Bump bundler install for chef-zero Travis to 1.10.6.

tylercloke:
Fixes broken master.


marcparadise:
:+1:

ksubrama:
:+1:

ksubrama:
Just FYI - you probably want 1.11.2 or later at some point in the future because it includes dependency resolution fixes when circular deps are involved (like chef needing ohai needing chef for chef-config).

PR #777, opened 2016-03-14
Add some more testing related info to README

ksubrama:


jmink:
:+1:

tylercloke:
:+1: Works for now. We should probably just make bifrost `make pedant`smarter long term though.

PR #781, opened 2016-03-15
[oc_erchef] Make the _status endpoints health check timeout configurable.

andy-dufour:
http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/98/


stevendanna:
We typically avoid putting `ChangeLog-Entry:`  in the first line and rather do it in the body of the commit message.  However, we aren't using omnibus's change log generator anymore so you can omit it entirely.

andy-dufour:
Gone from my commit, and the PR title!

andy-dufour:
I also tested this with ec-metal and some load. When the timeout is lower, we get _status failures but no failed client requests. I'm not sure **exactly** where the ceiling is on this, probably not much more than a couple thousand ms, but making it configurable should help high load customers reduce false failures from the _status endpoint. I think the default as configured was good for most Chef server users.

PR #794, opened 2016-03-31
Modernize Add-On Install

schisamo:
### Changes: 

* Leverage `Mixlib::Install` to resolve platform-appropriate packages.chef.io download URL and checksum for add-on.
* Fetch add-on packages using `remote_file` as opposed to APT/YUM. This greatly simplifies remote add-on install and ensures the install process doesn’t leak into system-wide config (e.g. no APT and YUM sources are created).
* Use a `Chef::Config[:file_cache_path]` that isn’t part of the install tree - All files in the install tree should be fully owned by the install process not the configuration process.
* Update to the latest version of the `omnibus` cookbook

This work is in service of [JEX-79] and fully removes Chef Server's dependency on packagecloud.io.

### Testing:

- [x] Wilson ad-hoc build: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/118/downstreambuildview/
- [x] Remote Install of each add-on: https://gist.github.com/schisamo/b8c8f1609fcbb0f30b33b2eacbab2e2a
- [x] Local Install of an add-on: https://gist.github.com/schisamo/94cdec61d8c1d39376bd3312585876f9

### Associated Doc Updates:

https://delivery.chef.co/e/chef/#/organizations/CIA/projects/chef-web-docs/changes/e6c82738-bee2-4a01-8a6b-e0365925472b

/cc @chef/engineering-services @chef/chef-server-maintainers @mmzyk


schisamo:
Test failures are un-related to any of my changes.

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -171,14 +171,20 @@ def ldap_authentication_enabled?
...
-      File.exists? "/etc/yum.repos.d/#{name}.repo"
     when 'debian'
-      File.exists? "/etc/apt/sources.list.d/#{name}.list"
+      'deb'
+    when 'rhel', 'suse'
+      'rpm'
     else
-     false
+      # TODO: probably don't actually want to fail out?
+      raise "I don't know how to install addons for platform family: #{node['platform_family']}"
```
tylercloke: Maybe we should provide a flag to this command to allow the user to override and specify which repo type they want, and if they hit this, we could tell them about the flag?

schisamo: @tylercloke Perhaps, this was the functionality/message from before so I didn't want to gut too much.

tylercloke:
:+1: Nice PR! Just had one comment.

stevendanna:
:+1: I think this should be a custom resource, but that is a job for another day.

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
tylercloke: Maybe I'll make a card.

PR #795, opened 2016-03-31
Update RELEASE_PROCESS.md to reflect the new announcement process

mmzyk:
Releases should now be announced on discourse.chef.io and no longer on
the blog.

### DIFF
--- a/RELEASE_PROCESS.md
+++ b/RELEASE_PROCESS.md
@@ -22,8 +22,8 @@ In order to release, you will need the following accounts/permissions:
 - Local checkouts of the chef-server and chef-web-downloads repositories
 - Push access to the chef-server github repository
 - Chef Software, Inc Slack account
-- blog.chef.io wordpress account
 - VPN account for Chef Software, Inc.
+- Account on [https://discourse.chef.io](https://discourse.chef.io) using your Chef email address
 - Login for wilson.ci.opscode.us (This is linked to your github
 account.)
 - Access to artifactory.chef.co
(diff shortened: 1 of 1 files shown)


RELEASE_PROCESS.md
```diff
@@ -132,10 +127,15 @@ based on the branch you just created.
...
-the release with a link to the blog post.
+- [ ] Write and then publish a Discourse post on https://discourse.chef.io
+  once the release is live. This post should contain a link to the downloads
+  page ([https://downloads.chef.io](https://downloads.chef.io)) and its contents
+  should be based on the information that was added to the RELEASE_NOTES.md file
+  in an earlier step. *The post should  be published to the Chef Release
+  Announcements category on https://discourse.chef.io/. If it is a security
+  release, it should also be published to the Chef Security Announcements
+  category.* Full details on the policy of making release announcements on
+  Discourse can be found on the wiki: [https://chefio.atlassian.net/wiki/display/ENG/Release+Announcements+and+Security+Alerts](https://chefio.atlassian.net/wiki/display/ENG/Release+Announcements+and+Security+Alerts)
```
tylercloke: Should we really be sticking our private wiki in this open source facing file? Not saying we shouldn't, just curious.

tylercloke:
:+1: Other than comment.

RELEASE_PROCESS.md
btm: If we leave it, I'd say like `Chef internal wiki:` so there's no question you're not supposed to have a login.

RELEASE_PROCESS.md
```diff
@@ -22,8 +22,8 @@ In order to release, you will need the following accounts/permissions:
 - Local checkouts of the chef-server and chef-web-downloads repositories
 - Push access to the chef-server github repository
 - Chef Software, Inc Slack account
-- blog.chef.io wordpress account
 - VPN account for Chef Software, Inc.
+- Account on [https://discourse.chef.io/](https://discourse.chef.io/) using your Chef email address
 - Login for wilson.ci.opscode.us (This is linked to your github
```
btm: http://wilson.ci.chef.co/

is this really github? I thought it as AD.

btm:
:+1: other than tyler's comment.

RELEASE_PROCESS.md
mmzyk: I figured it was locked down and not a big deal; but I've updated to remove it to be on the safer side.

RELEASE_PROCESS.md
mmzyk: Yeah, it's github. It uses the github oauth system. Eng Services should have full details.

sdelano:
Awesome. Thanks for the updates here.

metadave:
perhaps a notification to the support folks would be good as well?

PR #797, opened 2016-04-03
Work to support chef-server on IBM POWER platforms

edolnx:
These are all the chef-server specific changes to support Linux (RHEL and Ubuntu) on IBM POWER8. This also depends on chef/omnibus-software#642 which updates software components, adds the IBM JRE, and works around missing JITs on the POWER8 architecture.

/cc @chef/chef-server-maintainers


src/chef-mover/rebar.config
```diff
@@ -122,7 +122,6 @@
             ssh,
             uuid,
             debugger,
-            hipe,
```
stevendanna: Note to reviewers: I *believe* this was only here because during the "Big Erlang Build Refactor" we hit an issue where rebar was producing wildly different build artifacts so we took a hammer to the problem and listed every single app that was in the old artifact but not in the new artifact.  Since we don't ship native builds, I don't know of a reason we would need this in the release.

manderson26:
:+1: Looks reasonable to me. It might be nice to rebase/squash before merge.

stevendanna:
👍

PR #798, opened 2016-04-06
Add support for service credentials rotation

ryancragun:
This PR add's support for Chef Server service credential rotation.  It add's a new library gem called Veil that handles the loading, creating, hashing, rotating, and saving of the Chef Server secrets. It works by creating a shared secret and salt for the desired hashing algorithm (PBKDF2 or BCrypt) and it creates credentials for each service by using the hashing algorithm with the shared secret, shared salt, service name, credential name and version to create an encrypted value.

Prior to Veil, Chef Server secrets we're generated on the bootstrap backend node and we're long lived for the duration of the Chef Server install. The secrets generated were completely random so there was no way to add a new service or credential without copying the secrets file between each Chef Server in the cluster during upgrades. By requiring shared secrets between Chef Server nodes the new system can deterministically generate new credential versions or new credentials on during upgrades without having to re-copy the secrets file between nodes. If you want to rotate a credential you can run the rotate command on each Chef Server and it will generate same set of new credentials, or you can rotate on one Server and copy the secrets file between nodes and reconfigure.

It is also possible to rotate the shared secrets.  When the shared secrets are rotated  it will generate a new shared key and salt and rotate all of the service credentials with the new shared keys.  In this scenario you will need to copy the secrets file between nodes and reconfigure.

It's backwards compatible with the old system in that it writes the secrets and metadata to `/etc/opscode/private-chef-secrets.json` with the same general format.  It does add an additional key called 'veil' which tracks versions, length, the shared secrets, and other credential metadata. Other than the veil key, any existing parsing of that file should be the same.

If you use chef to manage the Chef Server's private-chef-secrets it is recommended that you update your template to include the veil metadata. If you don't update the template `chef-server-ctl reconfigure` will still work as expected though it will write the file with the additional metadata.

## Rotating All Service Credentials

Rotating all over the service credentials will iterate over all credentials, bump their version, and rehash their password with the new version.

  ```shell
  $: chef-server-ctl rotate-all-credentials
  ...
  All credentials have been rotated!
  Run 'chef-server-ctl rotate-all-credentials' on each Chef Server
  ```

## Individual Credential Rotation

Rotate the service credentials for a single Chef Server subsystem

  ```shell
  $: chef-server-ctl rotate-credentials $SERVICE_NAME
  ...
  All credentials have been rotated!
  Run 'chef-server-ctl rotate-credentials SERVICE_NAME' on each Chef Server
  ```

## Shared Secret Rotation

Regenerating the shared secrets regenerates the shared services, all credentials. In this scenario the user needs to to redistribute /etc/opscode/private-chef-secrets.json:

  ```shell
  $: chef-server-ctl rotate-shared-secrets
  ...
  The shared secrets and all service credentials have been rotated!
  Please copy /etc/opscode/private-chef-secrets.json to each Chef Server and
  run 'chef-server-ctl reconfigure'
  ```


charlesjohnson:
Docs?

src/veil/lib/veil/hasher/bcrypt.rb
```diff
@@ -0,0 +1,58 @@
...
+          if ::BCrypt::Engine.valid_secret?(opts[:secret]) && ::BCrypt::Engine.valid_salt?(opts[:salt])
+            @secret = opts.delete(:secret)
+            @salt = opts.delete(:salt)
+          elsif ::BCrypt::Engine.valid_secret?(opts[:secret])
+            raise Veil::InvalidSalt.new("#{opts[:salt]} is not valid salt")
+          else
+            raise Veil::InvalidSecret.new("#{opts[:secret]} is not valid secret")
+          end
+        else
+          # BCrypt is capped at 55 character input:
```
stevendanna: Is this comment correct?  `SecureRanom.hex(53)` will return 106 characters since the 53 refers to bytes and each byte will be represented by two hexadecimal digits in the returned string.

src/veil/lib/veil/hasher/bcrypt.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      # Hash data with the stored secret and salt
+      #
+      # @param [String] data
+      #   The service name and version to be encrypted with the shared key
+      #
+      # @return [String] SHA512 hex digest of hashed data
+      def encrypt(data)
+        # Append the secret to the data because BCrypt will truncate our data to 55 characters.
+        hex_digest(::BCrypt::Engine.hash_secret([data, secret].join, salt))
```
manderson26: Rather than risk diluting the secret by appending and possibly having most or all of it truncated off, why not take a sha512 of the full data, and take whatever appropriate prefix of that; here 55 chars at 8 bits a char would grab the first 220 bits.

src/veil/lib/veil/hasher/bcrypt.rb
```diff
@@ -0,0 +1,58 @@
...
+            raise Veil::InvalidSalt.new("#{opts[:salt]} is not valid salt")
+          else
+            raise Veil::InvalidSecret.new("#{opts[:secret]} is not valid secret")
+          end
+        else
+          # BCrypt is capped at 55 character input:
+          #   secret = 53
+          #   version >= 1
+          #   service >= 1
+          @secret = SecureRandom.hex(53)
```
manderson26: Don't reduce the security of the secret just to accommodate bcrypt; use a hash as described below.

jaymzh:
I like this! (in theory :)

A few questions:

* where does the shared-secret live?
* the assumption that "you just run the command to rotate the passwords" is flawed. People shouldn't man-handle their servers... so chef will run on my FEs and BEs over time... so what happens if an FE is rotated and then a BE, and then another FE.... rotation should ideally add an additional password, and then allow the old password to be removed.

ryancragun:
@jaymzh I added a README that will probably answer most of your questions: https://github.com/chef/chef-server/blob/ryan/veil/omnibus/files/veil/README.md

> where does the shared-secret live?
The shared secrets live in `/etc/opscode/private-chef-secrets.json`

> the assumption that "you just run the command to rotate the passwords" is flawed. People shouldn't man-handle their servers... so chef will run on my FEs and BEs over time... so what happens if an FE is rotated and then a BE, and then another FE.... rotation should ideally add an additional password, and then allow the old password to be removed.

Secrets rotation is definitely an admin style task so the front-end of this is `chef-server-ctl` command.  I love the idea that services should support multiple credentials but in many cases that would require quite a bit of additional logic in each service and is probably outside of the scope of this.

ryancragun:
@charlesjohnson docs https://github.com/chef/chef-server/blob/ryan/veil/omnibus/files/veil/README.md

stevendanna:
Pending our discussion of HMAC vs plain SHA512 digest, 👍 from me.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
```diff
@@ -75,10 +75,11 @@ def run!
   # backend components, allow data bootstrapping to be bypassed when
   # no chef-server-running.json is present but a secrets file is present.
   def bypass_bootstrap?
-    first_run? && secrets_exists?
+    first_run? && secrets_exists? && PrivateChef['postgresql']['external']
```
marcparadise: I'm ok with this change - I gave it as a workaround in a customer issue  -  but with reservations. The `bypass_bootstrap` exists as a best a attempt to determine when someone is using a completely external back-end, and is running multiple chef servers in 'standalone` mode connected to the same backend. (So effectively, they're just front-ends, without adding a new topo).   So even with this change, it could come up with the wrong answer since it's possible to have a tiered or HA setup with external postgresql. 

A better fix - and I think I saw something like this underway in another PR - is an explicit flag saying that's how we're running chef-server, and using instead of the external flag. .

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
```diff
@@ -44,7 +44,7 @@ def first_run?
   end
 
   def secrets_exists?
-    PrivateChef.existing_secrets.length > 0
+    File.exist?("/etc/opscode/private-chef-secrets.json")
```
marcparadise: The reason we checked for non-zero length was that we had a situation  where an empty file existed, which stopped the installation from proceeding because of the preflight check.  Why the change to only `exist`?

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -435,60 +426,74 @@ def gen_frontend_default
...
+      PrivateChef["nginx"]["server_name"] ||= PrivateChef["api_fqdn"]
       PrivateChef["nginx"]["url"] ||= "https://#{PrivateChef['api_fqdn']}"
     end
 
     def gen_secrets_default(node_name)
-      consume_existing_secrets
-      # Transition from erchef's sql_user/password etc living under 'postgresql' in older versions,
-      # to 'opscode_erchef' in newer versions.
-      if PrivateChef['postgresql'].has_key? 'sql_password'
-        PrivateChef['opscode_erchef']['sql_password'] ||= PrivateChef['postgresql']['sql_password']
```
marcparadise: This will still be a necessary thing if upgrading from older versions that kept the password in a different place, and I don't see that it's handled below?

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -435,60 +426,74 @@ def gen_frontend_default
...
+      credentials =
+        if File.exist?(secrets_json)
+          Veil::CredentialCollection::ChefSecretsFile.from_file(secrets_json)
+        elsif PrivateChef["topology"] == "ha" && !PrivateChef["servers"][node_name]["bootstrap"]
+          Chef::Log.fatal("In an H/A topology the secrets must be created on the bootstrap node. "\
+                          "Please copy the contents of /etc/opscode/ from your bootstrap Server " \
+                          "to complete the setup")
+          exit(44)
+        else
+          Veil::CredentialCollection::ChefSecretsFile.new(path: secrets_json)
```
marcparadise: This gets a little odd with the external-backends + standalone-as-frontends topo. We have not addressed the upgrade scenario for that case yet, but in the initial setup the first standalone node you bring up behaves as bootstrapping, while subsequent nodes will be expected to have secrets copied over. 

Worth mentioning here because the topology looks like 'standalone' in that case.

marcparadise:
For future updates, please separate out large-scale style changes into a separate PR. This helps reviewers by ensuring that we don't need to filter to determine which change sets are relevant to the functionality.

marcparadise:
Generally good, but I did have some concerns raised inline.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
ryancragun: I definitely think that's a much better way to fix this, but I had to add this check otherwise copying the secrets file between nodes on existing HA nodes would fail.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
ryancragun: I changed it initially because `existing_secrets` no longer exists with this implementation. If we think that having an invalid file is reason to stop then I can definitely load the file and ensure that the secrets count is non-zero.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
ryancragun: Should be resolved in https://github.com/chef/chef-server/pull/798/commits/f4121c56db68c858e06d7e32d5ffa73d2ccbbe4f

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
ryancragun: I think in that case we should probably either have a different topology for those scenarios or have helper methods to determine that we're in that sort of a state.  I had no idea 'standalone-as-frontend-without-being-in-frontend-topology' was a thing until now.

ryancragun:
I believe all feedback is addressed.  I'm unsure of the travis failure but it looks like it's not related to these code changes.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
marcparadise: Agreed, there's no easy way to tell if that's the setup now. I do think @stevendanna has a PR in flight (or already merged?) that adds a flag to indicate this is the state.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
marcparadise: Alright, it looks like that attribute has already landed. If you could change this to `PrivateChef['use_chef_backend']` instead of postgresql/external, that will do it.

marcparadise:
Thanks @ryancragun. I added one comment - there's a better flag available than external postgres, so let's change to that. 

Beyond that,, contingent on a run through CI and chef-acceptance ( package upgrade testing through delivery)  ,  :thumbsup:. I can help out with that tomorrow (Fri) morning. I'm also checking with the spool team to see if anyone there is available to help sooner if you need.

ryancragun:
http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/149/downstreambuildview/

manderson26:
Generally good. It might be nice to add a version field to the veil serialization format in case we have to make a breaking change and upgrade later.

jaymzh:
OK coming back to this, sorry I was traveling a lot.

If you store it in `/etc/opscode/private-chef-secrets.json` doesn't that provide an odd problem... I still now need to manage that file, so how does this change anything except add another ste.

* Chef templatizes `private-chef-secrets.json`, runs `chef-server-ctl` which gets updated with a bunch more secrets... which means the template doesn't match so then I write it again and then...

OK, so that doesn't work how about this:

* I put it it in private-chef-sercrets.json, run `chef-server-ctl` manually, then take the new `private-chef-sercrets.rb` and shove THAT in my cookbook.

Except I thought the whole point of this was to prevent that, so now I have to do  more manual effort and I'm still managing that file. What am I missing here?

ryancragun:
Many thanks for the feedback @marcparadise, @manderson26, and @stevendanna.  I've rebased and run the build through the 10 acceptance pipelines using the qa-chef-server-cluster repo in Delivery and they have all passed.  🎉

marcparadise:
:+1: :shipit:

PR #803, opened 2016-04-11
Preserve extra secrets in private-chef-secrets.json

andy-dufour:
This hash was previously built statically, but when a customer templates private-chef-secrets and adds an extra key (or has an obsolete one that we've removed) they may not want us to remove it from the JSON file otherwise, we could cause a fight between chef-server-ctl reconfigure and their automation.

Instead lets warn them that they should remove it from their automation, because it does nothing, and preserve it.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -631,7 +631,7 @@ EOF
     end
 
     def existing_secrets
-      @existing_secrets ||= if File.exists?("/etc/opscode/private-chef-secrets.json")
+      existing_secrets ||= if File.exists?("/etc/opscode/private-chef-secrets.json")
                               Chef::JSONCompat.from_json(File.read("/etc/opscode/private-chef-secrets.json"))
                             else
                               {}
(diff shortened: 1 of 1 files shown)


andy-dufour:
http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/131

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -642,7 +642,7 @@ def consume_existing_secrets
       existing_secrets.each do |k, v|
         v.each do |pk, p|
           if not PrivateChef[k]
-            Chef::Log.info("Ignoring unused secret for #{k}.")
+            Chef::Log.warn("The secret for #{k} doesn't appear to be valid, you should remove it.")
```
stevendanna: I think we want some combination of these two messages.  "doesn't appear to be valid" doesn't tell the user much about what is wrong.  I would say "is not used by the Chef Server, ..."

stevendanna:
Heads up: other work going on in this area: https://github.com/chef/chef-server/pull/798

stevendanna:
Thou shall not pass without tests that pass: https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/private_chef_spec.rb

:D Let me know if you have trouble adding a test for this.

andy-dufour:
Closing in favor of #798, for the time being. Will re-open with comments if #798 doesn't solve this use-case completely.

PR #804, opened 2016-04-12
Fix error in error handling for server admins permission migration

paulmooring:
This was failing on our migrations because it was trying to call nested `string:concat` calls on a combination of strings and binaries.

[Build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/135/downstreambuildview/)


src/chef-mover/src/mover_server_admins_existing_users_read_permissions_callback.erl
```diff
@@ -44,7 +44,7 @@ migration_action(UserRecord, _AcctInfo) ->
 add_permission_to_existing_user_for_server_admins(BifrostSuperuserId, Username, ServerAdminsAuthzId, UserAuthzid, Permission) ->
     case mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserId, group,  ServerAdminsAuthzId, actor, UserAuthzid, Permission) of
 	{error, Error} ->
-	    lager:error("Failed to update " ++ Permission ++ " permission for user " ++ Username ++ " with error:"),
+	    lager:error("Failed to update " ++ binary_to_list(Permission) ++ " permission for user " ++ binary_to_list(Username) ++ " with error:"),
```
stevendanna: Perhaps rather than worrying about the conversion, something more like this would be useful:

```
lager:error("Failed to update ~p permission for user ~p with error: ~p", [Permission, Username, Error]),
```

However, I'm still +1 on just merging this for now if ya'll have already tested it.

tylercloke: Yeah, definitely a better approach, but since this is tested, let's just use it.

stevendanna:
What's the story on this?  Seems good to merge.

paulmooring:
If it helps, we tested this is a one-off build for a deploy and it works

src/chef-mover/src/mover_server_admins_existing_users_read_permissions_callback.erl
marcparadise: :+1:

PR #808, opened 2016-04-14
Use HAProxy to route Postgresql and ElasticSearch connections

stevendanna:
This provides two new configuration options:

   use_chef_backend
   chef_backend_members

When use_chef_backend is true, HAProxy will be enabled with a
configuration that will route requests to Postgresql and Elasticsearch
to the current Chef Backend leader.

### DIFF
--- a/omnibus/config/software/haproxy.rb
+++ b/omnibus/config/software/haproxy.rb
@@ -1,0 +1,71 @@
+#
+# Copyright 2016 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+name "haproxy"
+default_version "1.6.4"
+
+dependency "zlib"
+dependency "pcre"
+dependency "openssl"
+
+license "GPLv2"
+license_file "LICENSE"
+
+# HTTPS is available but certificate validation fails on OS X
+source url: "http://www.haproxy.org/download/1.6/src/haproxy-#{version}.tar.gz",
+       sha256: "e5fa3c604f1fe9ecb6974ccda6705c105ebee14b3a913069fb08f00e860cd230"
+
+relative_path "haproxy-#{version}"
+
+build do
+  env = with_standard_compiler_flags(with_embedded_path)
+  #
+  # Many of these are the same environment variables that debian sets
+  # PREFIX and TARGET are mandatory to get it building under omnibus.
+  #
+  build_options = {
+   "PREFIX" => "#{install_dir}/embedded",
+   # Use libpcre for regex, libpcre > 8.32 required
+   # for JIT
+   "USE_PCRE" => "1",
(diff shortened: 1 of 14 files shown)


stevendanna:
This is a WIP PR to facilitate discussion.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_checks.rb
```diff
@@ -126,7 +126,12 @@ def initialize(node)
   def run!
     begin
       BootstrapPreflightValidator.new(node).run!
-      PostgresqlPreflightValidator.new(node).run!
+      # When Chef Backend is configured, this is too early to verify
+      # postgresql accessibility since we need to configure HAProxy
+      # first
+      if ! PrivateChef['use_chef_backend']
```
marcparadise: With this flag, we can at some point  get rid of the hacky "derive whether we really care about bootstrap" logic that's used to allow a front-end to do bootstrapping.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/haproxy.cfg.erb
```diff
@@ -0,0 +1,26 @@
...
+frontend elasticsearch
+         bind <%= @listen %>:<%= @local_elasticsearch_port %>
+         mode tcp
+         default_backend chef_backend_elasticsearch
+
+backend chef_backend_postgresql
+        mode tcp
+        option httpchk GET /leader HTTP/1.1\r\nHost:localhost:<%= @leaderl_healthcheck_port %>\r\n\r\n
+        <% @chef_backend_members.each do |name, ip| -%>
+        server <%= name %> <%= ip %>:<%= @remote_postgresql_port %> check port <%= @leaderl_healthcheck_port %> rise 1 fall 1
```
dmccown: I think we can specify defaults here up above for all the servers. Like:
```
default-server inter 3s fall 3 rise 2
server name ip:port check port
server name ip:port check port
...
```
https://www.haproxy.com/doc/aloha/7.0/haproxy/healthchecks.html

Reading the docs a little more there's actually built in support for postgres as well:

```
backend be_pgsql
 [...]
 option pgsql-check
 server srv1 10.0.0.1:5432 check
 server srv2 10.0.0.2:5432 check
```

stevendanna: We are using an HTTP check here because the /leader endpoint knows about the leader/follower state.  (postgresql could be erroneously up as a leader while leaderl is in the process of taking it down.  The /leader endpoint will return the correct thing in that case while a postgresql-based check may not).

stevendanna:
@dmccown Updated the PR to use the default-server config directive.

omnibus/config/software/haproxy.rb
```diff
@@ -0,0 +1,71 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+name "haproxy"
+default_version "1.6.4"
+
+dependency "zlib"
```
dmccown: Are all of the dependencies run time dependencies? Is it worth it to prune the non-runtime deps at this time? I'm OK if there's another task to clean this up later.

stevendanna: All of these are runtime deps since we don't do static linking, so zlib, pcre, openssl all have to be in the package.

dmccown:
Other than my comment on dependencies and the comment you cleaned up on your last PR nothing is really jumping out at me. So overall I'm 👍

PR #809, opened 2016-04-14
[omnibus] Add rb-readline to the build

stevendanna:
Omnibus-software no longer compiles ruby with readline support because
of various compatibility issues. Adding rb-readline ensures that irb
and other irb or pry-based REPLs have readline support.


stevendanna:
Resolved the build issue for this: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/140//downstreambuildview

PR #811, opened 2016-04-20
[expander] Set HOME in expander's runsv script

stevendanna:
Upstream omnibus-software no longer builds Ruby with readline. This
means that 'require "readline"' now picks up the rb-readline that oc_id
installs into our ruby environment. Bundler internally does a require on
readline during `bundle exec`. Unfortunately rb-readline fails if it is
required with HOME unset. This leads to expander failing to start.


schisamo:
:+1:

PR #817, opened 2016-04-21
Fix bug where requestor membership of public_key_read_access was not being properly tested for keys access.

tylercloke:
Access to the org scoped user and client keys read endpoints are controlled by membership on the public_key_read_access group (which defaults to the users and clients groups); however, a bug resulted in testing membership on public_key_read_access for the member being requested instead of the requestor.

This commit fixes that bug and adds regression tests.

[Passing build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/143/downstreambuildview/)

### DIFF
--- a/oc-chef-pedant/spec/api/keys/client_keys_spec.rb
+++ b/oc-chef-pedant/spec/api/keys/client_keys_spec.rb
@@ -774,6 +774,143 @@ describe "Client keys endpoint", :keys, :client_keys do
             get_client_key(org_name, org_client_name, current_requestor, 'key1').should look_like(:status => 401)
           end
         end
+
+        # public_key_read_access testing
+        shared_examples_for 'multiple actors READ access to the client keys endpoints depends on public_key_read_access membership' do
+          context 'when an actor is removed from the public_key_read_access group' do
+            before do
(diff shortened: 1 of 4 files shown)


oc-chef-pedant/spec/api/keys/client_keys_spec.rb
```diff
@@ -774,6 +774,143 @@ def rand_id
...
+              get_client_key(org_name, test_client_name_1, current_requestor, 'default').should look_like(:status => 403)
+              get_client_key(org_name, test_client_name_2, current_requestor, 'default').should look_like(:status => 403)
+            end
+
+            it 'the second actor can no longer list client keys, returning a 403', :authentication do
+              list_client_keys(org_name, test_client_name_1, other_requestor).should look_like(:status => 403)
+              list_client_keys(org_name, test_client_name_2, other_requestor).should look_like(:status => 403)
+            end
+
+            it 'the second actor an no longer get client keys, returning a 403', :authentication do
```
stevendanna: s/an/can/

stevendanna:
This change looks good to me 👍

PR #818, opened 2016-04-22
Bug fix: treat a successful PG conn and auth as a preflight success

adamleff:
Currently, if the PostgreSQL preflight check is able to connect to the
DB *and* authenticate (due to a pg_hba trust entry, for example) but is
unable to find the not-expected-to-be-there database name in question,
it treats the preflight connection check as a failure.

An example of such a condition would be configuring Chef Server and
Delivery on the same node and configuring Chef Server to use Delivery's
database. Delivery's pg_hba.conf trusts local connections, so a
`chef-server-ctl reconfigure` will fail the preflight checks because
it connects to the DB but fails to find the fake database.

This change treats a successful connection and auth as a success during
pre-flight and also adds tests for the `#connectivity_validation` method
which previously had none.


ryancragun:
🎯

PR #821, opened 2016-04-22
Dynamically generate the wait-for-rabbit script

adamleff:
The private-chef cookbook allows a user to change the rabbitmq nodename and data directories. However, the wait-for-rabbit script used in other processes' runit scripts does not account for this and uses hardcoded values. This prevents those services from starting properly if the user changes the rabbitmq nodename or data directory.

This change renders the wait-for-rabbit script as a template in the private-chef::rabbitmq recipe and removes it from the private-chef-scripts directory.


ryancragun:
👍

adamleff:
http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/146/downstreambuildview/

PR #826, opened 2016-04-28
Revert changes to sqitch plan files to avoid upgrade breakage

stevendanna:
Sqitch calculates hashes for each change based on the contents of the
plan file preceding that changes entry.  It stores that data in the
database to determine what has already been deployed. If you change the
plan file, you change the hashes, and nothing works.


stevendanna:
This reverts changes made here: 268a0c9967b3dfdb37e4b267b533312d1c95078f

PR #832, opened 2016-05-06
Ensure erchef is configured to use custom nginx ports

metavida:
This pull request is intended to fix Issue #50 & is based on [a workaround](https://github.com/chef/chef-server/issues/50#issuecomment-183827124) suggested by @wduncanfraser

This pull request should cover the case where a custom nginx ssl_port is specified,
and should also handle when nginx is configured to not use SSL.

Note: I'm not 100% sure what the `opscode-chef-mover.config.erb` template is used for, but I updated it to match the `oc_erchef.config.erb` changes since the two files were very similar.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
@@ -218,7 +218,7 @@
                   %% chef server can talk to bookshelf.
                   {s3_url, "http://<%= node['private_chef']['bookshelf']['listen'] %>:<%= node['private_chef']['bookshelf']['port'] %>"},
                   <% else -%>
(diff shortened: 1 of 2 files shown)


marcparadise:
I believe this is addressed by https://github.com/chef/chef-server/pull/833 and can be closed.

metavida:
@marcparadise Yup, looks like #833 might resolve what I was trying to fix. We'll give it a try once the next chef-server release comes out.

marcparadise:
Excellent, thanks for getting back to me!

dixon1234:
Not able to login if set custom  nginx ssl port number
nginx['ssl_port']=8444
bookshelf['vip_port'] = 8444

When i login in chef UI i get this error

500Smell something burning?Sorry we have a small fire in the kitchen.An unexpected error has occurred. Our staff are manning the fire extinguishers and have been alerted.

PR #833, opened 2016-05-09
Fixing pedant/bookshelf when nginx on non-standard port

adamleff:
This change addresses three primary issues:
 * bookshelf breaks when a custom SSL port is used for nginx
 * the bookshelf VIP port itself cannot be customized
 * pedant tests fail when a customer SSL port is used for nginx

When a user configures `nginx['ssl_port']` to be something other than 443, Pedant tests fail because the `nginx['url']` attribute assumes port 443 by default. That attribute is rendered in the `pedant_config.rb` file, so Pedant fails when you change the port from the standard 443. This change will correctly set `nginx['url']` with the ssl_port as well.

Additionally, the `s3_url` used in opscode-erchef for internal access to the bookshelf store assumes that the bookshelf vip is always listening on port 443. If you are using the bookshelf service and you have changed your config to have nginx listen on a port other than 443, cookbook-related tasks that use bookshelf will fail. This change introduces a new attribute (`vip_port`) that defaults to 443, and a helper method to properly render the s3_url in the erchef config with a port number if the user elects to change from the default port.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
@@ -1,0 +1,42 @@
+require_relative '../../libraries/helper.rb'
+
+describe OmnibusHelper do
+  describe '#bookshelf_s3_url' do
+    let(:node) do
+      {
+        'private_chef' => {
+          'nginx' => {
+            'x_forwarded_proto' => 'https'
+          },
+          'bookshelf' => {
+            'vip_port' => 8443
+          }
+        }
+      }
+    end
+
+    it 'returns a properly formatted URL' do
+      helper = described_class.new(node)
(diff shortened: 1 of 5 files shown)


adamleff:
http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/157/downstreambuildview/

jkeiser:
:+1:

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
```diff
@@ -0,0 +1,42 @@
+require_relative '../../libraries/helper.rb'
+
+describe OmnibusHelper do
```
marcparadise: Thank you for including tests!

marcparadise:
:+1: after a rewording of the commit message - this addresses two larger issues: 
1) can't use a custom ssl port without breaking bookshelf 
2) can't customize bookshelf port.

PR #835, opened 2016-05-11
[omnibus] bypass_bootstrap? should ensure both creds exist

stevendanna:
In addition to its use in the validator bypass_bootstrap? is called in
recipes/default.rb to determine whether or not we should run the
bootstrap cookbook.  The call to bypass_bootstrap? in recipes/default.rb
happens /after/ the private-chef-secrets.json file is written to disk.
Thus, this check was leading us to bypass the bootstrap even in cases
when this was the first node running reconfigure against an external
postgresql.

Additionally, given the current complexity of the bootstrap procedure, I
don't think we should actively use the `use_chef_backend` option for
anything but the "HAProxy routing enabled" bits until we ship the code
related to that feature.

Overall, I'm not thrilled with this solution. I've started to do some
code review of the bootstrap situation generally and will have a follow
up PR with some documentation on what I've found.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
@@ -50,7 +50,6 @@ class BootstrapPreflightValidator < PreflightValidator
       end
     end
 
-
     # TODO:
     # If the pivotal key file exists locally:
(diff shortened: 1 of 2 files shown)


stevendanna:
cc @marcparadise

marcparadise:
:+1: as a short-term fix. The sooner we can move towards explicit bootstrap vs recofigure, the better.

PR #838, opened 2016-05-11
Remove chef-sync from the known add on packages for the install command

mmzyk:
chef-sync is no longer being offered as a product to new users and should
be removed as an option to prevent unintended user confusion.

It has already been removed as an option on the doc site and from the downloads site.

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/install.rb
+++ b/omnibus/files/private-chef-ctl-commands/install.rb
@@ -3,7 +3,6 @@
 
 KNOWN_ADDONS = %w(
   chef-ha
-  chef-sync
   chef-manage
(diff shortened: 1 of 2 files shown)


marcparadise:
:+1:

omnibus/files/private-chef-ctl-commands/install.rb
```diff
@@ -3,7 +3,6 @@
 
 KNOWN_ADDONS = %w(
   chef-ha
-  chef-sync
```
stevendanna: There is a reference in attributes/default.rb that you might want to nuke too

mmzyk:
@stevendanna nuked.

PR #840, opened 2016-05-13
Add chef-server-ctl require-credential-rotation command

ryancragun:
# Requiring Chef Server credential rotation

As as Marketplace developer, I want a way to generate a Chef Server
image that requires credential rotation before starting services so that my Chef
Server does not start up insecurely.

## Current Problems

1. There's no way for us to force a credential rotation before the Chef Server's
  services start up.
1. There's no way to force a credential rotation before a `chef-server-ctl reconfigure`
1. There's no way to force a credential rotation before a `chef-server-ctl start`

## Solution

We solve the current problems by:

* Writing a `chef-server-ctl require-credential-rotation` command that:
  * Stops all Chef Server services
  * Disables all Chef Server services
  * Write a special prehook that will prevent service and reconfigure commands from running until rotation command has run.

OmnibusCtl prehook system: https://github.com/chef/omnibus-ctl/pull/43

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/chef-server-ctl
+++ b/omnibus/files/private-chef-ctl-commands/chef-server-ctl
@@ -216,6 +216,17 @@ EOM
         "/admin/ping?wt=json"
       end
     end
+
+    # Override reconfigure to skip license checking
+    def reconfigure(*args)
+      status = run_chef("#{base_path}/embedded/cookbooks/dna.json")
+      if status.success?
+        log "#{display_name} Reconfigured!"
+        exit! 0
+      else
+        exit! 1
+      end
+    end
   end
 end
 
--- a/omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
+++ b/omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
@@ -27,7 +27,7 @@ add_command_under_category "rotate-credentials", "credential-rotation", "Rotate 
     backup_file = backup_secrets_file
 
     log("Rotating #{service}'s credentials...", :notice)
-    credentials = Veil::CredentialCollection::ChefSecretsFile.from_file("/etc/opscode/private-chef-secrets.json")
+    credentials = Veil::CredentialCollection::ChefSecretsFile.from_file(secrets_file_path)
     credentials.rotate(service)
     credentials.save
   rescue => e
@@ -63,7 +63,7 @@ add_command_under_category "rotate-all-credentials", "credential-rotation", "Rot
     backup_file = backup_secrets_file
 
     log("Rotating all Chef Server service credentials...", :notice)
-    credentials = Veil::CredentialCollection::ChefSecretsFile.from_file("/etc/opscode/private-chef-secrets.json")
+    credentials = Veil::CredentialCollection::ChefSecretsFile.from_file(secrets_file_path)
     credentials.rotate_credentials
     credentials.save
   rescue => e
@@ -94,13 +94,12 @@ end
 add_command_under_category "rotate-shared-secrets", "credential-rotation", "Rotate the Chef Server shared secrets and all service credentials", 2 do
   ensure_configured!
 
-  secrets_file = "/etc/opscode/private-chef-secrets.json"
   backup_file = backup_secrets_file
 
   # Rotate and save the credentials
   begin
     log("Rotating shared credential secrets and service credentials...", :notice)
-    credentials = Veil::CredentialCollection::ChefSecretsFile.from_file(secrets_file)
+    credentials = Veil::CredentialCollection::ChefSecretsFile.from_file(secrets_file_path)
     credentials.rotate_hasher
     credentials.save
   rescue => e
@@ -117,7 +116,7 @@ add_command_under_category "rotate-shared-secrets", "credential-rotation", "Rota
     if status.success?
       remove_backup_file(backup_file)
       log("The shared secrets and all service credentials have been rotated!", :notice)
-      log("Please copy #{secrets_file} to each Chef Server and run 'chef-server-ctl reconfigure'", :notice)
+      log("Please copy #{secrets_file_path} to each Chef Server and run 'chef-server-ctl reconfigure'", :notice)
       exit(0)
     else
       log("Shared credential rotation failed", :error)
@@ -132,7 +131,7 @@ end
 add_command_under_category "show-service-credentials", "credential-rotation", "Show the service credentials", 2 do
   ensure_configured!
 
-  credentials = Veil::CredentialCollection::ChefSecretsFile.from_file("/etc/opscode/private-chef-secrets.json")
+  credentials = Veil::CredentialCollection::ChefSecretsFile.from_file(secrets_file_path)
   pp(credentials.legacy_credentials_hash)
   exit(0)
 end
@@ -173,6 +172,9 @@ add_command_under_category "require-credential-rotation", "credential-rotation",
     file.write <<-EOF.gsub(/^\s{6}/, "")
       add_global_pre_hook "require_credential_rotation" do
         # Allow running "chef-server-ctl rotate-shared-secrets"
+        # "chef-server-ctl" is a wapper that runs "omnibus-ctl opscode $command"
+        # so we'll look for our command and opscode in the args
+
         return true if ARGV.include?("opscode") && ARGV.include?("rotate-shared-secrets")
 
         raise("You must rotate the Chef Server credentials to enable the Chef Server. "\
@@ -186,17 +188,19 @@ end
 
 def backup_secrets_file(backup_file = nil)
   timestamp = Time.now.iso8601
-  secrets_file = "/etc/opscode/private-chef-secrets.json"
-  backup_file ||= File.expand_path(File.join(File.dirname(secrets_file), "#{File.basename(secrets_file)}#{timestamp}.json"))
-  log("Backing up #{secrets_file} to #{backup_file}...")
-  FileUtils.cp(secrets_file, backup_file)
+  backup_file ||= File.expand_path(File.join(File.dirname(secrets_file_path), "#{File.basename(secrets_file_path)}#{timestamp}.json"))
+  log("Backing up #{secrets_file_path} to #{backup_file}...")
+  FileUtils.cp(secrets_file_path, backup_file)
   backup_file
 end
 
 def restore_secrets_file(backup_file)
-  secrets_file = "/etc/opscode/private-chef-secrets.json"
-  log("Restoring #{backup_file} to #{secrets_file}...")
-  FileUtils.cp(secrets_file, backup_file)
+  log("Restoring #{backup_file} to #{secrets_file_path}...")
+  FileUtils.cp(secrets_file_path, backup_file)
+end
+
+def secrets_file_path
+  "/etc/opscode/private-chef-secrets.json"
 end
 
 def remove_backup_file(backup_file)
--- a/omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
+++ b/omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
@@ -194,15 +194,18 @@ add_command_under_category "require-credential-rotation", "credential-rotation",
       add_global_pre_hook "require_credential_rotation" do
         # Allow running "chef-server-ctl rotate-shared-secrets"
         # "chef-server-ctl" is a wapper that runs "omnibus-ctl opscode $command"
-        # so we'll look for our command and opscode in the args
+        # so we'll look for that in ARGV
 
-        return true if ARGV.include?("opscode") && ARGV.include?("rotate-shared-secrets")
+        return true if ARGV == %w{omnibus-ctl opscode rotate-shared-secrets}
 
         raise("You must rotate the Chef Server credentials to enable the Chef Server. "\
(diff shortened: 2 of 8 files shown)


omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -216,6 +216,17 @@ EOM
         "/admin/ping?wt=json"
       end
     end
+
+    # Override reconfigure to skip license checking
```
mmzyk: Why is this needed? As an open source project, the license acceptance code should not be triggered. If that's the case, this is working around a bug and this is not the right fix.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
```diff
@@ -85,22 +94,28 @@
 add_command_under_category "rotate-shared-secrets", "credential-rotation", "Rotate the Chef Server shared secrets and all service credentials", 2 do
   ensure_configured!
 
-  begin
-    timestamp = Time.now.iso8601
-    secrets_file = "/etc/opscode/private-chef-secrets.json"
-    secrets_file_backup = File.expand_path(File.join(File.dirname(secrets_file), "#{File.basename(secrets_file)}#{timestamp}"))
-    log("Backing up #{secrets_file} to #{secrets_file_backup}...")
-    FileUtils.cp(secrets_file, secrets_file_backup)
+  secrets_file = "/etc/opscode/private-chef-secrets.json"
```
mmzyk: You assign this path to the secrets_file here, but not in the command above. Should probably at least be consistent.

This path actually gets hard coded in a lot of places. I've debated if it's worth asking you to hard code it and then pass it around to the methods, but I'm not sure it's worth it. Another option would be to make secrets_file a method that just returns this path, since it seems to always be the same.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
```diff
@@ -122,6 +137,77 @@
   exit(0)
 end
 
+add_command_under_category "require-credential-rotation", "credential-rotation", "Disable the Chef Server and require credential rotation", 2 do
```
mmzyk: Disable the chef server sounds ominous. Is that the right description? Saying Stop the Chef Server is probably better, since that's the actually command being run. Disabling the chef server makes me think I'm about to royally screw something up.

mmzyk: You'll need to correct the wording in all the places you use it, fyi.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
```diff
@@ -122,6 +137,77 @@
...
+    end
+  end.parse!(ARGV)
+
+  # Agree to disable and require rotation
+  msg = "Are you sure you want to disable the Chef Server and require credential rotation?" \
+          "\n Type 'yes' to confirm: "
+  exit(1) unless @agree_to_disable || @ui.ask("<%= color(%Q(#{msg}), :yellow) %>") =~ /yes/i
+
+  # Shut down Chef Server
+  run_sv_command("stop")
```
mmzyk: So you stop the chef server, but don't start it again after the prehook has been applied? Is that by design? I can see that being very confusing to the user. Might want to output some instructions at the end that they need to restart their chef server.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
```diff
@@ -122,6 +137,77 @@
...
+    File.unlink("#{service_path}/#{service}") if service_enabled?(service)
+  end
+
+  # Create global pre hook that'll prevent ctl commands from running until the
+  # credentials have been rotated
+  File.open(credential_prehook_file_path, "a+") do |file|
+    file.write <<-EOF.gsub(/^\s{6}/, "")
+      add_global_pre_hook "require_credential_rotation" do
+        # Allow running "chef-server-ctl rotate-shared-secrets"
+        return true if ARGV.include?("opscode") && ARGV.include?("rotate-shared-secrets")
```
mmzyk: Why are we looking for opscode in the args? Where does that come from? Park of omnibus-ctl somehow? A comment on why that is here might be good.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
```diff
@@ -122,6 +137,77 @@
...
+
+        raise("You must rotate the Chef Server credentials to enable the Chef Server. "\
```
mmzyk: I'm not sure I understand the intent of this command. It seems like it just stops the chef server and locks it. How is this desirable from a user's standpoint?

omnibus/files/private-chef-ctl-commands/spec/rotate_credentials_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+        .to receive(:get_all_services)
+        .and_return(%w{postgresql oc_erchef})
+      allow(subject.ctl).to receive(:service_enabled?).and_return(true)
+      allow(File).to receive(:unlink).with(/postgresq/).and_return(true)
+      allow(File).to receive(:unlink).with(/oc_erchef/).and_return(true)
+      expect { subject.run_test_omnibus_command("require-credential-rotation", %w{--yes}) }
+        .to raise_error(SystemExit) { |e| expect(e.status).to eq(0) }
+    end
+
+    it "writes the pre-hook lock" do
```
mmzyk: Tests seem fine, although it's still hard for me when reading them to devine what the purpose of the command actually is. Ideally I'd be able to tell that from just reading the tests.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
mmzyk: We're traditionally had 1 ctl command per file, although I don't know how religiously we've stuck to that. While all the commands here are related to credentials, it might be best to separate our require-crediential-rotation, since it's in a different bucket from the others and doesn't seem to share much, if any code.

omnibus/files/private-chef-ctl-commands/chef-server-ctl
ryancragun: The current implementation in omnibus-ctl that does the license checking has no special cases for open source projects.  If _any_ license is present it requires acceptance.  I suggested several possible solutions to the omnibus maintainers and they suggested this solution. I'm more than happy to implement something else but this is what was suggested.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
ryancragun: Stopping the Chef Server and disabling the services are two different things and this command technically does both. It does disable the Chef Server in that runsvdir will no longer start any services which effectively renders them disabled. systemd and runit use that word to describe what we're doing here.  Perhaps I should say "safely disable the chef server services" or something to that effect?

As per a different file, I could do that but there this command and rotating-shared-secrets share some common code paths.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
ryancragun: You _cant_ restart the server until the secrets have been rotated which is very much intentional.  When you do rotate the secrets the Chef Server will come online again.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
ryancragun: yeah, chef-server-ctl is a wrapper that calls `omnibus-ctl opscode $command`

omnibus/files/private-chef-ctl-commands/chef-server-ctl
marcparadise: I seem to remember a reconfigure  flag that can be passed it  to suppress it?

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
marcparadise: After "yes" is typed, an informational "This Chef Server has been disabled until credentials have been rotated." message would be good.  

Even knowing that's the intent of the command, it's still helpful to show a positive response indicating that the requested thing was done.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
```diff
@@ -117,11 +145,94 @@
...
+  # Create global pre hook that'll prevent ctl commands from running until the
+  # credentials have been rotated
+  File.open(credential_prehook_file_path, "a+") do |file|
+    file.write <<-EOF.gsub(/^\s{6}/, "")
+      add_global_pre_hook "require_credential_rotation" do
+        # Allow running "chef-server-ctl rotate-shared-secrets"
+        # "chef-server-ctl" is a wapper that runs "omnibus-ctl opscode $command"
+        # so we'll look for our command and opscode in the args
+
+        return true if ARGV.include?("opscode") && ARGV.include?("rotate-shared-secrets")
```
marcparadise: This check would make it possible to do something like `chef-server-ctl start oc_erchef rotate-shared-secrets`.  I don't know how much a concern that is.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
```diff
@@ -22,21 +29,24 @@
...
-    log("Backing up #{secrets_file} to #{secrets_file_backup}...")
-    FileUtils.cp(secrets_file, secrets_file_backup)
+    backup_file = backup_secrets_file
 
     log("Rotating #{service}'s credentials...", :notice)
-    credentials = Veil::CredentialCollection::ChefSecretsFile.from_file(secrets_file)
+    credentials = Veil::CredentialCollection::ChefSecretsFile.from_file(secrets_file_path)
     credentials.rotate(service)
     credentials.save
+  rescue => e
```
marcparadise: I like the error handling improvement here

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
```diff
@@ -117,11 +145,94 @@
...
+  run_sv_command("stop")
+
+  # Disable services
+  get_all_services.each do |service|
+    File.unlink("#{service_path}/#{service}") if service_enabled?(service)
+  end
+
+  # Create global pre hook that'll prevent ctl commands from running until the
+  # credentials have been rotated
+  File.open(credential_prehook_file_path, "a+") do |file|
```
marcparadise: Source code that only exists sometimes makes me uncomfortable, and source code that writes itself to a place that's generally intended to be updated only by the package installation (within `/opt/opscode`).  We can't be sure that customers won't have read-only mounts to /opt. 

Instead, let's make a permanent pre-hook that checks for the existence of a guard file in `/var/opt/opscode`

omnibus/files/private-chef-ctl-commands/chef-server-ctl
ryancragun: @marcparadise you can indeed pass `--accept-license` but my understanding was that we didn't want that to happen for the Chef Server since it's fully open source.

ryancragun:
@mmzyk We could also override the `check_license_acceptance` instead of reconfigure if that's preferred but it's essentially doing the same thing :\

mmzyk:
@ryancragun Interesting - that might be clearly as to why it's being overwritten, as opposed to reconfigure. I double checked on things and apparently overriding is the only option we have right now, unless we rewrite how omnibus-ctl works.

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
```diff
@@ -117,11 +145,97 @@
...
+  msg = "Are you sure you want to disable the Chef Server and require credential rotation?" \
+          "\n Type 'yes' to confirm: "
+  exit(1) unless @agree_to_disable || @ui.ask("<%= color(%Q(#{msg}), :yellow) %>") =~ /yes/i
+
+  # Shut down Chef Server
+  run_sv_command("stop")
+
+  # Disable services
+  get_all_services.each do |service|
+    File.unlink("#{service_path}/#{service}") if service_enabled?(service)
```
marcparadise: I don't see anything that puts these back, which would mean a reconfigure is required after rotating secrets. Is that the intent, or am I not looking in the right place?

Too, do we need to disable them in this way if we're disabling all chef-server-ctl access to them via prehook?

ryancragun: The secrets rotation kicks off a reconfigure so they're good on that front. The reason that I disable them this way is so that on reboot runsvdir doesn't start them up.

marcparadise: Do we need to unlink since we disable access through chef-server-ctl? 

Too, I don't see that we recreate those links - are we relying on the post-rotate reconfigure for that?

marcparadise:
:+1: with clarification around File..unlink

omnibus/files/private-chef-ctl-commands/rotate_credentials.rb
ryancragun: runsvdir doesn't call out to chef-server-ctl so the user could reboot to re-enable the services. You're correct that the reconfigure re-enables the services by recreating the links.

PR #848, opened 2016-06-07
Use enterprise cookbook version that supports systemd on ubuntu 16.04

yzl:
Successful pipeline run on ubuntu 16.04 in a dev jenkins environment:  http://10.194.10.88/job/chef-server-12-trigger-ad_hoc/10/downstreambuildview/


mmzyk:
@yl Is there a reason to use the ref and to not tag it instead?

yzl:
@mmzyk Nope, just not sure what your process was.  Happy to create a tag instead.

yzl:
Updated to fetch enterprise cookbook by tag instead of git ref.

PR #852, opened 2016-06-08
oc_erchef users list: allow filtering by external_authentication_id

sdelano:
For the enablement of external authentication mechanisms that are not LDAP,
we require the ability to fetch a list of users based on their external
authentication identifier.

### DIFF
--- a/oc-chef-pedant/spec/api/user_spec.rb
+++ b/oc-chef-pedant/spec/api/user_spec.rb
@@ -64,6 +64,52 @@ describe "users", :users do
             })
         end
 
+        context "saml user" do
+          let(:username) do
+            "user_with_external_uid"
+          end
+
+          let(:external_auth_id) do
+            "sam.l.jackson"
+          end
+
+          let(:user_options) do
+            { :overrides => {
+                :first_name => "external",
+                :last_name => "user",
+                :display_name => "SAML USER",
+                :external_authentication_uid => external_auth_id
+              }
+            }
+          end
+
+          let(:filtered_external_users_body) do
+            { username => "#{request_url}/#{username}" }
+          end
+
+          before :each do
+            @user = platform.create_user(username, user_options)
+          end
+
+          after :each do
+            platform.delete_user(@user)
+          end
+
+          it "returns no users when filtering by non-existing external_authentication_uid", :smoke do
+            get("#{request_url}?external_authentication_uid=somenonexistingemail@somewhere.com", platform.superuser).should look_like({
+                :status => 200,
+                :body_exact => empty_users_body
+              })
+          end
+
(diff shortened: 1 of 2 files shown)


marcparadise:
I think there's some work needed in the query and possible `chef_user` to support this?

chefsalim:
@marcparadise can you give clarify your comment on the more work?

stevendanna:
The chef_user module already has a handler for external_authentication_ui being set in the StubRec:

```
list(#chef_user{external_authentication_uid = ExtAuthUid}, CallbackFun) when ExtAuthUid =/= undefined ->
    CallbackFun({list_users_by_ext_auth_uid, [ExtAuthUid], [username]});
list(#chef_user{email = undefined} = User, CallbackFun) ->
    CallbackFun({list_query(User), [], [username]});
list(#chef_user{email = EMail}, CallbackFun) ->
    CallbackFun({list_users_by_email, [EMail], [username]}).
```

So this is good to go I believe.  It will ignore the email parameter if that is supplied along with the external_authentication_uid, maybe we want to return an error if both are specified? Doesn't seem like a common use case since we are the only people who use this.

marcparadise:
Disregard - I took a look at the place I thought you'd need to modify (chef_user.erl, pgsql_statements,.config) and saw that we already support it, and it was just a matter of exposing as is done here.

stevendanna:
The failing test is because this endpoint hasn't also been implemented in chef-zero.  Options are to either implement it there or to add a tag to your rspec tests so that you can turn off those tests when run against chef-zero.

stevendanna:
Overall 👍. Before merging y'all will need to make a call on how to handle the chef-zero stuff and whether to return an error when both email and external_authentication_uid is specified.

marcparadise:
:+1: once chef-zero is handled.

oc-chef-pedant/spec/api/user_spec.rb
```diff
@@ -64,6 +64,52 @@
...
+          before :each do
+            @user = platform.create_user(username, user_options)
+          end
+
+          after :each do
+            platform.delete_user(@user)
+          end
+
+          it "returns no users when filtering by non-existing external_authentication_uid", :smoke do
+            get("#{request_url}?external_authentication_uid=somenonexistingemail@somewhere.com", platform.superuser).should look_like({
```
srenatus: Nitpick, looking at [chef-manage's test stubs](https://github.com/chef/chef-manage/blob/master/src/chef-manage/spec/controllers/sessions_controller_spec.rb#L49-L50), perhaps this should be `%40`... doesn't matter much for this test, though.

srenatus:
LGTM. @stevendanna: +1 re: supporting both filters -- we just don't need that now, so let's fix that when it's needed.

PR #854, opened 2016-06-09
[ET-202] Check for SAML enablement during reconfigure

chefsalim:
This change adds a sanity check to make sure we don't try and enable LDAP at the same time that SAML is enabled. It maps to the same sanity check we do during chef-manage-ctl reconfigure.

@chef/chef-server-maintainers

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
@@ -37,6 +37,13 @@ if File.exists?("/etc/opscode/chef-server-running.json")
   node.consume_attributes({"previous_run" => old_config['private_chef']})
 end
 
+# used for checking if LDAP and SAML are both enabled
+chef_manage = { "chef_manage" => {} }
+if ::File.exists?("/var/opt/chef-manage/etc/chef-manage-running.json")
+  chef_manage = JSON.parse(IO.read("/var/opt/chef-manage/etc/chef-manage-running.json"))
+end
+node.consume_attributes({"chef_manage" => chef_manage["chef-manage"]})
+
 if File.exists?("/etc/opscode/chef-server.json") &&
     !(File.exist?("/etc/opscode/private-chef.rb") || File.exist?("/etc/opscode/chef-server.rb"))
   Chef::Log.fatal("Configuration via /etc/opscode/chef-server.json is not supported. Please use /etc/opscode/chef-server.rb")
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
@@ -52,6 +52,14 @@ if OmnibusHelper.has_been_bootstrapped? or
   node.set['private_chef']['bootstrap']['enable'] = false
 end
 
+# Do a sanity check to make sure both SAML and LDAP are not enabled at the same time
+ldap_enabled = !(node['private_chef']['ldap'].nil? || node['private_chef']['ldap'].empty?)
+saml_enabled = node['chef_manage']['saml'] && node['chef_manage']['saml']['enabled']
+
+if ldap_enabled && saml_enabled
+  Chef::Log.fatal("Both SAML and LDAP auth are enabled at the same time - please enable only one of those auth types.")
+  exit!(1)
+end
 
 # Create the Chef User
 include_recipe "private-chef::users"


omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -52,6 +52,17 @@
   node.set['private_chef']['bootstrap']['enable'] = false
 end
 
+# Do a sanity check to make sure both SAML and LDAP not enabled at the same time
+chef_manage = { 'chef_manage' => {} }
```
stevendanna: This bit (56-60) might make more sense in the config recipe around here: https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb#L39

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -52,6 +52,17 @@
...
+if ::File.exists?('/var/opt/chef-manage/etc/chef-manage-running.json')
+    chef_manage = JSON.parse(IO.read("/var/opt/chef-manage/etc/chef-manage-running.json"))
+    node.consume_attributes({"chef_manage" => chef_manage['chef-manage']})
+end
+
+if !(node['private_chef']['ldap'].nil? || node['private_chef']['ldap'].empty?) && node['chef_manage'] && node['chef_manage']['saml'] && node['chef_manage']['saml']['enabled']
```
stevendanna: Maybe do a small refactor here to help with readability?

```
ldap_enabled =  !(node['private_chef']['ldap'].nil? || node['private_chef']['ldap'].empty?)
saml_enabled = node['chef_manage'] && node['chef_manage']['saml'] && node['chef_manage']['saml']['enabled']

if ldap_enabled && saml_enabled
  Chef::Log.fatal("Both SAML and LDAP auth are enabled at the same time - please enable only one of those auth types.")
  exit!(1)
end
```

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
```diff
@@ -37,6 +37,21 @@
...
 end
 
+# Do a sanity check to make sure both SAML and LDAP not enabled at the same time
+chef_manage = { 'chef_manage' => {} }
+if ::File.exists?('/var/opt/chef-manage/etc/chef-manage-running.json')
+  chef_manage = JSON.parse(IO.read("/var/opt/chef-manage/etc/chef-manage-running.json"))
+  node.consume_attributes({"chef_manage" => chef_manage['chef-manage']})
+end
+
+ldap_enabled = !(node['private_chef']['ldap'].nil? || node['private_chef']['ldap'].empty?)
```
stevendanna: Apologies, in my original comment I wasn't clear.  Lines 40-45 make sense here; but lines 48-53 should stay in the their original location since this check will only work after line 110 of this file.

srenatus: Having to re-do this is well-deserved for blindly following comments. ;)

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
```diff
@@ -37,6 +37,13 @@
   node.consume_attributes({"previous_run" => old_config['private_chef']})
 end
 
+# used for checking if LDAP and SAML are both enabled
+chef_manage = { 'chef_manage' => {} }
+if ::File.exists?('/var/opt/chef-manage/etc/chef-manage-running.json')
+  chef_manage = JSON.parse(IO.read("/var/opt/chef-manage/etc/chef-manage-running.json"))
+  node.consume_attributes({"chef_manage" => chef_manage['chef-manage']})
```
stevendanna: If you move this node.consume outside of the `if` then you can avoid checking that `node["chef-manage"]` exists before using it.

stevendanna:
Looks good to me: 👍

omnibus/files/private-chef-cookbooks/private-chef/recipes/config.rb
stevendanna: Hrm, /maybe/.  I guess there could be an odd case where chef-manage-running is poorly formed.

stevendanna:
NB.  You may want to put a similar check in chef-manage?

srenatus:
Right -- this should have already been taken care of, I think: [here](https://github.com/chef/chef-manage/blob/0222d3d049b957b31fdc837b722efb643b18f483/omnibus/files/cookbooks/omnibus-chef-manage/recipes/default.rb#L43-L46)

PR #859, opened 2016-06-15
Add ci/run_tests.sh to drive the CI process

jkeiser:
opscode-ci will get a parallel change to run this instead
of running `chef-server-ctl test -J $WORKSPACE/pedant.xml --all`
directly. This gives us per-checkin control of the test set and
procedure.


marcparadise:
:+1:

PR #862, opened 2016-06-16
Update misleading filename error message

MichaelPereira:
Since private-chef.rb has been renamed to private-chef.rb, these error messages are misleading as the file doesn't exist anymore.


stevendanna:
👍

PR #863, opened 2016-06-16
add common_name, cn ldap binding, to auth lookup fields

patrick-wright:
When LDAP auth is configured and the bindings do not include "displayname" this causes a UI bug in Manage. Once a user is authenticated the account is linked to a chef account. If the account doesn't exist a new account is created using a post to the /users rest api. Since display_name is mapped to "unknown" this is was gets displayed in the UI. Product's request is that we fall back to the cn (common name) binding in this case. Manage is handling the fall back logic, but we need to provide that info when the user is authenticated.

Manage PR https://github.com/chef/chef-manage/pull/830

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl
@@ -200,6 +200,7 @@ result_to_user_ejson(LoginAttr, UserName, [{eldap_entry, CN, DataIn}|_]) ->
     LookupFields = [{"displayname", <<"display_name">>},
                     {"givenname", <<"first_name">>},
                     {"sn", <<"last_name">>},
+                    {"cn", <<"common_name">>},
                     {"c", <<"country">>},
                     {"l", <<"city">>},
                     {"mail", <<"email">>} ],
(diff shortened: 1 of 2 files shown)


src/chef-mover/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl
```diff
@@ -171,6 +171,7 @@ result_to_user_ejson(LoginAttr, _, [{eldap_entry, CN, DataIn}|_]) ->
     LookupFields = [{"displayname", <<"display_name">>},
                     {"givenname", <<"first_name">>},
                     {"sn", <<"last_name">>},
+                    {"cn", <<"common_name">>},
```
marcparadise: Let's remove this one -- LDAP  auth components aren't used in chef-mover, and we want to keep the chef-mover vendored apps as frozen as we can.

marcparadise:
:+1:  once the chef-mover change is reverted.

Sgtpluck:
👍

stevendanna:
Doesn't prevent merge, but I'm wondering if it makes more sense to have the decision logic on what to set display_name to in this module rather than in manage.

patrick-wright:
@stevendanna I had thought the same thing. In the end, @charlesjohnson and I went with the approach of leaving that logic out of chef-server.

PR #865, opened 2016-06-22
[IPO-203] Update oc_chef_wm to send actions to the Data Collector

ryancragun:
Update oc_chef_wm to send actions to the Data Collector if it's enabled. This also slightly modifies oc_chef_action to export a function for message construction as the Data Collector uses the same schema.

### DIFF
--- a/src/oc_erchef/apps/data_collector/src/data_collector.erl
+++ b/src/oc_erchef/apps/data_collector/src/data_collector.erl
@@ -24,8 +24,8 @@
 -module(data_collector).
 -export([
          ping/0,
-         update/1,
-         is_enabled/0
+         is_enabled/0,
+         token/0
         ]).
 
 -spec ping() -> pong | pang.
@@ -35,13 +35,6 @@ ping() ->
         _ -> pang
     end.
 
--spec update(iolist() | binary()) -> ok | {error, term()}.
-update(Body) when is_list(Body) ->
-    update(iolist_to_binary(Body));
-update(Body) ->
-    %% TODO: Transform to data collector JSON
-    data_collector_http:post("/", Body).
-
 -spec is_enabled() -> boolean().
 is_enabled() ->
     case application:get_env(data_collector, root_url) of
@@ -49,4 +42,13 @@ is_enabled() ->
             true;
         undefined ->
             false
+    end.
+
+-spec token() -> list() | atom().
+token() ->
+    case application:get_env(data_collector, token) of
+        {ok, Token} ->
+            Token;
+        undefined ->
+            undefined
     end.
--- a/src/oc_erchef/apps/data_collector/src/data_collector_http.erl
+++ b/src/oc_erchef/apps/data_collector/src/data_collector_http.erl
@@ -16,8 +16,19 @@
 
 -define(DEFAULT_HEADERS, [{"Content-Type", "application/json"}]).
 
+default_headers() ->
+    maybe_add_data_collector_token_header(?DEFAULT_HEADERS).
+
+maybe_add_data_collector_token_header(Headers) ->
+    case data_collector:token() of
+        undefined ->
+            Headers;
+        Token ->
+            [{"x-data-collector-token", Token} | Headers]
+    end.
+
 request(Path, Method, Body) ->
-    request(Path, Method, Body, ?DEFAULT_HEADERS).
+    request(Path, Method, Body, default_headers()).
 
 request(Path, Method, Body, Headers) ->
     {ok, Timeout} = application:get_env(data_collector, timeout),
@@ -29,11 +40,11 @@ request(Path, Method, Body, Headers) ->
 %%
 -spec get(list()) -> ok | {error, term()}.
 get(Path) ->
-    get(Path, [], ?DEFAULT_HEADERS).
+    get(Path, [], default_headers()).
 
 -spec get(list(), iolist() | binary()) -> ok | {error, term()}.
 get(Path, Body) ->
-    get(Path, Body, ?DEFAULT_HEADERS).
+    get(Path, Body, default_headers()).
 
 -spec get(list(), iolist() | binary(), list()) -> ok | {error, term()}.
 get(Path, Body, Headers) ->
@@ -41,7 +52,7 @@ get(Path, Body, Headers) ->
 
 -spec post(list(), iolist() | binary()) -> ok | {error, term()}.
 post(Path, Body) ->
-    post(Path, Body, ?DEFAULT_HEADERS).
+    post(Path, Body, default_headers()).
 
 -spec post(list(), iolist() | binary(), list()) -> ok | {error, term()}.
 post(Path, Body, Headers) ->
@@ -49,7 +60,7 @@ post(Path, Body, Headers) ->
 
 -spec delete(list(), iolist() | binary()) -> ok | {error, term()}.
 delete(Path, Body) ->
-    delete(Path, Body, ?DEFAULT_HEADERS).
+    delete(Path, Body, default_headers()).
 
 -spec delete(list(), iolist() | binary(), list()) -> ok | {error, term()}.
 delete(Path, Body, Headers) ->
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
@@ -1,0 +1,62 @@
+-module(oc_chef_data_collector).
+
+-include("oc_chef_wm.hrl").
+
+-export([notify/2]).
+
+-spec notify(Req :: wm_req(), State :: #base_state{}) -> ok.
+notify(Req, #base_state{reqid = ReqId, resource_state = ResourceState, resource_mod = ResourceMod} = State) when
+    is_record(ResourceState, acl_state);
+    is_record(ResourceState, association_state);
+    is_record(ResourceState, client_state);
+    %% Skip controls because oc_chef_action can't extract them yet
+    %% is_record(ResourceState, control_state);
+    is_record(ResourceState, cookbook_state);
+    is_record(ResourceState, data_state);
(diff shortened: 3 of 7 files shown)


omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
```diff
@@ -253,6 +253,9 @@
     {data_collector, [
         {root_url,                "<%= node['private_chef']['data_collector']['root_url'] %>"},
         {token,                   "<%= node['private_chef']['data_collector']['token'] %>"},
+        <% if node['private_chef']['data_collector']['log_skipped_actions'] %>
```
stevendanna: Can we add a default value for this to `attributes/default.rb` in the reconfigure cookbooks with a comment about what it does?

src/oc_erchef/apps/data_collector/src/data_collector_http.erl
```diff
@@ -16,8 +16,19 @@
...
 
+default_headers() ->
+    maybe_add_data_collector_token_header(?DEFAULT_HEADERS).
+
+maybe_add_data_collector_token_header(Headers) ->
+    case data_collector:token() of
+        undefined ->
+            Headers;
+        Token ->
+            Headers ++ [{"x-data-collector-token", Token}]
```
stevendanna: Minor, but if we don't particularly care about the order: `[{"x-data-collector-token", Token} | Header]` is a bit cleaner I think.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
```diff
@@ -0,0 +1,60 @@
...
+            %% Notify any other 2XX message for matching resources
+            {_, true, _} ->
+                Msg = oc_chef_action:create_message(Req, State, true),
+                data_collector_http:post("/", Msg);
+            %% Maybe log actions that we're skipping
+            {ReqMethod, _, _} ->
+                case data_collector:log_skipped_actions() of
+                    true ->
+                        ResCode = wrq:response_code(Req),
+                        lager:info("Data Collection skipped for ~p ~p (~p)\n",
```
stevendanna: The escape for a newline is `~n`, but you shouldn't need it at the end of the line like this when using lager.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
```diff
@@ -0,0 +1,60 @@
...
+                                   [ReqMethod, ResourceState, ResCode]);
+                    false ->
+                        ok
+                end
+        end;
+
+notify(_Req, _State) ->
```
stevendanna: What do we expect to hit this clause?  Might be good to consider removing it or adding logging to it.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
```diff
@@ -0,0 +1,60 @@
...
+            %% Skip GET because we're not updating anything
+            {'GET', _, _} ->
+                ok;
+            %% authenticate_user should not be an action.
+            {_, _, oc_chef_wm_authenticate_user} ->
+                ok;
+            %% Notify any other 2XX message for matching resources
+            {_, true, _} ->
+                Msg = oc_chef_action:create_message(Req, State, true),
+                data_collector_http:post("/", Msg);
```
stevendanna: I'm I'm reading the control flow correctly, this is going to happen synchronously as part of the request.  Given that, I think we should consider some of the following:

- Logging errors that occur here rather than silently failing.
- Wrapping this call, at some level, in a stats_hero timer so that we generate statsd metrics and request log annotations with timing information about how long the post has taken.  Here is an example of how you might do that: https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_index/src/chef_index.erl#L142-L144

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
```diff
@@ -0,0 +1,60 @@
...
+            %% Maybe log actions that we're skipping
+            {ReqMethod, _, _} ->
+                case data_collector:log_skipped_actions() of
```
stevendanna: Moving this into a function might improve the readability of this notify function as it is getting a bit long.

stevendanna: Some documentation on the intent of log_skipped_actions would be helpful.  From reading this, it appears that we'll hit this case for requetss that failed?

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
```diff
@@ -0,0 +1,60 @@
...
+                end
+        end;
+
+notify(_Req, _State) ->
+    ok.
+
+-spec req_success(Req :: wm_req()) -> boolean().
+req_success(Req) ->
+    case wrq:response_code(Req) of
+        Code when Code =:= 200; Code =:= 201; Code =:= 204 ->
```
stevendanna: The comment at 34 says we log any 2XX message but here we only log a subset.  Do we want to be this strict?  If so, might want to update the comment.  If not, maybe change this to a match: `Code = [$2|_Rest]` or similar

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
```diff
@@ -0,0 +1,60 @@
+-module(oc_chef_data_collector).
+
+-include("oc_chef_wm.hrl").
+
+-export([notify/2]).
+
+-ifdef(TEST).
```
stevendanna: Do we need this?  We are exporting the major function in this module.

src/oc_erchef/apps/data_collector/src/data_collector.erl
```diff
@@ -24,8 +24,9 @@
 -module(data_collector).
```
stevendanna: Given that a lot of the meat of this ended up in the oc_chef_wm app, I wonder if this really needs to be its own app.  However, that feels like a refactor that can easily be made down the road if we want so this layout is fine by me for now.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
stevendanna: I see now that there are a number of resource_state's we aren't handling so we'll actually hit this clause often so logging here might be a bit too noisy.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
```diff
@@ -253,6 +253,9 @@
...
+        {log_skipped_actions,     <<"true">>},
```
stevendanna: It might be cleaner to use the atom true.

src/oc_erchef/apps/data_collector/src/data_collector.erl
```diff
@@ -50,3 +44,21 @@ is_enabled() ->
...
+    case application:get_env(data_collector, token) of
+        {ok, Token} ->
+            Token;
+        undefined ->
+            false
+    end.
+
+-spec log_skipped_actions() -> boolean().
+log_skipped_actions() ->
+    case application:get_env(data_collector, log_skipped_actions) of
```
stevendanna: Since we are returning a boolean, you might consider changing the item we render to the config file the atom `true` and then making this:

```
log_skipped_actions() ->
   application:get_env(data_collector, log_skipped_actions, false).
````

File: src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl (hunk 4 above)
Comment (ryancragun) [inline src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl]: That was my thought as well.

File: src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl [hunk 12]
```diff
@@ -0,0 +1,60 @@
...
+            %% Skip GET because we're not updating anything
+            {'GET', _, _} ->
+                ok;
+            %% authenticate_user should not be an action.
+            {_, _, oc_chef_wm_authenticate_user} ->
+                ok;
+            %% Notify any other 2XX message for matching resources
+            {_, true, _} ->
+                Msg = oc_chef_action:create_message(Req, State, true),
+                stats_hero:ctime(ReqId, {data_collector, update}, fun() -> data_collector_http:post("/", Msg) end);
```
stevendanna: Uff. Sorry about this, but I forgot to mention that I think in order to add a new upstream like this (rather than using an existing upstream), you'll need to modify the functions here:

https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl#L642-L666

ryancragun:
http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/190/downstreambuildview/

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_data_collector.erl
stevendanna: I'm also still thinking it might be good to handle non-success responses when posting to the data collector and at least log them so that a user could reconstruct from logs events that went missing from the data collector if, say, it was unavailable for some period of time.

ryancragun:
I decided to yank having specific config for logging and just put errors in debug logs. During design of 203 we decided that we wanted this to be a fire and forget style of updating. I think that with the combination of the status endpoint for data collector health, stats_hero metrics for all the updates, and optional debug logging of failed requests, we should be okay.

stevendanna:
👍 This is looking reasonable to me.

PR #866, opened 2016-06-27
Fix logging in server_admins_existing_users_read_permissions

stevendanna:
Permission is not a binary, it is an atom, calling binary_to_list on it
causes a badarg error.  Using the ~p format directive avoids these
issues.


markan:
👍

PR #867, opened 2016-06-29
[IPO-204] Send actions to the Data Collector before sending stats_her…

ryancragun:
This commit will send actions to the Data Collector/Analytics before the stats_hero worker sends it's data for the request. This allows us to gather additional stats_hero data during data collection that would otherwise be lost.


stevendanna:
👍 Good catch!

tylercloke:
LGTM! 👍

PR #868, opened 2016-06-29
[ET-221] Move SAML/LDAP check into pre-flight

chefsalim:
This change moves the check for SAML/LDAP into the pre-flight checks.
It adds a new AuthPreflightValidator class.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_auth_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_auth_validator.rb
@@ -1,0 +1,58 @@
+#
+# Copyright:: Copyright (c) 2016 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_auth_validator.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  private
+
+  def validate_sane_state
+    # used for checking if LDAP and SAML are both enabled
+    chef_manage = { "chef_manage" => {} }
+    if ::File.exists?("/var/opt/chef-manage/etc/chef-manage-running.json")
+      chef_manage = JSON.parse(IO.read("/var/opt/chef-manage/etc/chef-manage-running.json"))
+    end
+    node.consume_attributes({"chef_manage" => chef_manage["chef-manage"]})
```
stevendanna: If we only need this in the check, why merge this into the node object at all?

marcparadise:
:+1: pending ssd's suggestion to just pull the  attributes we need into the node, since there's just 1

chefsalim:
Updated w/ change to just pull the attributes instead of merging.

stevendanna:
:+1:

PR #870, opened 2016-06-30
Fix repo for manderson26->markan git change

markan:
The github handle manderson26 is now markan. The repo currently is forwarded, but might as well get this changed before the forwarding expires.


marcparadise:
:+1: once it's passing after the next PR merges

PR #871, opened 2016-06-30
New rack requires updating chef-zero to 4.7

markan:
Rack 2.0 requires ruby 2.2+, which we aren't using yet. Chef-zero
pulls in rack, but versions prior to 4.7 took latest. Chef-zero 4.7
pins rack to < 2.0, fixing the issue.


marcparadise:
:+1: on the change, but it looks like something else is still missed

PR #874, opened 2016-07-05
[travis] Save rebar3 cache between builds

stevendanna:


stevendanna:
This was a big help on chef-backend but doesn't seem to be doing much here.  Might be related to the version of rebar we are on, unsure atm.

marcparadise:
:+1:

PR #884, opened 2016-07-19
Add option to proxy insights data collector via the chef server

stephenbm:
Enterprise nodes won't necessarily have access to the chef automate server. This allows the nodes to report to the insights data collector via the chef server which they all already have access to instead of having to add more firewall rules.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -808,8 +808,11 @@ default['private_chef']['folsom_graphite']['retry_interval'] = 2000
 # data_collector configuration for erchef. These are used to configure an
 # opscoderl_httpc pool of HTTP connecton workers.
 # If a root_url and token are present the erchef will start the data_collector
-# application.
+# application. If proxy and root_url are present, nginx will send data_collector
+# events to the insights server
 
+# Proxy events to the insights data collector
+# default['private_chef']['data_collector']['proxy']
 # Fully qualified URL to the data collector server (e.g.: https://localhost/insights).
 # default['private_chef']['data_collector']['root_url']
 # The authentication token to pass via the header to the data collector server
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
@@ -45,6 +45,17 @@
       alias /opt/opscode/version-manifest.txt;
     }
 
+    <% if node['private_chef']['data_collector']['proxy'] -%>
+    location "/data-collector/" {
+      proxy_set_header        Host           $http_host;
+      proxy_set_header        X-Real-IP      $remote_addr;
+      proxy_set_header        X-Forwared-For $proxy_add_x_forwarded_for;
+      proxy_pass_request_headers on;
+      proxy_redirect off;
+      proxy_pass <%= node['private_chef']['data_collector']['root_url'] %>;
+    }
+    <% end -%>
+
     # bookshelf
     <% if node['private_chef']['opscode-erchef']['nginx_bookshelf_caching'] != :off -%>
     location ~ "/<%= node['private_chef']['opscode-erchef']['s3_bucket'] %>/{0,1}.*$" {


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -501,6 +501,12 @@
 default['private_chef']['nginx']['cache_max_size'] = '5000m'
 default['private_chef']['nginx']['server_names_hash_bucket_size'] = 128
 default['private_chef']['nginx']['enable_ipv6'] = false
+#
+# If your nodes don't have access to your chef automate server you can configure
+# a reverse proxy in nginx to forward requests to the data collectore via the 
+# chef server
+default['private-chef']['nginx']['insights_proxy'] = false
```
stevendanna: We already have a: `default['private_chef']['data_collector']['root_url']` option, maybe we reuse that instead?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -501,6 +501,12 @@
...
+default['private-chef']['nginx']['insights_server'] = nil
```
stevendanna: We already have a: default['private_chef']['data_collector']['root_url'] option, maybe we reuse that instead?

ryancragun: Agreed, I don't think we should add another namespace for data_collector related items.

stevendanna:
@stephenbm Thanks for the contribution!  Overall I think this is a nice feature to add.  The only change I would make is to use the `data_collector` config namespace that already exists rather than adding new configuration under `nginx`.

stephenbm:
Thanks for the feedback - I moved the attributes to the data_collector namespace

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -45,6 +45,17 @@
...
     }
 
+    <% if node['private_chef']['data_collector']['proxy'] -%>
+    location "/data-collector/v0/" {
+      proxy_set_header        Host           $http_host;
+      proxy_set_header        X-Real-IP      $remote_addr;
+      proxy_set_header        X-Forwared-For $proxy_add_x_forwarded_for;
+      proxy_pass_request_headers on;
+      proxy_redirect off;
+      proxy_pass <%= node['private_chef']['data_collector']['root_url'] %>/data-collector/v0/;
```
stevendanna: @ryancragun Can you confirm whether we generally expect the root_url to include the `/data-collector/v0` or not?

stephenbm: It took us a while to dig through and find the /data-collector/v0/ path so I think it's probably easier if people just set the the base url of the chef automate host. As far as I know the /data-collector/v0/ portion is static so I think it's a bit easier if the config excludes that.

ryancragun: @stevendanna @stephenbm The `root_url` needs to be the fully qualified URL to data_collector endpoint that you want to POST actions to. According to the RFC, the Data Collector implementations in Chef Server and Chef Client should be agnostic from the Visibility endpoint in Chef Automate, therefore it's of my opinion that we should avoid any implementation that assumes Chef Automate and/or it's endpoint.

I think it would make sense to make the nginx location `data-collector` and leave off the API version as that relates only to the Data Collector endpoint in Chef Automate, though I'm not really all that worried if we keep it as is.  I do think that we need to change the proxy pass of the location to be only `node['private_chef']['data_collector']['root_url']` and leave it up to the user to know about appending `/data-collector/v0` in the event that you wish to forward/proxy to Automate.  Currently, If you configured the `root_url` to exclude `/data-collector/v0/` then the Chef Server would fail to forward all the Chef Server actions as the `root_url` wouldn't point to the proper path. It sounds like we should also do another pass on the documentation to make sure it's clear how a user would configure the client, server, and automate for data collection.

stephenbm:
@ryancragun @stevendanna I checked the docs [here](https://docs.chef.io/release/automate/ingest_data_chef_automate.html) and updated the proxy_pass so it should support the client data and the Chef Server actions. I also changed the location to /data-collector/ as you suggested.

ryancragun:
👍 Looks good to me. Might want to squash the commits into one though.

stevendanna:
👍 If you get a chance to squash the commits it would be nice.  Otherwise this it good to merge.

PR #885, opened 2016-07-21
if pivotal.pem (and webui pair) exists, ensure permissions in default recipe

srenatus:


omnibus/files/private-chef-cookbooks/private-chef/recipes/private_keys.rb
```diff
@@ -0,0 +1,57 @@
+#
+# Author:: Adam Jacob <adam@chef.io>
```
stevendanna: :D maybe change this

srenatus: oh, I thought we'd just copy that around. ok, can do.

rmoshier:
LGTM 👍

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
```diff
@@ -49,18 +38,9 @@
...
-  content pivotal_key.to_pem.to_s
-  sensitive true
-end
-
 file OmnibusHelper.bootstrap_sentinel_file do
   owner "root"
   group "root"
   mode "0600"
-  content "You've been bootstrapped, punk. Delete me if you feel lucky. Do ya, Punk?"
+  content "You've been bootstrapped, punk. Delete me if you feel lucky. Do ya, punk?"
```
tylercloke: Heh maybe we should just update this message in general xD

tylercloke:
I don't think there is a use case for the user wanting to have custom perms on the pivotal key? Cause this would overwrite them. Seems unlikely / like a bad idea. If not, 👍

omnibus/files/private-chef-cookbooks/private-chef/recipes/bootstrap.rb
srenatus: Yeah, I've tried making it more meaningful now, including the time it was bootstrapped...

srenatus:
@tylercloke: I don't think this is an issue. You can change the owner using attributes, but you shouldn't have anything less strict than 0600 on it, I guess.

srenatus:
I've had wilson do a build and test run, [only ppc failed](http://wilson.ci.chef.co/view/Chef%20Server%2012/job/chef-server-12-test/290/). ([binaries here](http://artifactory.chef.co/repo/com/getchef/chef-server/12.8.0+20160722072334/))

PR #886, opened 2016-07-21
Server Admins Part Deux

tylercloke:
Finish up server-admins functionality by granting them permissions on orgs:
- Grant server-admins CRUD on the orgs container.
- Grant server-admins RUD on the org at org creation since the container defaults get disregarded there (should probably fix that...) via bootstrap and migration for existing installs.
- Grant server-admins CRUD on existing orgs via migration.
- Add server-admins into the admins group on org creation and for existing orgs via migration.
- Allow server-admins to bypass the org membership check for auth.
- Make all migrations idemopotent.

⚠️  ⚠️  ⚠️  ⚠️  ⚠️  
Would love to write pedant tests, but there is no global groups endpoint, so I can't programmatically add users into the server-admins group. Lots of manual testing was done. Additionally, our test coverage will at least show that normal users won't gain access to stuff they shouldn't which is the most important aspect of this work. 

⚠️ ⚠️  ⚠️  ⚠️  ⚠️ 

[Build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/212/downstreambuildview/)

### DIFF
--- a/omnibus/files/private-chef-upgrades/001/031_server_admins_add_perms_on_orgs_container.rb
+++ b/omnibus/files/private-chef-upgrades/001/031_server_admins_add_perms_on_orgs_container.rb
@@ -1,0 +1,19 @@
+define_upgrade do
+  if Partybus.config.bootstrap_server
+    must_be_data_master
+    # Make sure API is down
+    stop_services(["nginx", "opscode-erchef"])
+
+    start_services(["oc_bifrost", "postgresql"])
+    force_restart_service("opscode-chef-mover")
+    log "Graning CRUD on the orgs container to server admins."
+
+    run_command("/opt/opscode/embedded/bin/escript " +
+                "/opt/opscode/embedded/service/opscode-chef-mover/scripts/migrate " +
+                "mover_server_admins_add_perms_on_orgs_container_callback " +
+                "normal " +
+                "mover_transient_queue_batch_migrator")
+
+    stop_services(["opscode-chef-mover"])
+  end
+end
--- a/omnibus/files/private-chef-upgrades/001/031_server_admins_add_perms_on_orgs_container.rb
+++ b/omnibus/files/private-chef-upgrades/001/031_server_admins_add_perms_on_orgs_container.rb
@@ -6,7 +6,7 @@ define_upgrade do
 
     start_services(["oc_bifrost", "postgresql"])
     force_restart_service("opscode-chef-mover")
-    log "Graning CRUD on the orgs container to server admins."
+    log "Granting CRUD on the orgs container to server admins."
 
     run_command("/opt/opscode/embedded/bin/escript " +
                 "/opt/opscode/embedded/service/opscode-chef-mover/scripts/migrate " +
--- a/src/chef-mover/src/mover_server_admins_add_perms_on_orgs_container_callback.erl
+++ b/src/chef-mover/src/mover_server_admins_add_perms_on_orgs_container_callback.erl
@@ -1,0 +1,87 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
+%% ex: ts=4 sw=4 et
+%% @author Tyler Cloke <tyler@chef.io>
+%% @copyright 2015 Chef Software, Inc.
+%%
+%% This migration grants READ, CREATE, UPDATE, and DELETE perms to
+%% server admins on the orgs container.
+%%
+-module(mover_server_admins_add_perms_on_orgs_container_callback).
+
+-export([
+         migration_init/0,
+         migration_complete/0,
+         migration_type/0,
+         supervisor/0,
+         migration_start_worker_args/2,
+         migration_action/2,
+         next_object/0,
+         error_halts_migration/0,
+         reconfigure_object/2,
+         needs_account_dets/0
+        ]).
+
+-include("mover.hrl").
+-include("mv_oc_chef_authz.hrl").
+
+-define(GLOBAL_PLACEHOLDER_ORG_ID, <<"00000000000000000000000000000000">>).
+
+-record(container, {authz_id}).
+-record(group, {authz_id}).
+
+migration_init() ->
+    mv_oc_chef_authz_http:create_pool(),
+    mover_transient_migration_queue:initialize_queue(?MODULE, [?GLOBAL_PLACEHOLDER_ORG_ID]).
+
+migration_action(_GlobalOrgId, _AcctInfo) ->
+    BifrostSuperuserAuthzId = mv_oc_chef_authz:superuser_id(),
+    ServerAdminsAuthzId = get_server_admins_authz_id(),
+
+    %% Grant server-admins global group permissions on orgs container
+    OrgContainerAuthzId = get_org_container_authz_id(),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, group, ServerAdminsAuthzId, container, OrgContainerAuthzId, read),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, group, ServerAdminsAuthzId, container, OrgContainerAuthzId, create),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, group, ServerAdminsAuthzId, container, OrgContainerAuthzId, update),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, group, ServerAdminsAuthzId, container, OrgContainerAuthzId, delete),
+    ok.
+
+get_server_admins_authz_id() ->
+    {ok, [ServerAdmin]} = sqerl:select(get_server_admins_authz_id_sql(), [], rows_as_records, [group, record_info(fields, group)]),
+    ServerAdmin#group.authz_id.
+
+get_server_admins_authz_id_sql() ->
+    erlang:iolist_to_binary([<<"SELECT authz_id FROM groups WHERE name='server-admins' and org_id='">>, ?GLOBAL_PLACEHOLDER_ORG_ID , <<"'">>]).
+
+get_org_container_authz_id() ->
+    {ok, [Container]} = sqerl:select(org_container_query(), [], rows_as_records, [container, record_info(fields, container)]),
+    Container#container.authz_id.
+
+org_container_query() ->
+    <<"SELECT authz_id FROM containers WHERE name='organizations'">>.
+
+migration_complete() ->
+    mv_oc_chef_authz_http:delete_pool().
+%%
+%% Generic mover callback functions for
+%% a transient queue migration
+%%
+needs_account_dets() ->
+    false.
+
+migration_start_worker_args(Object, AcctInfo) ->
+    [Object, AcctInfo].
+
+next_object() ->
+    mover_transient_migration_queue:next(?MODULE).
+
+migration_type() ->
+    <<"server_admins_add_perms_on_orgs_container_callback">>.
+
+supervisor() ->
+    mover_transient_worker_sup.
+
+error_halts_migration() ->
+    true.
+
+reconfigure_object(_ObjectId, _AcctInfo) ->
+    ok.
--- a/src/chef-mover/src/mover_server_admins_add_perms_on_orgs_container_callback.erl
+++ b/src/chef-mover/src/mover_server_admins_add_perms_on_orgs_container_callback.erl
@@ -1,7 +1,7 @@
 %% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
 %% ex: ts=4 sw=4 et
 %% @author Tyler Cloke <tyler@chef.io>
-%% @copyright 2015 Chef Software, Inc.
+%% @copyright 2016 Chef Software, Inc.
 %%
 %% This migration grants READ, CREATE, UPDATE, and DELETE perms to
 %% server admins on the orgs container.
--- a/src/chef-mover/src/mover_server_admins_existing_orgs_permissions_callback.erl
+++ b/src/chef-mover/src/mover_server_admins_existing_orgs_permissions_callback.erl
@@ -1,7 +1,7 @@
 %% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
 %% ex: ts=4 sw=4 et
 %% @author Tyler Cloke <tyler@chef.io>
-%% @copyright 2015 Chef Software, Inc.
+%% @copyright 2016 Chef Software, Inc.
 %%
 %% This migration iterates through all existing orgs and grants
 %% the server-admins global group READ, UPDATE, and DELETE access on them,
@@ -32,7 +32,7 @@
 -record(group, {authz_id}).
 -record(mover_chef_group, {
           id,
-          org_id,	  
+          org_id,
           authz_id,
           name,
           last_updated_by,
@@ -56,8 +56,8 @@ migration_action(OrgRecord, _AcctInfo) ->
     OrgAdminsAuthzId =
         case get_org_admins_authz_group_if_exists(OrgId) of
             admins_group_missing ->
-                lager:warn("There was no admins group for org ~p with id ~p. Creating one from scratch.", [OrgName, OrgId]),
-                lager:warn("This probably means that this org is broken as it has no admins, but now you can add a new one via server-admins!"),
+                lager:warning("There was no admins group for org ~p with id ~p. Creating one from scratch.", [OrgName, OrgId]),
+                lager:warning("This probably means that this org is broken as it has no admins, but now you can add a new one via server-admins!"),
                 NewAdminsGroupAuthzId = case create_group_in_authz(BifrostSuperuserId) of
                                             AuthzId when is_binary(AuthzId) ->
                                                 AuthzId;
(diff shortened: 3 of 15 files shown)


src/chef-mover/src/mover_server_admins_global_group_callback.erl
```diff
@@ -92,14 +92,6 @@ chef_object_flatten_group(ObjectRec) ->
     end,
     Tail.
 
-throw_not_found(Type) ->
-    lager:error("Bifrost could not find the " ++ Type ++ " permission on the users container."),
-    throw(not_found).
-
-throw_server_error() ->
-    lager:error("There was an error communicating with bifrost."),
-    throw(server_error).
```
tylercloke: Not used. They were causing mover compile warnings.

src/chef-mover/src/mover_server_admins_existing_orgs_permissions_callback.erl
```diff
@@ -0,0 +1,228 @@
...
+    OrgId = OrgRecord#org.id,
+    OrgAuthzId = OrgRecord#org.authz_id,
+    BifrostSuperuserId = mv_oc_chef_authz:superuser_id(),
+    ServerAdminsAuthzId = get_server_admins_authz_id(),
+
+    % Get the admins group authz id for this org, or if it doesn't exist, create it with a warning.
+    OrgAdminsAuthzId =
+        case get_org_admins_authz_group_if_exists(OrgId) of
+            admins_group_missing ->
+                lager:warn("There was no admins group for org ~p with id ~p. Creating one from scratch.", [OrgName, OrgId]),
```
stevendanna: `lager:warn` -> `lager:warning`.  I don't think `lager:warn` exists.

src/chef-mover/src/mover_server_admins_existing_orgs_permissions_callback.erl
```diff
@@ -0,0 +1,228 @@
...
+    mv_oc_chef_authz_http:create_pool(),
+    mover_transient_migration_queue:initialize_queue(?MODULE, get_orgs()).
+
+
+migration_action(OrgRecord, _AcctInfo) ->
+    OrgName = OrgRecord#org.name,
+    OrgId = OrgRecord#org.id,
+    OrgAuthzId = OrgRecord#org.authz_id,
+    BifrostSuperuserId = mv_oc_chef_authz:superuser_id(),
+    ServerAdminsAuthzId = get_server_admins_authz_id(),
```
stevendanna: I wonder if there is a way to cache this between actions.  For large installations caching it would save a SQL query for every org to migrate.

stevendanna:
Regarding `mover_server_admins_existing_orgs_permissions_callback` I have a couple questions/concerns:

- Have we tested this against preprod data?  From my reading I don't see much that would be an issue, but considering the trouble we've had with other similar migrations it would probably be good to do.

- Currently this migration follows a patterns we started using in which any error halts the migration. This pattern makes sense for most installations but can be problematic for Hosted and other large installations.  By my reading there are at least 6 network calls for each orgs that we need to migrate (2 sql calls, 4 bifrost calls which will turn into sql calls).  If any one of those fails for transient problems, we have to restart the entire migration from the beginning.

omnibus/files/private-chef-upgrades/001/031_server_admins_add_perms_on_orgs_container.rb
```diff
@@ -0,0 +1,19 @@
+define_upgrade do
+  if Partybus.config.bootstrap_server
+    must_be_data_master
+    # Make sure API is down
+    stop_services(["nginx", "opscode-erchef"])
+
+    start_services(["oc_bifrost", "postgresql"])
+    force_restart_service("opscode-chef-mover")
+    log "Graning CRUD on the orgs container to server admins."
```
ryancragun: Typo with "Granting"

src/chef-mover/src/mover_server_admins_add_perms_on_orgs_container_callback.erl
```diff
@@ -0,0 +1,87 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
+%% ex: ts=4 sw=4 et
+%% @author Tyler Cloke <tyler@chef.io>
+%% @copyright 2015 Chef Software, Inc.
```
ryancragun: `s/2015/2016/` cause I'm a pedant

src/chef-mover/src/mover_server_admins_existing_orgs_permissions_callback.erl
tylercloke: Ah damn, I think I've used it in the past. Wouldn't I get a warning on compile time? I'll grep and update everything to `lager:warning`.

src/chef-mover/src/mover_server_admins_existing_orgs_permissions_callback.erl
tylercloke: Good idea! I think we've used ets tables or something similar before. I'll poke around.

tylercloke:
@stevendanna WRT `mover_server_admins_existing_orgs_permissions_callback`, the admins group missing queries shouldn't trigger that often, but I agree it is a lot of calls. Testing against preprod is a good idea, if you know where to get me started with that, let me know.

WRT the migration failing, I feel like that is the right thing to do for the customer for org based migrations. If it is an issue in hosted, we should instrument around it. We should add a `failures_as_warnings` mode or something...

src/chef-mover/src/mover_server_admins_existing_orgs_permissions_callback.erl
marcparadise: There's a table org_migration_state and a corresponding module `moser_state_tracker`. It'll require adding an enum type for a new migration, but it will let you track migration completion at the org level. iirc correctly the module can do most of that work for you.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -444,22 +444,30 @@ authorized_by_org_membership_check(Req, #base_state{requestor=#chef_requestor{ty
...
-                Error ->
-                    Msg = forbidden_message(unverified_org_membership, UserName, OrgName),
-                    {false, wrq:set_resp_body(chef_json:encode(Msg), Req),
-                     State#base_state{log_msg = {user_not_in_org_error, Error}}}
+            %% If user is a member of the global group server-admins,
+            %% we do not want to fail on org membership check, only acls.
+            case is_server_admin(DbContext, RequestorAuthzId) of
+                true -> {true, Req, State};
+                _ ->
+                    case chef_db:is_user_in_org(DbContext, UserName, OrgName) of
```
markan: I would guess for 90% of our requests the requestor is *not* a server admin, but is in the org. We should swap the two fetches and most likely save a request.

src/chef-mover/src/mover_server_admins_existing_orgs_permissions_callback.erl
markan: While it is an improvement, is it worth the effort? The query is likely going to be > 5ms, so even large installations with thousand's of orgs would only see a few seconds slowdown on a one time migration. I'd vote for skipping this optimization in service of getting this merged in.

markan:
I've pushed some fixes for review requests, and one minor change that I wanted. I'll rebase/squash before merging, but figured it might be easier to review the changes without that.

src/chef-mover/src/mover_server_admins_existing_orgs_permissions_callback.erl
stevendanna: @markan I'm 👍 on skipping this particular optimization and agree we need to get this merged.  

My thought about saving the sql query was for Hosted where it is less about time and more about the concerns I raised in my June 22 comment. Fewer requests reduces the surface area for transient errors.  In Hosted, per-org migrations that are this complex takes a long time (often through multiple load peaks).  Without the ability to resume, we have to restart the migration from the start if a single thing fails.

I'm still personally nervous about this migration because it looks very similar to the other migrations that got held up in Hosted for a long while.  However, I think the best thing to do is to work to get this merged and address those problems as we hit them.

stevendanna:
@markan Changes look good.  I'm happy for this to move forward (think it is a long overdue feature); let's just be sure to ping Ops to let them know another potentially long migration is in the pipe.

tylercloke:
These two files keep sneaking in: `src/oc_erchef/chef_s3.coverdata`. Can we throw them in gitignore?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl.erl
```diff
@@ -270,18 +275,30 @@ safe_fetch_ids(_Type, _OrgId, []) ->
...
-    case GivenNames -- FoundNames of
+safe_fetch_ids(Type, OrgId, Names) when not is_list(OrgId) ->
+    safe_fetch_ids(Type, [OrgId], Names);
+safe_fetch_ids(Type, OrgIds, Names) ->
+    {Missing, Records} = safe_fetch_ids_rec(Type, OrgIds, Names, []),
+    case Missing of
         [] ->
             ok;
-        Missing ->
+        _ ->
```
marcparadise: Why this change? `Missing` will always match `Missing`.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl.erl
```diff
@@ -270,18 +275,30 @@ safe_fetch_ids(_Type, _OrgId, []) ->
     % Save the cost of a query for an empty list -
     % fairly common for users and to a lesser extent groups.
     [];
-safe_fetch_ids(Type, OrgId, Names) ->
-    Records = oc_chef_authz_db:authz_records_by_name(Type, OrgId, Names),
-    FoundNames = lists:sort(names_from_records(Records)),
-    GivenNames = lists:sort(Names),
-    case GivenNames -- FoundNames of
+safe_fetch_ids(Type, OrgId, Names) when not is_list(OrgId) ->
```
marcparadise: This guard is a little ambiguous since a string will drop through to the `OrgIds` path.  `when is_binary` would be safer.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl.erl
```diff
@@ -254,7 +254,12 @@ names_to_ids(ACE, OrgId) ->
             safe_fetch_ids(user, undefined, ej:get({<<"users">>}, ACE)) ++
             safe_fetch_ids(client, OrgId, Clients)
     end,
-    GroupIds = safe_fetch_ids(group, OrgId, ej:get({<<"groups">>}, ACE)),
+    %% Global groups can be intermixed with org local groups. This creates some ambiguity as
```
marcparadise: I do see a potential problem here.  The scenario is where someone had permissions assigned to a same-named group within the org before the migration ran that created the global group.  The first time that they PUT back a body containing the global group , the org group would lose permissions.  It's obscure and not likely to be hit by many people - but for those who do hit it, it's   unavoidable if we don't enforce a policy preventing the assignment similar to our user/client disambiguation.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
```diff
@@ -151,6 +151,7 @@ check_json_validity(Part, Ace) ->
 update_from_json(#acl_state{type = Type, authz_id = AuthzId, acl_data = Data},
                  Part, OrgId) ->
     try
+        lager:error("update_part ~p -- ~p -- ~p", [Part, Data, Type]),
```
marcparadise: This should probably be `info` instead of `error`.

src/oc_erchef/apps/chef_db/src/chef_sql.erl
```diff
@@ -176,6 +178,16 @@ is_user_in_org(UserName, OrgName) ->
             {error, Error}
     end.
 
+-spec get_server_admins_authz_id() -> boolean() | {error, _}.
+get_server_admins_authz_id() ->
```
marcparadise: Let's keep this in `oc_chef_authz_db` since it's also used within an `oc_chef_authz` module - we already may have a circular dependency there, and the less we add to it, the easier it will be to untangle.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl.erl
```diff
@@ -270,18 +275,30 @@ safe_fetch_ids(_Type, _OrgId, []) ->
...
 
+% Recursively expands the types for each org id, removing the found names 
+safe_fetch_ids_rec(_Type, [], Names, Records) ->
+    {Names, Records};
+safe_fetch_ids_rec(_Type, _OrgIds, [], Records) ->
+    {[], Records};
+safe_fetch_ids_rec(Type, [OrgId | OrgIds], Names, RecordsAcc) ->
+    Records = oc_chef_authz_db:authz_records_by_name(Type, OrgId, Names),
+    FoundNames = lists:sort(names_from_records(Records)),
+    GivenNames = lists:sort(Names),
```
marcparadise: Could we keep this sort in `safe_fetch_ids` so that we don`t end up doing it more than once? (assuming that `GivenNames--FoundNames` below will preserve the sort order)

src/chef-mover/src/mover_server_admins_add_perms_on_orgs_container_callback.erl
```diff
@@ -0,0 +1,87 @@
...
+    mv_oc_chef_authz_http:create_pool(),
+    mover_transient_migration_queue:initialize_queue(?MODULE, [?GLOBAL_PLACEHOLDER_ORG_ID]).
+
+migration_action(_GlobalOrgId, _AcctInfo) ->
+    BifrostSuperuserAuthzId = mv_oc_chef_authz:superuser_id(),
+    ServerAdminsAuthzId = get_server_admins_authz_id(),
+
+    %% Grant server-admins global group permissions on orgs container
+    OrgContainerAuthzId = get_org_container_authz_id(),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, group, ServerAdminsAuthzId, container, OrgContainerAuthzId, read),
```
marcparadise: Might be a bit better as:

``` 
[mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, group,
                                     ServerAdminsAuthzId, container, OrgContainerAuthzId, Perm) || 
                                     Perm <- [read, create, update, delete] ]
 
```

src/chef-mover/src/mover_server_admins_add_perms_on_orgs_container_callback.erl
```diff
@@ -0,0 +1,87 @@
...
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, group, ServerAdminsAuthzId, container, OrgContainerAuthzId, create),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, group, ServerAdminsAuthzId, container, OrgContainerAuthzId, update),
+    mv_oc_chef_authz:add_ace_for_entity(BifrostSuperuserAuthzId, group, ServerAdminsAuthzId, container, OrgContainerAuthzId, delete),
```
marcparadise: Should we also be including `grant`?

src/chef-mover/src/mover_server_admins_add_perms_on_orgs_container_callback.erl
```diff
@@ -0,0 +1,87 @@
...
+    ok.
```
marcparadise: Should we log or capture  in any way if any of these adds fails?

src/chef-mover/src/mover_server_admins_add_perms_on_orgs_container_callback.erl
```diff
@@ -0,0 +1,87 @@
...
+migration_init() ->
+    mv_oc_chef_authz_http:create_pool(),
+    mover_transient_migration_queue:initialize_queue(?MODULE, [?GLOBAL_PLACEHOLDER_ORG_ID]).
+
+migration_action(_GlobalOrgId, _AcctInfo) ->
+    BifrostSuperuserAuthzId = mv_oc_chef_authz:superuser_id(),
+    ServerAdminsAuthzId = get_server_admins_authz_id(),
+
+    %% Grant server-admins global group permissions on orgs container
+    OrgContainerAuthzId = get_org_container_authz_id(),
```
marcparadise: Each one of the `add_ace_for_entity` calls will cause a GET of the full ACL - could we instead do a single fetch of the full ACL, then use `set_ace_for_entity`?

marcparadise:
Nit: commits don't have DCO signoff.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl.erl
markan: I think this is the argument for moving server-admins to a globally distinct namespace; a reserved prefix would help us distinguish the two.

stevendanna: > It's obscure and not likely to be hit by many people

Let's do a search against hosted data to see if we have any conflicting group names.

Also, until we have a solution that allows us to distinguish global vs local groups in user-provided acl input, maybe we should disallow people from creating groups or actors that match various reserved names in the system?

poliva83:
@marcparadise @stevendanna What is the ETA on merging this PR?

stevendanna:
Closing. We can reopen if we want to re-explore this work.

PR #891, opened 2016-07-26
Added db init and max pool size options

paulmooring:
Added options to configure db_pool_max and db_pool_init to chef-server.rb for bifrost and erchef.  db_pool_size is still the only default set and max and init are only used if set by the user.


stevendanna:
LGTM, let's remember to add this to the docs.

ryancragun:
👍

tylercloke:
👍 I can help you get the doc updated if you want.

marcparadise:
:+1: on merging though I'm curious what led to it? I seem to recall some difficulties that could happen when actually using a init size that was less than max.

paulmooring:
@marcparadise we've hit an issue with postgres/RDS where persistent connections continue to eat memory until postgres starts swapping and fails.  Pooler will recycle connections when they are inactive, but this doesn't happen if init and max are set to the same number (which is the only currently supported configuration via omnibus).  We've been running with this change via a manually editted sys.config file for a few weeks and so far all is well.

marcparadise:
Okay, thanks for the clarification

PR #895, opened 2016-08-02
Minor logrotate configuration fixes

mhorbul:
The following fixes in the logrotate configuration files allows to avoid issues on the boxes with SELinux enabled.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/opc_logrotate.cron
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/opc_logrotate.cron
@@ -3,4 +3,4 @@
 PATH=/opt/opscode/embedded/sbin:$PATH
 
 command -v logrotate >/dev/null 2>&1 || exit 0
-nice -n 19 ionice -c3 logrotate /etc/opscode/logrotate.conf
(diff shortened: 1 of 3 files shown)


marcparadise:
Thanks for submitting this @mhorbul! The changes look good.   ll run this through our CI and merge if there are no issues.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/opc_logrotate.cron
```diff
@@ -3,4 +3,4 @@
 PATH=/opt/opscode/embedded/sbin:$PATH
 
 command -v logrotate >/dev/null 2>&1 || exit 0
-nice -n 19 ionice -c3 logrotate /etc/opscode/logrotate.conf
+nice -n 19 ionice -c3 logrotate -s /var/log/chef/logrotate.status /etc/opscode/logrotate.conf
```
stevendanna: Overall this looks good, my only question is whether `/var/log/chef` is the most appropriate path for this.

mhorbul: good point. Let me move it to /var/log/opscode.

marcparadise:
This went through with no issue, and I've restarted the Travis job (looks a transient failure where the tests were terminated externally).  Will merge once that's clear.

PR #899, opened 2016-08-16
Secure oc-id session cookies

tduffield:
This change addresses SPOOL-219 by enforcing oc-id session cookies to be secure. 

It also fixes a small bug in dev as well as adding rails server support to dvm.


dev/Vagrantfile
```diff
@@ -118,7 +118,7 @@ def define_chef_server(config, attributes)
       chef.add_recipe("dev::user-env")
       chef.add_recipe("dev::dvm")
       chef.json = json || {}
-      #chef.nodes_path = "nodes"
+      chef.nodes_path = "nodes"
```
marcparadise: I actually tried this locally (lastest vagrant) and chef zero failed because of a 404 on the intiial /nodes/ request. I had to change the provider to chef-solo to get dvm online.  That said if it works for you, it's probably a version related thing.  I'll hunt it down and update the README. :+1:

tduffield: I think the nodes folder needs to exist, it doesn't automatically get created.

dev/dvm/lib/dvm/project/rails.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+    def start_server
+      say "Starting #{service['name']} Rails Server"
+      # We can't trigger a background process via run_command, and -d won't redirect STDOUT
+      exec "cd #{project_dir} && bundle exec rails server -p #{port} -b #{host} -e #{environment} > /var/log/#{omnibus_project}/#{service['name']}/current 2>&1 &"
+
+    end
+    def unload
+      run_command("kill -9 $(cat tmp/pids/server.pid)", "Stopping #{service['name']} Rails Server", cwd: project_dir)
+      run_command("chef-server-ctl start #{service['name']}", "Restarting #{omnibus_project} #{service['name']}")
```
marcparadise: Do we need to remove the symlinks?

marcparadise:
:+1: to the secure change, and just the one question about the dvm addition.

dev/dvm/lib/dvm/project/rails.rb
tduffield: No. tmp automatically get deleted when the process quits, and database.yml doesn't get synced because the included one is useless in dvm.

cnunciato:
:+1:

PR #900, opened 2016-08-16
force chef_password:verify to fail if salt is null

tduffield:
Also, run `crypto:hash` in a case statement to prevent any secure data
from leaking into the logs.

### DIFF
--- a/src/oc_erchef/apps/chef_objects/src/chef_password.erl
+++ b/src/oc_erchef/apps/chef_objects/src/chef_password.erl
@@ -99,9 +99,18 @@ verify(Password, {HashedPass, Salt, _}) ->
     end.
 
 sha1(Salt, Password, osc) ->
-    hexstring(crypto:hash(sha, ["--", Salt, "--", Password, "--"]));
+    hexstring(safe_crypto(["--", Salt, "--", Password, "--"]));
 sha1(Salt, Password, ec) ->
-    hexstring(crypto:hash(sha, [Salt, "--", Password, "--"])).
+    hexstring(safe_crypto([Salt, "--", Password, "--"])).
+
+%% Eat any exceptions so secure data doesn't escape
+-spec safe_crypto(list()) -> binary().
+safe_crypto(Data) ->
+    try crypto:hash(sha, Data) of
+        Digest -> Digest
+    catch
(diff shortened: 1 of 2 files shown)


marcparadise:
:+1:

src/oc_erchef/apps/chef_objects/src/chef_password.erl
```diff
@@ -99,9 +102,17 @@ verify(Password, {HashedPass, Salt, _}) ->
...
 sha1(Salt, Password, ec) ->
-    hexstring(crypto:hash(sha, [Salt, "--", Password, "--"])).
+    hexstring(safe_crypto([Salt, "--", Password, "--"])).
+
+%% Eat any exceptions so secure data doesn't escape
+safe_crypto(Data) ->
+    try crypto:hash(sha, Data) of
+        Digest -> Digest
+    catch
+        _ -> <<"crypto:hash failure">>
```
markan: Wonder if it's worth mocking crypto:hash and having it return an exception to cover this side of the branch?

src/oc_erchef/apps/chef_objects/src/chef_password.erl
```diff
@@ -99,9 +102,17 @@ verify(Password, {HashedPass, Salt, _}) ->
     end.
 
 sha1(Salt, Password, osc) ->
-    hexstring(crypto:hash(sha, ["--", Salt, "--", Password, "--"]));
+    hexstring(safe_crypto(["--", Salt, "--", Password, "--"]));
 sha1(Salt, Password, ec) ->
-    hexstring(crypto:hash(sha, [Salt, "--", Password, "--"])).
+    hexstring(safe_crypto([Salt, "--", Password, "--"])).
+
+%% Eat any exceptions so secure data doesn't escape
```
markan: maybe add a dialyzer typespec here; even
-spec safe_crypto(list()) -> binary().
would help.

markan:
👍  Other than the very minor nits above

src/oc_erchef/apps/chef_objects/src/chef_password.erl
marcparadise: With the null atom that is passed in in the test it does hit this path.

PR #903, opened 2016-08-17
Tweak rails DVM setup to be more like erlang

tduffield:
* Be able to start process in foreground or background
* Support symlinking of configurable files
* Fix bugs preventing oc-id from loading properly


marcparadise:
:+1: I like the consistency with erlang services

PR #904, opened 2016-08-18
SPOOL-322: Fix Elasticsearch Reindex

tduffield:
1. Adds chef-backend to DVM
2. Replaces use of deprecated delete-by-query API with recommended reindexing strategy: scroll query and then delete.

### DIFF
--- a/src/oc_erchef/apps/chef_index/src/chef_elasticsearch.erl
+++ b/src/oc_erchef/apps/chef_index/src/chef_elasticsearch.erl
@@ -31,7 +31,7 @@ update(Body) ->
                     {error, {solr_400, string()}} |
                     {error, {solr_500, string()}}.
 search(#chef_solr_query{} = Query) ->
-    Url = "/chef/_search",
+    Url = "/chef/_search?scroll=1m",
     {ok, Code, _Head, Body} = chef_index_http:request(Url, get, query_body(Query)),
     case Code of
         "200" ->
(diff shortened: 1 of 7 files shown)


src/oc_erchef/apps/chef_index/src/chef_elasticsearch.erl
```diff
@@ -80,20 +112,49 @@ query_string_query_ejson(QueryString) ->
...
+                            },
+    {ok, _, _, Ids} = search(Query),
+    delete_ids(Ids).
+
+delete_ids([]) ->
     ok = commit(),
-    ok.
+    ok;
+delete_ids([Id | Ids]) ->
+    ok = chef_index_http:delete("/chef/object/" ++ Id, []),
```
marcparadise: This will stop deletes on any one failure - is that the intent?

tduffield: That was not the intent, no. We should confirm with @charlesjohnson if thats the behavior he wants when we are re-indexing.

charlesjohnson: Our conversation in standup today resolved to:
- "Assuming infrequent failures, and assuming failures mean that false data could remain in the system, then let's start by tending toward fragile-but-true, and back that off if we need to."

Issue #111, opened 2015-02-25
400 Bad Request trying to grant permissions to client when user has same name


jkeiser:
When a user on the server (but not in an org) has the same name as a client in the org, you cannot add that client to the permissions of any object.  It gives you back a 400 instead.  I would expect it to add the permissions (assuming the user isn't also in the org and there is no ambiguity).

This affects chef-provisioning and any client (including the webui) that tries to add permissions to objects.

Detail
------

When I have:
- An organization "foo" with node "bar"
- A user "mario" on the Hosted or Enterprise Chef Server who is not in organization "foo"
- A client "mario" in organization "foo"

And I do a `PUT /organizations/foo/nodes/bar` with:

```json
{
  "read": {
    "actors": [
      "mario",
      "pivotal"
    ]
  }
}
```

I expect the request to succeed and `mario` to have `read` access to `bar`.

Instead, I get this:

```
[2015-02-24T11:38:58-08:00] DEBUG: ---- HTTP Status and Header Data: ----
[2015-02-24T11:38:58-08:00] DEBUG: HTTP 1.1 400 Bad Request
[2015-02-24T11:38:58-08:00] DEBUG: server: ngx_openresty
[2015-02-24T11:38:58-08:00] DEBUG: date: Tue, 24 Feb 2015 19:38:57 GMT
[2015-02-24T11:38:58-08:00] DEBUG: content-type: text/html
[2015-02-24T11:38:58-08:00] DEBUG: content-length: 156
[2015-02-24T11:38:58-08:00] DEBUG: connection: close
[2015-02-24T11:38:58-08:00] DEBUG: x-ops-api-info: flavor=hec;version=12.0.0;oc_erchef=0.29.2
[2015-02-24T11:38:58-08:00] DEBUG: ---- End HTTP Status/Header Data ----
[2015-02-24T11:38:58-08:00] DEBUG: ---- HTTP Response Body ----
[2015-02-24T11:38:58-08:00] DEBUG: <html><head><title>400 Bad Request</title></head><body><h1>Bad Request</h1>Bad Request<p><hr><address>mochiweb+webmachine web server</address></body></html>
[2015-02-24T11:38:58-08:00] DEBUG: ---- End HTTP Response Body -----
```

marcparadise:
[...]

PR #905, opened 2016-08-18
[SPOOL-197] [#111] clients can be added to ACL even if user exist

marcparadise:
This addresses three long-standing issues - see also #111 :

1. when a user exists in the Chef Server, it was not possible to add
   a client to an object's ACL if that client had the same name. The
   logic that retrieved actors based on names obeyed org constraints for
   clients, but looked at the global users list without consideration
   for whether or not the users were members of the organization.  This
   has been corrected, so that the presence of a user anywhere in the
   system can no longer block a same-named client from being added to an
   object's ACL.
2. when an actor being added does not exist or is not in the
   organization, the request would fail with a 400 'missing/invalid actor'
   message. It would not give any indication of which actor(s) caused a
   problem. This has been corrected, and the error message now includes the
   list of actor(s) that could not be added.
3. when an actor being added exists as both an org client and a user,
   the same "400 missing/invalid actor" message would be sent.
   Occurrences of this will be reduced now that we restrict the search
   to users in the org, but can still occur if an org-user shares a name
   with a client.   We have changed this to reply with "422 (Unprocessable Entity)".
   The error message explains that the actor name(s) are ambiguous and
   provides the list of names.

 The next round of updates will expand the acl API to accept and provide
`clients` and `users` attributes.  Using those attributes instead of
`actors` when updating ACLs will give a workaround for the scenario described above (3)
and will be the preferred method for updating ACLs via the API.

Remaining work: 
* [x] there exist no unit tests for `oc_chef_authz_acl`, at minimum add coverage for new  functions.


src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl.erl
```diff
@@ -182,32 +182,48 @@ has_grant_on(ObjectType, ObjectId, ActorId) ->
...
+        {_, _, Ambiguous}  -> {error, {ambiguous_actor, names_from_records(Ambiguous)}}
     end.
 
+names_from_records(Records) ->
+    [ Name || {Name, _, _} <- Records].
+
+id_from_record({_, UserAuthzId, null}) -> UserAuthzId;
+id_from_record({_, null, ClientAuthzId}) -> ClientAuthzId.
+
+is_missing_actor({_, UserAuthzId, ClientAuthzId}) ->
```
stevendanna: I think our usual style for something like this would be:

```
is_missing_actor({_, null, null}) ->
    true;
is_missing_actor({_, _, _}) ->
   false.
```

The present code is fine and correct though.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_db.erl
```diff
@@ -182,3 +183,18 @@ fetch_group(#oc_chef_authz_context{reqid = ReqId, server_api_version = ApiVersio
         {error, _} = Error ->
             Error
     end.
+
+find_org_actors_by_name(OrgId, ActorNames) ->
+    case sqerl:select(find_org_actors_by_name, [OrgId, ActorNames]) of
+        {ok, L} when is_list(L) ->
+            % WIP - change this to a record.
```
stevendanna: We are bad at coming back to WIP and TODO comments.  Let's remove this unless we think we are gonna do it in a soon-to-come follow up PR.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
```diff
@@ -111,15 +111,16 @@ from_json(Req, #base_state{organization_guid = OrgId,
...
-            % We don't actually know which one, so can't use a more specific
-            % message without checking every single object (which we don't yet
-            % do efficiently); we only get back a list of Ids that has a
-            % different length than the list of names we started with.
-            Msg = <<"Invalid/missing actor in request body">>,
-            Msg1 = {[{<<"error">>, [Msg]}]},
-            Req1 = wrq:set_resp_body(chef_json:encode(Msg1), Req),
-            {{halt, 400}, Req1, State#base_state{log_msg = bad_actor}};
+        {ambiguous_actor, Actors} ->
+            Msg = {[{<<"error">>, iolist_to_binary([<<"The actor(s)">>, Actors,
```
stevendanna: We may want to add spaces to the error message, I think as-is this will end up looking like: "The actors(s)actor_1actor_2actor_3exist as both..."

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
```diff
@@ -133,6 +134,8 @@ from_json(Req, #base_state{organization_guid = OrgId,
 
 %% Internal functions
 
+format_actor_list(Actors) ->
+    gTkkkkgT
```
stevendanna: What's this for?

src/oc_erchef/apps/oc_chef_authz/priv/pgsql_statements.config
```diff
@@ -244,3 +246,23 @@
...
+% THere is a special case for pivotal, because pivotal is not a member
+% of any org, but is commonly present in ACLs.
+{find_org_actors_by_name,
+  <<"WITH inputs   AS (SELECT unnest($2::text[]) AS name_in),
+          user_org AS (SELECT username, authz_id AS u_authz_id
+                         FROM org_user_associations, users
+                        WHERE org_id = $1 AND user_id = users.id
+                       UNION
+                       SELECT username, authz_id AS u_authz_id FROM users WHERE username = 'pivotal')
+   SELECT name_in, u_authz_id, clients.authz_id c_authz_id
```
stevendanna: Should this be `SELECT name_in, u_authz_id, clients.authz_id AS c_authz_id`

stevendanna:
Big fan of improved error messages from the API

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_db.erl
marcparadise: There's another 1 or 2 TODOs floating around that's going to wrapped up in a separate PR later today - when putting that one together I'll make sure they're all cleaned when I submit it.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
marcparadise: Yep, this is one of the times when a TODO would have helped me remember to come back :D Fix in flight.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_acl_permission.erl
marcparadise: Misbehaving editor artifact, apparently.

src/oc_erchef/apps/oc_chef_authz/priv/pgsql_statements.config
marcparadise: The `AS` is optional, but I'll add it for clarity.

stevendanna: TIL!

tduffield:
Looks sane to me.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_acl.erl
```diff
@@ -182,32 +185,48 @@ has_grant_on(ObjectType, ObjectId, ActorId) ->
...
+
+fetch_actors(OrgId, ActorNames) ->
+    {ok, Actors} = oc_chef_authz_db:find_org_actors_by_name(OrgId, ActorNames),
+    {Missing, Remaining} = lists:partition(fun is_missing_actor/1, Actors),
+    {Ambiguous, Valid} = lists:partition(fun is_ambiguous_actor/1, Remaining),
+    case {Valid, Missing, Ambiguous} of
+        {Ids, [], []}      -> {ok, [id_from_record(R) || R <- Ids]};
+        {_, Missing, []}   -> {error, {bad_actor, names_from_records(Missing)}};
+        {_, [], Ambiguous} -> {error, {ambiguous_actor, names_from_records(Ambiguous)}};
+        {_, _, Ambiguous}  -> {error, {ambiguous_actor, names_from_records(Ambiguous)}}
```
stevendanna: What's the difference between this case and the proceeding case?

marcparadise: In this case we also have missing values - but we're ranking 'ambiguous' higher  in terms of what we report first since it's harder to fix.  The idea being to prevent giving them a bunch of little things to fix (removing invalid actors), then coming back with the one big thing that either can't or requires a bunch more change - tell about the more complicated issue up front.

stevendanna:
👍 Looks good to me.

PR #906, opened 2016-08-18
Version Bumping

tduffield:
* Use Ubuntu 14.04 as default in omnibus-dev build
* Update Ruby deps to match Chef 12.14.13


.travis.yml
```diff
@@ -51,8 +51,8 @@ install:
   - luarocks install --local lpeg
   - luarocks install --local lua-cjson
   - eval $(luarocks path)
-  - rvm install --binary 2.1.4
-  - rvm use 2.1.4
+  - rvm install --binary 2.3.1
```
marcparadise: Do we need to update the `CHEFDK_GECODE_PATH` for the new version number?  (L39)

marcparadise:
:+1: save the one question, we'll see how travis & pipeline like it...

.travis.yml
tduffield: ChefDK on master is still using Ruby 2.1

omnibus/Gemfile.lock
```diff
@@ -1,19 +1,22 @@
 GIT
   remote: git://github.com/chef/omnibus-software.git
-  revision: 40fbb9d412a2ebc17c06ff34922682de28970127
+  revision: b1a068c75d4eb0833c6c3884f7e0692ab1f29fb1
+  branch: ksubrama/ruby23
```
stevendanna: I'm really not thrilled with the number of projects where we are now on this branch.  We should help get it merged if possible.

src/opscode-expander/Gemfile
```diff
@@ -2,7 +2,7 @@
 source "http://rubygems.org"
 
 gem "amqp", "~> 0.6.7"
-gem "eventmachine", '~> 0.12.10'
+gem "eventmachine", '~> 1.2.0'
```
stevendanna: This is a big jump.  I think we have to do it eventually, but we'll want to keep it in mind if there are expander issues in the near future.

omnibus/Gemfile.lock
tduffield: For sure! I'm going to keep in touch with maintainers :)

stevendanna:
Looks reasonable

src/opscode-expander/Gemfile
tduffield: Yep, I pulled this in from @sersut's #907 change.

PR #907, opened 2016-08-18
Fix chef-server build with clear caches

sersut:
Bumps the included Ruby version to 2.2.5 and eventmachine gem to 1.2.0.

The main reason for this is the version of berkshelf we are picking up requires buff-extensions which requires Ruby version >= 2.2.0. When we bump ruby, the older version of eventmachine can not build against ruby 2.2.X which needs a bump as well.

/cc: @chef/chef-server-maintainers


tduffield:
Closed in favor of #906

PR #908, opened 2016-08-18
Bring in omnibus branch with mysys2 fixes

tduffield:
This is temporary and should be reverted when merged.


stevendanna:
Merge at will :D

PR #911, opened 2016-08-22
Add licensing information for the transitive dependencies of Chef Server

sersut:
This PR contains the minor fixes to the software definitions which will enable Chef Server to collect licensing information of transitive dependencies. Fixes are generally in one of the two nature:

1-) Include a `bundle install` so that we create the `Gemfile.lock` which is used by license_scout to collect the license information of dependencies.
2-) Include `skip_transitive_dependency_licensing` for the software which does not need license collection for transitive dependencies.

You can read https://github.com/chef/omnibus/pull/705 about the details of the licensing feature. Also notice that https://github.com/chef/license_scout/pull/15 and https://github.com/chef/omnibus-software/pull/716 are the accompanying PRs and once approved I will fix the `Gemfile` before merging this. 

With these we will have 3 things remaining in Chef Server which I will fix with subsequent PRs. Wanted to get this big set in before it gets even bigger 😄 The remaining things are:

- [x] Implement license collection via Berkshelf in license_scout and fix this error.
- [x] Fix a bug in license_scout which can interpret `BUNDLE_PATH` while collecting licenses and fix this issue:
- [x] Add license overrides for perl dependencies and fix errors in https://gist.github.com/sersut/2f826a14db412f2321f29f4bbc0e7984

/cc: @danielsdeleo @ryancragun @chef/chef-server-maintainers

### DIFF
--- a/omnibus/config/software/private-chef-cookbooks.rb
+++ b/omnibus/config/software/private-chef-cookbooks.rb
@@ -29,8 +29,7 @@ build do
       " -v 2.1.3" \
       " --no-rdoc --no-ri", env: env
 
-  command "berks vendor #{install_dir}/embedded/cookbooks",
-          env: env, cwd: "#{project_dir}/private-chef"
+  command "berks vendor #{install_dir}/embedded/cookbooks", env: env
 
   block do
     fips_enabled = (project.overrides[:fips] && project.overrides[:fips][:enabled]) || false
--- a/omnibus/config/software/private-chef-cookbooks.rb
+++ b/omnibus/config/software/private-chef-cookbooks.rb
@@ -29,7 +29,6 @@ build do
       " -v 2.1.3" \
       " --no-rdoc --no-ri", env: env
 
-
   command "berks vendor #{install_dir}/embedded/cookbooks",
           env: env, cwd: "#{project_dir}/private-chef"
 
--- a/omnibus/Gemfile
+++ b/omnibus/Gemfile
@@ -8,7 +8,8 @@ gem 'bundler', '>1.10'
 # Install omnibus software
 group :omnibus do
   gem 'omnibus', github: 'chef/omnibus'
-  gem 'omnibus-software', github: 'chef/omnibus-software'
+  gem 'omnibus-software', github: 'chef/omnibus-software', branch: 'sersut/chef-server-dep-licensing'
+  gem 'license_scout', github: 'chef/license_scout', branch: 'sersut/chef-server-fixup'
 end
 
 group :test do
--- a/omnibus/Gemfile
+++ b/omnibus/Gemfile
@@ -8,8 +8,8 @@ gem 'bundler', '>1.10'
 # Install omnibus software
 group :omnibus do
   gem 'omnibus', github: 'chef/omnibus'
-  gem 'omnibus-software', github: 'chef/omnibus-software', branch: 'sersut/chef-server-dep-licensing'
-  gem 'license_scout', github: 'chef/license_scout', branch: 'sersut/chef-server-fixup'
+  gem 'omnibus-software', github: 'chef/omnibus-software'
+  gem 'license_scout', github: 'chef/license_scout'
 end
 
 group :test do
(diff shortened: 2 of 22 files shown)


sersut:
http://wilson.ci.chef.co/job/chef-server-12-build/395/ is the build with these changes.

danielsdeleo:
changes lgtm 👍

omnibus/config/software/private-chef-cookbooks.rb
```diff
@@ -29,7 +29,6 @@
       " -v 2.1.3" \
       " --no-rdoc --no-ri", env: env
 
-
```
marcparadise: Do we need it here as well?

marcparadise:
Just the one question, but lgtm 

:+1:

omnibus/config/software/private-chef-cookbooks.rb
sersut: We have a berkshelf based license collector in the works too so we need to leave this definition alone for now 😄

stevendanna:
> The reason we are making this change is because licensing. license_scout gem that we use to gather licenses of transitive dependencies expects to find the files of the used dependency manager (berkshelf in this case) at the root of the repo. 

> This project seems to be the only project that does not place these files under the root so instead of creating special options in omnibus and license_scout, this commit reorders the directory structure of 'files/private-chef-cookbooks'.

I might be misunderstand how this project is different from others.  In my view, the old organizations (the cookbooks in named directories under the cookbook dir) matches a number of other projects:

https://github.com/chef/opscode-pushy-server/tree/master/omnibus/files/pushy-server-cookbooks
https://github.com/chef/oc_reporting/tree/master/omnibus/files/cookbooks
https://github.com/chef/chef-manage/tree/master/omnibus/files/cookbooks
https://github.com/chef/chef-analytics/tree/master/omnibus/files/cookbooks

However, I am the first to admit that I hate the current organization of the cookbooks.  If we are going to go through the trouble of moving them, maybe lets move them to the top of the directory tree or as a subdirectory of `src/` so that they are more clearly next to the other core application code?

sersut:
Thanks for checking this out @stevendanna. Did not know the directory structure of other projects 😄 

From licensing point of view what we need is to have the `Berksfile` and `Berksfile.lock` under the source directory specified in the software DSL:
```
source path: "#{project.files_path}/#{name}"
```
Thinking about it now, instead of moving the cookbooks around I can update `source` to be `path: "#{project.files_path}/#{name}/private-chef"`.

I can also relocate them under the `src` or another directory too. Let me know 😄

sersut:
Actually on second thought, I will need to put the `chef-ha-drbd` cookbook into a place where I can refer with a partial path in the `Berksfile`. So I need to at least relocate `chef-ha-drbd` into a sub directory of `private-chef`.

stevendanna:
@sersut I wonder if it would be possible to write a Berksfile that could just live next to both ie.

```
some_root_dir
| - Berksfile
| - private-chef/
| - chef-ha-drbd/
```

sersut:
Hrmm sweet idea. Will 👀 into it.

sersut:
@stevendanna LMK what you think about this. This should be the final set. CI run is [here](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/230/downstreambuildview/).

stevendanna:
👍 (Reminder: Gemfile still needs to be fixed up)

danielsdeleo:
Changes since last review lgtm 👍

sersut:
Found an issue with rebar based license collection and update the `Gemfile.lock` one more time to be able to test the rebar fix. It is at https://github.com/chef/license_scout/pull/19 incase you are interested 😄

sersut:
http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/242/downstreambuildview/ is a full green run with this change. 

Since the required omnibus changes are bumping the `SERIAL_NUMBER` of the git cache,  I will be merging these changes towards EOD today so that git caches can warm up during the silent hours.

PR #914, opened 2016-08-24
Run Ruby tests on Ruby 2.3 (main Ruby for chef/chef-dk now)

jkeiser:


tduffield:
Do we want to wait until 2.3.1 is available for Omnibus as well? Just so that the Ruby we test on is the same that we ship with?

jkeiser:
Sure! Is omnibus running 2.2.5 right now? I thought it was running 2.1.9.

jkeiser:
This isn't really blocking anything, FYI.

stevendanna:
Going to close this.  We'll bump these ruby versions when we bump it in the main omnibus config as well.  Thanks!

PR #919, opened 2016-08-29
Add LDAP server to DVM

tduffield:


dev/Vagrantfile
```diff
@@ -110,6 +127,17 @@ def define_chef_server(config, attributes)
       json = simple_deep_merge(json, { "provisioning" => { "opscode-reporting-config" => pg } })
     end
 
+    if attributes['vm']['ldap']['start'] and attributes['vm']['ldap']['use-external']
```
marcparadise: In this case, the additional external flag isn't buying us anything - I kind of see this as more we're using LDAP and need the ldap server, or we aren't/don't.  

'use-external' originally came about indicate that server should be configured to use external versions of things that ship with it (postgres).

dev/cookbooks/provisioning/files/default/ldap-data/user-child.ldif
```diff
@@ -0,0 +1,12 @@
+dn: cn=Julia Child,ou=chefs,dc=chef-server,dc=dev
```
marcparadise: request for `user-user1` ldif file too, with password `password`.  Keeps it simple and easy to remember when we want to test something quickly.

marcparadise:
:+1: other than a couple of minor nits.  I'd like to take a close rlook at our Vagrantfile, I think it's getting a bit unwieldy - I'll probably do that sometime this week. 

Thanks for taking the time to do this - we've been needing it for years.

tylercloke:
👍 Nice! :poolparty: present and future will thank you.

dev/cookbooks/provisioning/files/default/ldap-data/user-child.ldif
tduffield: Done :)

PR #920, opened 2016-08-30
New webmachine to fix multi-host header, and tests to cover it.

marcparadise:
When X-Forwarded-Host or X-Forwarded-Host contains a list of hosts
instead of a single host, erchef would crash [via webmachine].

This updates to the latest webmachine which contains the fix for that,
and adds some pedant coverage.


marcparadise:
the WM change  for reference: https://github.com/webmachine/webmachine/pull/270

PR #921, opened 2016-08-31
Support special characters in LDAP bind password

tduffield:
This addresses an issue where a valid LDAP password that contained special characters would not work as expected when configured in chef-server.rb.


marcparadise:
:+1: 

It may be worth using this helper function in any of our bits that write passwords to config, since this can affect any service where the user provides a password.  In particular, postgres and rabbitmq can be provided externally, with passwords we have no control over.

tduffield:
@marcparadise I'm all for that, but I think supporting special characters in postgres and rabbitmq will require more work than just adding this helper in the sys config. There are libs that would need to be updated as well. I think coming back to address special characters in those places should be a separate change.

PR #924, opened 2016-09-01
Enable verification of email address when its changed in the profile.

ksubrama:


src/oc-id/app/mailers/email_verify_mailer.rb
```diff
@@ -0,0 +1,25 @@
...
+    @user = user
+    @email = email
+    @expires = 1.day.from_now.to_i
+    @signature = Signature.new(
+      @user.username,
+      email,
+      @expires,
+      Settings.secret_key_base
+    )
+    mail from: Settings.email_from_address,
```
tduffield: I'm not seeing where a body gets set? Am I missing something?

ksubrama: I was initially confused too..  see view/email_verify.text.erb

convention over configuration wheeeee :)

src/oc-id/app/views/email_verify_mailer/email_verify.text.erb
```diff
@@ -0,0 +1,16 @@
+Hello <%= @user.first_name %>,
+
+It looks like you've just tried to change the email address associated with
+your Chef user account. If you did not initiate this request, please ignore
+this email. If you are indeed trying to update your account, follow
+this link to confirm your new email address.
+
+<%= email_profile_url(
```
ksubrama: And in case you are confused reading this...  email_profile_url is a dynamically generated method based on <subroute>_<route>_url(...) because...  magic...

src/oc-id/config/routes.rb
```diff
@@ -16,6 +16,7 @@
 
     resource :profile, only: [:show, :update] do
       put  "password" => "profiles#change_password"
+      put  "email" => "profiles#change_email"
```
ksubrama: This is the bit that we decided to change from a put to a get - just to make accessing a link from an email easier.

src/oc-id/app/controllers/profiles_controller.rb
```diff
@@ -41,6 +49,32 @@ def change_password
   end
 
   #
+  # GET /id/profile/change_email
+  #
+  # Changes the user's email.
+  #
+  # This should really be a PUT but we are using GET to make it easier to access
+  # the end-point from an email without having to dive into javascript.
```
marcparadise: For more context, it might be worth mentioning that the initiating action is a form submit/POST, and that this GET is completion/verification of that action

src/oc-id/spec/controllers/password_resets_controller_spec.rb
```diff
@@ -133,35 +133,42 @@
...
       it 'succeeds when given a valid signature' do
-        get :show, expires: expires, signature: signature, username: name
+        get :show, expires: expires, signature: signature, username: name, email: email
         expect(response).to be_success
       end
 
       describe 'invalid params' do
-        it 'requires a username' do
+        it 'requires an email' do
           get :show, expires: expires, signature: signature
```
marcparadise: Let's include the username here too, so that we're sure it's passing for the right reason.

src/oc-id/spec/controllers/password_resets_controller_spec.rb
```diff
@@ -249,6 +259,35 @@
               expect(response).to render_template('show')
             end
           end
+
+          describe 'email mismatch' do
+            let!(:user) do
```
marcparadise: I didn't know that `let!` was a thing, that's handy.

marcparadise:
:+1: with minor comments.

src/oc-id/spec/controllers/password_resets_controller_spec.rb
ksubrama: It is - it's right below it.  git diff is just being weird :-P

ksubrama: Oh nvm - I see what you mean.  Fixing.

PR #925, opened 2016-09-06
Don't 500 if search attribute is invalid

tduffield:
In the case where a user, when providing a search, provides an attribute
constraint (-a flag) that is invalid, we currently throw a 500. An
invalid attribute constraint most commonly is specifying a nested hash
using dot-notation when one or more values in that constraint are, in
fact, not a hash. This change prevents the system from throwing a 500
and presents the user with a more correct result.

For example, `knife search node "*:*" -a uptime.seconds` is invalid
because the key `'uptime'` does exist but it is a string, not a
hash. Thus, ej attempts to fetch the key `'seconds'` from that string,
throwing an error within ej for an invalid function clause. With this change,
we simply return "ERROR - Attribute does not exist".

The reason there are no tests is that there isn't a great way to test this code path as it is a private function nested inside several helpers. The time it would take to setup mocks to exercise this code path likely isn't worth the time given how easily this is to replicate manually.

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
@@ -360,7 +360,19 @@ make_query_from_params(Req) ->
 extract_path(_Item, []) ->
     null;
 extract_path(Item, Path) ->
-    ej:get(list_to_tuple(Path), Item, null).
+    try ej:get(list_to_tuple(Path), Item, null) of
+        Result -> Result
+    catch
+        %% Don't throw a 500 when the user input is invalid
+        %%   This typically happens when the user provides an invalid attribute
+        %%   key, typically trying to access a value as a hash that is actually
+        %%   a string (i.e. uptime.seconds)
+        error:function_clause ->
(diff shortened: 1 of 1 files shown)


src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
```diff
@@ -360,7 +360,16 @@ make_query_from_params(Req) ->
...
-    ej:get(list_to_tuple(Path), Item, null).
+    try ej:get(list_to_tuple(Path), Item, null) of
+        Result -> Result
+    catch
+        %% Don't throw a 500 when the user input is invalid
+        %%   This typically happens when the user provides an invalid attribute
+        %%   key, typically trying to access a value as a hash that is actually
+        %%   a string (i.e. uptime.seconds)
+        error:function_clause -> null;
+        Error:Reason -> {Error, Reason}
```
marcparadise: If the term that comes back can't be converted as ejson -> json, we'll still end up with a 500 as we prepare the body response.  Maybe a generic error string for the return value in that case, and use `error_logger`to capture the details.

marcparadise:
:+1:

PR #928, opened 2016-09-08
Use recent versions of activesupport

jkeiser:
This is preventive maintenance to avoid us colliding if chef-zero or other deps end up needing activesupport.


marcparadise:
:+1:

markan:
👍

PR #929, opened 2016-09-10
Fix bug where we can update a different user during an email change

ksubrama:
If the logged in user was not the username mentioned in the email-update link, we were accidentally updating the email of the current user instead of the username mentioned in the link.

Further, in this PR, we no longer allow email change links to update the email of accounts whose email was already changed from the one that existed at the time the link was created.

There is also a small fix to redirect accesses to https://id.chef.io/id/profile to the signin page if nobody is signed in already.  Currently, accessing that page within first signing in yields a 500 error page.


marcparadise:
:+1:

PR #930, opened 2016-09-12
Make activesupport dependency less restrictive

rhass:
There is a transitive dependency in fog and chef-provisioning-fog which creates a conflict with ActiveSupport 5.x. This should resolve the conflict by allowing us to pull in the appropriate version wherever possible. Additionally, we are regenerating the Gemfile.lock as the previous version was corrupted.


jkeiser:
:+1:

mwrock:
:+1:

PR #932, opened 2016-09-15
License scout build fix

marcparadise:


marcparadise:
merging under the "it fixes the broken build"  rule

PR #935, opened 2016-09-20
Allow non-pivotal members of server-admins to create orgs

markan:
DO NOT MERGE

This PR is against tc/server-admins-org-perms, as depends on that being present. I'll rebase against master after it merges.

The organization creation endpoint expects the creator to be pivotal;
more specifically it uses the creating user to create many subsidary
objects that we later expect have been created by pivotal (or at least
have the permissions set as if pivotal were the creator)

This enhances the org creation endpoint by using the superuser
(pivotal) to create the objects, even when the originator is not the
superuser.


oc-chef-pedant/spec/api/organization_spec.rb
```diff
@@ -68,7 +69,7 @@
   end
 
   describe "POST /organizations" do
-    let(:orgname) { "test-#{Time.now.to_i}-#{Process.pid}" }
+    let(:orgname) { "test-#{Time.now.to_i}-#{SecureRandom.hex}-#{Process.pid}" }
```
marcparadise: Can we remove the time prefix?

oc-chef-pedant/spec/api/organization_spec.rb
```diff
@@ -91,24 +92,46 @@
...
+        )
+      end
+    end
 
-        it "it rejects the new org as conflicting" do
-          post("#{platform.server}/organizations", superuser, :payload => request_body).should look_like( :status => 409)
-        end
+    context "when a non-superuser in server-admins posts a new organization with a valid body and name" do
+      it "should respond with a valid newly created organization" do
+        require 'pp'; pp server_admin_user: platform.server_admin_user, platform_keys: platform.methods.sort - Object.methods
```
marcparadise: temporary I assume?

stevendanna:
Closing. Please reopen if we want to explore this work in the future.

PR #936, opened 2016-09-21
Re-enable fatal licensing warnings

sersut:
Last week we have disabled fatal licensing warnings because we had a dependency on cpan build service which is slow & unreliable and cause lots of failures during the builds.

Now license_scout does not have a network dependency on the cpan build service anymore. This change picks up the latest license_scout with this functionality and re-enables fatal licensing checks to catch any missing license as our software evolves.

It is tested in wilson (3 new errors popped and they were fixed). 

http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/261/downstreambuildview/

/cc: @chef/chef-server-maintainers @marcparadise


markan:
👍

PR #937, opened 2016-09-21
Update README so changes to oc_reporting get synced correctly.

rmoshier:
https://github.com/chef/chef_index/pull/19

### DIFF
--- a/src/oc_erchef/apps/chef_index/README.md
+++ b/src/oc_erchef/apps/chef_index/README.md
@@ -22,7 +22,7 @@ You will also need to set up the repository with the appropriate branches. We do
 
     http://wiki.chef.io/display/chef/Working+with+git
 
-Once your repository is set up, you can start working on the code.
+Once your repository is set up, you can start working on the code. **If your changes will affect oc_reporting, make sure to sync your changes in the [chef_index](https://github.com/chef/chef_index) repository.**
 
 # LINKS:
 


src/oc_erchef/apps/chef_index/README.md
```diff
@@ -22,7 +22,7 @@ You will also need to set up the repository with the appropriate branches. We do
 
     http://wiki.chef.io/display/chef/Working+with+git
 
-Once your repository is set up, you can start working on the code.
+Once your repository is set up, you can start working on the code. **If your changes will affect oc_reporting, make sure to sync your changes in the [chef_index](https://github.com/chef/chef_index) repository.**
```
tylercloke: The repository you linked to is deprecated. `chef_index` now lives in `oc_erchef` in this repo:

https://github.com/chef/oc_erchef/tree/master/apps/chef_index

rmoshier: Right, so oc_reporting still relies on the deprecated chef_index and we are working on making oc_reporting work with the new chef_backend. To do that we basically had to copy over most of the chef_index stuff in chef_server and update it in the old chef_index repo. I am pointing to the depreciated repo because if anyone needs their changes in chef_server to also work in reporting, that old repo needs to be updated other wise reporting will break. Obviously, this isn't an ideal way of doing this, but I think this is the best we can do with reporting the way it is currently. @tylercloke

tylercloke: Was not aware oc_reporting was using that repo still even though we listed it as deprecated. Should probably update the repo header, but this PR is g2g!

PR #945, opened 2016-09-26
Updates omnibus software to latest

jonsmorrow:
This change updates omnibus software to the latest
version so that we bring in the latest version of
openssl.


marcparadise:
:+1:

PR #946, opened 2016-09-26
Prepare for 12.9.1 release

jonsmorrow:


RELEASE_NOTES.md
```diff
@@ -7,6 +7,14 @@ in the release. For a detailed list of changed components, refer to
 This document contains release notes for the current major release and all patches.
 For prior releases, see [PRIOR\_RELEASE\_NOTES.md](PRIOR_RELEASE_NOTES.md).
 
+## 12.9.1 (2016-9-26)
+
+### Chef Server
+
+* Fix CHEFDK_GECODE_PATH for latest ChefDK
```
marcparadise: We can get rid of these two, they aren't customer-facing

RELEASE_NOTES.md
```diff
@@ -7,6 +7,14 @@ in the release. For a detailed list of changed components, refer to
...
+* Fixed typo
+* Updated openssl
```
marcparadise: If you take a look in PRIOR_RELEASE_NOTES there's a good example of how to break this into a separate ``Security Updates`` section with links to the relevant CVEs

marcparadise: In the 12.9.0 release notes in this file, we call out 12.9.1 as a release that is receiving a fix related to users acls.  Could you update that to 12.9.2 and add a note here that the users ACLs fix is delayed to 12.9.2 so that we can get the security fix out the door?

jonsmorrow: Ok updated and new commit. I hope this is what you want.

PR #948, opened 2016-09-30
Update contributing docs

tas50:


README.md
```diff
@@ -43,22 +39,44 @@ Server.
...
+- Runit for service supervision
 
 If you're looking to contribute to certain parts of the Chef server, familiarity with the following related tools is also beneficial, depending on the area.
 
-* rebar (used for dependency management in Erlang)
-* keepalived (used for HA setups in the server)
-* sqitch (database migrations)
-* lua (routing rules in openresty)
+- rebar (used for dependency management in Erlang)
+- keepalived (used for HA setups in the server)
```
danielsdeleo: Should we purge this?

danielsdeleo:
FWIW, the travis issue is:

```
$ rvm install --binary 2.2.5
Searching for binary rubies, this might take some time.
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
```

Would be 💯 if you could take a second to get that sorted, but I personally wouldn't require for merge.

👍 overall

README.md
tas50: Probably worth nuking

tas50:
Builds should be nice a green show with a small travis change

PR #949, opened 2016-10-03
Install nodejs from binary packages

stevendanna:
Nodejs is a build-time dependency of oc_id. Previously, we were
installing this from source. Now, we install it from upstream-provided
binary packages, avoiding the python2.7 dependency. Further nodejs is
now contained in an easy-to-clean-up folder. We aren't yet cleaning it
up because oc_id's Gemfile setup will need some modification to allow us
to bundle exec without coffee-rails and other gems that require a
javascript runtime.

In terms of diskspace, overall this nodejs installation ends up being
larger than the nodejs built from source, but the increase is offset by
python no longer being needed. We can likely reduce the disk space used
by node further in follow-up efforts.

### DIFF
--- a/omnibus/config/software/oc_id.rb
+++ b/omnibus/config/software/oc_id.rb
@@ -22,7 +22,7 @@ license "Apache-2.0"
 license_file "LICENSE"
 
 dependency "postgresql92" # for libpq
-dependency "nodejs"
+dependency "nodejs-binary"
 dependency "ruby"
 dependency "bundler"
 
@@ -30,6 +30,7 @@ relative_path "oc-id"
 
 build do
   env = with_standard_compiler_flags(with_embedded_path)
+  env['PATH'] = "#{env['PATH']}:#{install_dir}/embedded/nodejs/bin"
 
   bundle "config build.nokogiri --use-system-libraries" \
          " --with-xml2-config=#{install_dir}/embedded/bin/xml2-config" \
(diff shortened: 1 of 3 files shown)


stevendanna:
Depends on: https://github.com/chef/omnibus-software/pull/739

omnibus/Gemfile
```diff
@@ -8,7 +8,7 @@ gem 'bundler', '>1.10'
 # Install omnibus software
 group :omnibus do
   gem 'omnibus', github: 'chef/omnibus'
-  gem 'omnibus-software', github: 'chef/omnibus-software'
+  gem 'omnibus-software', github: 'chef/omnibus-software', branch: 'ssd/nodejs-binary'
```
stevendanna: Reminder: DO NOT MERGE WITH THIS HERE.

charlesjohnson:
I love you, man. Thanks for doing this.

omnibus/config/software/cleanup.rb
```diff
@@ -0,0 +1,8 @@
+name 'cleanup'
+default_version '1.0.0'
+skip_transitive_dependency_licensing true
+license :project_license
+
+build do
+  delete "#{install_dir}/embedded/nodejs"
```
marcparadise: Will this also clean up the downloaded npm packages that are used at build time?

stevendanna: Yes, I believe it does.  I am unable to find any node_modules folder in the final package

stevendanna:
Still have some stuff to work out, it turns out we may need nodejs at runtime afterall. Will ping when it is ready for review.

stevendanna:
@marcparadise Still running it through wilson, but I've updated this to much move to the binary nodejs install but not do the cleanup for now. I think that the cleanup will require reworking the gemfile so it probably makes sense to do that in another PR.

PR #950, opened 2016-10-03
Add cleanup build step

stevendanna:
This is similar to the cleanup done in delivery and the cleanup proposed
in #169. However, we are a bit less aggressive than the approaches
there since there were previously concerns about loosing too much
debugging information.

Currently we:

- Remove the megaco application from the erlang install
- Remove the wx application from the erlang install
- Strip all gecode related shared libraries

While we have avoided stripping executables in the past, the gecode
libraries are very large without stripping.


stevendanna:

stevendanna:
Passing wilson build: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/268/downstreambuildview/

PR #951, opened 2016-10-04
Policy data collector

danielsdeleo:
Adds policyfile pushes and cookbook artifacts to the data collector

### DIFF
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_cookbook_artifact_version.erl
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_cookbook_artifact_version.erl
@@ -35,6 +35,7 @@
          org_id/1,
          type_name/1,
          authz_id/1,
+         identifier/1,
          create_query/1,
          update_query/1,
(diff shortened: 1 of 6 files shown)


danielsdeleo:
http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/269/downstreambuildview/

danielsdeleo:
Updated omnibus build with changes as above: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/274/downstreambuildview/

danielsdeleo:
Chef Server is working correctly with the last commit. Would be working end-to-end if it weren't for this: https://github.com/elastic/elasticsearch/pull/12068 which makes elasticsearch choke on the `solution_dependencies` section of the policyfile data, because it has a structure like `"cookbook (version)": [ [dependency], ... ]` (ex: `"apt (4.0.2)": [["compat_resource", ">= 12.10.0"]]`). Currently automate works around this limitation with log stash config, so nothing further needs to be done here.

PR #953, opened 2016-10-04
Bump omnibus-software

stevendanna:
The newest version of omnibus software builds an erlang without megaco
or wx, so their removal is no longer needed.  Further, depending on
the berkshelf recipe should only depend on 1 version of depselector
now.


stevendanna:

stevendanna:
Wilson passed: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/270/downstreambuildview/

PR #956, opened 2016-10-05
Make loading reporting in dvm easier

ksubrama:


smith:
You need to do the DCO. You can do `git commit --amend -s` and force push to fix it.

PR #958, opened 2016-10-06
Don't include source in erlang releases

stevendanna:
Delivery also uses this config without an issue.  For us, it saves 50-60MB after unpacking (about 10MB in ubuntu deb size).

### DIFF
--- a/src/bookshelf/rebar.config
+++ b/src/bookshelf/rebar.config
@@ -98,7 +98,9 @@
 
 {profiles, [
     {dev, [
-        {relx, [{dev_mode, true}]}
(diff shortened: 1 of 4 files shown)


stevendanna:
Wilson passed here: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/275/downstreambuildview/

danielsdeleo:
I'm pretty ignorant about `relx` but if wilson likes it, 👍 from me. Would be nice to add to Automate.

marcparadise:
:+1:

PR #964, opened 2016-10-11
Do not fail complete reindex when one object fails

smith:
There are cases when we do a reindex where the object name in the
serialized JSON does not match the name we get when selecting `(id, name)` from
the database.

Currently when this lookup fails, the entire reindex crashes. This
change makes it so that if we fail to find the name, instead of failing
entirely, we output a list of objects that failed to reindex to the
user.

img

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/chef_reindex.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/chef_reindex.erl
@@ -44,11 +44,11 @@
 %% reindexing.  Does not drop existing index entries beforehand; this
 %% is explicitly a separate step.
 -spec reindex(Ctx :: chef_db:db_context(),
-              OrgInfo :: org_info()) -> ok.
+              OrgInfo :: org_info()) -> {ok, list()}.
 reindex(Ctx, {OrgId, _OrgName}=OrgInfo) ->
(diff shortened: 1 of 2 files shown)


src/oc_erchef/apps/oc_chef_wm/src/chef_reindex.erl
```diff
@@ -216,12 +215,18 @@ send_to_index_queue(OrgId, Index, [SO|Rest], NameIdDict) ->
...
-    %% indexable.
-    ok = oc_chef_object_db:add_to_solr_async(StubRec, PreliminaryEJson),
-    send_to_index_queue(OrgId, Index, Rest, NameIdDict).
+    NewMissingList = case dict:find(ItemName, NameIdDict) of
+        {ok, ObjectId} ->
+            StubRec = stub_record(Index, OrgId, ObjectId, ItemName, PreliminaryEJson),
+            %% Since we get here via valid `Index', we don't have to check that the objects are
+            %% indexable.
+            oc_chef_object_db:add_to_solr_async(StubRec, PreliminaryEJson),
+            [];
```
marcparadise: This should be the unmodified `MissingList` getting returned - otherwise we empty our existing list

smith: Fixed in latest commit.

PR #970, opened 2016-10-14
Don't delete ctl symlink for amazon linux

srenatus:
Before, the check in there made it _not_ delete the symlink for packages
updates undertaken by RPM. However, amazon linux uses RPM, too, but
doesn't have any /etc/redhat-release.

Now, the test is using extended tests ([[ ... ]]) and uses this to
properly express the condition we want.


srenatus:
@stevendanna :+1: totally

PR #975, opened 2016-10-19
bump redis version ~> 3.0.7

srenatus:
3.0.7 is the latest version of 3.0.x.

[These are the release notes](https://raw.githubusercontent.com/antirez/redis/3.0/00-RELEASENOTES), and they include some high urgency upgrades we've so far missed, using the [default_version 3.0.4](https://github.com/chef/omnibus-software/blob/master/config/software/redis.rb#L24).

### DIFF
--- a/omnibus/config/projects/chef-server.rb
+++ b/omnibus/config/projects/chef-server.rb
@@ -29,6 +29,7 @@ build_iteration 1
 
 override :rabbitmq, version: "3.3.4"
 override :erlang, version: "17.5"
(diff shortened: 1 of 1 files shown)


stevendanna:
@srenatus I'd prefer we change the default version to latest and not pin here.  In general, I think that if we need to pin to get the latest version, the real problem we need to fix is whatever is holding back the default in omnibus-software.

srenatus:
Ok, sure! omnibus-software PR in progress. Closing this.

PR #977, opened 2016-10-20
Pin lua to 5.1

stevendanna:
The openresty lua plugin only supports 5.1.


stevendanna:
Yes, the goal of this is to unblock the build until we find a better solution.  Note that this is only used on ppc, x86 users are using luajit instead.

PR #978, opened 2016-10-21
Adds retry behavior for expander

paulmooring:
Currently expander will drop jobs when solr is unavailable.  This PR adds configurable options to expander for `max_retries` and `retry_wait` to allow expander to retry writing to solr for a configurable number of times.  This allows Solr restarts without dropping tasks.


src/opscode-expander/lib/opscode/expander/solrizer.rb
```diff
@@ -201,22 +201,30 @@ def pointyize_delete
...
           else
             log.error { "Failed to post to solr: #{indexed_object}" }
           end
         end
         http_req.errback do
-          completed
           log.error { "Failed to post to solr (connection error): #{indexed_object}" }
+
+          if retries < max_retries
+            log.info { "Retrying solr connection: #{indexed_object} attempt #{retries}" }
```
paulmooring: I changed the message to

```
Retrying solr connection in RETRY_WAIT seconds: OBJECT attempt ATTEMPT_NUMBER
```

PR #982, opened 2016-10-26
Add validation endpoint to Chef Server.

markan:
Add validation endpoint to Chef Server.

We sometimes have the need to be able to determine if a request signed
with a client key is valid.  This adds an endpoint
/organizations/ORGNAME/validate/PATH that accepts a forwarded post,
and checks the signing as if it were sent to the PATH resource (PATH
can be any resource string). If the request validates, it returns 200,
and some potentially useful data. If it doesn't it returns 401 like
any other request with a bad signature.
- [x] Verify it works with the nginx code
- [x] Add RAML docs
- [x] Wait for chef server release to finish.
- [x] Rebase/Squash

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_validate.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_validate.erl
@@ -1,0 +1,113 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil; fill-column: 92 -*-
+%% ex: ts=4 sw=4 et
+%% @author Mark Anderson <mark@chef.io>
+%% Copyright 2016 Chef Software, Inc. All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+
+-module(chef_wm_validate).
+
+-include("oc_chef_wm.hrl").
+
+
+%% Webmachine resource callbacks
+-mixin([{oc_chef_wm_base, [content_types_provided/2,
+                           content_types_accepted/2,
+                           finish_request/2,
+                           malformed_request/2,
+                           ping/2,
+                           forbidden/2,
+                           {allow_all/2, auth_info},
+                           service_available/2]}]).
+
+-export([allowed_methods/2,
+         to_json/2,
+         from_json/2]).
+
+%% chef_wm behavior callbacks
+-behaviour(chef_wm).
+-export([init/1,
+         init_resource_state/1,
+         malformed_request_message/3,
+         is_authorized/2,
+         request_type/0,
+         validate_request/3,
+         delete_resource/2,
+         delete_completed/2,
+         process_post/2
+        ]).
+
+
+init(Config) ->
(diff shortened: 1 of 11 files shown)


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -987,7 +999,8 @@ verify_request_signature(Req,
             PublicKey = select_user_or_webui_key(Req, Requestors),
             Body = body_or_default(Req, <<>>),
             HTTPMethod = method_as_binary(Req),
-            Path = iolist_to_binary(wrq:path(Req)),
+            Path = Extractor(path, Req, State),
+            io:format("reqsig: ~p~n", [Path]),
```
stevendanna: Let's use `lager` if we want to keep this in.

stevendanna:
Overall I like this idea as I've wanted this feature for a while so that add-ons don't need to do the "go to the principals endpoint, then do the auth yourself" dance.  A few high level questions and comments:

- Do we have a planned use for this currently or is this just scratching an itch?
- If we have a planned use, did we consider whether to implement this RFC instead: https://github.com/chef/chef-rfc/blob/master/rfc038-token-authentication.md

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_validate.erl
```diff
@@ -0,0 +1,94 @@
...
+
+process_post(Req, State) ->
+    {Result, Req1, State1} = to_json(Req, State),
+    case Result of
+        {halt, _} -> {Result, Req1, State1};
+        _ -> {true, wrq:set_resp_body(Result, Req1), State1}
+    end.
+
+to_json(Req, #base_state{requestor = #chef_requestor{name = Name, type = Type}} = State) ->
+    %% return org, client/user auth'd as.
```
stevendanna: Doesn't appear that we are returning the org.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -414,10 +416,18 @@ check_permission(Perm, AuthzObjectType, AuthzId, Req, #base_state{requestor_id=R
...
+
+is_authorized(Req, State) ->
+    is_authorized(Req, State, fun authorization_data_extractor/3).
+
 %% part of being authorized is being a member of the org; otherwise we
 %% fail out early.
-is_authorized(Req, State) ->
-    case verify_request_signature(Req, State) of
+is_authorized(Req, State, Extractor) ->
+    case verify_request_signature(Req, State, Extractor) of
```
stevendanna: I don't see a verify_request_signature/3 function yet

stevendanna:
Passing Wilson build: http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/288/downstreambuildview/

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_validate.erl
```diff
@@ -0,0 +1,113 @@
...
+    {true, Req2, State1}.
+
+to_json(Req, State) ->
+    make_body(Req, State).
+
+from_json(Req, State) ->
+    make_and_set_body(Req, State).
+
+validate_extractor(path, Req, _State) ->
+    %% wrq:disp_path/1 returns the path portion matched by "*", ie.
```
stevendanna: 👍 Nice comment, had to look this up the first time I read this.

stevendanna:
As a follow up, I just remembered that we now have raml docs for API endpoints, perhaps we should add some docs for this endpoint once we are done with the nginx-side feature?

srenatus:
@stevendanna It would make sense, yes. However, we should probably discuss if this endpoint is ever going to be outside facing, or if it's just an internal thing to nginx offloading signed header authn for other services... So we'd maybe rather just document the external API endpoint that is going to be protected by this? 🤔

srenatus:
LGTM?

srenatus:
Rebased and squashed... ✅

PR #984, opened 2016-10-27
Security features for /data-collector/ endpoint

sersut:
Signed-off-by: Serdar Sutay serdarsutay@gmail.com

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
@@ -73,6 +73,12 @@ http {
   }
   <%- end -%>
 
+  <% if node['private_chef']['data_collector']['root_url'] && node['private_chef']['data_collector']['token'] -%>
+  upstream data-collector {
+    server <%= URI.parse(node['private_chef']['data_collector']['root_url']).host %>:<%= URI.parse(node['private_chef']['data_collector']['root_url']).port %>;
+  }
+  <% end -%>
+
   # Include upstream definitions for addons
   include <%= node['private_chef']['nginx']['dir'] %>/etc/addon.d/*_upstreams.conf;
 
@@ -169,5 +175,3 @@ http {
 
   include <%= node['private_chef']['nginx']['dir'] %>/etc/nginx.d/*.conf;
 }
-
-
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
@@ -55,6 +55,15 @@
       proxy_pass <%= node['private_chef']['data_collector']['root_url'] %>;
     }
     <% end -%>
+
+    <% if node['private_chef']['data_collector']['root_url'] && node['private_chef']['data_collector']['token'] -%>
+    location ~ /organizations/([^/]+)/data-collector {
+      set $request_org $1;
+      access_by_lua_file '<%= @script_path %>/check_token_using_endpoint.lua';
+      proxy_pass $upstream;
+    }
+    <% end -%>
+
     # bookshelf
     <% if node['private_chef']['opscode-erchef']['nginx_bookshelf_caching'] != :off -%>
     location ~ "/<%= node['private_chef']['opscode-erchef']['s3_bucket'] %>/{0,1}.*$" {
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
@@ -1,0 +1,24 @@
+ngx.req.read_body()
+local res
+
+if ngx.var.request_method == "POST" then
+  meth = ngx.HTTP_POST
+else
+  meth = ngx.HTTP_GET
+end
+
+res = ngx.location.capture("/organizations/" .. ngx.var.request_org .. "/validate" .. ngx.var.request_uri,
+    { method = meth,
+      always_forward_body = true,
+      copy_all_vars = true,
+      vars = { request_org = ngx.var.request_org, upstream = "http://opscode_erchef" }
+    })
+
+if res.status == ngx.HTTP_OK then
+  ngx.var.upstream = '<%= URI.parse(node['private_chef']['data_collector']['root_url']).scheme %>://data-collector<%= URI.parse(node['private_chef']['data_collector']['root_url']).path %>'
+  ngx.req.set_header("x-data-collector-token", "<%= node['private_chef']['data_collector']['token'] %>")
+  ngx.req.set_header("x-data-collector-auth", "version=1.0")
+  return
+end
+
+ngx.exit(res.status)
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
@@ -1,24 +1,24 @@
 ngx.req.read_body()
 local res
 
-if ngx.var.request_method == "POST" then
-  meth = ngx.HTTP_POST
-else
-  meth = ngx.HTTP_GET
+if ngx.var.request_method ~= "POST" then
+  ngx.exit(ngx.HTTP_NOT_ALLOWED)
 end
 
 res = ngx.location.capture("/organizations/" .. ngx.var.request_org .. "/validate" .. ngx.var.request_uri,
-    { method = meth,
+    { method = ngx.HTTP_POST,
       always_forward_body = true,
       copy_all_vars = true,
-      vars = { request_org = ngx.var.request_org, upstream = "http://opscode_erchef" }
+      vars = { upstream = "http://opscode_erchef" }
     })
 
+<% if node['private_chef']['data_collector']['root_url'] && node['private_chef']['data_collector']['token'] -%>
 if res.status == ngx.HTTP_OK then
   ngx.var.upstream = '<%= URI.parse(node['private_chef']['data_collector']['root_url']).scheme %>://data-collector<%= URI.parse(node['private_chef']['data_collector']['root_url']).path %>'
   ngx.req.set_header("x-data-collector-token", "<%= node['private_chef']['data_collector']['token'] %>")
   ngx.req.set_header("x-data-collector-auth", "version=1.0")
   return
 end
+<% end -%>
 
 ngx.exit(res.status)
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
@@ -1,11 +1,12 @@
+-- ngx.req.read_body() is recommended when using ngx.location.capture()
+-- Also we need to pass the body of the request to the /validate call.
 ngx.req.read_body()
-local res
 
 if ngx.var.request_method ~= "POST" then
   ngx.exit(ngx.HTTP_NOT_ALLOWED)
 end
 
-res = ngx.location.capture("/organizations/" .. ngx.var.request_org .. "/validate" .. ngx.var.request_uri,
+local res = ngx.location.capture("/organizations/" .. ngx.var.request_org .. "/validate" .. ngx.var.request_uri,
     { method = ngx.HTTP_POST,
       always_forward_body = true,
       copy_all_vars = true,
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
@@ -115,7 +115,8 @@ lbconf = node['private_chef']['lb'].to_hash.merge(nginx_vars).merge(
  'resolver.lua',
  'route_checks.lua',
  'routes.lua',
- 'dispatch_route.lua'].each do |script|
+ 'dispatch_route.lua',
+ 'check_token_using_endpoint.lua'].each do |script|
   template File.join(nginx_scripts_dir, script) do
     source "nginx/scripts/#{script}.erb"
     owner 'root'


ryancragun:
I feel like this should just be added to the data collector proxy route, since that's what all existing clients will use.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -45,8 +45,30 @@
...
-    <% if node['private_chef']['data_collector']['proxy'] -%>
+    <% if node['private_chef']['data_collector']['root_url'] && node['private_chef']['data_collector']['token']-%>
     location "/data-collector/" {
+      access_by_lua '
+        ngx.req.read_body()
+        local token = "<%= node['private_chef']['data_collector']['token'] %>"
+
+        if ngx.var.request_method ~= "POST"
+        then
+          ngx.exit(ngx.HTTP_NOT_FOUND)
```
srenatus: nitpicking, but this could also be 405, i.e., `ngx.HTTP_NOT_ALLOWED`

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -45,8 +45,30 @@
...
+        end
+
+        -- TODO: auth with /validate_request endpoint
+
+        local h = ngx.req.get_headers()
+
+        if h["x-data-collector-token"] == nil
```
srenatus: Could this lead to weird issues? For example, when the client sends a x-data-collector-token header with an invalid token, and [the check](https://github.com/chef/chef-server/pull/984/files#diff-e61ee1a4529c1a04df73e9febe4bbe70R59) passes, the request would still fail (because of that wrong token in the initial request). I'm not sure if there's a valid use case behind _not overriding the client's headers_ here, so I'd propose we just unconditionally set them (if the check passes, of course), so there's one possibility less to worry about when debugging data collector related issues...

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -45,8 +45,30 @@
       alias /opt/opscode/version-manifest.txt;
     }
 
-    <% if node['private_chef']['data_collector']['proxy'] -%>
+    <% if node['private_chef']['data_collector']['root_url'] && node['private_chef']['data_collector']['token']-%>
```
stevendanna: The old configuration value needs to work still or this will break people using the proxying feature.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
srenatus: Thinking about this, it's probably needed for backwards-compat (as @stevendanna pointed out [here](https://github.com/chef/chef-server/pull/984/files#diff-e61ee1a4529c1a04df73e9febe4bbe70R48)):

If the condition was changed to
```
<% if node['private_chef']['data_collector']['proxy'] || (node['private_chef']['data_collector']['root_url'] && node['private_chef']['data_collector']['token']) -%>
```
and the `TODO` block is only used when _the header is not already present_, this could work for both old setups and new ones. However, we'll ask for the trouble hinted at in my previous comment ^

srenatus:
I'm afraid this is not using the right place for that logic:

* Before (i.e., old clients) it was using `/data-collector` when the corresponding node attr was set (`node['private_chef']['data_collector']['proxy']`)
* Now (new clients), [we need this to be org-scoped](https://github.com/chef/chef/pull/5487/files#diff-44fe5ab727624aef8691893516fb199aR1136): `/organization/ORG/data-collector`

Thus we can gather the org from the request path, maybe like this:

```
location ~ /organization/([^/]+)/data-collector {
    set $request_org $1;
    rewrite ^/organization/(?:.*)/data-collector$ /data-collector break;
    access_by_lua_file 'check_token_using_endpoint.lua'; # needs to be put someplace
    proxy_pass http://visibility; # right value here
}
```

and use this to forward the proper request to chef-server... (this is `check_token_using_endpoint.lua`):

```
ngx.req.read_body()
local res
if ngx.var.request_method == "POST" then
  meth = ngx.HTTP_POST
else
  meth = ngx.HTTP_GET
end
res = ngx.location.capture("/organization/" .. ngx.var.request_org .. "/validate" .. ngx.var.request_uri,
    { method = meth,
      always_forward_body = true,
      vars = { request_org = ngx.var.request_org }
    })

if res.status == ngx.HTTP_OK then
  ngx.log(ngx.CRIT, res.body)
  ngx.req.set_header("x-token-response", res.body)
  return
end
ngx.exit(res.status)
```

Note that this sends a the validation (sub)request to chef-server against `/organization/secured/validate/organization/secured/data-collector`. This seems redundant, but we better keep the `/organization/secured/data-collector` part since it's what's been signed...

sersut:
@srenatus @stevendanna thanks for checking this out. Updated per your comments. Would appreciate another 👀 into this. 

lua side got a bit complicated because of the fact that we are making the `/validate` call against erchef but the actual POST to a potentially different upstream. @srenatus if you have any ideas to make the lua side more clean, feel free to update.

/cc: @danielsdeleo @ryancragun

ryancragun:
👍 💯 🎉

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
```diff
@@ -73,6 +73,12 @@ http {
   }
   <%- end -%>
 
+  <% if node['private_chef']['data_collector']['root_url'] && node['private_chef']['data_collector']['token'] -%>
+  upstream data-collector {
+    server <%= URI.parse(node['private_chef']['data_collector']['root_url']).host %>:<%= URI.parse(node['private_chef']['data_collector']['root_url']).port %>;
```
srenatus: This would mean multiple Automate (Visibility) instances to be proxied by one chef-server? I doubt it.... 🤔 

We'd have to have some more logic in there then -- to which one should the request be routed...?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -55,6 +55,15 @@
       proxy_pass <%= node['private_chef']['data_collector']['root_url'] %>;
     }
     <% end -%>
+
+    <% if node['private_chef']['data_collector']['root_url'] && node['private_chef']['data_collector']['token'] -%>
+    location ~ /organizations/([^/]+)/data-collector {
+      set $request_org $1;
+      access_by_lua_file '<%= @script_path %>/check_token_using_endpoint.lua';
+      proxy_pass $upstream;
```
srenatus: This is a neat idea: having the lua script decide which upstream is the right one. :+1:

However, since it's currently hard-coded to data-collector-ish stuff only anyways, can't we put [that logic](https://github.com/chef/chef-server/pull/984/files/bb0efc646b250eac89d39eb7010265d7b7dbc5e0#diff-9376b5038260ef7c52869f5018a6feceR18) here?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
```diff
@@ -0,0 +1,24 @@
+ngx.req.read_body()
+local res
+
+if ngx.var.request_method == "POST" then
```
srenatus: Currently, this is not generic anyways, so do we need GET or is data-collector only accepting POSTs anyways?

I'm just saying that we might want to go either way:

1. making this truely generic (e.g. adding HEAD, OPTION, DELETE)
2. baking in data-collector specific assumptions

I'd lean towards (2.) now and (1.) later. (Creating something, maybe a proper lua module, that is tested properly and used in various places might make much sense.)

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
```diff
@@ -0,0 +1,24 @@
...
+  meth = ngx.HTTP_POST
+else
+  meth = ngx.HTTP_GET
```
stevendanna: Why are we changing the method to GET here?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
```diff
@@ -0,0 +1,24 @@
...
+end
+
+res = ngx.location.capture("/organizations/" .. ngx.var.request_org .. "/validate" .. ngx.var.request_uri,
+    { method = meth,
+      always_forward_body = true,
+      copy_all_vars = true,
```
srenatus: From my understanding, if we have `copy_all_vars = true`, the next line is superfluous, isn't it?

> `vars` take a Lua table which holds the values to set the specified Nginx variables in the subrequest as this option's value.
> `copy_all_vars` specify whether to copy over all the Nginx variable values of the current request to the subrequest in question. modifications of the nginx variables in the subrequest will not affect the current (parent) request. ([src](https://github.com/openresty/lua-nginx-module#ngxlocationcapture))

Since `upstream` is in `vars`, I don't think we need this line. (But see my other comment, maybe we don't even need upstream.)

We also don't need `request_org` here, since it's passed to the validate endpoint as part of the scoped URL. (This was a relict from trying to get this work in the mocked script/config used with Test::Nginx.)

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
srenatus: When I was playing with this, I had an issue with `proxy_pass http://something/data-collector` and the regexp-match in `location ~ /organizations/([^/]+)/data-collector`. I had worked around this with something that would _here_ look like this:

```
rewrite ^/organization/(?:.*)/data-collector$ <%= URI.parse(node['private_chef']['data_collector']['root_url']).path %> break;
proxy_pass <%= URI.parse(node['private_chef']['data_collector']['root_url']).scheme %>://data-collector
```

I've just gotten the impression that the `$upstream` trick might have addressed the same issue. This would be a more localised way around it, maybe? (The less _huh?_ the better...)

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
```diff
@@ -0,0 +1,24 @@
...
+      vars = { request_org = ngx.var.request_org, upstream = "http://opscode_erchef" }
+    })
+
+if res.status == ngx.HTTP_OK then
+  ngx.var.upstream = '<%= URI.parse(node['private_chef']['data_collector']['root_url']).scheme %>://data-collector<%= URI.parse(node['private_chef']['data_collector']['root_url']).path %>'
+  ngx.req.set_header("x-data-collector-token", "<%= node['private_chef']['data_collector']['token'] %>")
+  ngx.req.set_header("x-data-collector-auth", "version=1.0")
+  return
+end
```
stevendanna: Should we have an else clause here where we log what happened or does nginx already handle that well for us?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
srenatus: Scratch that comment about `copy_all_vars`, it's necessary (adam leff pointed out that this includes the original headers, of course we need those!). But `upstream` and `request_org` could still be unnecessary.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
sersut: We will get the `res.status` as the nginx response @stevendanna but we will not have extra details logged. However we should be able to find more info from the erchef logs. 

I feel pretty good with the current level but it's possible to add an additional log with the body or the response from `/validate`.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
sersut: `upstream` is necessary for sure but request_org is probably not necessary. I will remove that. 👍

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
danielsdeleo: Ok, I just wanted to be extra certain about this, since we would have to change the user-facing config from a string to an array which would make upgrading painful.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
srenatus: This is just replacing the String with the symbol IIRC

sersut: The idea is to be able to have a generic way of doing `/validate` that can handle other types of requests as well. This generic way is planned to be used by other features (e.g. compliance, etc..) in the future.

danielsdeleo:
@stevendanna do you know any more about the person who contributed the existing proxy stuff? Would they be happier with this implementation?

What I'm getting at is, should we deprecate the old one?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
sersut: Copy https://github.com/chef/chef-server/pull/984#discussion_r85969687 ...

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
sersut: Fully agreed. I will remove the `GET` path for now. We can add that later on when we actually need it.

marcparadise:
I have some comments incoming shortly, the gist of which is that we can fit this into the existing routing pretty well without needing to add reg-ex'd endpoints.Details to be posted shortly.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
stevendanna: Sorry, I'm not completely understanding the responses to this.  Given this code, what will happen if the method is DELETE and how does that fit with the goal of it being generic.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
marcparadise: If `token` were set but no `root_url` were set, nginx woudl fail to start. Maybe modify the conditional above to check for both?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
```diff
@@ -0,0 +1,24 @@
+ngx.req.read_body()
+local res
```
sersut: 👍

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint.lua.erb
```diff
@@ -0,0 +1,24 @@
...
+
+if ngx.var.request_method ~= "POST" then
+  ngx.exit(ngx.HTTP_NOT_ALLOWED)
+end
+
+res = ngx.location.capture("/organizations/" .. ngx.var.request_org .. "/validate" .. ngx.var.request_uri,
+    { method = ngx.HTTP_POST,
+      always_forward_body = true,
+      copy_all_vars = true,
+      vars = { upstream = "http://opscode_erchef" }
```
marcparadise: This works, thanks for the context.

sersut: 👍

PR #985, opened 2016-10-28
omnibus: Add arm build support for opscode-solr4

elthariel:
Following the merge of https://github.com/chef/omnibus-software/pull/751, this PR use the new `jre-from-jdk` software for opscode-solr4, allowing to build chef-server package for armhf, as demonstrated here: https://github.com/elthariel/chef-server/releases/tag/v12.5.0-arm-alpha

I've been running the chef server on a very small box (Scaleway.com C1 instance: 2GB RAM, Marvell Armada 370/XP CPU) for a few weeks, and it works well. For later reference, here's my `/etc/opscode/chef-server.rb` file:

```
# This line is *required*, without it, the keygen timeouts and starts new 'openssl genrsa' processes
# Completely slamming the box, preventin `reconfigure` from ever succeeding
opscode_erchef['keygen_timeout'] = 30000
# Tweak to your tastes
opscode_expander['nodes'] = 3
opscode_erchef['keygen_cache_workers'] = 1
# There's only 2GB of memory on my box, let's be conservative here
opscode_solr4['heap_size'] = 256
postgresql['shared_buffers'] = '256MB'
postgresql['shmall'] = 2097152
postgresql['shmmax'] = 4294967295
```

Signed-off-by: Julien 'Lta' BALLET contact@lta.io

### DIFF
--- a/omnibus/Gemfile.lock
+++ b/omnibus/Gemfile.lock
@@ -8,7 +8,7 @@ GIT
 
 GIT
   remote: git://github.com/chef/omnibus-software.git
-  revision: e72cafb03e6a7ee38f78dfda9feb9d683c60901a
+  revision: f1aa0f05e01438f3d6e3bc73f6fe60f105d466db
   specs:
     omnibus-software (4.0.0)
       chef-sugar (>= 3.4.0)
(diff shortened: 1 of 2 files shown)


elthariel:
By the way, this build is likely to work OOB on Raspberry Pi, but I havent' access to any right now to test

stevendanna:
If you do try to get this going on a Raspberry Pi, you may want to generate the certs and dhparam before hand and feed them in using these config items:

https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb#L477-L479

Otherwise the reconfigure might take forever.

elthariel:
I confirm, it does take forever.

elthariel:
Restarting the travis build. It fails in seemingly unrelated places, are some of the tests brittle ?

stevendanna:
@elthariel We've been seeing some issues with Travis with respect to installing the version of ruby some of the tests require.  If they fail again I'll try to restart the individual tests.  This PR should't influence any of the Travis tests at all.  I'll put this branch through our internal build system which runs the build and a set of end-to-end tests on all supported platforms.  Once it passes there, I think we are good to go.

elthariel:
@stevendanna, thanks for the update, I hope the tests will pass but there isn't any reasons for them not to pass. Also I closed/reopened the PR to restart the build and it passed, as you can see.

PR #988, opened 2016-10-28
Make the ElasticSearch shard and replica count configurable

irvingpop:
The default values for ElasticSearch's shard count (5) and replica count (1) may not be best for all customer topologies (3 nodes).    This PR makes them configurable, but reserves opinions on what our default values should be. 

Further reading:  https://www.elastic.co/guide/en/elasticsearch/guide/current/_scale_horizontally.html

Signed-off-by: Irving Popovetsky irving@getchef.com


stevendanna:
@irvingpop Have any time to chat about this.  This particular PR looks fine but I'm wondering if we need some chef-backend features with respect to changing the replica count config as nodes join and leave the cluster.

irvingpop:
hey @stevendanna I haven't done it with a build incorporating these changes yet, but I've been making extensive use of this bit of code to pre-create the `chef` index using a custom shard and replica count (based on what our bootstrap does):  https://github.com/irvingpop/cloudformation_automate/blob/master/backendless_chef.yaml#L413-L470

PR #994, opened 2016-11-02
Simplify MAINTAINERS.md

stevendanna:
Remove the components breakdown. This is a small group, we know who is
good at what, and trust each other to consider our own limitations when
reviewing code.


stevendanna:
@chef/chef-server-maintainers We should probably all vote on this change.

stevendanna:
I've gotten confirmation in chat from most people on this list.

PR #995, opened 2016-11-02
Update runit cookbook to a current version

marcparadise:


stevendanna:
We should definitely run this through wilson before merging.  The runit cookbook has had a TON of changes between those versions I think.

marcparadise:
In exciting news, our first wilson run has already failed! (Licensing)

tas50:
1.6 -> 1.7 was a complete rewrite of the LWRP. After that it's mostly just deprecating support for old platforms.

marcparadise:
superceded by https://github.com/chef/chef-server/pull/1200

PR #996, opened 2016-11-02
Add retries to rvm install in travis

stevendanna:
This has been failing a lot lately, maybe retrying will help.


stevendanna:
I was also considering caching the `.rvm` folder but that explodes the cache to 1GB.  Hopefully the retries will make the tests a bit less flakey.

PR #998, opened 2016-11-03
Sqitch is available via apt in TravisCI

stevendanna:


stevendanna:
It turns out installing debian sid on top of ubuntu isn't so easy after all. :(

PR #999, opened 2016-11-03
Add profiles forwarding to Automate endpoint

alexpop:
Similar to https://github.com/chef/chef-server/pull/984 but for Compliance Profiles retrieval from Chef Automate.

Once connected with Automate, it can be tested using:
```
[07:01:53 ~/chef-repo]$
$ knife raw /owners/admin/compliance/linux/tar
```

or via the `audit` 2.0 cookbook.


stevendanna:
> Let's circle back around and follow up after we get this in for the short term.

How big is the task to just do it now?  In my experience, we never circle back.

marcparadise:
I'm OK with this, subject to the same limitation as https://github.com/chef/chef-server/pull/984 - let's come back to it in the very near term and work it into the existing routing framework so that we don't need to maintain  multiple paths to pre-validate a request before sending it to an upstream.

srenatus:
Looks merge'able, then, doesn't it? We've scheduled a discussion around improving this on Wednesday (please holler if you'd like to join but didn't get the invite); and this is on par with the other endpoint atm...

Issue #675, opened 2015-12-17
LDAP user name mojibake


takkawai:
https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl#L176

About this line, when the value of eladp_entry includes multibyte characters, it will be mojibake (corrupted).
(1) Could you fix it not to be mojibake?
(2) I think that this code uses "displayname" or "display_name" attribute of LDAP.
Can we use "displayNamePrintable" in stead of it?

echohack:
To +1 what @takkawai is saying here: When you have an Active Directory user that is populated into Chef via LDAP(s) that has Japanese characters (mojibake) it causes the username to become corrupted. It appears that Chef Server is using US-ascii encoding, and not UTF-8 to translate these user names.

@takkawai Can you provide an example (screenshot) of the problem to help highlight this issue?

PR #1001, opened 2016-11-05
[ldap] Allow user customization of field mapping

### LINKED ISSUE #104: Mapping Capability Between LDAP and Chef Server 12 Attributes
As a customer, I would like the Chef Server 12 LDAP module to be able to map between attributes appearing in an LDAP user record, and the required attributes like "mail:"

For example, having the Chef Server 12 LDAP module understand a mapping like this between the required attribute "mail:" and the attribute that I have in my LDAP user record "UserPrincipalName:" would be extremely helpful

```
mail: -> UserPrincipalName:
```

Making this capability available for all required fields would be ideal.

Zendesk 3083

### LINKED ISSUE #151: LDAP: Synthesize a displayname from surname, givenname
As a customer, I would like the chef server LDAP module to create a `displayname` field by copying the `name` attribute, or concatenating the `givenname` and `surname` attributes together with a space.

For example, my LDAP record might have this, but no `displayname`. Name is a good substitute in this case.

```
name: Sean Horn
```

Alternatively, my LDAP record might have this and no `displayname`

```
sname: Horn
givenname: Sean
```

The Chef server's ldap module should just concatenate those values with a space into `displayname`and create the chef server user record using that.

If neither of these avenues are taken by the chef server against an LDAP server lacking `displayname` attributes for users, their name shows up as `"unknown"` in the Manage interface.

### LINKED ISSUE #800: Allow setting LDAP "displayName" attribute in Chef Server
In our LDAP schema, the user's "name" isn't `displayName` (like in Active Directory) but rather `cn`.

When using the Management console, it shows the user as "unknown" in the upper-right.

I think this is code in question: https://github.com/chef/chef-server/blob/bb28b489960c8fae6ac061bf2dab5800142b22a3/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl#L200

In my experience, `displayName` is Active Directory only and everyone else uses `cn` (common name).

stevendanna:
Attributes from a user's LDAP record are used during account-linking to
populate the erchef user record when it is created. Previously, the
mapping between LDAP attributes and chef user attributes were fixed.
Now, they are configurable.

For example, if the user's LDAP record stores their email address in a
field named 'address' instead of 'mail', then you could set the
following in private-chef.rb:

    ldap['email_attribute'] = "address"

Fixes #151
Fixes #800
Fixes #104
Partially addresses #675

Issue #800 was also addressed in #863 which allowed common_name to
service as a fallback for display name. The fallback is still in place
but now any field can be used for the display_name.

Issue #675 is an issue which our unicode handling. The unicode handling
is still broken; however, this would allow users to use a different
field that might not contain multi-byte characters.

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl
@@ -110,14 +110,15 @@ find_and_authenticate_user(Session, User, Password, Config) ->
                  bind(Session, BindDN, BindPass)
          end,
 
-    % And then search
+    %% And then search
     {ok, Result} = search_result(eldap:search(Session, [Base, Filter])),
     case result_to_user_ejson(LoginAttr, User, Result) of
         {error, Reason} ->
             {error, Reason};
         {CN, UserName, Data} ->
-            % We found the user identified by username, now we need to
-            % see if we can authorize as that user, using the provided password.
+            %% We found the user identified by username, now we need
+            %% to see if we can authorize as that user, using the
+            %% provided password.
             case eldap:simple_bind(Session, CN, Password) of
                 ok -> {UserName, Data};
                 {error, Error} ->
@@ -129,8 +130,9 @@ find_and_authenticate_user(Session, User, Password, Config) ->
 search_result({ok, {eldap_search_result, Result, _}}) ->
     {ok, Result};
 search_result({error, Reason}) ->
-    % An error response means some kind of failure occurred - no matching results
-    % would not result in an error tuple, but rather an empty result set.
+    %% An error response means some kind of failure occurred - no
+    %% matching results would not result in an error tuple, but rather
+    %% an empty result set.
     lager:error("LDAP search failed unexpectedly: ~p", [Reason]),
     error.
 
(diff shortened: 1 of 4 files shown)


sean-horn:
:thumbsup: LGTM

stevendanna:
Rebased this against master.  @chef/chef-server-maintainers

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl
```diff
@@ -197,20 +201,27 @@ result_to_user_ejson(LoginAttr, UserName, [{eldap_entry, CN, DataIn}|_]) ->
...
+    Terms = [ {Name, value_of(Key, Data, "unknown") } || {Key, Name} <- ldap_attribute_map() ],
     Result = Terms ++ [ { <<"username">>, CanonicalUserName },
                         { <<"external_authentication_uid">>, UserName },
                         { <<"recovery_authentication_enabled">>, false } ],
     {CN, CanonicalUserName, {Result}}.
 
+
+ldap_attribute_map() ->
+    Config = envy:get(oc_chef_wm, ldap, list),
+    Map = [{proplists:get_value(display_name_attribute, Config, "displayname"), <<"display_name">>},
```
srenatus: It's a little curious to have the defaults be there twice... or am I missing something? We shouldn't end up with a config that doesn't have at least the default values in it... (But it doesn't hurt, I guess)

stevendanna: I agree it is a bit sub-optimal. I was mostly thinking of a case where someone has upgraded the package but hasn't run `reconfigure` yet.  This isn't in general case we support but it seemed easy enough to support in this case.

PR #1002, opened 2016-11-05
[ldap] Don't mangle multi-byte characters

### LINKED ISSUE #675: LDAP user name mojibake

stevendanna:
Previously, if a user's ldap data contained multi-byte characters, we
would mangle it.

For example, if a user's displayname was "张伟" it would get returned as
"å¼ ä¼".  This was being caused by the call to characters_to_binary.

The UTF-8 representation of "张伟" is

    229,188,160,228,188,159

2 characters, each 3 bytes. unicode:characters_to_binary would transform
this into:

    <<195,165,194,188,194,160,195,164,194,188,194,159>>

It does this because characters_to_binary assumes that the input is a
list of unicode code points. In our example, it would expect

    [24352,20255]

and would then return

    <<229,188,160,228,188,159>>

Since we are sending it raw UTF-8 data rather than code points, our data
gets mangled.

Since we are already getting UTF-8 encoded data from eldap, it appears
that we can just forward it on without modification and it will display
correctly.

Fixes #675

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_authn_ldap.erl
@@ -231,9 +231,10 @@ value_of(Key, Data, Default) ->
     [R|_] = proplists:get_value(Key, Data, [Default]),
     characters_to_binary(R).
 
-characters_to_binary(Characters) when is_list(Characters);
-                                      is_binary(Characters) ->
-    unicode:characters_to_binary(Characters);
+characters_to_binary(Characters) when is_list(Characters) ->
+    list_to_binary(Characters);
+characters_to_binary(Characters) when is_binary(Characters) ->
+    Characters;
 % In case of unexpected value, don't crash the auth process:
 characters_to_binary(Atom) when is_atom(Atom) ->
     atom_to_binary(Atom, utf8).


stevendanna:
This is built on top of https://github.com/chef/chef-server/pull/1001, will rebase once that is merged.

stevendanna:
See https://github.com/chef/chef-server/pull/1002/commits/6e86e8045a491a5460b2c9af2a1e0d3404799c87 for the relevant change.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
```diff
@@ -89,6 +89,15 @@
             {base_dn, "<%= node['private_chef']['ldap']['base_dn'] || "" %>" },
             {group_dn, "<%= node['private_chef']['ldap']['group_dn'] || "" %>" },
             {login_attribute, "<%= node['private_chef']['ldap']['login_attribute'] || "samaccountname" %>" },
+            %% LDAP Attribute Mappings
+            {display_name_attribute, "<%= node['private_chef']['ldap']['display_name_attribute'] || 'displayname' %>" },
```
smith: Why are we using `||`s here instead of just setting the defaults in the attributes?

stevendanna: This is a good question that I also asked myself.  It is because the entire ldap tree is nil when ldap is disabled so we can't set defaults in the attribute file.  Well, we probably good, I was just a bit lazy and did't want to wrap my head around solving that issue in a backwards compatible way.

stevendanna:
Rebased this. cc @chef/chef-server-maintainers The relevant change is: https://github.com/chef/chef-server/pull/1002/commits/8ddf9795118d720accd3a458eaba456d44e99a09

PR #1003, opened 2016-11-06
a wip story

marcparadise:


src/bookshelf/src/bksw_io_names.erl
```diff
@@ -61,7 +61,7 @@ sha_to_hex_str(<<SHA:160/big-unsigned-integer>>) ->
 -spec write_path(string() | binary(), string() | binary()) -> binary().
 write_path(Bucket, Path) ->
     Root = bksw_conf:disk_store(),
-    {T1, T2, T3} = erlang:now(),
+    {T1, T2, T3} = erlang:timestamp(),
```
stevendanna: Because `erlang:timestamp()` doesn't provide the same uniqueness guarantees as erlang:now(), I think we might actually want to add some randomness or use `erlang:unique_integer()` here otherwise we may get collisions.  Same for your use in `generate_id` in bksw_req.sql

stevendanna:
@marcparadise I'm going to close this PR but leave the branch.  I've split out some of the cleanups into new PRs.  We should totally move to erlang 18 or 19.  I also have an outstanding branch for this work.  Maybe we could pair up sometime over the next few weeks and figure out what is left to do.

PR #1006, opened 2016-11-08
[OPS-174] expose compliance profiles configurables in chef-server.rb

srenatus:
With this change, you can set

    data_collector['token'] = 'yeahrightsecure'
    profiles['root_url'] = 'https://automate.test'

to allow for signed requests to fetch compliance profiles from automate. (This should have been part of #999.)


alexpop:
Thanks Stephan, I tried the change and works as expected

PR #1010, opened 2016-11-10
Rewrite the POST to the correct Automate endpoint

alexpop:
Without this, the URI received by chef-server(ex: `/organizations/:org/data-collector`) is forwarded to Automate where a 404 is returned.


danielsdeleo:
I'm not a maintainer here, but 👍 for getting this out now, and we should after that take a moment to consider if we can add some testing around it.

stevendanna:
The TravisCI failures look unrelated and possibly related to the environment no longer having a USER variable set.  Going to merg this for now.

markan:
:+1: Looks like this is more correct

PR #1011, opened 2016-11-10
Release note version update

tas50:
There is no 12.9.2. It's 12.10


tas50:
No DCO needed and my readme change isn't breaking travis

stevendanna:
@marcparadise Is this correction true? I thought this is still an outstanding bug?

PR #1012, opened 2016-11-11
data-collector rewrite based on the path of root_url attribute

alexpop:
After reading [this](https://docs.chef.io/release/automate/setup_visibility_chef_automate.html#configure-chef-server-to-send-server-object-data) I'm proposing a different solution for fix: https://github.com/chef/chef-server/pull/1010


stevendanna:
Do we need this to go out in 12.11?

srenatus:
@stevendanna nope, we don't. This change makes something configurable that was hardcoded before, but the hardcoded version is ok for super-near-future.

markan:
We probably should make sure the compliance profile is changed to match; it was modeled after the data collector. 
Also, the release notes may need updating to fit with this change.

PR #1014, opened 2016-11-14
Update 12.10.0 release notes to mention push 1 compat issue

stevendanna:


stevendanna:
Note, I haven't tested this config change yet.  But if someone already has an knows it to be correct perhaps they can vouch for it.  I imagine we might also need to change ciphers?

PR #1022, opened 2016-11-23
Mp/merge bifrost 3

marcparadise:
Not for merge yet, but I did want to get this through travis and see how it behaves.


omnibus/config/projects/chef-server.rb
```diff
@@ -70,10 +70,6 @@
 dependency "keepalived"
 dependency "bookshelf"
 
-# the front-end services
-dependency "oc_bifrost"
-dependency "oc_id"
```
stevendanna: Did you mean to remove oc_id?

marcparadise: d'oh, nice catch

omnibus/omnibus.rb
```diff
@@ -11,12 +11,12 @@
...
+build_retries 0 #TODO NOPE
 fetcher_read_timeout 120
 
 # Load additional software
 # ------------------------------
 # software_gems ['omnibus-software', 'my-company-software']
 # local_software_dirs ['/path/to/local/software']
 
-fatal_transitive_dependency_licensing_warnings true
+fatal_transitive_dependency_licensing_warnings  false
```
stevendanna: We need to leave this to true in the long run.  What is failing currently?

stevendanna:
Overall I think this is probably a good idea.  One less unix-level service from the user's perspective.  Do you have a sense of what the TODO list here is @marcparadise?

marcparadise:
Just some likely squashing and more eyes. Upgrade testing to make sure services are properly cleaned up.

There's also a sqerl branh that's part of this that needs to land, I'm going to get that in this week.

stevendanna:
I still think this could make a lot of sense, but I'm going to close out the PR (will leave branch though) as this is now over a year old.

PR #1027, opened 2016-12-02
Update docs for development environment

thommay:
:construction: I'm updating the docs as I go, but here's the first bit :construction:

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -50,6 +50,31 @@ In a separate terminal session/pane/window:
     sudo -i
     dvm quickstart oc_erchef
 
+To use your running Chef Server through standard commands such as knife,
+you'll need to create an organisation and a user, and then create a
+knife config on your workstation.
+
+    vagrant ssh
+    sudo -i
+    # create a user to access chef with
+    chef-server-ctl user-create -f /tmp/admin.pem admin Admin User admin@example.com password 
+    # create an organization
+    chef-server-ctl org-create -f /tmp/test-validator.pem test Test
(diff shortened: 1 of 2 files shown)


dev/README.md
```diff
@@ -50,6 +50,31 @@ In a separate terminal session/pane/window:
...
+To use your running Chef Server through standard commands such as knife,
+you'll need to create an organisation and a user, and then create a
+knife config on your workstation.
+
+    vagrant ssh
+    sudo -i
+    # create a user to access chef with
+    chef-server-ctl user-create -f /tmp/admin.pem admin Admin User admin@example.com password 
+    # create an organization
+    chef-server-ctl org-create -f /tmp/test-validator.pem test Test
```
stevendanna: You can add `-a admin` and the user will automatically be added to the org so you don't need the step below.

stevendanna:
Left a minor comment, but this looks good to me.  Thanks for taking the time to update these.

dev/README.md
```diff
@@ -50,6 +50,31 @@ In a separate terminal session/pane/window:
...
+    # associate the user with the organization
+    chef-server-ctl org-user-add test admin
+
+Now on your workstation, create `.chef/knife.rb` in the root of your
```
marcparadise: Maybe we could pre-create this in dev/.chef as part of this PR?

stevendanna:
> Maybe we could pre-create this in dev/.chef as part of this PR?

I think let's merge this.  It is better than no documentation and we can do the precreation when we have time.

PR #1028, opened 2016-12-05
Return correct number of rows when searching with Elasticsearch

smith:
Recent changes to the way we use the Elasticsearch paging APIs to do
Chef searches had caused searches to always return all rows when given a
`rows` query parameter.

This change fixes the Elasticsearch behavior.

It also adds Pedant tests to test that the `rows` parameter returns the
correct number of rows.

(Note that our internal CI testing infrastructure only runs the Pedant
test suite against a default installation with Solr as the search
provider. These new tests should pass on either provider given the
changes included in this commit, or fail against an Elasticsearch-backed
earlier version.)

tumblr_o6gi6gpjx21uluepno1_400

### DIFF
--- a/oc-chef-pedant/lib/pedant/rspec/search_util.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/search_util.rb
@@ -1,4 +1,4 @@
-# Copyright: Copyright (c) 2012 Opscode, Inc.
+# Copyright: Copyright (c) 2012-2016 Chef Software, Inc.
 # License: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


smith:
Note that Travis is failing because the Pedant tests are failing on Chef Zero. This should be fixed with chef/chef-zero#247.

stevendanna:
This change looks reasonable to me.

oc-chef-pedant/lib/pedant/rspec/search_util.rb
```diff
@@ -323,6 +323,25 @@ def can_perform_basic_searches_for(object_type)
           end
         end # can_perform_basic_searches_for
 
+        # Search for limited rows to see if we get the correct number back
```
marcparadise: Let's also include a test that shows (for example) if we have three expected results, our search returns them paged correctly (2, then 1).

marcparadise: Similarly even though it's implicit, we should verify that a search that returns less than the expected number of results will return no results for the next page.

smith:
@marcparadise cc20459 should address your concerns.

src/oc_erchef/apps/chef_index/src/chef_elasticsearch.erl
```diff
@@ -45,33 +45,11 @@ search(#chef_solr_query{} = Query) ->
...
-            AllIds = lists:append([ Ids, NewIds ]),
-            scroll(ScrollId, NumFound, length(AllIds), AllIds);
-        %% For now keep these error messages
-        %% consistent with chef_solr
-        "400" ->
-            {error, {solr_400, Url}};
-        "500" ->
-            {error, {solr_500, Url}}
-    end.
+    {ok, undefined, NumFound, Ids}.
```
stevendanna: I think this was an existing problem, but since we return undefined here, it seems like our spec for search() above is a bit wrong.

marcparadise: I'm surprised dialyzer gave a pass to this one.

src/oc_erchef/apps/chef_index/src/chef_elasticsearch.erl
```diff
@@ -45,33 +45,11 @@ search(#chef_solr_query{} = Query) ->
     end.
 
 handle_successful_search(ResponseBody) ->
-    EjsonBody = jiffy:decode(ResponseBody),
-    ScrollId = ej:get({<<"_scroll_id">>}, EjsonBody),
-    Response = ej:get({<<"hits">>}, EjsonBody),
+    Response = ej:get({<<"hits">>}, jiffy:decode(ResponseBody)),
     NumFound = ej:get({<<"total">>}, Response),
     DocList  = ej:get({<<"hits">>}, Response),
```
smith: Apparently yes. See:

https://github.com/chef/chef-server/blob/18c5684338f72379fe0c280f6fce8dee4fc355d2/src/oc_erchef/apps/chef_index/src/chef_elasticsearch.erl#L64

https://github.com/chef/chef-server/blob/18c5684338f72379fe0c280f6fce8dee4fc355d2/src/oc_erchef/apps/chef_index/src/chef_elasticsearch.erl#L50

on current master.

marcparadise:
LGTM

PR #1029, opened 2016-12-05
replace node[fqdn] with server_name in nginx.conf

srenatus:
This should mean we end up with the api_fqdn setting in those places,
irregardless of whether the node this is running on has an FQDN or not.

Tested by running an ad-hoc build, and replacing an OpsWorks instance's chef-server with this one; running reconfigure, and adding&&converging a node against it. ✅ 


danielsdeleo:
LGTM but I'm not a maintainer here. @chef/chef-server-maintainers ?

markan:
LGTM

srenatus:
rebased. will merge when green

PR #1031, opened 2016-12-07
Ignore `-w` option on `chef-server-ctl reindex` with Elasticsearch

smith:
Elasticsearch does not use RabbitMQ to reindex, and the `--wait` option
currently fails.

This makes it so the option does nothing and tells the user it will do
nothing.

Also changes to the omnibus/files/private-chef-ctl-commands Gemfile:

* Update to latest Chef (fixes problem with bundling the older version
  that was locked)
* Use HTTPS URLs for GitHub dependencies
* Sort things
* Include Pry (and rb-readline to make sure it works)
* Include Redis (needed by reindex ctl command)

tumblr_mv33l3bh451re0c97o1_400

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/reindex.rb
+++ b/omnibus/files/private-chef-ctl-commands/reindex.rb
@@ -102,7 +102,17 @@ add_command_under_category "reindex", "general", "Reindex all server data for a 
 
   OptionParser.new do |opts|
     opts.on("-w", "--wait", "Wait for reindex queue to clear before exiting") do |w|
-      options[:wait] = w
+      # Don't attempt to wait if the search_queue_mode is "batch"
+      search_queue_mode = running_config["private_chef"]["opscode-erchef"]["search_queue_mode"]
+      if search_queue_mode == "batch"
+        $stderr.puts <<-EOF
+The search queue mode is currently configured to be "#{search_queue_mode}."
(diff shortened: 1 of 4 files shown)


omnibus/files/private-chef-ctl-commands/reindex.rb
```diff
@@ -102,7 +102,16 @@ def do_reindex(orgs_to_reindex, options)
 
   OptionParser.new do |opts|
     opts.on("-w", "--wait", "Wait for reindex queue to clear before exiting") do |w|
-      options[:wait] = w
+      # Elasticsearch doesn't do any queue waiting, so -w doesn't do anything
+      if running_config["private_chef"]["opscode-erchef"]["search_provider"] == "elasticsearch"
+        $stderr.puts <<-EOF
+Elasticsearch is currently configured as the search provider.
+It does not support waiting for reindex queue. Ignoring "wait" option.
```
marcparadise: I think this is more whether search mode is `rabbitmq` vs `batch`/`inline`.  The `rabbitmq` mode is the only time this flag is applicable.
ref: https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb#L320

stevendanna: Good catch!

PR #1034, opened 2016-12-14
Segment-free cookbook data (RFC 67)

thommay:
https://chef.github.io/chef-rfc/rfc067-cookbook-segment-deprecation.html
This bumps API version for cookbook requests to 2.0, and transforms
cookbook metadata to provide all_files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ matrix:
       script: bundle exec rake chef_zero_spec
       env:
         - "PEDANT_OPTS=\"--skip-oc_id\""
-        - "GEMFILE_MOD=\"gem 'rake'; gem 'chef-zero', github: 'chef/chef-zero', tag: 'v14.0.5' \""
+        - "GEMFILE_MOD=\"gem 'rake'; gem 'chef-zero', github: 'chef/chef-zero' \""
       # Remove things only used by erlang
       install:
       after_failure:
@@ -102,7 +102,7 @@ matrix:
       script: bundle exec rake chef_zero_spec
       env:
         - "PEDANT_OPTS=\"--skip-oc_id\""
-        - "GEMFILE_MOD=\"gem 'rake'; gem 'chef-zero', github: 'chef/chef-zero', tag: 'v14.0.5' \""
+        - "GEMFILE_MOD=\"gem 'rake'; gem 'chef-zero', github: 'chef/chef-zero' \""
         - CHEF_FS=1
       # Remove things only used by erlang
       install:
--- a/oc-chef-pedant/lib/pedant/rspec/cookbook_util.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/cookbook_util.rb
@@ -176,12 +176,11 @@ module Pedant
         end
       end
 
-      def new_cookbook_artifact(name, identifier, opts = {})
+      def new_cookbook_artifact_v0(name, identifier, opts = {})
         result = {
           "name" => "#{name}",
           "identifier" => identifier,
           "version" => opts[:version] || default_version, # version doesn't matter for cookbook_artifacts
-          "json_class" => "Chef::CookbookVersion",
           "chef_type" => "cookbook_version",
           "frozen?" => false,
           "recipes" => opts[:recipes] || [],
@@ -212,6 +211,50 @@ module Pedant
         result
       end
 
+      def new_cookbook_artifact_v2(name, identifier, opts = {})
+        if opts.key?(:recipes)
+          recipes = opts[:recipes].map do |r|
+            r["name"] = "recipes/#{r["name"]}"
+            r
+          end
+        end
+
+        all_files = recipes || []
+
+        result = {
+          "name" => "#{name}",
+          "identifier" => identifier,
+          "version" => opts[:version] || default_version, # version doesn't matter for cookbook_artifacts
+          "chef_type" => "cookbook_version",
+          "frozen?" => false,
+          "all_files" => all_files,
+          "metadata" => {
+            "version" => opts[:version] || default_version,
+            "name" => name, # not actually used
+            "maintainer" => opts[:maintainer] || default_maintainer,
+            "maintainer_email" => opts[:maintainer_email] || default_maintainer_email,
+            "description" => opts[:description] || default_description,
+            "long_description" => opts[:long_description] || default_long_description,
+            "license" => opts[:license] || default_license,
+            "dependencies" => opts[:dependencies] || {},
+            "attributes" => opts[:attributes] || {},
+            # this recipies list is not the same as the top level list
+            # this is a list of recipes and their descriptions
+            "recipes" => opts[:meta_recipes] || {}
+          }
+        }
+        result["metadata"]["providing"] = opts[:providing] if opts[:providing]
+        result
+      end
+
+      def new_cookbook_artifact(name, identifier, opts = {})
+        if platform.server_api_version >= 2
+          new_cookbook_artifact_v2(name, identifier, opts)
+        else
+          new_cookbook_artifact_v0(name, identifier, opts)
+        end
+      end
+
       def delete_cookbook_artifact(requestor, name, identifier)
         res = delete(api_url("/#{cookbook_url_base}/#{name}/#{identifier}"),
                requestor)
@@ -237,7 +280,7 @@ module Pedant
         opts = {
           recipes: recipes,
           meta_recipes: recipes.each_with_object({}) { |r,h| h["#{cookbook_name}::#{r["name"][0..-4]}"] = "" },
-          providing: recipes.each_with_object({}) { |r,h| h["#{cookbook_name}::#{r["name"][0..-4]}"] = ">= 0.0.0" }
+          providing: recipes.each_with_object({}) { |r,h| h["#{cookbook_name}::#{r["name"][0..-4]}"] = ">= 0.0.0" },
         }
         make_cookbook_artifact(admin_user, cookbook_name, identifier, opts)
       end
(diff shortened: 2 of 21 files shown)


stevendanna:
@thommay Any way the maintainers can help out here?  Maybe we could set up some time to pair on this.

thommay:
OK, here's the state of the game. The work in chef and chef-zero is complete and works. This now needs pedant updates to:
 - stop testing for deprecated cookbook metadata
 - test for v2 cookbook manifests
once that's done, the trifecta should all be mergeable at the same time.

stevendanna:
@thommay I'm going through old PRs and this feels like something that we still definitely want.  (I'm assuming here, given that we haven't needed it so far, I'm guessing we just have client-side code that deals with the old format still).  Any chance this is something that could end up on communities board this year?

thommay:
I would very much like to get this on our board this year but the habitat work is (rightly!) taking cycles right now

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
```diff
@@ -589,6 +662,66 @@ list_query(_ObjectRec) ->
...
+                                    _ ->
+                                        add_segment_to_filename(Segment, File)
+                                end,
+                        ej:set_p({<<"all_files">>, new}, CB, File1)
+                end,
+                Metadata,
+                Data).
+
+populate_segments(Data, Metadata) ->
+    Md1 = [ ej:set({Segment}, Metadata, []) || Segment <- ?COOKBOOK_SEGMENTS ],
```
thommay: This is clearly wrong but I can't figure out how to do what I want, which is to create each segment as an empty list before populating them. Is there a better way?

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
```diff
@@ -290,10 +331,11 @@ file_list_spec() ->
 
 -spec extract_checksums(ejson_term()) -> [binary()].
 extract_checksums(CBJson) ->
+    Segments = [ <<"all_files">> | ?COOKBOOK_SEGMENTS ],
```
thommay: We only add `all_files` to the list here because we don't want to have it treated as a segment generally (because it's not one!)

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
```diff
@@ -589,6 +672,84 @@ list_query(_ObjectRec) ->
...
+populate_segments(Data, Metadata) ->
+    CB1 = lists:foldl(fun(Segment, CB) ->
+                              ej:set({Segment}, CB, [])
+                      end,
+                      Metadata,
+                      ?COOKBOOK_SEGMENTS),
+
+    lists:foldl(fun(File, CB) ->
+                        { Segment, Record } = remove_segment_from_filename(File),
+                        case lists:member(Segment, ?COOKBOOK_SEGMENTS) of
```
markan: What happens to files that are uploaded via a V2 call that don't fit in the V0 segment pattern? The RFC doesn't mention that, but I see @stevendanna commented about it (https://github.com/chef/chef-rfc/pull/161#issuecomment-156105423). What was the decision? It looks like we silently drop them; I wonder if we might want to error instead, or otherwise we could silently be breaking cookbooks.

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
```diff
@@ -589,6 +672,84 @@ list_query(_ObjectRec) ->
...
+        % if Name is nil, then we have a root file (like metadata.rb) - so we'll set the name to the segment, and the segment to <<"root_files">>
+        [] ->
+            { <<"root_files">>, ej:set({<<"name">>}, File, Segment)};
+        _ ->
+            Record = ej:set({<<"name">>}, File, lists:last(Name)),
+            { Segment, Record }
+    end.
+
+populate_all_files(Segment, Data, Metadata) ->
+    lists:foldl(fun(File, CB) ->
```
markan: Since Segment is a constant, It might be easier to invert the order; e.g. 
```
    ExpandedFiles = case Segment of 
        <<"root_files">> -> Data;
    _ ->
         [ add_segment_to_filename(Segment, File) || File <- Data ]
    end,
    ej:set({<<"all_files">>}, Metadata, ExpandedFiles)
```

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
```diff
@@ -589,6 +672,84 @@ list_query(_ObjectRec) ->
...
+    CB1 = lists:foldl(fun(Segment, CB) ->
+                              ej:set({Segment}, CB, [])
+                      end,
+                      Metadata,
+                      ?COOKBOOK_SEGMENTS),
+
+    lists:foldl(fun(File, CB) ->
+                        { Segment, Record } = remove_segment_from_filename(File),
+                        case lists:member(Segment, ?COOKBOOK_SEGMENTS) of
+                            true -> ej:set_p({Segment, new}, CB, Record);
```
markan: I tend to do much of my building of JSON structures outside of EJSON, then insert in the end. IMHO it can lead to clearer code. In particular, the ej:set_p pattern with new caught me by surprise, (it's not used anywhere else in chef-server or a1 that I can see.
I might have done this by folding over a map of segments to arrays, inserting into that, and then for each recognized COOKBOOK_SEGMENT extracting the built array from the map and ej:set-ting the field.
That's a matter of taste though.
As a secondary thing, I do like splitting things into smaller functions when possible; mostly because redbug doesn't really let you trace an anonymous function easily.

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
```diff
@@ -271,6 +290,31 @@ cookbook_spec(CBName, CBVersion) ->
...
+                        {{opt, <<"maintainer">>}, string},
+                        %% do we want to regex on email address or at least verify that we
+                        %% have a '@'?
+                        {{opt, <<"maintainer_email">>}, string},
+                        {{opt, <<"license">>}, string},
+
+                        {{opt, <<"platforms">>}, constraint_map_spec(cookbook_name)},
+
+                        {{opt, <<"dependencies">>}, constraint_map_spec(cookbook_name)}
+                       ]}},
```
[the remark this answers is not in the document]
thommay: RFC 85 removes those fields, and no v2 client will send them.

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
thommay: I think the only reasonable thing we can do is drop them; our choice is to drop them - and have cookbooks that only encompass the functionality that chef 12 can provide, whilst upgrading to 13 will handle everything; or we error, at which point cookbooks uploaded on 13 will almost always entirely stop working on 12, because inevitably we'll upload something (like a test directory) that 12 doesn't understand but that doesn't affect running the cookbook.

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
thommay: So you'd break out the function the second map takes (assuming the structure remains the same) into a real function?

oc-chef-pedant/lib/pedant/rspec/cookbook_util.rb
```diff
@@ -330,44 +373,75 @@ def new_cookbook(name, version, opts = {})
...
+            af["name"] = "recipes/#{r['name']}"
+            acc << af
+          end
+        end
+
+        if opts.key?(:payload)
+          opts[:payload].each do |part, files|
+            all_files += files.each_with_object([]) do |f, acc|
+              file = f.dup
+              file["name"] = "#{part}/#{file["name"]}"
```
markan: Looking at the usage: https://github.com/chef/chef-server/pull/1034/files#diff-d5911d79b4d2796916287911d6a428fcR69 it seems like we'd need to do something different for root_files?

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
```diff
@@ -589,6 +672,79 @@ list_query(_ObjectRec) ->
...
+
+ensure_v2_metadata(Ejson) ->
+    case ej:get({<<"all_files">>}, Ejson) of
+        undefined ->
+            Ejson1 = ej:set({<<"all_files">>}, Ejson, []),
+            lists:foldl(fun(Segment, CB) ->
+                                case ej:get({Segment}, CB) of
+                                    undefined -> CB;
+                                    Data ->
+                                        CB1 = populate_all_files(Segment, Data, CB),
```
markan: It looks like we replace all_files in populate_all_files above, so this fold would only keep the last segment.

markan:
I pushed a branch ma/kill_segment_tweaks with some ideas on refactoring the code.

src/oc_erchef/apps/chef_objects/src/chef_cookbook_version.erl
markan: Understand the reasoning, but it strikes me as creating a really painful debug experience where those files are needed in the cookbook, but silently dropped on an older version of chef. 
This is where something like a 'required_chef_version' cookbook metadata field would be nice; you could allow uploading non-standard paths only if the required_chef_version was set for a version that supported , and error on downloading cookbooks to chef versions older than supported.

.travis.yml
```diff
@@ -83,7 +83,7 @@ matrix:
       script: bundle exec rake chef_zero_spec
       env:
         - "PEDANT_OPTS=\"--skip-oc_id\""
-        - "GEMFILE_MOD=\"gem 'rake'; gem 'chef-zero', github: 'chef/chef-zero', tag: 'v14.0.5' \""
+        - "GEMFILE_MOD=\"gem 'rake'; gem 'chef-zero', github: 'chef/chef-zero'; gem 'chef', github: 'chef/chef' \""
```
markan: The pin to 14.0.5 was due to the latest chef-zero having issues with chef 13; it might be better to wait on this unpinning until we no longer pin the rest of chef server to chef 13.

thommay: it actually does no good anyway, so yeah, i'll revert it.

markan:
I think this is really close to merging. 
TODO
[] One last review by prior reviewers (@stevendanna @srenatus)
[] Squash/rebase once master is green
[] Make sure commit message makes sense in terms of auto changelog generation.

thommay:
http://wilson.cd.chef.co/job/chef-server-trigger-ad_hoc/85/downstreambuildview/

PR #1038, opened 2017-01-05
[bookshelf] Don't send Cache-Control headers for 404 responses

stevendanna:
The Chef Backend tests run Chef Server with nginx cookbook caching
enabled. This led to test failures because of cached responses from
previous test cases being returned by nginx.

In a few cases, we found that nginx was caching 404 responses because
bookshelf was always sending a Cache-Control header. Since we never
want to cache 404s, the Cache-Control header is now removed from the
response before it is sent.

For a case where the problem was caused by a cached 200 response
(which we normally want to be cached), I've changed the test so that
it is unlikely that the tests are referring to the file.

### DIFF
--- a/src/bookshelf/test/bkswt_api_SUITE.erl
+++ b/src/bookshelf/test/bkswt_api_SUITE.erl
@@ -249,12 +249,12 @@ sql_tests() ->
       put_object_sql,
       object_roundtrip_sql,
       object_delete_sql,
+      cache_control_sql,
       sec_fail_sql,
       signed_url_sql,
       signed_url_fail_sql,
       at_the_same_time_sql,
       noop ].
-
 
 filesystem_tests() ->
     [ bucket_basic,
@@ -265,6 +265,7 @@ filesystem_tests() ->
       put_object,
       object_roundtrip,
       object_delete,
+      cache_control,
       sec_fail,
       signed_url,
       signed_url_fail,
@@ -293,6 +294,8 @@ head_object_sql(Arg) ->
     head_object(Arg).
 put_object_sql(Arg) ->
     put_object(Arg).
+cache_control_sql(Arg) ->
+    cache_control(Arg).
 object_roundtrip_sql(Arg) ->
     object_roundtrip(Arg).
 object_delete_sql(Arg) ->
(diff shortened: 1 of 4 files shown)


markan:
👍

src/bookshelf/test/bkswt_api_SUITE.erl
```diff
@@ -432,6 +433,42 @@ put_object(Config) when is_list(Config) ->
     % Cleanup
     ?assertEqual(ok, mini_s3:delete_bucket(Bucket, S3Conf)).
 
+has_header(_Header, []) ->
+    false;
+has_header(Header, [{Name, _Value}|T]) ->
+    Needle = string:to_lower(Header),
```
srenatus: [nit] This is called for every list element (i.e. the second argument), but doesn't change -- but since this is a test file and this doesn't get called often, I suppose that's fine.

stevendanna:
> It seems like this check, used to determine if it's a GET request, is also something nginx supports via proxy_cache_methods GET;

Unfortunately, `proxy_cache_methods` *always* adds GET and HEAD to the list of acceptable methods and I believe we never want to cache HEAD requests.  From the nginx docs:

> “GET” and “HEAD” methods are always added to the list, though it is recommended to specify them explicitly.

> Near that, wouldn't we want to clear Cache-Control headers for other @uncached destinations, too? I don't oversee if, say, a POST would have the bookshelf respond with a Cache-Control header, though.

I think the original intent was that there a few ways to run this:

- No caching window turned on in erchef so erchef geneates signed URLs that change with every request and thus aren't very cachable.
- Caching window turned on in erchef but nginx caching disabled so erchef generates cachable URLs and caching is handled by an external LB.
- Caching window turned on and nginx caching turned on.

In the third case we clear the Cache-Control header from the cached destination so that we don't get some double caching going on but leave it in on the uncached destination in case the user wants to cache some other set of requests.  I'm not really sure that makes sense though.

srenatus:
@stevendanna thanks for sharing your insights, making this perspicuous. :)

Re `proxy_cache_methods`: [This is the line, in case someone wonders.](https://github.com/nginx/nginx/blob/6d9023f7b435e41a72b7729fc0a7afdd5fa59c5a/src/http/modules/ngx_http_scgi_module.c#L1477)

PR #1041, opened 2017-01-06
[CON-514] chef-server ctl install for chef reporting warning added

Shadae:


omnibus/files/private-chef-ctl-commands/install.rb
```diff
@@ -18,6 +18,19 @@
     package = "chef-manage"
   end
 
+  if package == "opscode-reporting"
+    STDERR.puts "
+     Chef Reporting is a legacy product, and is no longer receiving new feature updates.
```
stevendanna: It may be good to wrap these paragraphs at 80 columns to make them more readable in the terminal.  What do you think?

Shadae: good idea! I'll change that.

tylercloke: 👍

xenolinguist:
👍

PR #1044, opened 2017-01-18
[erchef] Improve error message on user conflict

### LINKED ISSUE #59: Confusing error message with duplicate email address
When creating a user with an email address already taken by an existing user, you will receive the following confusing error message:

```
vagrant@private-chef:~$ sudo chef-server-ctl user-create foobar John Q Public "john.public@example.com" some_password
ERROR: Conflict
Response: User 'foobar' already exists
```

Note that there is no user `foobar` but we still receive this generic conflict message because the email address is already taken.

In the near term, we should present a better error message to the user.  In the long term, we should consider whether the uniqueness constraint on email addresses is necessary outside of Hosted Chef.

stevendanna:
We can get a conflict from the DB in two cases, username conflict or
email conflict. This fixes the error message to alert the user of both
possibilities.

The old error message confused a lot of people:

Fixes #59


tylercloke:
:rocket:

stevendanna:
With this change in place:
```
root@api:~# chef-server-ctl user-create testuser2 test user test@example.com test1234
ERROR: Conflict
Response: Username or email address already in use.
```

PR #1045, opened 2017-01-18
update index to include the reconfigure steps...

larryebaum:
if existing steps followed verbatim, the user is stuck... and needs to figure out (by reading through docs) that the two reconfigures are necessary...
Signed-off-by: Larry Eichenbaum <larryebaum@chef.io>


stevendanna:
Thanks @larryebaum! This looks good but you need the DCO sign-off.  You should be able to do something like:

```
git commit --amend -s
git push origin  larryebaum/manageindex --force-with-lease
```

Let me know if that doesn't work or if you would like some help.

PR #1050, opened 2017-01-20
[SPOOL-490] Make java handle GC log rotation

srenatus:
With logrotate's copytruncate handling of
/var/log/opscode/opscode-solr4/gclog.log, the system could end up with
(truncated) sparse files that had a huge apparent size (`du
--apparent-size FILE`). Non-sparse-file-aware tools could then create a
mess when copying those logs.

See also http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7055309

Now, the rotation of those gclog files is handled by java (by passing
the right arguments). Also, it can be disabled in case you don't need
those files. For backwards compatibility, we've defaulted it to true.

Also, this adds another recipe for `chef-server-ctl cleanup` that
removes old rotated log files (when it was handled by logrotate).

### DIFF
--- a/omnibus/config/software/private-chef-cookbooks.rb
+++ b/omnibus/config/software/private-chef-cookbooks.rb
@@ -53,7 +53,8 @@ build do
       f.write FFI_Yajl::Encoder.encode(
         run_list: [
           'recipe[private-chef::post_11_upgrade_cleanup]',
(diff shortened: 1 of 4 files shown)


stevendanna:
@chef/chef-server-maintainers (cc @jeremiahsnapp) Looking for another maintainer vote on this.

PR #1061, opened 2017-01-25
[ER-444] Enable FIPS configuration at runtime via chef-server.rb

tyler-ball:
Instead of at build time. Setting 'fips true' in your chef-server.rb and
reconfiguring will run the existing fips recipe. If FIPS is enabled at
the Kernel level this value is defaulted to true and can be overridden
with 'fips false'.

Signed-off-by: tyler-ball <tyleraball@gmail.com>
Signed-off-by: rmoshier <rmoshier@gmail.com>

For the `chef-server-12` pipeline this should not cause any differences. The `fips_mode?` method was returning false and will keep returning false. It does default the `fips_enabled` attribute to true if FIPS is enabled in your Kernel. This could potentially break customers until we start including the OpenSSL-FIPS module (the next card in our backlog).

- [x] Update README or docs to show the new `fips true` flag available in `chef-server.rb`
- [x] Update tests

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,6 +8,19 @@ This document contains release notes for the current major release and all patch
 For prior releases, see
 [PRIOR\_RELEASE\_NOTES.md](PRIOR_RELEASE_NOTES.md).
 
+## Unreleased
+
+### FIPS runtime flag exposed
+
+We are updating the Chef Server package to expose a `fips` configuration flag
+in the `chef-server.rb`. Setting `fips true` and reconfiguring will start the
+server in FIPS mode. The default value of this flag is `false` except
+on systems where FIPS is enabled at the Kernel where it defaults to `true`.
+
+The only supported systems at this time for FIPS mode are RHEL. Package for
+other systems will be missing the required OpenSSL FIPS module and will fail
+to start if reconfigured with `fips true`.
+
 ## 12.12.0 (2017-01-26)
 
 This release addresses a number of bugs, the most notable are describe
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -39,11 +39,7 @@ default['private_chef']['license']['upgrade_url'] = "http://www.chef.io/contact/
 
 default['private_chef']['default_orgname'] = nil
 
-# Enable fips mode (openssl)
-# This requires the chef-server-fips package. If you do not have this,
-# your chef-server will probably not work. If you have to manually
-# change this, you're doing it wrong.
-default['private_chef']['fips_enabled'] = false
+default['private_chef']['fips_enabled'] = ChefConfig.fips?
 
 ###
 # Options for installing addons
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -82,6 +82,7 @@ module PrivateChef
   from_email nil
   role nil
   user Mash.new
+  fips nil
 
   ldap Mash.new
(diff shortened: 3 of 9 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -244,6 +246,7 @@ def generate_hash
       results["private_chef"]["oc-chef-pedant"] = PrivateChef["oc_chef_pedant"]
       results["private_chef"]["notification_email"] = PrivateChef["notification_email"]
       results["private_chef"]["from_email"] = PrivateChef["from_email"]
+      results["private_chef"]["fips_enabled"] = PrivateChef["fips"] || PrivateChef["fips_enabled"]
```
marcparadise: Why two exposed flags for a single toggle?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -43,7 +43,7 @@
 # This requires the chef-server-fips package. If you do not have this,
```
marcparadise: Is this comment still true?

marcparadise:
I had thought there were specific components that got built differently for fips (openssl and maybe erlang?).  Was that incorrect or had it changed elsewhere in some way so that we no longer need to?

marcparadise:
What's the time frame for the fips build card?  If we're looking at a long gap it might be best to hold off until that is also ready for merge.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
tyler-ball: Our Chef configuration expects `fips true` so I wanted to provide the same interface to all our products that are going to allow fips enabling at runtime. But because Chef Server internally references `fips_enabled` I thought we could fall back to that.

The more I think about this though, the worse that is. Lets define the interface (`fips true`) and just stick to that.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
tyler-ball: No, we'll update this comment

stevendanna:
gif-keyboard-9842703361763374856

stevendanna:

RELEASE_NOTES.md
```diff
@@ -8,6 +8,19 @@ This document contains release notes for the current major release and all patch
...
+## Unreleased
+
+### FIPS runtime flag exposed
+
+We are updating the Chef Server package to expose a `fips` configuration flag
+in the `chef-server.rb`. Setting `fips true` and reconfiguring will start the
+server in FIPS mode. The default value of this flag is `false` except
+on systems where FIPS is enabled at the Kernel where it defaults to `true`.
+
+The only supported systems at this time for FIPS mode are RHEL. Package for
```
dmccown: Do we need to specify which versions of RHEL we support?

tyler-ball: No - Chef Server is built for RHEL 5, 6 and 7. We support FIPS in all those versions.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -39,11 +39,7 @@
 
 default['private_chef']['default_orgname'] = nil
 
-# Enable fips mode (openssl)
-# This requires the chef-server-fips package. If you do not have this,
-# your chef-server will probably not work. If you have to manually
-# change this, you're doing it wrong.
-default['private_chef']['fips_enabled'] = false
+default['private_chef']['fips_enabled'] = ChefConfig.fips?
```
marcparadise: When you come back for cleanup tasks, it would be good to add a preflight check that we're on a box that supports fips if it has been enabled in the config.

PR #1065, opened 2017-01-26
[ER-459] Add crypto2 to the Erlang load path in FIPS mode

rmoshier:
This templatizes `vm.args` to include the crypot2 module at the front of the load path if FIPS mode is enabled. This means that after adding `fips true` to `chef-server.rb` and reconfiguring, the Erlang app will be using the crypto2 module to access the OpenSSL FIPS module.

We include the `erlang-crypto2` module in all omnibus distributions, but users chose to activate it at runtime.

- [x] Manual testing to make sure this works
- [x] Update the other Erlang modules to use a similar pattern


omnibus/config/software/erlang-crypto2.rb
```diff
@@ -0,0 +1,41 @@
...
+default_version "master"
+
+source git: "https://github.com/jaym/erlang-crypto2.git"
+
+# TODO get Jay to convert this to Apache 2
+license "Reserved"
+license_file "LICENSE"
+
+dependency "erlang"
+dependency "rebar"
```
stevendanna: This will be rebar 2.  Does this compile with rebar3?  If so, it might be good to make a rebar3 software definition (we can pull their prebuilt binary and copy it into embedded/bin) so we don't have to deal with both rebar 2 and rebar 3 when figuring out build issues in the future.

omnibus/config/software/erlang-crypto2.rb
```diff
@@ -0,0 +1,41 @@
...
+
+build do
+  env = with_standard_compiler_flags(with_embedded_path)
+
+  env['USE_SYSTEM_GECODE'] = "1"
```
stevendanna: Are these environment variables necessary for erlang-crypto2?  I don't seem them referenced in Jay's repo.

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
```diff
@@ -59,6 +59,12 @@
   notifies :restart, 'runit_service[opscode-erchef]' unless backend_secondary?
 end
 
+template "/opt/opscode/embedded/service/opscode-erchef/vm.args" do
+  source "oc_erchef.vm.args.erb"
```
stevendanna: I believe we'll eventually want to do this for the other erlang apps as well so we'll likely want this to be a more generic template that we can pass the app name into.

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
```diff
@@ -59,6 +59,12 @@
   notifies :restart, 'runit_service[opscode-erchef]' unless backend_secondary?
 end
 
+template "/opt/opscode/embedded/service/opscode-erchef/vm.args" do
```
stevendanna: For sys.config we write the template into `/var/opt/opscode/opscode-erchef` and then symlink that into `/opt/opscode/embedded/service/opscode-erchef/`.  Perhaps we should do the same here for symmetry?

omnibus/config/software/erlang-crypto2.rb
tyler-ball: It compiled successfully with rebar 2 in my tests. So I figured I would try that instead of trying to bring in rebar 3.

I thought that chef-server was using rebar 2? Thats the default version of the rebar definition (and there is no `rebar3` binary)

omnibus/config/software/erlang-crypto2.rb
tyler-ball: I have no idea - I can try removing them and seeing if it still compiles. I just copied the erchef definition and left in whatever was there

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
tyler-ball: I saw that. I didn't know the reasoning behind symlinking instead of just writing the file. If you think thats the right thing to do, then I'll do that!

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
tyler-ball: But the other `vm.args` I looked at was pretty different. I'll see if I can make some common template parts I will. If they are all different, I'll leave them as separate templates.

omnibus/config/software/erlang-crypto2.rb
stevendanna: @tyler-ball Chef Server uses rebar3 everywhere.  We moved to rebar3 during the beta so we vendored rebar3 into each app at the time of conversion.  We have a long-standing TODO to get to a single version now that it is stable. (You'll see src/oc_erchef/rebar3, src/bookshelf/rebar3, etc).

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
stevendanna: @tyler-ball I am honestly unsure on the original intent of these symlinks.  I think we've been carting them around for a while.  @marcparadise Do you know?

tyler-ball:
I need to test this some more and see if there are any tests I can add/update, but I would love to start getting eyes on this.  /cc @chef/chef-server-maintainers

omnibus/config/projects/chef-server.rb
```diff
@@ -97,6 +97,7 @@
 dependency "private-chef-upgrades"
 dependency "private-chef-cookbooks"
 dependency "chef-ha-plugin-config"
+dependency "erlang-crypto2"
```
stevendanna: Let's move this further up.  My guess is this code will not be changing often after we have it stable.

omnibus/config/software/oc_bifrost.rb
```diff
@@ -25,7 +25,7 @@
 
 build do
   env = with_standard_compiler_flags(with_embedded_path)
-  profile_name = fips_mode? ? "fips" : "default"
+  profile_name = "default"
```
stevendanna: Not important right now, but we can probably remove the profile_name stuff completely and just assume "default" which is what rebar3 will build by...default.

src/oc_erchef/rebar.config
```diff
@@ -101,11 +101,6 @@
                 ]},
         {erl_opts, [export_all]}
        ]},
-    {fips, [
```
stevendanna: Can we do similar cleanups in other erlang apps?

omnibus/files/private-chef-cookbooks/private-chef/recipes/oc_bifrost.rb
```diff
@@ -60,4 +60,18 @@
...
 end
 
+vmargs_config = File.join(oc_bifrost_dir, "vm.args")
+
+template vmargs_config do
+  source "oc_bifrost.vm.args.erb"
+  owner OmnibusHelper.new(node).ownership['owner']
+  group OmnibusHelper.new(node).ownership['group']
+  mode "644"
+  notifies :restart, 'runit_service[oc_bifrost]' unless backend_secondary?
```
stevendanna: Note to other reviewers: The differences in the restart strategies across the various apps is matching the already existing differences from the sys.config files for these apps.  My guess is that this can be sorted out so they are all the same, but I don't think it should be a blocker for this.

omnibus/config/projects/chef-server.rb
tyler-ball: Done, moved into the `global` commented block

omnibus/config/software/oc_bifrost.rb
tyler-ball: I assumed it would do that but wanted to introduce as few changes as possible. I'll go ahead and remove this though since it does `default` by... default 😁

src/oc_erchef/rebar.config
tyler-ball: :doh: Incoming!

tyler-ball:
@stevendanna Updated per your comments!

tyler-ball:
Pedant tests passed with a local build

omnibus/config/software/erlang-crypto2.rb
```diff
@@ -0,0 +1,39 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name "erlang-crypto2"
+default_version "er-459/update-crypto"
```
marcparadise: Reminder to change this once the branch lands

omnibus/config/software/erlang-crypto2.rb
```diff
@@ -0,0 +1,39 @@
...
+
+source git: "https://github.com/chef/erlang-crypto2.git"
+
+license "BSD-3-Clause"
+license_file "LICENSE"
+# https://github.com/chef/license_scout/issues/61
+skip_transitive_dependency_licensing true
+
+dependency "erlang"
+dependency "rebar"
```
marcparadise: This will pull in rebar2 which we're trying to eradicate. It's not great, but our current practice is to include the rebar3 binary within the project.

tyler-ball:
> Since we're moving the vm.args files [back] into templates, let's also remove them from their original locations in $project/config/vm.args, and the overlay references in $project/rebar.config

@marcparadise There is still $project/config/sys.config even though those are templatized. I thought the local copies of `sys.config` and `vm.args` were used for local testing?

omnibus/config/software/erlang-crypto2.rb
tyler-ball: I'm fine with updating to rebar3. How is that binary included in `chef-server` right now? I don't see an omnibus software dependency on it. I definitely want to build this using whatever other rebar tools we use.

tyler-ball: I see that it is just copied into the 4 Erlang applications. I'm going to use the `.../src/bookshelf/rebar3` instance to build this application.

tyler-ball:
@marcparadise I had to restore the `vm.args` for the oc_bifrost application. Without this file present in the root of the release, the `bin/oc_bifrost start` command loads `vm.args` and `sys.config` from `releases/12.12.0-blah/`. Those files are defaulted / empty so they are missing necessary config values to run.

Restoring `vm.args` to the root of the release dir causes the `bin/oc_bifrost start` to load `vm.args` and `sys.config` from the root dir. This `sys.config` has the correct configurations needed and the applicaiton starts.

It looks like the `bin/oc_bifrost` command is generated by rebar (or relx) so I could not change that. Instead I opted to restore the `vm.args` file that I had deleted.

doc/HOW_THINGS_WORK.md
```diff
@@ -226,3 +226,36 @@ object types and organizations.
...
+
+### FIPS Implementation Details
+
+The erlang-crypto2 app provides `crypto` module implementation. To support
+switching to this crypto module at runtime we perform the following:
+
+* Build the erlang-crypto2 app as a separate omnibus definition using the same
+Erlang libraries used to build all other Erlang apps. Copy the `ebin` and `priv`
+folders from the build into a custom location inside the omnibus package.
+* If `fips true` is set and the server reconfigured, we update the `vm.args`
```
rmoshier: s/server reconfigured/server is reconfigured

PR #1066, opened 2017-01-27
Log service start exit code. Use status to verify started service

marcparadise:


omnibus/files/private-chef-cookbooks/private-chef/templates/default/cluster.sh.erb
```diff
@@ -45,7 +45,9 @@ function transition
 
 <%- keepalive_services.each do |service| -%>
       log_me "Starting service <%= service %>"
-      /opt/opscode/bin/private-chef-ctl start <%= service %> || error_exit "Cannot start <%= service %>"
+      /opt/opscode/bin/private-chef-ctl start <%= service %>
```
stevendanna: Maybe redirect stdout and stderr here to the logfile?  I'm not sure we capture the output of this script by default, hence the log_me and error_exit functions.

PR #1068, opened 2017-01-31
Use xz compression (level 1) for RPM and DEBs

srenatus:
This reduces the package size considerably, with a similar run time.

[Ad-hoc build is here](http://wilson.ci.chef.co/view/Chef%20Server%2012/job/chef-server-12-trigger-ad_hoc/325/downstreambuildview/).

Glancing at the numbers, we find for el-7/x86_64 (size in MB, packaging time in seconds)

old: 327M, 458s
new: 275M, 369s

and for ubuntu-12.04/amd64:

old: 324M, 299s
new: 270M, 248s


stevendanna:
gif-keyboard-12878899955496040301

PR #1072, opened 2017-02-03
Add Habitat plan to bookshelf

elliott-davis:
* added Erlang17 plan to source since most of the chef-server source depends on it and it's not available in core.

Signed-off-by: Elliott Davis <edavis@chef.io>


src/erlang17/plan.sh
```diff
@@ -0,0 +1,54 @@
+pkg_name=erlang17
+pkg_origin=core
```
stevendanna: Should this be `chef`?

src/erlang17/plan.sh
```diff
@@ -0,0 +1,54 @@
...
+  ./configure --prefix="${pkg_prefix}" \
+              --enable-threads \
+              --enable-smp-support \
+              --enable-kernel-poll \
+              --enable-dynamic-ssl-lib \
+              --enable-shared-zlib \
+              --enable-hipe \
+              --with-ssl="$(pkg_path_for openssl)/lib" \
+              --with-ssl-include="$(pkg_path_for openssl)/include" \
+              --without-javac
```
stevendanna: It isn't vital but we can save a bit of space by not including `wc` and `megaco`: https://github.com/chef/omnibus-software/blob/master/config/software/erlang.rb#L137-L139

elliott-davis:
I agree it would be cool to not have to duplicate config. I wonder if we could do something with env-var substitution 🤔

src/erlang17/plan.sh
elliott-davis: We should probably port this to the other core habitat plans as well

elliott-davis:
@stevendanna @marcparadise I've updated this to use erlang 18 from habitat core. Would you mind giving it a 2nd look?

stevendanna:
@elliott-davis Sorry we weren't responsive here.  I'm closing this one out in favor of the ongoing chef-server-habitat work.  Thanks again for getting this work started.

PR #1076, opened 2017-02-08
[ER-455] Run pedant tests with FIPS enabled

tyler-ball:
This only happens on FIPS supported environments.

Signed-off-by: tyler-ball <tyleraball@gmail.com>

\cc @chef/chef-server-maintainers

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -11,5 +11,5 @@ then
   echo "Sleeping 120 seconds to allow the Chef Server to reconfigure in FIPS mode"
   echo ""
   sleep 120
-  sudo chef-server-ctl test -J $WORKSPACE/pedant-fips.xml
+  sudo chef-server-ctl test -J $WORKSPACE/pedant-fips.xml --smoke
 fi
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -1,4 +1,15 @@
 #!/bin/sh
 set -evx
 
-chef-server-ctl test -J $WORKSPACE/pedant.xml --all
+sudo chef-server-ctl test -J $WORKSPACE/pedant.xml --all
+
+if [ $OMNIBUS_FIPS_MODE == "true" ]
+then
+  echo "fips true" >> /etc/opscode/chef-server.rb
+  sudo chef-server-ctl reconfigure
+  echo ""
+  echo "Sleeping 120 seconds to allow the Chef Server to reconfigure in FIPS mode"
+  echo ""
+  sleep 120
+  sudo chef-server-ctl test -J $WORKSPACE/pedant-fips.xml
+fi


tyler-ball:
This depends on https://github.com/chef-cookbooks/opscode-ci/pull/638 being merged and applied to the chef-server testers before it will be leveraged

ci/run_tests.sh
```diff
@@ -2,3 +2,11 @@
 set -evx
 
 chef-server-ctl test -J $WORKSPACE/pedant.xml --all
+
+if [ $OMNIBUS_FIPS_MODE == "true" ]
+then
+  echo "fips true" >> /etc/opscode/chef-server.rb
+  chef-server-ctl reconfigure
+  sleep 120
```
marcparadise: Can we log that we're sleeping here before the sleep command?

tyler-ball:
@marcparadise Good Question!  EngServices suggested using the existing builders because it is a much higher cost to create new builders and modify the shape of the pipeline.

ci/run_tests.sh
tyler-ball: Updated!

stevendanna:
Maybe we could run the smoke tests on the FIPS run? Would that give us enough confidence?

tyler-ball:
@ssd what about a combination of smoke tests and `chef-server-ctl test` without the `--all`?

stevendanna:
> chef-server-ctl test without the --all?

That is what I meant by the smoke tests.  The default of `chef-server-ctl test` is to run as if the `--smoke` flag was passed.

tyler-ball:
@stevendanna I've been running some local tests and seeing some uncomfortable behavior. When I turn on FIPS mode and run `chef-server-ctl --all` I get a few tests (5-10 random ones) failing with a 503. If I then run `chef-server-ctl --rerun` to re-run those failed tests they succeed. I'm thinking that the server operates slower in FIPS mode and times at at the number of requests it is trying to handle during testing.

I have never seen a failure when running only the smoke tests. I want to get to the point where I am comfortable suggesting we only run those but the strange timing out bothers me. I don't think the server actually has a problem, I think it just runs slower in FIPS mode.

schisamo:
@marcparadise @stevendanna @tyler-ball If we are worried about test times we can also re-provision larger test slaves which should speed things up. What I don't want to do is muck with the underlying matrix shape and pipeline code too much. I would like to only explore that option as a last resort.

ci/run_tests.sh
```diff
@@ -1,4 +1,15 @@
 #!/bin/sh
 set -evx
 
-chef-server-ctl test -J $WORKSPACE/pedant.xml --all
+sudo chef-server-ctl test -J $WORKSPACE/pedant.xml --all
```
schisamo: The `test` subcommand does not need to be executed with elevated privileges.

ci/run_tests.sh
```diff
@@ -1,4 +1,15 @@
...
+
+if [ $OMNIBUS_FIPS_MODE == "true" ]
+then
+  echo "fips true" >> /etc/opscode/chef-server.rb
+  sudo chef-server-ctl reconfigure
+  echo ""
+  echo "Sleeping 120 seconds to allow the Chef Server to reconfigure in FIPS mode"
+  echo ""
+  sleep 120
+  sudo chef-server-ctl test -J $WORKSPACE/pedant.xml --all
```
schisamo: Ditto here on not needing elevated privileges.

ci/run_tests.sh
stevendanna: While it may be true that the `test` command doesn't need root (although, it may need to run as the service user to read some config), the chef-server-ctl command itself has a simple check:

https://github.com/chef/chef-server/blob/master/omnibus/config/software/private-chef-ctl.rb#L65-L69

so in practice you need to run all of its commands as root.

tyler-ball:
@stevendanna I bumped the keygen start size and cache size, then the `--all` tests passed with flying colors.

I'm now 💯 in favor of only running smoke tests for FIPS. I updated this PR to reflect that.

ci/run_tests.sh
schisamo: Good to know.

ci/run_tests.sh
```diff
@@ -1,4 +1,15 @@
 #!/bin/sh
 set -evx
 
-chef-server-ctl test -J $WORKSPACE/pedant.xml --all
+sudo chef-server-ctl test -J $WORKSPACE/pedant.xml --all
+
+if [ $OMNIBUS_FIPS_MODE == "true" ]
```
marcparadise: Based on discussion in Slack, let's only run the second round of test if the first pass succeeds.  That will prevent overwriting/losing the pedant.xml output that jenkins uses to report on the test results.  Let's add a FIXME note here to look into what it'd take to have jenkins support multiple test outputs (maybe a simple opscode-ci update), but that shouldn't block merging.

stevendanna: We `set -e` so I think we've already got early-exiting on failure covered.

ci/run_tests.sh
```diff
@@ -1,4 +1,15 @@
...
+then
+  echo "fips true" >> /etc/opscode/chef-server.rb
+  sudo chef-server-ctl reconfigure
+  echo ""
+  echo "Sleeping 120 seconds to allow the Chef Server to reconfigure in FIPS mode"
+  echo ""
+  sleep 120
+  sudo chef-server-ctl test -J $WORKSPACE/pedant.xml
```
schisamo: Per @marcparadise's concern about overwriting the non-FIPS test results we should write these results into a different file:

```ruby
sudo chef-server-ctl test -J $WORKSPACE/pedant-fips.xml
```

marcparadise: I'm +1 on a different file, but do we need to update the pipeline to look for both?

schisamo: @marcparadise Yep: https://github.com/chef-cookbooks/opscode-ci/pull/643

PR #1078, opened 2017-02-09
Remove the default admin page from solr4.

marcparadise:
Do not ship the solr for admin page. Any configuration is done
through the API.

Signed-off-by: Marc Paradise <marc.paradise@gmail.com>

### DIFF
--- a/omnibus/config/software/opscode-solr4.rb
+++ b/omnibus/config/software/opscode-solr4.rb
@@ -40,9 +40,12 @@ relative_path "solr-#{version}"
 service_dir = "#{install_dir}/embedded/service/opscode-solr4"
 
 build do
-  env = with_standard_compiler_flags(with_embedded_path) 
+  env = with_standard_compiler_flags(with_embedded_path)
+
+  embedded_jre_path = "#{install_dir}/embedded/jre/bin"
+
   unless @ibm_jre
-    env['PATH'] = "#{env['PATH']}:#{install_dir}/embedded/jre/bin"
+    env['PATH'] = "#{env['PATH']}:#{embedded_jre_path}"
   end
 
   # copy over the licenses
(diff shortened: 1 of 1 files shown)


omnibus/config/software/opscode-solr4.rb
```diff
@@ -39,8 +39,6 @@
 service_dir = "#{install_dir}/embedded/service/opscode-solr4"
 
 build do
-  env = with_standard_compiler_flags(with_embedded_path)
```
marcparadise: Reviewer note: I removed this because we're not referencing `env`

omnibus/config/software/opscode-solr4.rb
```diff
@@ -54,4 +52,10 @@
   delete "#{service_dir}/jetty/example*"
   delete "#{service_dir}/jetty/multicore"
   delete "#{service_dir}/jetty/solr"
+
+  # Replace the built-in admin.html with a blank page to effectively disable the admin
+  # interface to solr.  Note that this does not disable API access.
+  touch "admin.html"
+  command "jar -uf #{service_dir}/jetty/webapps/solr.war admin.html"
```
marcparadise: Reviewer note - looks like I'll put that `env` back in and use it here so that `jar` can be found at build time...

omnibus/config/software/opscode-solr4.rb
```diff
@@ -54,4 +58,18 @@
   delete "#{service_dir}/jetty/example*"
   delete "#{service_dir}/jetty/multicore"
   delete "#{service_dir}/jetty/solr"
+
+  # Replace the built-in admin.html with a blank page to effectively disable the admin
+  # interface to solr.  Note that this does not disable API access.
+  touch "admin.html"
+
+  if ppc64? || ppc64le? || s390x?
+    # s390, PPC et all have `zip` but do not include `jre` in the JRE
```
srenatus: [nit] this be "et al" and "do not include `jar`", I think

PR #1080, opened 2017-02-10
Make chef_index_batch_tests a less flaky

srenatus:
Before, these often failed on travis -- and we've established that could
have been the case because the flush() we call in the test has been
executed _before_ the item as added. This change makes add_item wait for
the queue size to change to something non-empty. (This needs adjustments
if we add test cases that care for more than one item added and waited
for.)

### DIFF
--- a/src/oc_erchef/apps/chef_index/test/chef_index_batch_tests.erl
+++ b/src/oc_erchef/apps/chef_index/test/chef_index_batch_tests.erl
@@ -39,8 +39,8 @@ chef_index_batch_test_() ->
                WrapperSize = maps:get(wrapper_size, State),
                meck:expect(chef_index, update,
                            fun(solr, Payload) ->
-                                   ExpectedSize = byte_size(iolist_to_binary(Payload)),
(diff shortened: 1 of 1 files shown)


src/oc_erchef/apps/chef_index/test/chef_index_batch_tests.erl
```diff
@@ -119,3 +126,14 @@ wait_for_started() ->
...
             ok
     end.
+
+%% Note that this cannot work if the item added is bigger than max_size -- these
+%% will be flushed immediately, and this _could_ have happened before we started
+%% waiting/looping.
+wait_for_added() ->
+    case chef_index_batch:status() of
+        #{ item_queue := [] } ->
+            timer:sleep(10),
```
marcparadise: Is a 10ms sleep the intent? Maybe something like 500?

stevendanna: This was the intent.  We believe the race is literally between two *very fast most of the time* calls, so we didn't want to slow tests down too much.  I cold see going to 100ms though.

PR #1081, opened 2017-02-10
[erchef] Improve reindex reliability when using ElasticSearch

stevendanna:
The fire-and-forget reindex process make sense for
RabbitMQ+opscode-expander but does not make sense for the direct-to-ES
indexing methods.  In Hosted and other large environments, a reindex
may quickly overwhelm the HTTP connection pool, leading to reindex
failures that go unreported to the user.

This change collects all errors so that they can be reported at the
end of the reindex run. It also introduces retries with randomized
sleep to help reduce the chance of overall failures.

For large organizations, this new strategy may reduce the throughput
of reindexing. Such users may need to tune the reindex_batch_size and
the search_batch_max_wait parameters. Further, to finish reindexing in
a reasonable time, users may still need to adjust the HTTP connection
pool size and the batch size.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -277,6 +277,42 @@ default['private_chef']['opscode-erchef']['sql_user'] = "opscode_chef"
 default['private_chef']['opscode-erchef']['sql_password'] = "snakepliskin"
 default['private_chef']['opscode-erchef']['sql_ro_user'] = "opscode_chef_ro"
 default['private_chef']['opscode-erchef']['sql_ro_password'] = "shmunzeltazzen"
+#
+# Reindex configurables
+#
+# NOTE: The semantics of reindexing are very different between
+# ElasticSearch and Solr. Solr reindexing is only placing an item on a
+# queue, thus it is unlikely to fail at the erchef level.  Rather, you
+# will want to tune opscode-expander.
+#
+# These configuration items are consulted during reindex requests.
+# Such requests originate from users running chef-server-ctl reindex.
+#
+#   reindex_batch_size - Number of items to fetch from the database
+#                        and send to the search index at a time.
+#
+# reindex_sleep_min_ms - Minimum number of milliseconds to sleep
+#                        before retrying a failed attempt to index an
+#                        item. Retries are delayed a random number of
+#                        miliseconds between reindex_sleep_min_ms and
+#                        reindex_sleep_max_ms. Set both this and
+#                        reindex_sleep_max_ms to 0 to skip the sleep.
+#
+# reindex_sleep_max_ms - Maximum number of milliseconds to sleep
+#                        before retrying a failed attempt to index an
+#                        item. Retries are delayed a random number of
+#                        miliseconds between reindex_sleep_min_ms and
+#                        reindex_sleep_max_ms. Set both this and
+#                        reindex_sleep_min_ms to 0 to skip the sleep.
+#
+# reindex_item_retries - Number of times to retry indexing an object
+#                        if it fails.
+#
+default['private_chef']['opscode-erchef']['reindex_batch_size'] = 10
+default['private_chef']['opscode-erchef']['reindex_sleep_min_ms'] = 500
+default['private_chef']['opscode-erchef']['reindex_sleep_max_ms'] = 2000
+default['private_chef']['opscode-erchef']['reindex_item_retries'] = 3
+#
 # Pool configuration for postgresql connections
 #
 # db_pool_size - the number of pgsql connections in the pool
--- a/src/oc_erchef/apps/chef_index/src/chef_index.erl
+++ b/src/oc_erchef/apps/chef_index/src/chef_index.erl
@@ -110,9 +110,9 @@ add_batch(Batch) ->
     [chef_wait_group:add(BatchWorker, {TypeName, Id, DbName}, Item) || {TypeName, Id, DbName, _} = Item <- Batch],
     case chef_wait_group:wait(BatchWorker) of
         {ok, Results}  ->
-            case all_ok(Results) of
-                true -> ok;
-                false -> {error, not_ok(Results)}
+            case not_ok(Results) of
+                [] -> ok;
+                R -> {error, R}
             end;
         {error, Results, FailedJobs} ->
             {error, [not_ok(Results)|FailedJobs]}
@@ -150,17 +150,13 @@ wait_before_retry(0, 0) ->
 wait_before_retry(Min, Min) ->
     lager:info("chef_index: waiting ~B ms before retry", [Min]),
     timer:sleep(Min);
+wait_before_retry(Min, Max) when Min > Max ->
+    lager:error("chef_index: reindex_sleep_max_ms less than reindex_sleep_min_ms. Sleeping ~B", [Max]),
+    timer:sleep(Max);
 wait_before_retry(Min, Max) ->
-    Rand01 = random:uniform(),
-    %% Tranform (0, 1) -> (Min, Max)
-    %% Round to an integer since timer:sleep only takes an integer
-    RandMinMax = round(Min + (Rand01*(Max - Min))),
+    RandMinMax = Min + random:uniform(Max - Min),
     lager:info("chef_index: waiting ~B ms before retry", [RandMinMax]),
     timer:sleep(RandMinMax).
-
-all_ok(Results) ->
-    Ok = fun({_, Res}) -> Res =:= ok end,
-    lists:all(Ok, Results).
 
 not_ok(Results) ->
     NotOk = fun({_, Res}) -> Res =/= ok end,
--- a/src/oc_erchef/apps/chef_index/src/chef_index.erl
+++ b/src/oc_erchef/apps/chef_index/src/chef_index.erl
@@ -28,8 +28,7 @@
          transform_data/2,
          delete/4,
          add/5,
-         add_async/4,
-         add_async/5,
+         add_batch/1,
          search_provider/0
         ]).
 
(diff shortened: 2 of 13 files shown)


src/oc_erchef/apps/chef_index/src/chef_batch_worker.erl
```diff
@@ -0,0 +1,207 @@
...
+%% If someone calls wait and we are already done, return immediately
+handle_call(gather, _From, #chef_batch_worker_state{waiting_pid = undefined,
+                                                    waiting_workers = [],
+                                                    done_jobs = Done,
+                                                    failed_jobs = Failed}) ->
+    {stop, normal, make_gather_reply(Done, Failed), #chef_batch_worker_state{}};
+%% If someone calls wait and we aren't done, store their pid so we can reply later
+handle_call(gather, From, State = #chef_batch_worker_state{waiting_pid = undefined,
+                                                           waiting_workers = _StillWaiting}) ->
+    {noreply, State#chef_batch_worker_state{waiting_pid = From}};
```
srenatus: Could this stall the caller for a very long time? (I.e., do we need timeouts?)

I can see that the `waiting_pid` would eventually get a reply, when all jobs are done, so this would complicate things a bit, I suppose...

src/oc_erchef/apps/chef_index/src/chef_batch_worker.erl
```diff
@@ -0,0 +1,207 @@
...
+%% If someone is already waiting, error
+handle_call(gather, _From, State = #chef_batch_worker_state{waiting_pid = _Pid}) ->
+    {reply, {error, already_waiting}, State};
+%% If a job is done, remove it from the waiting list.  If this is the
+%% last job, rely to the waiting pid if we have one.
```
srenatus: Got the hunch that this would read nice if "last job" was part of the pattern match in the function head. (But I believe you've probably explored that route already.)

src/oc_erchef/apps/chef_index/src/chef_batch_worker.erl
stevendanna: Yes and maybe are my answers.  The goal of wait is to stall the caller for as long as it takes for the jobs to finish.  Right now we wait without a timeout.  It might be prudent to allow the caller to set a timeout in some way in case it doesn't wait forever, but the use case I was thinking of is the caller definitely wants to wait until all the jobs are done, however long that takes.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -273,6 +273,42 @@
...
+#
+# NOTE: The semantics of reindexing are very different between
+# ElasticSearch and Solr. Solr reindexing is only placing an item on a
+# queue, thus it is unlikely to fail at the erchef level.  Rather, you
+# will want to tune opscode-expander.
+#
+# These configuration items are consulted during reindex requests.
+# Such requests originate from users running chef-server-ctl reindex.
+#
+#   reindex_batch_size - Number of items to fetch from the database
```
srenatus: [nit] indentation is a little off here

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -273,6 +273,42 @@
 default['private_chef']['opscode-erchef']['sql_password'] = "snakepliskin"
 default['private_chef']['opscode-erchef']['sql_ro_user'] = "opscode_chef_ro"
 default['private_chef']['opscode-erchef']['sql_ro_password'] = "shmunzeltazzen"
+#
+# Reindex configurables
+#
+# NOTE: The semantics of reindexing are very different between
```
srenatus: So are the settings below good for ES, for Solr, or good enough for both?

src/oc_erchef/apps/chef_index/src/chef_index.erl
```diff
@@ -111,15 +104,67 @@ add(TypeName, Id, DbName, IndexEjson, ReqId) ->
...
+    wait_before_retry(Min, Max).
+
+wait_before_retry(0, 0) ->
+    ok;
+wait_before_retry(Min, Min) ->
+    lager:info("chef_index: waiting ~B ms before retry", [Min]),
+    timer:sleep(Min);
+wait_before_retry(Min, Max) ->
+    Rand01 = random:uniform(),
+    %% Tranform (0, 1) -> (Min, Max)
```
srenatus: [nit] "Transform"

src/oc_erchef/apps/chef_index/src/chef_index.erl
```diff
@@ -111,15 +104,67 @@ add(TypeName, Id, DbName, IndexEjson, ReqId) ->
...
-    %% part of a request, But we provided a add_async/5 to make it
-    %% easier to call in oc_chef_object_db.
-    add_async(TypeName, Id, DbName, IndexEjson).
+-spec add_batch(list()) -> ok | {error, list()}.
+add_batch(Batch) ->
+    {ok, BatchWorker} = chef_wait_group:start_link(fun add_batch_item_with_retries/1, []),
+    [chef_wait_group:add(BatchWorker, {TypeName, Id, DbName}, Item) || {TypeName, Id, DbName, _} = Item <- Batch],
+    case chef_wait_group:wait(BatchWorker) of
+        {ok, Results}  ->
+            case all_ok(Results) of
```
srenatus: We could also pattern match `not_ok(Results)` here and save a call. (Don't think it matters much, though.)

src/oc_erchef/apps/chef_index/src/chef_wait_group.erl
```diff
@@ -0,0 +1,205 @@
...
+    end.
+
+-spec maybe_reply_to_waiter(#chef_wait_group_state{}) -> boolean().
+maybe_reply_to_waiter(#chef_wait_group_state{waiting_pid = undefined}) ->
+    false;
+maybe_reply_to_waiter(#chef_wait_group_state{waiting_pid = Pid,
+                                             waiting_workers = [],
+                                             done_jobs = Done,
+                                             failed_jobs = Failed}) ->
+    gen_server:reply(Pid, make_gather_reply(Done, Failed)),
```
srenatus: Is it possible that `waiting_pid` has gone away? (I don't think it's expected in the current set up, but if this would be used more generically, would we have to deal with that here?)

src/oc_erchef/apps/chef_index/test/chef_wait_group_tests.erl
```diff
@@ -0,0 +1,45 @@
...
+    {ok, Pid} = chef_wait_group:start_link(fun(MyArg) -> MyArg end, []),
+    chef_wait_group:add(Pid, test_id1, testarg),
+    ?assertEqual({ok, [{test_id1, testarg}]}, chef_wait_group:wait(Pid)).
+
+wait_returns_error_if_the_worker_fails_test() ->
+    {ok, Pid} = chef_wait_group:start_link(fun() -> throw(badfun) end, []),
+    chef_wait_group:add(Pid, test_id1, []),
+    ?assertEqual({error, [], [{test_id1, {throw, badfun}}]}, chef_wait_group:wait(Pid)).
+
+wait_returns_error_with_finished_jobs_and_faild_jobs_test() ->
```
srenatus: [typo] "failed"

src/oc_erchef/priv/reindex-opc-organization
```diff
@@ -98,20 +98,34 @@ perform(drop, _Context, {OrgId, OrgName}) ->
...
-    case Missing of
-        [] ->
-            ok;
-        _ ->
-            io:format(standard_error, "~nThe following objects in organization \"~s\" were unable to be reindexed:~n~n", [OrgName]),
-            [io:format(standard_error, "\t~s ~s~n", [Type, Name]) || {Type, Name} <- Missing],
-            io:format(standard_error, "~n", [])
+    case rpc:call(?ERCHEF, chef_reindex, reindex, [Context, OrgInfo]) of
+        {ok, Missing} ->
+            case Missing of
```
srenatus: I think this would also read nice if collapsed:

```erlang
case rpc:call(...) of
    {ok, []} -> ok;
    {ok, Missing} -> print_missing(Missing, OrgName);
    {error, Failed, Missing} ->
            ...
end,
```

srenatus: Looking at `print_missing/2`, this would do, too:

```erlang
case rpc:call(...) of
    {ok, Missing} -> print_missing(Missing, OrgName);
    {error, Failed, Missing} ->
            ...
end,
```

src/oc_erchef/apps/chef_index/src/chef_index.erl
srenatus: Also, I think we could do `Min + random:uniform(Max-Min)`

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
stevendanna: I think that depends on whether you think the configurable names should be left or right justified ;)

srenatus: hah! right. sorry

stevendanna:

Overall, the reindex system still needs a lot of work, but I think this moves us in the right direction as we will no longer silently fail.

src/oc_erchef/apps/chef_index/src/chef_index_batch.erl
```diff
@@ -120,12 +120,17 @@ init([]) ->
     end.
 
 spawn_flusher(Time) ->
+    %% FIXME: this process will stay around even after someone calls
+    %% chef_index_batch:stop(). It would be nice if this process also
+    %% died on stop. Potentially we could send a message to it from
+    %% the parent in terminate() and receive that message here
+    %% or..something.
```
marcparadise: Maybe `monitor(ParentPid)` and handle the `DOWN` message?

PR #1083, opened 2017-02-13
[solr4] Clean up unused and unwanted configuration

stevendanna:
We had a lot of configuration for request handlers and other things
that we don't use at all as part of the application. Much of this
looks like it was example configuration that was copied in and then
modified.

This also deletes unused template files at the top level of the
template directory.

Finally, it attempts to clean up some of the verbose commenting while
leaving the comments we thought might be useful.


stevendanna:
TODO:
- [ ] Actually implement new enable_full_admin_api variable

PR #1089, opened 2017-02-17
[pedant] Fix PATH in queues_empty? helper, add debug output

stevendanna:


stevendanna:
This appears to be working in the ad-hoc build, I think we should leave the debug output in for a bit, at least until we are sure this test is stable.

PR #1090, opened 2017-02-17
[just trying] try rebar3 with fold marks

srenatus:
Built from [this](https://github.com/erlang/rebar3/compare/master...srenatus:add-travis-fold-marks). Trying to upstream this [here](https://github.com/erlang/rebar3/pull/1484) 😃


srenatus:
*cleanup*

PR #1092, opened 2017-02-17
[omnibus] Allow user to override ssl cert generation

stevendanna:
Previously we would generate certs and dhparams even if the user had
specified files for us to use.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
@@ -49,36 +49,45 @@ end
   end
 end
 
-ssl_keyfile = File.join(nginx_ca_dir, "#{node['private_chef']['nginx']['server_name']}.key")
-ssl_crtfile = File.join(nginx_ca_dir, "#{node['private_chef']['nginx']['server_name']}.crt")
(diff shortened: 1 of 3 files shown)


omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
```diff
@@ -49,36 +49,45 @@
...
+
+# Generate self-signed SSL certificate unless the user has provided one
+if (node['private_chef']['nginx']['ssl_certificate'].nil? &&
+    node['private_chef']['nginx']['ssl_certificate_key'].nil?)
+
+   ssl_keyfile = File.join(nginx_ca_dir, "#{node['private_chef']['nginx']['server_name']}.key")
+   ssl_crtfile = File.join(nginx_ca_dir, "#{node['private_chef']['nginx']['server_name']}.crt")
+
+   openssl_x509 ssl_crtfile do
+     common_name node['private_chef']['nginx']['server_name']
```
srenatus: If this value changes, the user would have to delete the generated cert and key files and reconfigure to have a matching cert regenerated. Do we need to catch this edge case?

srenatus: Oh dear it's too early. Nevermind, the `server_name` in those filenames of course does the trick ✅

PR #1094, opened 2017-02-20
Use OTP 18.3

srenatus:
This gets all components on Erlang/OTP 18.3 and bumps a few dependencies. 🎉


stevendanna:
gif-keyboard-3275793917616347467

stevendanna:

src/bookshelf/rebar.config
```diff
@@ -22,7 +22,7 @@
          {git, "git://github.com/mochi/mochiweb", {tag, "v2.12.2"}}},
         %% TODO 2017/02/09 sr: upgrade -- hex doesn't have that old version
         {erlware_commons, ".*",
-         {git, "https://github.com/erlware/erlware_commons", {tag, "v0.11.1"}}},
+         {git, "https://github.com/erlware/erlware_commons", {tag, "0.22.0"}}},
```
marcparadise: I think this one has been stable enough to use master? We still get pinned when we `rebar3 upgrade`

Issue #694, opened 2016-01-15
Organizations starting with "bookshelf" are unusable


stevendanna:
If you try to create an organization named "bookshelf" or "bookshelf-test" or "bookshelfANYTHING", the organization will be largely unusable.  For example, you won't be able to associate users to it:

```
root@api:/tmp# chef-server-ctl org-user-add bookshelf-test uploader -a
ERROR: Method Not Allowed
Response: <html><head><title>405 Method Not Allowed</title></head><body><h1>Method Not Allowed</h1>Method Not Allowed<p><hr><address>mochiweb+webmachine web server</address></body></html>
```

The root cause is a bad nginx rule that routes all the traffic to the bookshelf service rather than nginx:

```
    # bookshelf
    location ~ "/bookshelf/{0,1}.*$" {
      proxy_pass http://bookshelf;
    }
```

ameir:
Per our discussion over the mailing list, this also impacts cookbooks whose names match the bookshelf or bucket name:
https://discourse.chef.io/t/issues-with-new-chef-server-and-s3-bookshelf/10462

It sounds like it'd also affect creating an organization that matches an S3 bucket name, when not using the built-in bookshelf.

For the S3 case, simply removing that block should fix it.

stevendanna:
This is fixed and should be available in the next release.

PR #1096, opened 2017-02-21
Don't proxy to local bookshelf if it's disabled.

ameir:
Issue described in https://github.com/chef/chef-server/issues/694.

It's not clear whether `nginx_bookshelf_caching` works for remote bookshelfs; if so, this would need an update to factor that in.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
@@ -77,8 +77,9 @@
     }
     <% end -%>
 
+    <% if node['private_chef']['bookshelf']['enable'] -%>
     # bookshelf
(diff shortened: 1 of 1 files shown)


stevendanna:
Thanks for following up on this!

I don't think nginx_bookshelf_caching works for remote cookbook storage. I suppose it is possible if someone had customized their nginx configuration via the .d configuration directory.  But I think that is an unlikely case and that this is fine.

ameir:
Great; thanks for clarifying.  If there's anything outstanding for this PR, let me know and I'll work on it when I can scrape up some time.

stevendanna:
@ameir The only thing we were thinking of adding was potentially doing *both* this change and tightening up the regular expression.

ameir:
@stevendanna, I updated the regex to what I believe is a safe value.  You may have insights that I don't have, though.  On a default Chef Server install, the `location` block now resolves to:

```
location ~ "^/bookshelf/organization-.+/" {
  proxy_pass http://bookshelf;
}
```

In my tests, things look good.  Let me know what you think.

stevendanna:
gif-keyboard-13008948372314378437

PR #1101, opened 2017-02-24
Wire chef_secrets into erchef, bifrost, bookshelf

srenatus:
Trying to establish an abstraction for accessing secrets used in chef-server (database and rabbitmq credentials for example), this introduces chef_secrets into chef-server's erlang components.

Our immediate goal is to reduce the number of files on disk that contain secrets. This is the first step into that direction, creating an interface and using it. This currently directly reads `/etc/opscode/private-chef-secrets.json`.


src/bookshelf/rebar.config
```diff
@@ -117,16 +119,17 @@
...
-     {sync, load},
-     {eunit, load},
-     {mixer, load},
-     syntax_tools,
-     compiler,
-     eper,
-     {pooler, load},
-     {sqerl, load}
+  {release,{bookshelf,{cmd,"cat VERSION"}}, [
+    bookshelf,
```
stevendanna: Spacing issue

PR #1107, opened 2017-02-28
[SPOOL-528] update chef-server to use Veil for pivotal/webui keys

marcparadise:
Do not merge.

IMPORTANT:  This PR will break upgrades.  See first task below

This set of changes does the following:

* replace usage of the files pivotal.pem and webui pem with the Vail key store in chef-server-ctl reconfigure/cookbooks
* update pedant the same

There are some tasks left:
* [x] IMPORTANT: a partybus migration must be created to put the existing keys into Veil. The changes in this PR assume that if the key is not there, it must be created.
* [x] omnibus test cleanup is needed  to get them passing again, this was expected
* [ ] embedded knife - ship a plugin that teaches knife how to pull the creds from Veil.

### DIFF
--- a/oc-chef-pedant/lib/pedant/rspec/common.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/common.rb
@@ -488,6 +488,7 @@ module Pedant
 
         shared(:superuser) { platform.superuser }
         shared(:superuser_key) { platform.superuser_key }
+        shared(:webui_key) { platform.webui_key }
 
         # Given a requestor, create a new one with the same name, but
         # with the web UI's private key for 'impersonation' tests
--- a/oc-chef-pedant/lib/pedant/rspec/common.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/common.rb
@@ -487,13 +487,12 @@ module Pedant
         # Platform object
 
         shared(:superuser) { platform.superuser }
-        shared(:superuser_key_file) { platform.superuser_key_file }
-        shared(:webui_key_file) { Pedant::Config.webui_key || (fail "Missing webui_key in Pedant config!") }
+        shared(:superuser_key) { platform.superuser_key }
 
         # Given a requestor, create a new one with the same name, but
         # with the web UI's private key for 'impersonation' tests
         def impersonate(requestor_to_impersonate)
-          Pedant::Requestor.new(requestor_to_impersonate.name, webui_key_file)
+          Pedant::Requestor.new(requestor_to_impersonate.name, platform.webui_key)
         end
 
         # Need a well-formed yet invalid key for a requestor to test authentiction
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
@@ -22,7 +22,8 @@ class BootstrapPreflightValidator < PreflightValidator
 
   def run!
     # Sanity checks: for any configuration we should expect that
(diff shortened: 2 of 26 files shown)


marcparadise:
Pedant state:  currently re-running locally, but at last check two tests remain to be corrected.

oc-chef-pedant/lib/pedant/rspec/common.rb
```diff
@@ -487,13 +487,12 @@ def set_opt(opt, default)
...
         shared(:superuser) { platform.superuser }
-        shared(:superuser_key_file) { platform.superuser_key_file }
-        shared(:webui_key_file) { Pedant::Config.webui_key || (fail "Missing webui_key in Pedant config!") }
+        shared(:superuser_key) { platform.superuser_key }
 
         # Given a requestor, create a new one with the same name, but
         # with the web UI's private key for 'impersonation' tests
         def impersonate(requestor_to_impersonate)
-          Pedant::Requestor.new(requestor_to_impersonate.name, webui_key_file)
+          Pedant::Requestor.new(requestor_to_impersonate.name, platform.webui_key)
```
marcparadise: So the lack of webui_key defined in platform at present - and the lack of failing tests as a result - are a thing that makes me wonder.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
```diff
@@ -105,9 +106,6 @@ def pivotal_user_exists?
     end
     connect_as(:superuser, 'silent' => true,'db_name' => 'opscode_chef') do |conn|
       begin
-        # The opscode_chef database may not exist - so we'll explicitly
```
marcparadise: This was pretty much just a lie, so it went away.

omnibus/files/private-chef-cookbooks/private-chef/recipes/private_keys.rb
```diff
@@ -23,35 +23,43 @@
...
-node.set['bootstrap']['superuser_public_key'] = pivotal_key.public_key.to_s if new_pivotal
-
-file pivotal_key_path do
-  owner OmnibusHelper.new(node).ownership['owner']
-  group "root"
-  mode "0600"
-  content pivotal_key.to_pem.to_s if new_pivotal
-  sensitive true
+unless PrivateChef.credentials.exist?('chef-server', 'superuser_key')
+  pivotal_key = OpenSSL::PKey::RSA.generate(2048)
```
srenatus: Maybe we should take this chance to step this up a notch? How about 4096?

omnibus/files/private-chef-cookbooks/private-chef/recipes/private_keys.rb
```diff
@@ -23,35 +23,43 @@
...
 
-webui_priv_key_path = "/etc/opscode/webui_priv.pem"
-new_webui = !File.exists?(webui_priv_key_path)
-webui_key = OpenSSL::PKey::RSA.generate(2048) if new_webui
+unless PrivateChef.credentials.exist?('chef-server', 'webui_key')
+  webui_key = OpenSSL::PKey::RSA.generate(2048)
+  PrivateChef.credentials.add('chef-server', 'webui_key',
+                              value: webui_key.to_pem,
+                              frozen: true)
+  PrivateChef.credentials.save
```
srenatus: This makes me wonder if `save` shouldn't be part of `add`. But I suppose this is fine.

marcparadise: I had the same conversation with myself.  

I think for safety, it should get committed to disk after every add. But that does not match what someone would expect based on the name if they just called `add`...

PR #1110, opened 2017-03-02
Don't render redis password to redis.config

stevendanna:
This uses a patched version of redis which can now read the password
from the environment.


stevendanna:
Waiting up omnibus-software PR

PR #1113, opened 2017-03-03
Clean up redis_password from lua configuration

stevendanna:
The main change here is cleaning up the redis password from the nginx
lua scripts. It also ensures all access to the redis password happens
via veil.


stevendanna:
This is built on top of the SPOOL-528/keys branch, will rebase once that is merged.

PR #1121, opened 2017-03-07
Ingest user-supplied passwords

stevendanna:
The ldap bind_password and (potentially) the postgresql superuser password can be set by the user in the configuration file.  For users who have policies against passwords in plaintext configuration files, we offer 2 new commands:

    chef-server-ctl set-ldap-bind-password
    chef-server-ctl set-db-superuser-password

which can be used to set those passwords on the command line.

NOTE: The passwords are still stored in plain-test in the private-chef-secrets.json file.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -519,9 +519,9 @@ module PrivateChef
           credentials.add("postgresql", "db_superuser_password",
                           value: PrivateChef["postgresql"]["db_superuser_password"],
                           frozen: true, force: true)
-        else
-          credentials.add("postgresql", "db_superuser_password", length: 100)
         end
+      else
+        credentials.add("postgresql", "db_superuser_password", length: 100)
       end
     end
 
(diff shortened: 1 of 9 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -10,6 +10,7 @@
 require "chef/json_compat"
 require "chef/mixin/deep_merge"
 require "veil"
+require_relative "./warnings.rb"
```
srenatus: [nit] I think we could drop the suffix here?

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -518,6 +512,48 @@ def gen_secrets_default(node_name)
...
 
+    def migrate_and_check_db_superuser_password
+      if PrivateChef["postgresql"]["external"]
+        if PrivateChef["postgresql"] && PrivateChef["postgresql"]["db_superuser_password"]
+          warn_if_cred_mismatch("postgresql", "db_superuser_password", "set-db-superuser-password")
+          credentials.add("postgresql", "db_superuser_password",
+                          value: PrivateChef["postgresql"]["db_superuser_password"],
+                          frozen: true, force: true)
+        else
+          credentials.add("postgresql", "db_superuser_password", length: 100)
```
stevendanna: Yikes! Yes, this should be outside of the next else!

PR #1122, opened 2017-03-07
moving mover to moving moving to remove secrets

marcparadise:
WIP that's winding down.  This desc to be updated shortly.


src/chef-mover/rebar.config
```diff
@@ -2,7 +2,7 @@
 %% ex: ts=4 sw=4 ft=erlang et
 
 %% Plugin usage
-{cover_enabled, true}.
+{cover_enabled, false}.
```
marcparadise: Will turn this back on, it was causing local rebar internal failures.

src/chef-mover/src/mover_eredis_sup.erl
```diff
@@ -21,7 +21,7 @@ start_link() ->
 init([]) ->
     Host = envy:get(mover, eredis_host, string),
     Port = envy:get(mover, eredis_port, integer),
-    Password = envy:get(mover, eredis_password, "", string),
+    Password = chef_secrets:get(<<"redis_lb">>, <<"password">>),
```
srenatus: this would return `{ok, Password}` on success -- it looks like we'd want to match that here

PR #1123, opened 2017-03-07
make private-chef::rabbitmq and partybus use veil

srenatus:
This doesn't check the password, just sets it unconditionally

Rationale: checking if it's ok, and setting it if not is probably just
as expensive as just setting it unconditionally.

Also, when upgrading to 3.6.6, it takes an extra password change to
upgrade hashes from md5 to sha256 (new default).

Also changes the execute's name to "ensure password for rabbitmq user
..." so it's a little less confusing to have this executed on each
reconfigure.


irvingpop:
I've noticed that the `not_if` guards on those password resources will prevent this from ever firing after bootstrap (unless rabbit's data is blown away)

srenatus:
@irvingpop new installs should be covered by the add_user calls -- they include a password, too. or am I missing something? :)

irvingpop:
My point was about password rotation -   the way those not_if guards are written, they passwords will never be changed.

srenatus:
@irvingpop I'm afraid I still don't get it. As of now, the passwords would be changed unconditionally (the change removes the `only_if`s). But this is part of why there's a WIP, so please disregard, I'll push some updates tomorrow)

irvingpop:
Ah, I was tripped up by the `not_if` statements in the add user blocks.   Your changes (removing the `only_if` statements) actually fix the issue I was seeing!   Thank you!

PR #1132, opened 2017-03-08
Do not configure secrets or content for analytics

marcparadise:
This is the first part of removing Analytics integration.

It just updates the release notes, and removes the recipe that configures actions.json and shared secrets.


marcparadise:
@srenatus I'd rather document it - deleting needed files without warning and with no fallback makes me nervous. While this release makes it so that new installs won't be compatible, leaving the files behind will - for a time at least - let people keep running their existing analytics installs without any nasty surprises.

PR #1137, opened 2017-03-10
Ingest S3 access key id/secret for bookshelf

srenatus:
Now, if for example chef-server.rb contains

    redis_lb['password'] = 'foo'

This would

1. be ingested in the secret store
2. trigger a warning if the value in the secret store is different from
   the one in the config
3. trigger a warning that it's in plain in chef-server.rb and should
   rather be taken from the secret store (i.e., removed from the config)

and could be set interactively via

    chef-server-ctl set-secret redis_lb password

or from the environment via

    REDIS_LB_PASSWORD=foobar chef-server-ctl set-secret redis_lb password

Replaces old ctl-commands by calls to set-secrets; with a whitelist.

Without the whitelisting, it would be quite easy to mistype the name of
a secret, and thus store it under the wrong key. Now, only anticipated
group/name combinations are allowed.

Note that the non-generic secret ingestion commands, i.e. those with
extra warnings and confirmations,

- postgresql/db_superuser_password
- rabbitmq/actions_password

are not retired, since `set-secret` does not support extras.

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/secrets.rb
+++ b/omnibus/files/private-chef-ctl-commands/secrets.rb
@@ -1,25 +1,49 @@
-add_command_under_category "set-ldap-bind-password", "Secrets Management", "Add or change LDAP bind password", 2 do
-  ARGV.delete("--yes") # remove --yes so users can give it in all cases in automation
-  bind_password = get_secret("BIND_PASSWORD", "LDAP bind password")
-  set_secret("ldap", "bind_password", bind_password)
-end
-
 add_command_under_category "set-db-superuser-password", "Secrets Management", "Add or change DB superuser password", 2 do
 
   confirm_continue!("WARN: Manually setting the DB superuser password is only supported for external postgresql instances")
-  password = get_secret("DB_PASSWORD", "DB superuser password")
-  set_secret("postgresql", "db_superuser_password", password)
+  password = get_secret("DB_PASSWORD", "DB superuser password", ARGV[3])
+  set_secret_("postgresql", "db_superuser_password", password)
 end
 
 add_command_under_category "set-actions-password", "Secrets Management", "Add or change the rabbitmq actions queue password", 2 do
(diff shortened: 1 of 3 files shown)


omnibus/files/private-chef-ctl-commands/secrets.rb
```diff
@@ -22,6 +22,16 @@
   set_secret("data_collector", "token", password)
 end
 
+add_command_under_category "set-s3-access-key-id", "Secrets Management", "Set or change the S3 access key id", 2 do
```
marcparadise: I'm starting to wonder if it wouldn't be a better UX to have something like:

```
chef-server-ctl set-secret NAME VALUE
And validate the names/provide a list of available names.
```

srenatus: :+1: Working on it -- either as part of this PR or the next 😃

omnibus/files/private-chef-ctl-commands/secrets.rb
```diff
@@ -42,16 +51,15 @@ def confirm_continue!(message)
   end
 end
 
-def set_secret(group, key, secret)
+def set(group, key, secret)
```
marcparadise: The only thing to keep in mind is that these all get `eval`d into the ChefServerCtl class - so `set` will become a member function of the class.  

If there's any chance of another command (there are none) or existing function of the same name (I'm less sure about this without looking closer), odd things will happen without clear indication of what went wrong. 

This one is generic enough that we're probably better off keeping it as `set_secret`, even if only to prevent accidental future collisions.

srenatus: `set_secret` did indeed clash with [this](https://github.com/chef/omnibus-ctl/blob/master/lib/omnibus-ctl.rb#L199) :D

srenatus:
@marcparadise [here's the extra warning](https://github.com/chef/chef-server/pull/1137/files#diff-c40634c5386397a69256fc985e3495d7R557)

PR #1139, opened 2017-03-10
Do not request secrets that don't exist

marcparadise:


srenatus:
An alternative would be to not have veil-env-helper go :boom: if the requested secret doesn't exist. But I think I prefer _this_. 😃

PR #1140, opened 2017-03-10
Remove sensitive attributes

marcparadise:
This PR is the result of the first passa through chef-server to identify and remove all secrets stored as node attributes, and any references to them.

Note: Merging this will break nightlies of reporting, which still expects at least some of this info in chef-server-running.json


omnibus/files/private-chef-cookbooks/private-chef/recipes/erchef_database.rb
```diff
@@ -3,12 +3,16 @@
 erchef = node['private_chef']['opscode-erchef']
 
 private_chef_pg_user erchef['sql_user'] do
-  password erchef['sql_password']
+  password PrivateChef.credentials.get('opscode_erchef', 'sql_password')
   superuser false
+  not_if {
```
srenatus: This is a debugging left-over, I think?

srenatus: ✅ I've removed it.

Issue #1111, opened 2017-03-03
chef-backend-ctl create-cluster creates broken cluster if ipv6 is turned on


voroniys:
chef-backend-1.3.2-1.el7.x86_64 on RHEL7
If ipv6 is enebled in the kernel 
chef-backend-ctl create-cluster 
produces broken cluster - leaderl is constantly failing with the following error message:
`2017-03-02_22:51:09.89234 Root: /opt/chef-backend/embedded/service/leaderl
2017-03-02_22:51:11.03097 {error_logger,{{2017,3,2},{23,51,11}},"Protocol: ~tp: register/listen error: ~tp~n",["inet_tcp",econnrefused]}
2017-03-02_22:51:11.03104 {error_logger,{{2017,3,2},{23,51,11}},crash_report,[[{initial_call,{net_kernel,init,['Argument__1']}},{pid,<0.473.0>},{registered_name,[]},{error_info,{exit,{error,badarg},[{gen_server,init_it,6,[{file,"gen_server.erl"},{line,344}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]}},{ancestors,[net_sup,kernel_sup,<0.460.0>]},{messages,[]},{links,[#Port<0.1119>,<0.470.0>]},{dictionary,[{longnames,true}]},{trap_exit,true},{status,running},{heap_size,610},{stack_size,27},{reductions,821}],[]]}
2017-03-02_22:51:11.03140 {error_logger,{{2017,3,2},{23,51,11}},supervisor_report,[{supervisor,{local,net_sup}},{errorContext,start_error},{reason,{'EXIT',nodistribution}},{offender,[{pid,undefined},{id,net_kernel},{mfargs,{net_kernel,start_link,[['leaderl@127.0.0.1',longnames]]}},{restart_type,permanent},{shutdown,2000},{child_type,worker}]}]}
2017-03-02_22:51:11.03169 {error_logger,{{2017,3,2},{23,51,11}},supervisor_report,[{supervisor,{local,kernel_sup}},{errorContext,start_error},{reason,{shutdown,{failed_to_start_child,net_kernel,{'EXIT',nodistribution}}}},{offender,[{pid,undefined},{id,net_sup},{mfargs,{erl_distribution,start_link,[]}},{restart_type,permanent},{shutdown,infinity},{child_type,supervisor}]}]}
[...]

PR #1141, opened 2017-03-13
Add preflight check for half-disabled ipv6 systems

srenatus:
As @voroniys noticed with one of our add-ons (#1111), there's a
hard-to-debug system configuration issue around epmd on ipv6-enabled
systems lacking `::1` on their loopback interface.

This preflight checks tries to catch that, and inform the user of their
possible ways forward.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_ipv6_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_ipv6_validator.rb
@@ -1,0 +1,65 @@
+#
+# Copyright:: Copyright (c) 2017 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative './preflight_checks.rb'
+
+class Ipv6PreflightValidator < PreflightValidator
+  def initialize(node)
+    super
+  end
+
+  def run!
+    verify_ipv6_for_lo
+  end
+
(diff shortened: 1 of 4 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_ipv6_validator.rb
```diff
@@ -0,0 +1,53 @@
...
+address.
+
+You must either pass `ipv6.disable=1` to your kernel command line,
+or ensure the loopback interface has an `::1` address by running
+
+    sysctl net.ipv6.conf.lo.disable_ipv6=0
+
+or
+
+    echo 0 > /proc/sys/net/ipv6/conf/lo/disable_ipv6
```
stevendanna: Since both of these do the same thing, I think we should just offer on set of instructions.  I'm partial to `sysctl` because it is more easily prepended with `sudo`

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_ipv6_validator.rb
```diff
@@ -0,0 +1,53 @@
...
+class Ipv6PreflightValidator < PreflightValidator
+  def initialize(node)
+    super
+  end
+
+  def run!
+    verify_ipv6_for_lo
+  end
+
+  def verify_ipv6_for_lo
```
stevendanna: I think we should put a comment about why we require IPv6 on lo.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_ipv6_validator.rb
srenatus: Alright! :+1:

srenatus:
@stevendanna it's hard to tell if this is common. However, before 18.3, we'd have not choked on it, with 18.3. I don't think it should be too common, but setting the epmd listen address env var also doesn't seem too common (so, we can't expect rabbitmq users to have hit this before very often, as an indicator)...

voroniys:
> template('audit/audit.rules.rhel7.extra.fragment.erb')
And they did - that is how I came to solution to disable ipv6.

srenatus:
@voroniys I'm afraid I don't get that reference -- what/where is that template? 😃

srenatus:
For the record, I've convinced myself using google that on Linux systems, the loopback interface is always named `lo`. 😄

PR #1143, opened 2017-03-13
Add specific config for webui auth tests

stevendanna:


stevendanna:
Actually, I think we might not need this.

stevendanna:
I think all we need to do to fix up pushy is this: https://github.com/chef/opscode-pushy-server/pull/160 with no changes locally.

PR #1146, opened 2017-03-14
ctl command secrets access cleanup

marcparadise:
Cleans up direct password access from ctl commands.

I have not yet verified the open source upgrade changes.


omnibus/files/private-chef-ctl-commands/chef-server-ctl
```diff
@@ -19,6 +18,11 @@ module Omnibus
       d
     end
 
+    def credentials
```
stevendanna: Not required in this PR, but we should clean up commands we already fixed to use this function rather than instantiating Veil objects.

PR #1154, opened 2017-03-15
Upgrade chef_secrets to latest

stevendanna:


marcparadise:
Could we remove the 'drop me' commits?

stevendanna:
@mp Yup, my plan was to squash this into a single commit once master of chef_secrets is the right thing to point at

omnibus/files/private-chef-cookbooks/private-chef/templates/default/sv-bookshelf-run.erb
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
 exec 2>&1
 
+export PATH=/opt/opscode/embedded/bin:$PATH
```
srenatus: ❓ There's nothing to gain from adding this to the `env` call below, is there?

stevendanna: Since we don't call `env` with `-i` I don't think there is really a difference and that line was getting pretty long.

PR #1159, opened 2017-03-17
Enhance API to handle global groups in local contexts

markan:
Support nonlocal group names

The server_admins group introduces an issue with ACLS and groups; we
have a global group in the context of a organization scoped
object. Not only do we have the potential of conflicts with a org
scoped group with the same name, we also lack a mechanisim to look up
groups that aren't in the org we're looking at.

This PR adds new syntax '::' to indicate scoping. ::GROUPNAME without
a prefix indicates a global (across multiple orgs) entity, while
ORGNAME::GROUPNAME refers to a group in an another org.

This should unblock the work around server admins, including improved
org creation.

There remain a few rough spots around error handling, and a few
opportunities to streamline and simplify the logic around 'lowering'
qualified names to authz_ids, but this should be enough to kickstart
the discussion.

This will need to be implemented in chef_zero to maintain feature parity.

**Ticket:** POOL-570

### DIFF
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
@@ -435,7 +435,7 @@ render_names_in_context(OrgId, ScopedNames, Context) ->
     GroupedScopedNames = group_by_key(ScopedNames),
     {Expanded, _Cache} = lists:foldl(fun(E, A) -> render_names_in_context_f(OrgId, E, A) end,
                                      {[], Context}, GroupedScopedNames),
-    lists:flatten(Expanded).
+    lists:sort(lists:flatten(Expanded)).
 
 %% We are in the same scope, omit qualifier
 render_names_in_context_f(OrgId, {OrgId, Names}, {Expanded, Context}) ->
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
@@ -24,24 +24,16 @@
 -include("oc_chef_types.hrl").
 
 -export([names_to_authz_id/3,
-         filter_by_error_type/2,
-         authz_id_to_names/3]).
-
--export([make_scoped_names/2,
-         parse_scoped_names/2,
-         parse_unscoped_names/2,
-         convert_ids_to_names/3,
+         initialize_context/2,
+         initialize_context/1,
+         org_id_to_name/1,
+         full_name/1,
 
          %% Used by oc_chef_group
          find_client_authz_ids/2,
          find_user_authz_ids/2,
          find_group_authz_ids/2,
-
-         make_sql_callback/0,
-         initialize_context/1,
-         initialize_context/2,
-         make_name/2,
-         org_id_to_name/1
+         convert_ids_to_names/3
         ]).
 
 -export([parse_scoped_name/3]).
@@ -75,6 +67,13 @@
                }).
 
 -type lookup_type() :: 'actor' | 'user' | 'client' | 'group'.
+
+%%
+%% Accessors for Scope name record
+%%
+-spec full_name(#sname{}) -> binary().
+full_name(#sname{full = FullName}) ->
+    FullName.
 
 %%
 %% Context is used to thread external state information through the system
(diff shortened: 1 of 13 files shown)


markan:
@chef/chef-server-maintainers 

I'm still working on dialyzer typespecs (coming shortly), but other than that this should be ready to roll.

stevendanna:
@markan Thanks for the heads up. Right now this is looking like it will probably land in release+1 since we've already got a ton in this release to test.

markan:
@ssd No worries ( @marcparadise has kept me up to date on your status ). I suspect this will go through a few review cycles before it's done anyways.  

This should unblock #886 & #935, so we'll probably want to bring this in early in the release cycle so we have a chance to tackle those, as well as maybe review whether #163 is still useful and relevant.

afiune:
Im rebasing and pushing since we have merged a couple of commits from this PR.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
```diff
@@ -0,0 +1,495 @@
...
+names_to_authz_id(Type, Names, MapperContext) ->
+    %% Lower to fully qualified orgname, name
+    ScopedNames = parse_scoped_names(Names, is_scoped_type(Type), MapperContext),
+    {ProperNames, Errors} = lists:foldl(fun filter_errors/2, {[], []}, ScopedNames),
+
+    OrgCache = init_org_name_cache(),
+
+    %% Map org names to org ids
+    {NamesWithOrgIds, Errors2, _} =
+        lists:foldl(fun(Name, Acc) -> lookup_org_id(Name, Acc, MapperContext) end,
```
afiune: I might be missing something but why do we pass the `MapperContext` to this lookup fun if it doesn't use it?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
```diff
@@ -0,0 +1,495 @@
...
+%% we are in a different scope, but it's the global scope. Use abbreviated version.
+render_names_in_context_f(_OrgId, {?GLOBAL_PLACEHOLDER_ORG_ID, Names}, {Expanded, Context}) ->
+    ENames = [ make_name(<<>>, Name) || Name <- Names],
+    { [ENames | Expanded], Context};
+render_names_in_context_f(_OrgId, {AnotherOrgId, Names}, {Expanded, Context}) ->
+    %% Design note: we drop missing orgs silently. Org deletion leaks many objects and we must
+    %% be robust to that.
+    case org_id_to_name(AnotherOrgId) of
+        not_found ->
+            {Expanded, Context};
```
afiune: @markan I think here is where we are dropping the errors when an organization doesn't exist.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
```diff
@@ -0,0 +1,495 @@
...
+    [{OrgName, Name} || Name <- Names].
+
+%%
+%% Expansion of authz ids into scoped names
+%% Takes {OrgName, Name} pairs in ScopedNames and returns
+%% list of names with scoping metacharacter inserted
+-spec render_names_in_context(binary(),[{binary(), [binary()]}], #context{}) -> [binary()].
+render_names_in_context(OrgId, ScopedNames, Context) ->
+    GroupedScopedNames = group_by_key(ScopedNames),
+    {Expanded, _Cache} = lists:foldl(fun(E, A) -> render_names_in_context_f(OrgId, E, A) end,
```
afiune: We do not use the `Context` in this fun. This is the only place we consume `render_names_in_context_f` and we drop it in `_Cache`. The fun itself is not using it so maybe we should remove it?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
```diff
@@ -0,0 +1,467 @@
...
+    [Names | Expanded];
+%% we are in a different scope, but it's the global scope. Use abbreviated version.
+render_names_in_context_f(_OrgId, {?GLOBAL_PLACEHOLDER_ORG_ID, Names}, Expanded) ->
+    ENames = [ make_name(<<>>, Name) || Name <- Names],
+    [ENames | Expanded];
+render_names_in_context_f(_OrgId, {AnotherOrgId, Names}, Expanded) ->
+    %% Design note: we drop missing orgs silently. Org deletion leaks many objects and we must
+    %% be robust to that.
+    case org_id_to_name(AnotherOrgId) of
+        not_found ->
```
stevendanna: Should we log here?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
```diff
@@ -0,0 +1,467 @@
...
+find_group_authz_ids(GroupNames, Context) ->
+    {AuthzIds, _Missing} = names_to_authz_id(group, GroupNames, Context),
+    AuthzIds.
+
+%%
+%% Lookup org name
+%%
+-spec org_id_to_name(binary()) -> not_found | binary().
+org_id_to_name(OrgId) ->
+    %% TODO maybe rework this; it bypasses a bunch of our statistics gathering code.
```
stevendanna: I think we should probably wrap this in a stats hero timer.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
```diff
@@ -0,0 +1,467 @@
...
+authz_id_to_names(client, AuthzIds, #context{db_callback_fun = CallbackFun}) ->
+    query_and_diff_authz_ids(find_client_name_in_authz_ids, AuthzIds, CallbackFun);
+authz_id_to_names(user, AuthzIds, #context{db_callback_fun = CallbackFun}) ->
+    query_and_diff_authz_ids(find_user_name_in_authz_ids, AuthzIds, CallbackFun).
+
+query_and_diff_authz_ids(_QueryName, [], _) ->
+    %% Sometimes the list of authz ids is empty; shortcut that and save a DB call.
+    {[], []};
+query_and_diff_authz_ids(QueryName, AuthzIds, CallbackFun) ->
+    case CallbackFun({QueryName, [AuthzIds]}) of
```
stevendanna: Probably a stats_hero timer around this as well.

afiune:
@chef/maintainers We are ready to review this change! cc/ @markan @stevendanna

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
afiune: an `erlang::error()` would be fine?

stevendanna: Let's use `lager` instead of `error_logger`

afiune: cool!

markan:
@stevendanna: Re: stats hero. I agree we want it, but there might be a few yaks there. We're in general inconsistent about gathering DB access stats in the authz code, and in many cases we've don't carry the ref for the request into the API deeply enough to have it handy. So there might be some yaks attached to wiring that up.

afiune:
After doing the spike to wrap the DB calls into a `stat_hero:ctime` we have discovered that it is indeed a change a bit big. Actually it is big enough that it would be a better idea to separate it from this change. At the end of the day we are making the same calls we were making before which wasn't being measured, so we are not introducing new ones. 

I have committed my changes to the branch: 
**=>** https://github.com/chef/chef-server/tree/afiune/revisit/stat_hero

In which I've got as far as I could without making a major refactor of the core modules in the app, things like `chef_sql:update()`, `chef_object:update()` and `chef_db:update()`.

markan:
👍 LGTM

Issue #727, opened 2016-02-01
pivotal.pem Missing on New Chef Server Install (HA Topology)


nathan-lochala:
When setting up a new high-availability chef-server installation, `chef-server-ctl reconfigure` fails claiming it is missing the /etc/opscode/pivotal.pem file. Below is the output when running `chef-server-ctl reconfigure` a second time on a new chef server installation.

```
Starting Chef Client, version 12.6.0
resolving cookbooks for run list: ["private-chef::default"]
Synchronizing Cookbooks:
  - enterprise (0.5.1)
  - chef-sugar (3.1.1)
  - private-chef (0.1.0)
  - apt (2.7.0)
  - yum (3.6.0)
  - openssl (4.4.0)
  - runit (1.6.0)
  - packagecloud (0.0.18)
Compiling Cookbooks...
/opt/opscode/embedded/cookbooks/cache/cookbooks/private-chef/libraries/preflight_checks.rb:139: warning: already initialized constant PreflightChecks::LINE_SEP
/opt/opscode/embedded/cookbooks/cache/cookbooks/private-chef/libraries/preflight_checks.rb:139: warning: previous definition of LINE_SEP was here
Recipe: private-chef::default
  * directory[/etc/opscode] action create
    - change owner from 'vagrant' to 'root'
    - change group from 'vagrant' to 'root'
    - restore selinux security context
  * directory[/etc/opscode/logrotate.d] action create
    - change owner from 'vagrant' to 'root'
    - change group from 'vagrant' to 'root'
    - restore selinux security context
[2016-02-01T18:41:20+00:00] FATAL:

-----------------------------------------------------------------------
BOOT007: The secrets file (/etc/opscode/private-chef-secrets.json) is present
         but the file /etc/opscode/pivotal.pem is missing.

         Ensure that private-chef-secrets.json is copied into /etc/opscode from the
         first Chef Server node that you brought online, then run
         'chef-server-ctl reconfigure' again.
-----------------------------------------------------------------------
```

thanizebra:
root@chs:~# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.3 LTS
Release:	14.04
Codename:	trusty
root@chs:~# 

root@chs:~# chef-server-ctl reconfigure
Starting Chef Client, version 12.7.0
resolving cookbooks for run list: ["private-chef::default"]
Synchronizing Cookbooks:
  - private-chef (0.1.0)
[...]

PR #1160, opened 2017-03-20
fix preflight check error message boot007, add fixture key removal to cleanup

srenatus:
This is to reduce the potential for confusion (see for example #727).

While this preflight check will probably change soon (there will be no need for copying the `pivotal.pem`), this is a small change, let's just get this merged anyways (if only for the cleanup addition).

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
@@ -78,27 +78,50 @@ class BootstrapPreflightValidator < PreflightValidator
     first_run? && all_creds_exist? && PrivateChef["postgresql"]["external"]
   end
 
+  def validate_sane_state
+    # If nothing is there, this is probably a first run
+    return true unless (secrets_file_exists? || pivotal_pem_exists?)
+
+    unless (new_secrets_layout? || old_secrets_info? || valid_mixed_state?)
+      fail_with err_BOOT006_invalid_secrets_state
+    end
+
+    true
+  end
+
   private
 
   def all_creds_exist?
     pivotal_key_exists? && secrets_exists?
   end
 
-  def pivotal_key_exists?
-    # on upgrades, we haven't had a chance to ingest the pivotal key, so we
-    # still need to check if the file is on disk
-    ::File.exist?("/etc/opscode/pivotal.pem") ||
-      PrivateChef.credentials.exist?('chef-server', 'superuser_key')
+
+  def new_secrets_layout?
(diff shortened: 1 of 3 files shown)


stevendanna:
On `current` shouldn't there be no pivotal.pem at all right now?

srenatus:
@ssd this is true. I'll remove the check.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_bootstrap_validator.rb
```diff
@@ -92,14 +92,8 @@ def pivotal_key_exists?
...
 
   def validate_sane_state
-    if pivotal_key_exists?
-      unless secrets_exists?
+    if pivotal_key_exists? && !secrets_exists?
         fail_with err_BOOT006_pivotal_with_no_secrets
-      end
-    else
-      if secrets_exists?
-        fail_with err_BOOT007_secrets_with_no_pivotal
```
marcparadise: Continuing the thought, the question there is.. is it an error?  I think it still can be, but it's a condition we might force to occur in the "ctrl+c to do something external and then run reconfigure again" instruction for HA.

srenatus: > it seems it is possible (through Ctrl+C) to end up in a situation where the secrets file exists containing no secrets other than the pivotal key, because we create it in the 'compile' phase of the chef client run.

This is true, but the next (successful) configure run would fix this. And it would be odd to have the preflight check _fail_ because of something fixable, wouldn't it?

> Continuing the thought, the question there is.. is it an error? I think it still can be, but it's a condition we might force to occur in the "ctrl+c to do something external and then run reconfigure again" instruction for HA.

I'm not sure I follow, would you want to _keep_ BOOT007? I'd assume we could blow this up, to have preflight checks around the existence of secrets in the store, but since stuff would be put there _later_, I'm not sure that there's a valid thing we could catch with it.

stevendanna:
@marcparadise @sr and I have updated this branch based our understanding of what the known-valid states are:

- No secrets exist (may be first run, no way to tell so we carry on)
- All secrets exist in the new format (private-chef-secrets.json exists and has the pivotal.pem)
- All secrets exist in the old format (private-chef-secrets.json exists, doesn't have pivotal.pem and is on disk
- A limited mixed format where the secrets file is in the new format (same as previous but technically private-chef-secrets.json is in an older format

Does this seem correct to you?  Our hope is the code is a bit clearer even though it got a big longer.

Overall, I think this is a HUGE indication of how much "clean up bootstrapping" needs to get in someone's backlog.

PR #1162, opened 2017-03-21
[dvm] Make sure to purge nodes/*.json

marcparadise:
This will ensure that node attributes are not carried over
between DVM loads.  THis prevents attributes that had been
set then cleared in config.yml from being retained.

### DIFF
--- a/dev/Vagrantfile
+++ b/dev/Vagrantfile
@@ -6,16 +6,32 @@ require "yaml"
 require "fileutils"
 load "dvmtools.rb"
 CS_VM_ADDRESS="192.168.33.100"
+# Used for external db configuration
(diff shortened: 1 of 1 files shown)


dev/Vagrantfile
```diff
@@ -15,7 +15,18 @@ DB_SUPERPASS="i1uvd3v0ps"
 LDAP_PASSWORD='H0\/\/!|\/|3tY0ur|\/|0th3r'
 
 nodes_dir = File.join(File.expand_path(File.dirname(__FILE__)), 'nodes')
-unless File.directory?(nodes_dir)
+if File.directory?(nodes_dir)
+	# This prevents attributes from previous runngs from getting
```
srenatus: `s/runngs/runs/`

PR #1167, opened 2017-03-22
[oc-id] Don't write application configuration to disk by default

stevendanna:
The user-configured applications previously resulted in files written
to disk with the oauth2 secret for that application.  These files are
how users currently integrate supermarket, compliance, and other
applications with oc-id.

To replace them, we've provided an oc-id-show-app ctl command that
will emit the same JSON that would have been in the file.


srenatus:
Note that the "by default" of this commit message is going to change with #1174.

PR #1172, opened 2017-03-23
Use the FD provider to facilitate root:root ownership of the secrets file

stevendanna:
The FD provider from veil/chef-secrets expects that the application has been passed a file descriptor containing the secrets it needs.  This file descriptor can be owned by the supervisor before we drop permissions, allowing us to keep the secrets store owned as root.

oc-id was the most problematic application to move to this method since ruby's `exec` method closes all file handles by default.

&lt;srenatus> *Update* Since it'd be a bigger change to get opscode-expander to use the FD method, it's still using the env provider. This allows us to keep the secrets file owned by root:root, but leaks the secret into procfs. However, that (synthetic) file only contains the one password expander needs, and is only readable by opscode, so from a security perspective, this leaves us in a much better place than having _the complete secrets file_ readable by opscode. &lt;/srenatus>


stevendanna:
Depends on https://github.com/chef/chef_secrets/pull/20

src/oc-id/config/initializers/00_secrets.rb
```diff
@@ -7,7 +7,17 @@ def self.get(group, name)
   end
 
   def initialize
-    @veil = Veil::CredentialCollection.from_config(provider: 'chef-secrets-fd')
+    # TODO(ssd) 2017-03-24: This should maybe go in chef-secrets itself
```
srenatus: :+1: I agree, but this is ok for now

PR #1174, opened 2017-03-23
Set insecure_addon_compat to true by default

stevendanna:
This change has the potential to break a lot of existing automation.
Setting this to true for now will give users time to update.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -107,7 +107,7 @@ module PrivateChef
 
   registered_extensions Mash.new
 
-  insecure_addon_compat false
+  insecure_addon_compat true
 
   class << self
     def from_file(filename)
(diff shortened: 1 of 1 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -571,7 +571,8 @@ def warn_if_cred_mismatch(opts)
     #
     def save_credentials_to_config
       credentials.legacy_credentials_hash.each do |service, creds|
-        next if service == "chef-server"
+        # Ignore secrets added by add-ons and the keys
+        next if PrivateChef[service].nil?
```
marcparadise: Will this do the right thing for opscode-erchef vs opscode_erchef?

stevendanna: I think any problems with that need to be cleared up before we get here, otherwise it would blow up with a NoMethodError [] on NilClass.  However, this is a good thing to double check on.

srenatus: @marcparadise I've discussed and double-checked this with @stevendanna, it's catching a bug we had earlier when you'd do the following:

1. install (a current package of) chef-server, set `insecure_addon_compat true`, reconfigure
2. install a current addon, run its reconfigure
3. run chef-server-ctl reconfigure again

it would then try to ingest _all the secrets_ from veil into `PrivateChef` (to end up writing them to running.json), and it would fail when a service didn't exist with a key in PrivateChef.

PR #1186, opened 2017-03-28
Add reporting sql_ro_password to known creds, remove keepalived default

srenatus:


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -794,8 +794,6 @@
 default['private_chef']['keepalived']['vrrp_instance_virtual_router_id'] = "1"
 default['private_chef']['keepalived']['vrrp_instance_priority'] = "100"
 default['private_chef']['keepalived']['vrrp_instance_advert_int'] = "1"
-# TODO mp 2017-03-10 this one too
-default['private_chef']['keepalived']['vrrp_instance_password'] = "sneakybeaky"
```
stevendanna: I think this reveals a bug in the current code.  We are still reading the vrrp_instance_password from the node attributes rather than `PrivateChef.credentials`.

stevendanna: I believe the same is true of the DRBD shared-secret.

srenatus: Ok, I was mislead by [this](https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb#L484-L485) then...

PR #1196, opened 2017-03-29
dvm - caching and minor refactor

marcparadise:
The biggest change here is to introduce host IO cache for the SATA device. 
This has improved the average VM 'up' time from 6m24s to 4m10s

Along the way I also refactored some chunks of the Vagrantfile that were repeated entirely too many times.

### DIFF
--- a/dev/Vagrantfile
+++ b/dev/Vagrantfile
@@ -114,24 +114,19 @@ def define_chef_server(config, attributes)
   config.vm.hostname = "api.chef-server.dev"
   config.vm.network "private_network", ip: IPS[:cs]
 
-  config.vm.provider :virtualbox do |vb|
-    vb.customize ["modifyvm", :id,
-                  "--name", "chef-server",
-                  "--memory", attributes["vm"]["memory"],
-                  "--cpus", attributes["vm"]["cpus"],
-                  "--usb", "off",
-                  "--usbehci", "off"
-    ]
-  end
+  vmattr = attributes["vm"]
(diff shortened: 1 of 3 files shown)


marcparadise:
With this change, experimentally changing the /host dir to be a normal vboxsf share shows improvement. It's still not as fast to load a project (oc_erchef: 1m30s using /host from host, 1m10s using vm hdd + sync)  - an improvement over the 4-5m time to build  from the host before caching was enabled.

It might be close enough that it's worth making the switch just to avoid continuing to need the external `sync`.

dev/Vagrantfile
```diff
@@ -114,24 +114,19 @@ def define_chef_server(config, attributes)
...
-                  "--cpus", attributes["vm"]["cpus"],
-                  "--usb", "off",
-                  "--usbehci", "off"
-    ]
-  end
+  vmattr = attributes["vm"]
+	customize_vm(config, memory: vmattr["memory"], cpus: vmattr["cpus"], name: "chef-server")
   if provisioning
+	  autoload = vmattr["omnibus-autoload"]
+		show_autoload_banner(autoload)
```
marcparadise: Spacing got weird here and elsewhere, will fix.

marcparadise:
One thing that isn't clear to me is if this is platform-specific - some early (3+ yo discussions) suggest *disabling* hostio cache on Mac OS for better performance.

marcparadise:
@stevendanna sync removal will be the next PR now that this is landed.

PR #1197, opened 2017-03-29
Only warn about passwords in c-s.rb when it matters.

marcparadise:

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -557,8 +557,9 @@ WARN2
       elsif pass_in_config
         unless PrivateChef["insecure_addon_compat"]
           warning = <<WARN
-#{config_key_desc} is configured in plain in chef-server.rb -- remove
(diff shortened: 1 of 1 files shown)


marcparadise:
Note on this one: we might want to look at using the node attribute here and elsewhere instead of `PrivateChef['insecure_addon_compat']`.  Otherwise we get into a place where we have three potential values:
 - attributes/default.rb - ignored currently
 - private_chef.rb - default value is set to true
 - chef-server.rb - overrides private_chef.rb and attributes/default.rb if set.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -555,10 +555,12 @@ def warn_if_cred_mismatch(opts)
 WARN2
         end
       elsif pass_in_config
-        warning = <<WARN
+        unless PrivateChef["insecure_addon_compat"]
+          warning = <<WARN
 #{config_key_desc} is configured in plain in chef-server.rb -- remove
```
stevendanna: should this say "plaintext" rather than "plain"?  Existing bug..but while we are here.

PR #1200, opened 2017-03-30
post 12.4 upgrades

marcparadise:
So far this upgrades enterprise-chef-common, removes the `= 1.6.0` constraint for the runit cookbook, and upgrades erlang dependencies.

🚧 🚧 🚧


omnibus/files/private-chef-cookbooks/Berksfile
```diff
@@ -1,9 +1,9 @@
 source "https://supermarket.chef.io"
 
-cookbook 'enterprise', :git => 'https://github.com/chef-cookbooks/enterprise-chef-common', :tag => '0.10.1'
+cookbook 'enterprise', :git => 'https://github.com/chef-cookbooks/enterprise-chef-common', :tag => '0.10.3'
 cookbook 'chef-ha-drbd', :path => './chef-ha-drbd'
 cookbook 'private-chef', :path => './private-chef'
 cookbook 'apt', '~> 2.0'
 cookbook 'yum', '~> 3.0'
```
tas50: If you're using a modern chef client you can remove yum and apt

marcparadise: ❗️ that's great, will do

PR #1207, opened 2017-04-03
Remove deprecated metadata

thommay:


thommay:
@marcparadise rebased

PR #1214, opened 2017-04-04
[CLOUD-296] Add required_recipe endpoint

ryancragun:
See [Chef RFC 89](https://github.com/chef/chef-rfc/blob/master/rfc089-server-enforced-recipe.md) for context.

Add the ability to serve a required recipe file to chef-clients.

* `/organizations/<orgname>/required_recipe` returns `404` for all organizations
  by default.
* `/organizations/<orgname>/required_recipe` returns `401` when the request is
  not made by a client from the requested org and the feature is enabled.
* `/organizations/<orgname>/required_recipe` returns the required recipe and
  `200` when the endpoint is enabled and requested by an authorized client.
* `required_recipe["enable"]` in `chef-server.rb` enables the required recipe
  feature.
* `required_recipe["path"]` in `chef-server.rb` specifies the recipe file to
  serve.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -41,6 +41,9 @@ default['private_chef']['default_orgname'] = nil
 
 default['private_chef']['fips_enabled'] = ChefConfig.fips?
 
+default['private_chef']['required_recipe']['enable'] = false
+default['private_chef']['required_recipe']['path'] = nil
+
 ###
 # Options for installing addons
 ###
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_required_recipe_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_required_recipe_validator.rb
@@ -1,0 +1,84 @@
+#
+# Copyright:: Copyright (c) 2017 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative './preflight_checks.rb'
+
+class RequiredRecipePreflightValidator < PreflightValidator
+  def initialize(node)
+    super
+    @required_recipe = PrivateChef['required_recipe']
+  end
+
+  def run!
+    return unless @required_recipe['enable']
+
+    verify_required_recipe_exists
+    verify_required_recipe_owner
+    verify_required_recipe_group
+    verify_required_recipe_mode
+  end
+
+  def stat
+    @stat ||= ::File.stat(@required_recipe['path'])
+  end
+
+  def verify_required_recipe_exists
+    unless ::File.exist?(@required_recipe['path'])
+      fail_with <<EOF
+Server enforced required recipe is enabled but the recipe file does not exist or
+is misconfigured. Please set the `required_recipe["path"] = /path/to/recipe` in
+`/etc/opscode/chef-server.rb` and run:
+
+    chef-server-ctl reconfigure
+EOF
+    end
+  end
+
+  def verify_required_recipe_owner
+    unless stat.uid == 0
+      fail_with <<EOF
+The required_recipe file must be owned by root. Please change the owner to root
+and reconfigure the Chef server:
+
+    chown root:root #{@required_recipe['path']}
+    chef-server-ctl reconfigure
+EOF
+    end
+  end
+
+  def verify_required_recipe_group
+    unless stat.gid == 0
+      fail_with <<EOF
+The required_recipe file must be in the root group. Please change the group to
+root and reconfigure the Chef server:
+
+    chown root:root #{@required_recipe['path']}
+    chef-server-ctl reconfigure
+EOF
+    end
+  end
+
+  def verify_required_recipe_mode
+    unless %w(600 644).include?(format('%o', stat.mode)[3..-1])
+      fail_with <<"EOF"
+The required_recipe must have a mode of 644 or 600. Please set a compatible mode
+and reconfigure the Chef server:
+
+    chmod 600 #{@required_recipe['path']}
+    chef-server-ctl reconfigure
+EOF
+    end
+  end
+end
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
@@ -49,8 +49,6 @@ end
   end
 end
 
-
-
 # Generate self-signed SSL certificate unless the user has provided one
 if (node['private_chef']['nginx']['ssl_certificate'].nil? &&
     node['private_chef']['nginx']['ssl_certificate_key'].nil?)
@@ -72,6 +70,17 @@ if (node['private_chef']['nginx']['ssl_certificate'].nil? &&
 
   node.default['private_chef']['nginx']['ssl_certificate'] = ssl_crtfile
   node.default['private_chef']['nginx']['ssl_certificate_key'] = ssl_keyfile
+end
+
+# Copy the required_recipe source into the nginx static root directory and
+# ensure that it's only modifiable by root.
+if node['private_chef']['required_recipe']['enable']
+  remote_file ::File.join(nginx_html_dir, 'required_recipe') do
+    source "file://#{node['private_chef']['required_recipe']['path']}"
+    owner 'root'
+    group 'root'
+    mode '0644'
+  end
 end
 
 # Generate dhparam.pem unless the user has provided a dhparam file
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
@@ -76,6 +76,26 @@
     }
     <% end -%>
 
+    location ~ "^/organizations/([^/]+)/required_recipe$" {
+    <% if node['private_chef']['required_recipe']['enable'] -%>
+      set $request_org $1;
+      access_by_lua_block { validator.validate("GET") }
+      types { }
+      default_type text/plain;
+      add_header Content-MD5 <%= Digest::MD5.base64digest(::File.read(node['private_chef']['required_recipe']['path'])) %>;
+      alias <%= ::File.join(@dir, "html", "required_recipe") %>;
+    <% else -%>
+      # This endpoint is unique because it is defined via nginx and is not
+      # handled by an upstream like oc_erchef. In order to make responses
+      # consistent between Chef server and chef-zero we'll always enable
+      # the location but explicitly return a 404 when the feature is disabled,
+      # rather than leave it undefined. If we were to leave it undefined any
+      # non-signed requests would be routed to the main index page and return
+      # a 200 instead of 404.
+      return 404;
+    <% end -%>
+    }
+
     <% if node['private_chef']['profiles']['root_url'] && node['private_chef']['data_collector']['token'] -%>
     # Compliance endpoint to forward profiles calls to the Automate API:
     #   /organizations/ORG/owners/OWNER/compliance[/PROFILE]
--- a/oc-chef-pedant/lib/pedant/command_line.rb
+++ b/oc-chef-pedant/lib/pedant/command_line.rb
@@ -149,7 +149,7 @@ module Pedant
                 principals acl containers groups association omnibus organizations
                 usags controls keys cookbook-artifacts license headers server-api-version
                 policies pedantic self-test api-v0 api-v1 object-identifiers
-                multiuser universe chef-zero-quirks user-keys client-keys)
+                multiuser universe chef-zero-quirks user-keys client-keys required-recipe)
       export_options(opts, tags)
     end
 
--- a/oc-chef-pedant/spec/api/required_recipe_spec.rb
+++ b/oc-chef-pedant/spec/api/required_recipe_spec.rb
@@ -1,0 +1,96 @@
+# Author:: Ryan Cragun <me@ryan.ec>
+# Copyright:: Copyright (c) 2017 Chef, Inc.
+
+require "pedant/rspec/common"
+
+describe "Required Recipe Endpoint", :required_recipe do
+  if Pedant::Config.required_recipe_enabled
+    describe "when required_recipe is enabled" do
+      describe "with a valid client" do
+        it "POST to /required_recipe returns 405" do
+          post(api_url("/required_recipe"), normal_client, payload: {}) do |response|
+            response.should look_like(status: 405)
+          end
+        end
+
+        it "GET to /required_recipe returns 200" do
+          get(api_url("/required_recipe"), normal_client) do |response|
+            response.should look_like(status: 200)
+          end
+        end
+      end
+
+      describe "with an invalid client" do
+        it "POST to /required_recipe returns 405" do
+          post(api_url("/required_recipe"), platform.bad_client, payload: {}) do |response|
+            response.should look_like(status: 405)
+          end
+        end
+
+        it "GET to /required_recipe returns 401" do
+          get(api_url("/required_recipe"), platform.bad_client, {}) do |response|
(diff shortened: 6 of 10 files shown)


omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -64,6 +64,14 @@
     }
     <% end -%>
 
+    <% if node['private_chef']['required_recipe']['enable'] -%>
+    location ~ /organizations/([^/]+)/required_recipe {
```
srenatus: I can't imagine how this would ever become an issue, but since we all _think_ that this location regex starts at the beginning of the request path, can you add a `^` there?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint_required_recipe.lua.erb
```diff
@@ -0,0 +1,28 @@
+<% unless node['private_chef']['required_recipe']['enable'] -%>
```
srenatus: Since the route won't be rendered if `node['private_chef']['required_recipe']['enable']` is false, is this block necessary?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint_required_recipe.lua.erb
```diff
@@ -0,0 +1,28 @@
...
+-- required_recipe mode is disabled to always return a 404
+
+if true then
```
srenatus: I'm not sure I understand what is gained from `if true then` here 😉

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
```diff
@@ -125,6 +130,7 @@
  'routes.lua',
  'dispatch_route.lua',
  'check_token_using_endpoint.lua',
+ 'check_token_using_endpoint_required_recipe.lua',
```
srenatus: I wonder if there would be a way to make the existing `check_token_using_endpoint.lua` generic enough to cover this case, too?

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
```diff
@@ -74,6 +72,13 @@
   node.default['private_chef']['nginx']['ssl_certificate_key'] = ssl_keyfile
 end
 
+# Link the required recipe file into the HTML static files root.
```
stevendanna: Should we consider checking the permissions on the file provided by the user?  That is, should we require that the file has restricted permissions to avoid a situation where a random user on the machine can modify content served to all chef-clients?

stevendanna:
Per @srenatus:

> Hmm this also has HA downsides, doesn’t it? you’d have to put that on each FE

Since this is being presented as a "super advanced" feature, perhaps we can just document that you have to keep them in sync; but things could get /very/ confusing if they were out of sync.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -41,6 +41,9 @@
 
 default['private_chef']['fips_enabled'] = ChefConfig.fips?
 
+default['private_chef']['required_recipe']['enable'] = false
```
stevendanna: I'm wondering if we want both of these?  We could alternatively check if path was non-nil.  If we do want both of these, we might want to add a check that path is set when enable is true to avoid a potentially confusing error message during reconfigure.

stevendanna:
I'm wondering if we want to send a header such as `Content-MD5` or similar that the client could use to verify the contents weren't modified/corrupted in transit?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/check_token_using_endpoint_required_recipe.lua.erb
ryancragun: I modified to always render the route and serve a 404 if it's disabled.

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
ryancragun: I thought about that, but all of them have very different post validation actions so making it generic might not be all that useful. Furthermore, I uncovered a bug with the validation endpoint where it'll return a 200 for _any_ properly formatted `GET` request. I'm working to resolve that issue before this PR moves ahead. That issue likely impacts the compliance profile endpoint as well.

```
root@api:/tmp# !curl
curl -kLI -X GET https://api.chef-server.dev/organizations/test/validate/i_like_bananas
HTTP/1.1 200 OK
Server: openresty/1.11.2.1
Date: Wed, 05 Apr 2017 17:35:22 GMT
Content-Type: text/html
Content-Length: 2474
Last-Modified: Tue, 04 Apr 2017 22:40:36 GMT
Connection: keep-alive
ETag: "58e420e4-9aa"
Accept-Ranges: bytes
```

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
ryancragun: I like this idea. Are we thinking readable only by `opscode` or `root`?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
ryancragun: Yeah, I can see why having two might not be necessary. I just wonder if having an on/off switch makes more usability sense than just having a path.

ryancragun:
@stevendanna @srenatus Sorry y'all, this is very much a WIP PR that I hadn't cleaned up, added documentation for, or added tests for. That said, I appreciate the feedback.

> When reviewing the RFC, it occurred to me that doing this in lua means we don't (currently) have access to the API versioning stuff (and also don't return API version information), which could make this harder to deprecate in the future if we ever replace it with a more robust feature.

> However, I can't think of a reason we couldn't move this into erchef at the point of wanting to deprecate it.

That's exactly the plan. I don't anticipate that happening any time soon though.

> Hmm this also has HA downsides, doesn’t it? you’d have to put that on each FE
Since this is being presented as a "super advanced" feature, perhaps we can just document that you have to keep them in sync; but things could get /very/ confusing if they were out of sync.

Yep, this was definitely a consideration but we've decided that the scope is limited to standalone installs. If we ever chose to support HA we'd probably move it into erchef and make the create API limited to server admins/pivotal.

> I'm wondering if we want to send a header such as Content-MD5 or similar that the client could use to verify the contents weren't modified/corrupted in transit?

I don't really see a way to have nginx add that header for static content. It should be easy enough to calculate it during the reconfigure and inject it into the nginx location block though.

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
ryancragun: > @srenatus  644 (root:root)

marcparadise:
Overall this looks okay, but I see this as something that belongs more properly in the erchef API in the longer term.  I know that we're in a situation where that may not be an option right now.  

To that end, I'd like to create a sustainable interface contract now so that we can address this in the future without further client updates being necessary. 

This should behave not as a recipe endpoint, but as a cookbook endpoint.  Chef Client already knows how to to load a cookbook in full.  Let's have our response look the same. Some modifications will still be needed  on the client (may not be a metadata.rb that's downloadable). 

On the nginx side, the endpoint would give a JSON response that looks like a versioned cookbook get.  It  would contain the storage URL for the recipe we want the client to grab. The recipe file we care about would come out of the backing file store (S3/bookshelf) , but still proxied via nginx and able to take advantage of caching if it's enabled.    This would be something we could more easily expand in the future, with support for cookbook files and dependencies if we wanted  to go that route.  

On the nginx side, this should  mean constructing the JSON response and generating the S3 url. 

On the client side, some further changes required because this isn't really a full cookbook, and won't have the files it expects (specifically metadata.rb).  

More importantly than the potential to expand it in the future though: I am increasingly less comfortable with the path of adding one-off behaviors into the reverse proxy instead of into the application that houses our API. 
If we want to have a separate lightweight API service for some of these tasks , we should design it and build it (even if in nginx) - but accumulating one-off `by_lua` behaviors isn't a sustainable approach. 

Separately: we will also need to add pedant coverage for this endpoint.

danielsdeleo:
To give a little more context here:

A few customers have asked for the ability to enforce certain items being in a node's run list at the server level, but this isn't that feature. There are a lot of possible designs that include having a side channel for a single cookbook, validating the run list at create/edit time to enforce required content, or having multiple run lists on the node. Each of these has variations on how the cookbook content could be uploaded/managed. On top of that, there are a lot of trade-offs related to attribute visibility between the various pieces and so on that would need to be decided upon in the client implementation. We would also need to look at the end to end development/testing/deployment story. To have a better chance at getting all of this right, we would want to engage some development partners and do a fair bit of research with them to work through the various trade-offs that various approaches could have. However, we haven't seen enough demand for such a feature that we moved it to the top of the priority list, and some folks (me especially) think that such a feature could discourage users from taking a more productive route to solving their problems, such as collaborating via shared Ci infrastructure.

Given all that, we're inclined to go hard on the YAGNI principle here. A lot of our strategy for long term scalability (in functionality terms) of the feature is to implement as little functionality as possible so it's not a huge cost to throw away and replace. **We know so little about what shape a replacement would look like that a lot of work we might try to do to anticipate future improvements would very likely end up as wasted work.**

In terms of the client work, Chef Client also already knows how to load single recipes in addition to its run list (see `chef-client -z path/to/recipe.rb`) so we are re-using that code path.

All that said, while we are targeting standalone installations for now, we do want to write the code so that an HA-compatible implementation could be written in the future without changing the configuration interface or Chef Client's download mechanism.

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
srenatus: Looks like you're right, and it's because that response is my favourite piece of html ever

    <h1>Are You Looking For the Chef Server?</h1>
    <p>Hello! It looks like you were trying to browse to your Chef Server but you haven't installed the Management Console.</p>

@ryancragun are you looking into this or do you want me to file a big somewhere? 😉

srenatus: @ryancragun nevermind, I've found the JIRA card. Thanks :)

marcparadise:
@ryancragun @dan: I am sorry - I should have done a much better job at not being an ass in my initial comment.  Next time our paths cross, :beer: (or other drink of choice) is on me.

ryancragun:
@marcparadise @stevendanna @srenatus @sdelano 

Howdy y'all. I think that this and https://github.com/chef/chef-server/pull/1226 are ready for 👀. All of the feedback should be addressed and an integration test plan has been formed on #1226.

Thanks!

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
ryancragun: Circling back on this, I definitely think sticking with both is ideal. If we ever move to a place where this file is stored in a database then we're unlikely to have a `path`. In that scenario we'd need would be left without a toggle.

sdelano: But if the file is stored in the database, then we'll need no config because if it's there it should be returned and if it's not it wont. But that's jumping to implementation details.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -76,6 +76,19 @@
...
+    location ~ "^/organizations/([^/]+)/required_recipe$" {
+    <% if node['private_chef']['required_recipe']['enable'] -%>
+      set $request_org $1;
+      access_by_lua_block { validator.validate("GET") }
+      types { }
+      default_type text/plain;
+      add_header Content-MD5 <%= Digest::MD5.base64digest(::File.read(node['private_chef']['required_recipe']['path'])) %>;
+      alias <%= ::File.join(@dir, "html", "required_recipe") %>;
+    <% else -%>
+      return 404;
```
sdelano: This will return 404 if the location is not configured. It seems a bit more straightforward to have the whole location block within the `if` block and just not render it when it's not enabled.

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
sdelano: This is all handled by the preflight check. Thanks!

sdelano:
@ryancragun I left 2 very minor comments related to dual config (`enabled` and `recipe_path`) and how we render the nginx config. In the end, the functionality remains the same, so I'll leave it up to you on whether you'd like to address those two issues. If not, merge away!

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
markan: Putting the location inside also seems to be more in keeping with the patterns elsewhere.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
ryancragun: I dunno, the file existing and the feature being enabled/disabled are two different things. But yeah, YAGNI. The idea here is to have an implementation agnostic switch and the `path` is definitely not agnostic.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
ryancragun: 👍

stevendanna:
@markan 

> I think we're probably better off all around copying the required recipe over to a secure location than guarding it with checks, but if we are diligent with guards that could be made to work.

The only problem I see is that the user might want to change the content, so it will always be the case that we need to read it in from the user-provided location so we will always want that user-provided location to have safe permissions.

ryancragun:
Checking the permissions at reconfigure time is a good idea. I agree that copying it instead of linking is needed. The content md5 header has to be regenerated if you intend to change the file content, a case the previous link method did not address. By validating and copying we ensure that:
* the file is always in a directory we own
* can only be changed by root
* source has to use a file owned by root
* that the md5 header is always in sync

markan:
👍  LGTM

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_required_recipe_validator.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def verify_required_recipe_mode
+    mode = ::File.stat(@required_recipe['path']).mode
+    unless format('%o', mode)[3..-1] == '644'
+      fail_with <<"EOF"
+The required_recipe must have a mode of 644. Please set mode to 644 and
+reconfigure the Chef server:
+
+    chmod 644 #{@required_recipe['path']}
```
stevendanna: Since we are now copying the contents, then we might consider making this 600 and/or making the check accept either 644 or 600.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_required_recipe_validator.rb
```diff
@@ -0,0 +1,81 @@
...
+root and reconfigure the Chef server:
+
+    chown root:root #{@required_recipe['path']}
+    chef-server-ctl reconfigure
+EOF
+    end
+  end
+
+  def verify_required_recipe_mode
+    mode = ::File.stat(@required_recipe['path']).mode
```
stevendanna: Very minor, but we could save a few calls to stat by moving this into a memoized function.

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
```diff
@@ -74,6 +72,17 @@
   node.default['private_chef']['nginx']['ssl_certificate_key'] = ssl_keyfile
 end
 
+# Copy the required_recipe source into the nginx static root directory and
+# ensure that it's only modifiable by root.
+if node['private_chef']['required_recipe']['enable']
+  remote_file ::File.join(nginx_html_dir, 'required_recipe') do
+    source "file://#{node['private_chef']['required_recipe']['path']}"
```
stevendanna: Interesting! I've always used `file` with `content ::File.read(blah)`, didn't know file_urls worked here.

PR #1215, opened 2017-04-05
[do not merge yet] post 12.14 upgrades

marcparadise:
This PR does the following:

* upgrades erlang deps to their latest versions for bifrost/bookshelf/erchef
* updates release # to 12.14.1 ahead of hte enxt release
* pulls in updated omnibus-software and berksfile
* additional changes related issues uncovered with those updates.


omnibus/Gemfile
```diff
@@ -13,7 +13,7 @@ end
 group :omnibus do
   gem 'omnibus', git: 'https://github.com/chef/omnibus'
   gem 'omnibus-software', git: 'https://github.com/chef/omnibus-software'
-  gem 'license_scout', git: 'https://github.com/chef/license_scout'
+  gem 'license_scout', git: 'https://github.com/chef/license_scout', branch: "mp/yum-epl-and-compat-resource-license"
```
afiune: any reason why we point to your branch?

omnibus/config/software/oc_id.rb
```diff
@@ -21,6 +21,7 @@
 license "Apache-2.0"
 license_file "LICENSE"
 
+dependency "nokogiri" # for
```
afiune: I see you tried to put a comment to explain why we are adding this. Is this part and intent to fix the build process?

omnibus/config/software/server-complete.rb
```diff
@@ -0,0 +1,67 @@
+name "server-complete"
+
+skip_transitive_dependency_licensing true
+
+license :project_license
+
+# global
+dependency "libiconv"
```
stevendanna: This makes me a bit nervous as I feel like this means we have under-specified software definitions here.  Perhaps a comment about why this is at the top level?

omnibus/config/software/oc_id.rb
srenatus: I also think adding nokogiri here means we'll probably end up with two versions of it? Since it's getting installed in the bundle install with the bundle-defined version, and the software definition has another version (latest on rubygems?).

On the other hand, [copying those here](https://github.com/chef/omnibus-software/blob/master/config/software/nokogiri.rb#L27-L30) might be more ugly than just installing nokogiri twice 🤔

srenatus:
I take it mover doesn't need to depsolve anything, so the depsolver bits that got dragged over from erchef when mover was created got nuked?

The two omnibus software definition related comments aside, let's get this (maybe squashed and) merged. :+1:

Issue #1220, opened 2017-04-06
Failure upgrading RabbitMQ from Migration 1.30 to 1.31


StuporHero:
I'm running chef-server-ctl upgrade, and it's failing when partybus tries to change the password of the chef user on the RabbitMQ instance during the migration from 1.30 to 1.31. Below is the log.

```
ok: down: bookshelf: 1606s, normally up
ok: down: nginx: 1605s, normally up
ok: down: oc_bifrost: 1605s, normally up
ok: down: oc_id: 1603s, normally up
ok: down: opscode-chef-mover: 1750s, normally up
ok: down: opscode-erchef: 1602s, normally up
ok: down: opscode-expander: 1599s, normally up
ok: down: opscode-reporting: 3164s, normally up
ok: down: opscode-solr4: 1599s, normally up
ok: down: postgresql: 0s, normally up
timeout: run: rabbitmq: (pid 22670) 765s, want down
ok: down: redis_lb: 1s, normally up
ok: run: postgresql: (pid 28855) 0s
[private-chef-upgrade] - Latest Migration Available: 1.32
[private-chef-upgrade] - Migrations to Run: [#<Partybus::MigrationFile:0x0000000156e018 @path="/opt/opscode/embedded/upgrades/001/031_migrate_rabbitmq_password.rb", @major=1, @minor=31>, #<Partybus::MigrationFile:0x0000000156cc40 @path="/opt/opscode/embedded/upgrades/001/032_oc_id_logdir_perms.rb", @major=1, @minor=32>]
[private-chef-upgrade] - Current Migration Version: 1.30
[private-chef-upgrade] - Starting Migration 1.31
[private-chef-upgrade] -        Starting Services ["rabbitmq"]
down: rabbitmq: 113s, normally up; run: log: (pid 570) 17401s
ok: run: rabbitmq: (pid 29386) 0s
run: rabbitmq: (pid 29386) 66s; run: log: (pid 570) 17525s
[private-chef-upgrade] - Sleeping for 15 seconds while services comes online...
[private-chef-upgrade] -        Running Command: /opt/opscode/embedded/bin/rabbitmqctl change_password chef aafc8632588d5dede202bd04e1d07845051182c927e324e41ff41a234d7b8961a0ed62a5c84c992430f3ef10f9e439540c94 with options {}
Changing password for user "chef" ...
Error: {noproc,{gen_server2,call,
                            [worker_pool,{next_free,<10245.172.0>},infinity]}}
[...]

PR #1221, opened 2017-04-06
Don't log rabbitmq password during migration 1.31

stevendanna:


srenatus:
:+1: Another fix caused by #1220. 🎉

PR #1224, opened 2017-04-07
Update oc-id

afiune:


stevendanna:
This looks reasonable.  How confident are we in our tests here?  It might be good to test this end to end and make sure that users are still able to log into oc_id with this change in place.

afiune:
I would be +1 on that thought @stevendanna - I would like to see this end-to-end as well.

I wonder if we should update (https://github.com/chef/chef-server/blob/master/src/oc-id/lib/chef_resource.rb#L13) to use `::Chef::ServerAPI.new` instead? 🤔

stevendanna:
@afiune I think that would be good to convert as it will future proof us against it's Chef 13 removal.

PR #1226, opened 2017-04-10
[CLOUD-306] Add validate upstream location to nginx

ryancragun:
The authorization validation for proxied `data-collector` and `compliance-profile`
requests utilized recursive location matching that made understanding the flow of
requests difficult, made anchoring location regular expressions impossible, and
required that all validated requests proxy to an upstream server.

This change introduces an `organizations/org/validate` nginx location that is only
accessible to local requests. This will allow us to:
  * Use anchored location regular expressions for validated endpoints
  * Not require recursively matching locations during validation
  * Allow using the `/validate` endpoint for requests that don't
  proxy to a different upstream, e.g.: `required_recipe`
  * Utilize a shared lua library for authorizing requests

The previous flow:
```
chef-client POST /organizations/org/data-collector
nginx location match for /organizations/org/data-collector
access_by_lua_file ->
  set $upstream to http://opscode_erchef
  sub-request POST /orgs/org/validate/orgs/org/data-collector
    nginx location match for /organizations/org/data-collector
    validation is bypassed
    proxy_pass to $upstream
      erchef does validation
  exit if erchef validation response is not 200
  configure data collector headers
  set $upstream to http://data-collector
proxy_pass to $upstream
```

The updated flow:
```
chef-client POST /organizations/org/data-collector
nginx location match for ^/organizations/org/data-collector$
access_by_lua_block ->
  sub-request POST /orgs/organizations/validate/orgs/org/data-collector
    nginx location match for ^/organizations/org/validate
    proxy_pass to http://opscode_erchef
      erchef does validation
  exit if erchef validation response is not 200
set data collector headers
proxy_pass to upstream http://data-collector
```

Signed-off-by: Ryan Cragun <me@ryan.ec>

There were no existing pedant or integration tests for these routes, presumably
because they are authorized proxied requests to Automate API's that are not
included in Chef server, therefore tests would require either stubbed Automate
API's that could change or require Chef server to be configured with an
Automate installation in a shared test environment.

We'd like to temporarily hold off implementing integration tests as the
ACC integration test framework is currently undergoing a big transition and the
work would likely end up being thrown away in a few weeks. The proposed plan
would require a pedant test for [CLOUD-296](https://github.com/chef/chef-server/pull/1214) which would test the
`/required_recipe` endpoint. That would give us coverage of `GET` requests using
the `/validate` location and the shared validation library. It's not an ideal
long term solution as the `/data-collector` and `/compliance-profiles` endpoints
would not be fully covered. To solve that we will implement integration tests
for both Chef Automate (CLOUD-310) and Chef server (CLOUD-311) when they are live
in the ACC environment.


danielsdeleo:
LGTM and I am on board with the plan for testing. But I'm not a maintainer here, FYI.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -71,9 +83,11 @@
     #   /compliance/organizations/ORG/owners/OWNER/compliance[/PROFILE]
     location ~ (?:/compliance)?/organizations/([^/]+)/owners/([^/]+)/compliance(.*) {
       set $request_org $1;
-      set $new_uri "/compliance/profiles/$2$3";
-      access_by_lua_file '/var/opt/opscode/nginx/etc/scripts/check_token_using_endpoint_compliance.lua';
-      proxy_pass $upstream;
+      access_by_lua_block { validator.validate("GET") }
+      proxy_set_header x-data-collector-token $data_collector_token;
```
srenatus: I keep wondering what the data collector token has to do with compliance profile storage, but this is unrelated to this change.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -71,9 +83,11 @@
     #   /compliance/organizations/ORG/owners/OWNER/compliance[/PROFILE]
     location ~ (?:/compliance)?/organizations/([^/]+)/owners/([^/]+)/compliance(.*) {
```
chris-rock: I preparation for the next version of our compliance backend, I like to make the url more flexible. Could we just pass everything to automate and let automate handle url mapping?

cc @alexpop

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
chris-rock: Chef Automate is verifying the data collector token in nginx. Otherwise the response is denied.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
chris-rock: Discussed with @alexpop we're updating the url in another PR. We like to simplify the url handling and migrate all url handling to Automate

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
alexpop: the DC token is used by Automate the authenticate the request:

```
[root@ap-ca6 ~]# curl -k "https://localhost/compliance/profiles"
<html>
<head><title>401 Authorization Required</title></head>
<body bgcolor="white">
<center><h1>401 Authorization Required</h1></center>
<hr><center>nginx</center>
</body>
</html>

[root@ap-ca6 ~]# curl -k "https://localhost/compliance/profiles" -H "x-data-collector-auth: version=1.0" -H "x-data-collector-token: $AUTOMATE_DC_TOKEN" -H "chef-delivery-enterprise: default"
{"admin":{"linux-baseline":{"id":"linux-baseline","owner_id":"admin","owner":"admin","name":"linux-baseline","title":"DevSec Linux Security Baseline","maintainer":"Hardening Framework Team","copyright":"Hardening Framework Team","copyright_email":"hello@hardening.io","license":"Apache 2 license","summary":"Test-suite for best-preactice Linux OS hardening","version":"2.0.1"}}}
```

The use-cases that I remember:
* chef-server -> Automate auth 
* chef-client solo -> Automate auth

srenatus: @alexpop @chris-rock thanks for reminding me. What I meant to point out was that this is *another thing* we use _the data collector token_ for, and it's got not much to do with data collection 😉

chris-rock: @srenatus Yes, I agree. We need to find a way to make the internal communication safe.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -57,10 +66,13 @@
     <% end -%>
 
     <% if node['private_chef']['data_collector']['root_url'] && node['private_chef']['data_collector']['token'] -%>
-    location ~ /organizations/([^/]+)/data-collector {
+    location ~ "^/organizations/([^/]+)/data-collector$" {
```
sdelano: we post to `/data-collector/v0`, IIRC, we also do a GET to that endpoint to check that it's there. If I'm reading this correctly, this will break posts to `/v0` and will also prevent GETs as well

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
markan: It sounds like we're going to match something like "^/compliance" and ignore the details going forward; let's make sure we '^' anchor the new regular expression.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/validator.lua.erb
```diff
@@ -0,0 +1,33 @@
...
+      method = HTTP_REQ_MAP[ngx.req.get_method()],
+      always_forward_body = true,
+      copy_all_vars = true
+    }
+  )
+
+  if res.status == ngx.HTTP_OK then
+    return
+  else
+    ngx.exit(res.status)
```
markan: Is it worth returning the body as well? We do try to return some useful error messages, even from our signing validation routines.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
sdelano: Awesome. I was totally incorrect here:

During a `chef-client` run, we do a POST to the `data-collector` endpoint at the start of the run. If that returns 404, then we disable the reporter ([link](https://github.com/chef/chef/blob/master/lib/chef/data_collector.rb#L125-L133))

Also, when using the chef-server as a data-collector endpoint, we leave off the v0 portion of the URL ([link](https://github.com/chef/chef/blob/b2b87c35ceb2d6e05c2be2f381b3e3dbe86a3dd1/chef-config/lib/chef-config/config.rb#L874-L880))

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -57,10 +66,13 @@
...
       set $request_org $1;
-      access_by_lua_file '<%= @script_path %>/check_token_using_endpoint.lua';
-      proxy_pass $upstream;
+      access_by_lua_block { validator.validate("POST") }
+      proxy_set_header x-data-collector-token $data_collector_token;
+      proxy_set_header x-data-collector-auth "version=1.0";
+      rewrite ^ <%= URI.parse(node['private_chef']['data_collector']['root_url']).path %> break;
+      proxy_pass <%= URI.parse(node['private_chef']['data_collector']['root_url']).scheme %>://data-collector;
```
sdelano: I'm assuming that we're potentially using HTTPS here because the Chef Server and the Automate server could be on different machines or in different networks even? We don't typically do the double-https dance, havning nginx terminate it once, but I think this makes sense here, I just want to be clear why.

sdelano: And perhaps we should comment that fact.

ryancragun: When the user configures their upstream data-collector we _dont_ enforce https, therefore the scheme changes depending on whether it's https or not. Since the upstream data collector (usually Automate) and Chef server can have (and in most cases do have) different SSL config, I think this is pretty much unavoidable.

ryancragun:
@markan We now return the validation sub-request message body

before
```html
 <html>
<head><title>401 Authorization Required</title></head>
<body bgcolor="white">
<center><h1>401 Authorization Required</h1></center>
<hr><center>openresty/1.11.2.1</center>
</body>
</html>
```

after
```json
 {"error":["Invalid signature for user or client 'ryan'"]}
```

PR #1227, opened 2017-04-11
fix typo => fix actions_password migration for external rmq

srenatus:


stevendanna:
gif-keyboard-4418692930698038724

PR #1230, opened 2017-04-14
Tests run in proxied systems

markan:
NOTE: Rebase and merge after CLOUD-294/fix_solr_for_pedant (#1233) lands, as it contains and depends on it.

Passes pedant in the CI environment. (But we have some intermittent test failures on ppc64le Ubuntu 12 testers)

This PR encompasses a series of changes to allow chef-server-ctl test to operate when the main chef-server endpoint is proxied behind another name, such as the automate/chef server all in one install. 

When opscode-erchef base_resource_url is set differently from the server address that nginx is listening at (as is often the case when proxying) we construct the expected values in pedant incorrectly, using the server address when we should be using base_resource_url

In some cases :host_headers isn't sufficient, and we need to be flexible on what we accept for input.

I've broken it up into a series of commits for readability. I'd especially like feedback on the approach of modifying PedantHashComparator; the alternate seems to be modifying a large number of tests to disambiguate the URI we use to access the system and the resource we expect to return. See the CLOUD-294/fix_tests_mechanical branch for a sense of what that might look like.

### DIFF
--- a/oc-chef-pedant/lib/pedant/rspec/matchers.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/matchers.rb
@@ -69,19 +69,7 @@ module RSpec
         @expected.keys.all? do |key|
           spec = @expected[key]
           value = actual[key]
-
-          case spec
-          when Regexp
-            spec =~ value
-          when Array
-            array_matches?(spec, value)
-          when Hash
-            PedantHashComparator.new(spec, @mode).matches?(value)
-          when Proc then
-            spec.call(value)
-          else
-            spec == value
-          end
+          compare_value(spec, value)
         end
 
       end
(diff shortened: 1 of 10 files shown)


markan:
@chef/chef-server-maintainers

oc-chef-pedant/lib/pedant/rspec/matchers.rb
```diff
@@ -95,12 +83,38 @@ def array_matches?(spec, value)
         # Note this is going to be slow for large arrays,
         # though our actual usage shows we don't have any arrays on a
         # scale that would matter.
-        return true if spec.all? { |s| value.include? s }
-        # Our comparisons are more complex than simple value
-        spec.each_with_index  do |newspec, x|
-          return false unless PedantHashComparator.new(newspec, @mode).matches? value[x]
+        return true if spec.all? do |s|
```
srenatus: [nit] isn't this `return true if ...` followed by `return false` equivalent to ending the method with

    spec.all? do ...

? 😉

oc-chef-pedant/lib/pedant/rspec/matchers.rb
```diff
@@ -95,12 +83,38 @@ def array_matches?(spec, value)
...
+          spec.call(value)
+        else
+          spec == value
+        end
+      end
+
+      #
+      def normalize(candidate)
+        if candidate.is_a?(String)
+          candidate.sub(/#{Pedant::Config.pedant_platform.server}/, Pedant::Config.pedant_platform.base_resource_url)
```
stevendanna: Is there an advantage of coercing both the expected and configured value at the time of the comparison versus changing the code that generates the expected URL to always be correct:

https://github.com/chef/chef-server/pull/1230/files#diff-c681a9689ad6f141cb970701a07cbbc3R71

One of my worries is that since we coerce both the expected and actual value, it will be hard to catch cases were we are generating the wrong url.

stevendanna:
Whelp, I just finally read this part of your commit message fully:

> the alternate seems to be modifying a large number of tests to disambiguate the URI we use to access the system and the resource we expect to return. See the CLOUD-294/fix_tests_mechanical branch for a sense of what that might look like

That is indeed a large change; however, what I don't quite see is why we would need to add the new `resource_url` function vs changing `api_url`.  If `api_url` was changed to do the right thing, your example branch would probably be a much smaller change.

I'm OK with merging this PR given that if fixes the problem, it just seems to me that we have a number of config items and functions trying to do the same thing in these pedant tests.

oc-chef-pedant/lib/pedant/rspec/matchers.rb
stevendanna: You already commented on this in your PR message, apologies for not seeing that. Follow-up comment is below.

marcparadise:
I'm a little uncomfortable with this because it adds another level of hidden behaviors that aren't apparent when trying to track down failures - but given the situation, it's a reasonable fix.

markan:
I should go a little deeper into the use case to explain why I ended up here. 
When we're operating in a proxied configuration, the configuration value
opscode_erchef['base_resource_url'] might be different than the actual host/port that erchef
is listening on. For example, in an automate/opsworks configuration the server might be
listening on FQDN:8443), but the correct URI to return would be
based off of the automate FQDN:443. However, the chef server tests universally use the listening
port for the server as the base for the API resources, and worse, they conflate the API to
access and the resource expected.

So we get many, many failures where we construct an expected return value using the listen
address (e.g FQDN:8443), but the return has resources correctly constructed using
base_resource_url.

Properly disambiguating that usage requires a large number of changes, and is fragile in that
we don't test proxied configurations in our mainline release path.

Pedant could be configured to address all API calls in tests to the proxied base url, but
sometimes we need the un-proxied URL (for example, when we are testing endpoints that are
not externally exposed).

Normalizing out the difference is the least fragile change at the cost of some possibly
confusing behavior.

A possible enhancement would be to add 'superstrict' as an option for the comparator, where
we would suppress this normalization and do a strict equality test.

PR #1232, opened 2017-04-14
Pull latest omnibus to fix SLES (Suse) x86 builds

markan:
We were failing when building keepalived on x86 SUSE. Previously we had only been building 390 SUSE, but we just added x86 as well.

@chef/chef-server-maintainers


danielsdeleo:
I'm not a maintainer here but the omnibus change made sense. 👍

markan:
I'm merging based on the minor nature of the change.

PR #1233, opened 2017-04-14
Fix chef-server-ctl test for opsworks

markan:
Passes pedant in the CI environment. Should be good to merge.

The AWS Opsworks Automate offering has Chef Server using Automate's
elasticsearch instead of the built in Solr. This is done via the
external database feature of Chef Server.

This is implemented by setting the following in chef_server.rb:
opscode_solr4['external_url'] = http://localhost:8080/elasticsearch'

And then having nginx proxy that to the elasticsearch backend.

However that breaks pedant because  it uses the solr_root helper, which
would take the above external_url and render http://localhost:8080.

If we use the solr_url helper instead, the full URI is rendered, and
the tests work.

### DIFF
--- a/oc-chef-pedant/lib/pedant/config.rb
+++ b/oc-chef-pedant/lib/pedant/config.rb
@@ -118,8 +118,8 @@ module Pedant
     maximum_search_time 65
 
     # Default URL for to commit/refresh before a search
-    search_commit_url "/solr/update?commit=true"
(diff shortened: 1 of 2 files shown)


markan:
@chef/chef-server-maintainers

markan:
The changes pass in pedant now @sdelano

sdelano:
@markan - approved!

PR #1238, opened 2017-04-20
Return 401 for requests from the future

stevendanna:
The real fix is upstream in chef_auth, this pulls in that fix and adds
a pedant test for it.  Previously, we would return 200 for any request
from the future, even if it was beyond the allowed time skew.


stevendanna:
Depends on https://github.com/chef/chef_authn/pull/34

PR #1240, opened 2017-04-21
Fix "chef-server-ctl reconfigure" for ldap configuration changes

jaymalasinha:
Worked with @stevendanna on this to create the set_ldap_enabled
validation check before generating the config.

Tested this successfully in a local dev environment

@stevendanna

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -780,9 +780,13 @@ EOF
       credentials.save if did_something
     end
 
+    def set_ldap_enabled
+      PrivateChef["ldap"]["enabled"] = !( PrivateChef["ldap"].nil? || PrivateChef["ldap"].empty? )
+    end
 
     def generate_config(node_name)
       assert_server_config(node_name) if server_config_required?
+      set_ldap_enabled
       gen_secrets(node_name)
       migrate_keys
 
@@ -810,9 +814,7 @@ EOF
 
       generate_config_for_topology(PrivateChef["topology"], node_name)
 
-      unless PrivateChef["ldap"].nil? || PrivateChef["ldap"].empty?
-        gen_ldap
-      end
+      gen_ldap if PrivateChef["ldap"]["enabled"]
 
       generate_hash
     end
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -780,6 +780,12 @@ EOF
       credentials.save if did_something
     end
 
+    # When insecure_addon_compat is true, PrivateChef["ldap"] may be
+    # non-empty after the secrets from veil are merged into the
+    # configuration. The user may have inadvertantly left the password
+    # in the secrets file, even when all other LDAP configuration has
+    # been removed. This function is called before the secrets merge
+    # allowing us to detect whether the user wants ldap or not.
     def set_ldap_enabled
       PrivateChef["ldap"]["enabled"] = !PrivateChef["ldap"].empty?
     end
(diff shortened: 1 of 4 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -810,9 +814,7 @@ def generate_config(node_name)
 
       generate_config_for_topology(PrivateChef["topology"], node_name)
 
-      unless PrivateChef["ldap"].nil? || PrivateChef["ldap"].empty?
-        gen_ldap
-      end
+      gen_ldap if PrivateChef["ldap"]["enabled"]
```
afiune: We wouldn't want to use `ldap_authentication_enabled?` here? or where is that fun being use? 🤔

stevendanna: The ldap_authentication_enabled? function is used at a number of call sites:

```
sdanna@thrace ~/oc/code/opscode/chef-server (master %) > rg 'ldap_authentication_enabled'
omnibus/files/private-chef-ctl-commands/password.rb
36:    if password == '' && ldap_authentication_enabled?
43:    message = if ldap_authentication_enabled?
53:    if ldap_authentication_enabled?
72:def ldap_authentication_enabled?

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
192:  def ldap_authentication_enabled?

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
27:ldap_authentication_enabled = OmnibusHelper.new(node).ldap_authentication_enabled?
29:enable_ssl = ldap_authentication_enabled ? node['private_chef']['ldap']['enable_ssl'] : nil
30:ldap_encryption_type = ldap_authentication_enabled ? node['private_chef']['ldap']['encryption_type'] : nil
47:  variables(node['private_chef']['opscode-erchef'].to_hash.merge(:ldap_enabled => ldap_authentication_enabled,
```

We don't use it here because at this point in the process, the PrivateChef configuration object hasn't been merged into the node object.

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/private_chef_spec.rb
```diff
@@ -42,8 +42,7 @@ def config_for(hostname)
 
   # Example content of /etc/opscode/private-chef-secrets.json
   # used when testing non-bootstrap config parsing.
-  let(:secrets) {<<EOF
-  {
+  let(:default_secrets) {<<EOF
   "redis_lb": {
```
srenatus: interesting, looks like github doesn't like that 😄

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -780,9 +780,13 @@ def migrate_keys
       credentials.save if did_something
     end
 
+    def set_ldap_enabled
+      PrivateChef["ldap"]["enabled"] = !( PrivateChef["ldap"].nil? || PrivateChef["ldap"].empty? )
```
srenatus: Is one of these checks maybe redundant? I'm wondering because _if `PrivateChef["ldap"].nil?` was true_, wouldn't `PrivateChef["ldap"]["enabled"] = ...` fail with `[]=` not defined for NilClass? 🤔

stevendanna: Good call, also since we set it ldap to Mash.new above, I'm not really sure we need the nil check.

stevendanna:
> I'm a little confused about that test --rendered_config isn't what gets put onto disk, is it? (Because that would mean the bind_password gets from the secrets store into a file it shouldn't be in...)

"Rendered" is a bit wrong in this context.  rendered_config is basically just the final node object after all of the parsing is done.  This test only works right now because insecure_addon_compat is true by default.

PR #1242, opened 2017-04-26
Do not return 400 for valid /users/USERNAME/_acl/PERM calls

stevendanna:
Previously, if PUT was made to update a user ACL, the request would
return 400 if the request body contained a user. This was because the
code expected an OrgId to be set. Now, we assume that requests in the
global context (no org id), can only refer to users.

An alternative approach would be to extend the implementation of the
new scoping syntax so that it works everywhere.

A number of disabled pedant tests have been re-enabled.

### DIFF
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
@@ -103,11 +103,9 @@ initialize_context(OrgId, DbContext, CallBackFun) ->
 %%   (if the later is present, otherwise inserting the current in-context org)
 %% * Mapping the org name to id
 %% * Looking up the org id, name pair in the appropriate table(s)
-
 %%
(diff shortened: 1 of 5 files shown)


src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz_scoped_name.erl
```diff
@@ -238,6 +236,14 @@ extract_full_names(ScopedNames, UnscopedNameSet) ->
 %% Abstract away difference in lookups between actors, clients, users and groups.
 %%
 %% Returns AuthzIds, missing names and names that are judged ambiguous
+%%
+
+%% The first two function heads are to account of this being called from a global context
```
stevendanna: Typo: "account of this" -> "account for this being"

jaymalasinha:
@stevendanna I added this commit to disable user acl tests for chef-zero.  Paired with @marcparadise on this

PR #1244, opened 2017-04-27
Fix Chef 12 oc_id regression

stevendanna:
The upgrade of oc_id to Chef 12 in commit

6824d90

introduced 2 issues:

(1) Incorrect arguments to ServrAPI.new
(2) A change in ssl verification behavior

This commit

- Fixes issue (1),
- Make the ssl_verify_mode configurable
- Updates omniauth-chef to a version that contains similar fixes.

* SSL Verify Mode

The ssl verify mode has been set to verify_none. This isn't a behavior
change from previous versions since this was the default in the old
version of Chef we were previously pinned to.

Because we want to talk to oc_id via 'localhost' to avoid cross-node
requests, changing this to verify_peer is problematic since, even if
we did the work to add /var/opt/opscode/nginx/ca (or the user-provided
cert location) to the trusted_cert_dir, the cert would likely still
mismatch since in the common case such certs ar for the `api_fqdn` and
not "localhost".

### DIFF
--- a/oc-chef-pedant/spec/api/oc_id_spec.rb
+++ b/oc-chef-pedant/spec/api/oc_id_spec.rb
@@ -1,0 +1,65 @@
+require 'pedant/rspec/common'
+
+describe "oc_id API", :oc_id do
+  context "status endpoint" do
+    let(:request_url) { "#{platform.server}/id/v1/status" }
+    let(:good_status) do
+      {"status" => "ok",
+       "erchef" => { "status" => "reachable" },
+       "postgres"=>{ "status" => "reachable" }}
+    end
+
+    context "GET /id/v1/status" do
+      it "retuns 200" do
+        get(request_url, platform.superuser).should look_like({:status => 200,
+                                                               :body => good_status})
+      end
+    end
+  end
+
+  context "signin" do
+    let(:username) { platform.non_admin_user.name }
+    let(:request_url) { "#{platform.server}/id/auth/chef/callback" }
+    let(:request_body) { "username=#{username}&password=#{password}&authenticity_token=#{CGI.escape(csrf[:token])}&commit=Sign+In" }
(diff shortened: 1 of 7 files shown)


stevendanna:
Depends on https://github.com/chef/omniauth-chef/pull/16

src/oc-id/Gemfile
```diff
@@ -13,14 +13,14 @@ gem 'unicorn-rails', '~> 1.1.0'
...
-gem 'mixlib-authentication', '~> 1.3.0'
+gem 'mixlib-authentication', '~> 1.3'
 gem 'sentry-raven'
 gem 'responders', '~> 2.0'
 gem 'newrelic_rpm'
 gem 'doorkeeper', '~> 1.4.0'
 
 gem 'veil', git: 'https://github.com/chef/chef_secrets'
-gem 'omniauth-chef', git: 'https://github.com/chef/omniauth-chef'
+gem 'omniauth-chef', git: 'https://github.com/chef/omniauth-chef', branch: 'ssd/POOL-581'
```
marcparadise: TODO

oc-chef-pedant/spec/api/oc_id_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    let(:good_status) do
+      {"status" => "ok",
+       "erchef" => { "status" => "reachable" },
+       "postgres"=>{ "status" => "reachable" }}
+    end
+
+    context "GET /id/v1/status" do
+      it "retuns 200" do
+        get(request_url, platform.superuser).should look_like({:status => 200,
+                                                               :body => {}})
```
stevendanna: I think we meant for this to be checking against good_status

stevendanna:
@chef/chef-server-maintainers Ready for review.

PR #1246, opened 2017-05-01
Fix regression in bookshelf preflight check

stevendanna:
The commit 0f4ded20d740d2a771da13d7b095d4575577cbc2 removed the to_s
coercion from the Bookshelf preflight checks. However, those checks
happen *before* the type coercion in the library, so we still need the
string coercion there since some configs have symbols in them.


echohack:
👍 I can confirm the use case here. This also effects the current version of `chef-backend-ctl gen-server-config` since that generates `bookshelf['storage_type'] = :sql` using the symbol.

stevendanna:

markan:
:+1: LGTM

PR #1250, opened 2017-05-03
replace 12.15.1 with 12.15.2.

marcparadise:
This was due to a bug caught after 12.15.1's release build
was completed, but prior to publishing.


btm:
👍

:D

PR #1253, opened 2017-05-04
[nginx] regression: proxied compliance requests get HTTP 404

stevendanna:
The compliance proxy no longer included the owner or profile name in
the upstream requests because of a bad rewrite rule.  This corrects
the rewrite rule.

*Testing*

Since there are no tests for this endpoint, I've tested it manually
using the following procedure:

1. Configure /etc/opscode/chef-server.rb as follows:

       data_collector['token'] = 'foobar'
       profiles['root_url'] = 'http://localhost:9998'

2. Use `socat` to start an SSL listener on 9998:

       socat OPENSSL-LISTEN:9998,reuseaddr,fork,verify=0,\
       cert=/var/opt/opscode/nginx/ca/api.chef-server.dev.crt,\
       key=/var/opt/opscode/nginx/ca/api.chef-server.dev.key,\
       dhparam=/var/opt/opscode/nginx/ca/dhparams.pem - | grep 'GET'

3. Use `knife raw` to make reqeusts and check the URL recieved by (2)

       knife raw /organizations/testorg/owners/foobar/compliance/bob -c /etc/opscode/pivotal.rb
       knife raw /compliance/organizations/testorg/owners/foobar/compliance/bob -c /etc/opscode/pivotal.rb

Before this change, the following two requests were recieved:

    GET /compliance/profiles/ HTTP/1.0
    GET /compliance/profiles/ HTTP/1.0

After this change:

    GET /compliance/profiles/foobar/bob HTTP/1.0
    GET /compliance/profiles/foobar/bob HTTP/1.0

### DIFF
--- a/oc-chef-pedant/spec/api/compliance_proxy_spec.rb
+++ b/oc-chef-pedant/spec/api/compliance_proxy_spec.rb
@@ -1,0 +1,73 @@
+describe "Compliance Proxy Tests", :compliance_proxy_tests do
+  if !Pedant.config.compliance_proxy_tests
+    it "Compliance Proxy Tests disabled since config.compliance_proxy_tests = false" do
+      expect(true).to eq(true)
+    end
+  else
+    before(:all) do
+      # Start stub compliance server in a thread
+      require 'webrick'
+      require 'webrick/https'
+      require 'openssl'
+      require 'thread'
+      cert_name = [
+        %w[CN localhost],
+      ]
+      puts "Starting stub compliance server on #{Pedant.config.compliance_proxy_port}"
+      Thread.new do
+        server = WEBrick::HTTPServer.new(:Port => Pedant.config.compliance_proxy_port,
+                                         :SSLEnable => true,
+                                         :SSLCertName => cert_name)
+        server.mount_proc '/' do |req, res|
+          if req.path =~ %r{/compliance/profiles/([^/]+)/?}
+            res.body = '{ "response": "ok" }'
+          elsif req.path =~ %r{/compliance/profiles/([^/]+)/([^/]+)/?}
+            res.body = '{ "response": "ok" }'
+          else
+            res.body = '{}'
+            res.status = 404
+          end
+        end
+        server.start
+      end
+    end
+
+    context "/organizations/ORG/owners/OWNER/compliance[/PROFILE]" do
+      let(:good_status) { {"response" => "ok"} }
+
+      context "GET /organizations/ORG/owners/OWNER/compliance" do
+        let(:request_url) { api_url("owners/foobar/compliance") }
+          it "retuns 200" do
+            get(request_url, admin_user).should look_like({:status => 200,
+                                                           :body => good_status})
+          end
+      end
+
+      context "GET /organizations/ORG/owners/OWNER/compliance/PROFILE" do
+        let(:request_url) { api_url("owners/foobar/compliance/testprofile") }
+        it "retuns 200" do
+          puts request_url
+          get(request_url, admin_user).should look_like({:status => 200,
+                                                         :body => good_status})
+        end
+      end
+
+      context "GET /compliance/organizations/ORG/owners/OWNER/compliance" do
+        let(:request_url) { "#{platform.server}/compliance/organizations/#{platform.test_org.name}/owners/foobar/compliance" }
+        it "retuns 200" do
+          puts request_url
+          get(request_url, admin_user).should look_like({:status => 200,
+                                                         :body => good_status})
+        end
+      end
+
+      context "GET /compliance/organizations/ORG/owners/OWNER/compliance/PROFILE" do
+        let(:request_url) { "#{platform.server}/compliance/organizations/#{platform.test_org.name}/owners/foobar/compliance/testprofile" }
+        it "retuns 200" do
+          get(request_url, admin_user).should look_like({:status => 200,
+                                                         :body => good_status})
+        end
+      end
+    end
+  end
+end
--- a/oc-chef-pedant/spec/api/compliance_proxy_spec.rb
+++ b/oc-chef-pedant/spec/api/compliance_proxy_spec.rb
@@ -32,41 +32,33 @@ describe "Compliance Proxy Tests", :compliance_proxy_tests do
       end
     end
 
-    context "/organizations/ORG/owners/OWNER/compliance[/PROFILE]" do
-      let(:good_status) { {"response" => "ok"} }
+    let(:response) { get(request_url, admin_user) }
 
-      context "GET /organizations/ORG/owners/OWNER/compliance" do
-        let(:request_url) { api_url("owners/foobar/compliance") }
-          it "retuns 200" do
-            get(request_url, admin_user).should look_like({:status => 200,
-                                                           :body => good_status})
-          end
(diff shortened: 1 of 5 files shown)


stevendanna:
cc @chef/chef-server-maintainers @chef/compliance-core 

This is still WIP as I work out how we want to test this in a more auotmated way, but I'd love for someone to verify those are the redirects we want now.

stevendanna:
Now with tests

stevendanna:
To enable these tests in wilson, we will need to modify the test config in opscode-ci and add `--compliance-proxy-tests` to https://github.com/chef/chef-server/blob/master/ci/run_tests.sh

oc-chef-pedant/spec/api/compliance_proxy_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+          puts request_url
+          get(request_url, admin_user).should look_like({:status => 200,
+                                                         :body => good_status})
+        end
+      end
+
+      context "GET /compliance/organizations/ORG/owners/OWNER/compliance/PROFILE" do
+        let(:request_url) { "#{platform.server}/compliance/organizations/#{platform.test_org.name}/owners/foobar/compliance/testprofile" }
+        it "retuns 200" do
+          get(request_url, admin_user).should look_like({:status => 200,
```
marcparadise: Non-blocking nit: We're moving generally `expect` instead of should

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -101,12 +101,12 @@
     #   /organizations/ORG/owners/OWNER/compliance[/PROFILE]
     # Supports the legacy(chef-gate) URLs as well:
     #   /compliance/organizations/ORG/owners/OWNER/compliance[/PROFILE]
-    location ~ (?:/compliance)?/organizations/([^/]+)/owners/([^/]+)/compliance(.*) {
+    location ~ <%= @compliance_proxy_regex -%> {
```
marcparadise: Can ee keep this in $variable in the config template itself? Or .maybe assign the ruby variable value in here instead of passing it in?  I'd prefer to keep this in code , since to my knowledge none of the other expressions come out oof the recipe - this way there's one source of truth.

oc-chef-pedant/spec/api/compliance_proxy_spec.rb
stevendanna: Moving /very slowly/ it seems :D: 
```
sdanna@thrace ~/oc/code/opscode/chef-server/oc-chef-pedant (ssd/compliance-proxy-regression) > rg '\.should ' spec/ | wc -l
    1272
sdanna@thrace ~/oc/code/opscode/chef-server/oc-chef-pedant (ssd/compliance-proxy-regression) > rg 'expect\w*[({]' spec/ | wc -l
     252
```

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
stevendanna: Unfortunately, it doesn't seem like we can use variables in the location block, unless I've messed up in my local tests.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -101,12 +101,12 @@
...
       set $request_org $1;
       access_by_lua_block { validator.validate("GET") }
       proxy_set_header x-data-collector-token $data_collector_token;
       proxy_set_header x-data-collector-auth "version=1.0";
-      rewrite ^ "/compliance/profiles/$2$3" break;
+      rewrite ^<%= @compliance_proxy_regex -%> "/compliance/profiles/$2$3" break;
```
sdelano: @stevendanna I'd also remove the quotes from the rewritten route. That was another inconsistency that triggered warning signs for me. It's probably fine, but you don't see that pattern anywhere in our code or in nginx docs. I think it's a holdover from translating this bit out of the Lua code.

stevendanna:
opscode-ci config update https://github.com/chef-cookbooks/opscode-ci/pull/707

sdelano:
I've verified that this is in fact the standard way to rewrite URLs for a proxy. I personally think it's ugly, but oh well.

sdelano:
Just for posterity,  here are the routes that we expose in automate:

```
  # GET:  /compliance/profiles
  # GET: /compliance/profiles/available
  # GET:  /compliance/profiles/john/linux
  # GET:  /compliance/profiles/john/linux/tar
  # POST: /compliance/profiles/john
```

@stevendanna :eyes: here to doublecheck that these line up?

stevendanna:
@sdelano Hrm. The previous comment doesn't mention supporting all of those.  I think these all work:

```
  # GET:  /compliance/profiles
  # GET:  /compliance/profiles/john/linux
  # GET:  /compliance/profiles/john/linux/tar
```

For this one is `available` a special endpoint? You can get the proxy to generate a request that looks like this, but only if you happen to know how the rewrite happens.
```
  # GET: /compliance/profiles/available
```

I don't think the post will pass validation.

sdelano:
@chris-rock @arlimus @vjeffrey - Can any of you chime in here about what needs to be supported by the chef-server proxy?

stevendanna:
Oh, I lied, this one doesn't work `/compliance/profiles`.

sdelano:
@stevendanna to be clear, I don't think it has to. We saw this change working end-to-end yesterday with the audit cookbook, so I'm fairly sure it works, but since we're now in the process of translating comments to enforceable tests, we might as well make sure we're testing the correct thing.

vjeffrey:
k all those endpoints you mentioned need to be supported; ie what's here: https://github.com/chef/automate/blob/master/cookbooks/delivery/templates/default/nginx-internal.conf.erb#L52  
```
 # GET:  /compliance/profiles                => /user/compliance
  # GET: /compliance/profiles/available => /profiles
  # GET:  /compliance/profiles/john/linux     => /owners/john/compliance/linux
  # GET:  /compliance/profiles/john/linux/tar => /owners/john/compliance/linux/tar
  # POST: /compliance/profiles/john           => /owners/john/compliance
```
(in the next week that's going to be changing to something more along the lines of https://github.com/chef/automate/blob/master/nginx/conf/nginx.conf#L114)

sdelano:
@vjeffrey do they need to be supported by the Chef Server, which IIRC is only supporting the audit cookbook?

vjeffrey:
oh geez. we're getting to the edges of my knowledge base atm....
i'm pretty sure what ya'll have is fine...let me ping chris so i don't mess things up lol

chris-rock:
I would prefer to keep the Chef Server dump and add additional routes to automate. But we can go with the current approach for now. I have not tested it in a real setup though.

sdelano:
@chris-rock last time we chatted about this you made a story in compliance to address it on the automate end: https://github.com/chef/chef-compliance/issues/1171

We're going to move forward with this change.

PR #1261, opened 2017-05-08
Added validation of oc-id profile page for proper character escaping.

piyushawasthi:


src/oc-id/config/initializers/constants.rb
```diff
@@ -0,0 +1,4 @@
+module Constants
+  REGEXP_NAME = /\A[^0-9`!@#\$%\^&*<+_=]+\z/
+  REGEXP_EMPTY_OR_NAME = /^(\s*|[\A[^0-9`!@#\$%\^&*<+_=]+\z]+)$/
```
stevendanna: Could this regex be:

```
/\A[^0-9`!@#\$%\^&*<+_=]*\z/
```

or is there something subtle I'm missing?

src/oc-id/app/controllers/profiles_controller.rb
```diff
@@ -118,4 +121,16 @@ def valid_signature?
       params[:email],
     ).valid_for?(params[:signature])
   end
+
+  def verify_user_params
+    @user = current_user
+    begin
+      raise TypeError, "Invalid First Name." unless Constants::REGEXP_NAME.match(params[:user][:first_name])
```
stevendanna: Does the underlying /users API require the firstname and last name?  The spec here indicates that they are optional: https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_objects/src/chef_user.erl#L190-L200  But I haven't read through all the code to see if they may be require elsewhere.

stevendanna: Should the error message tell the user what is invalid about their input? Otherwise they might be left to guess what they need to change.

marcparadise:
Thank you for contributing this.

I'm not sure what is being addressed though - is this intended to prevent cross-site script injections?

marcparadise:
I have found the context (MSYS-555).  We will also want to sanitize data on the way out, to ensure that existing fields won't render this content.

piyushawasthi:
@marcparadise, @stevendanna - Sanitizer only remove HTML tags. I am using REGEX instead of `rails-html-sanitizer` as there's a possibility of SQL injections if we go with the Sanitizer. Please check the sample inputs given to `sanitizer` below. Those inputs shouldn't be accepted as the Name, but `sanitizer` allows them. 
 
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John O'Malley")
=> "John O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John < O'Malley")
=> "John &lt; O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John > O'Malley")
=> "John &gt; O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John = O'Malley")
=> "John = O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John & O'Malley")
=> "John &amp; O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John \ O'Malley")
=> "John  O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John / O'Malley")
=> "John / O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John \" O'Malley")
=> "John \" O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John <> O'Malley")
=> "John &lt;&gt; O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John </script> O'Malley")
=> "John  O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John </ O'Malley")
=> "John  O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John <\ O'Malley")
=> "John &lt; O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("John # O'Malley")
=> "John # O'Malley"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("~ /(\/*\/)/ # //")
=> "~ /(/*/)/ # //"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("text.gsub(/(\/*.?*\/)/, "")")
=> "text.gsub(/(/*.?*/)/, “”)"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("[ /(\bdata:\b|eval|cookie|\bwindow\b|\bparent\b|\bthis\b)/i, # suspicious javascript-type words /behaviou?r|expression|moz-binding|@import|@charset|(java|vb)?script|[\<]|\\w/i, /[\<>]/")
=> "[ /(data:|eval|cookie|window|parent|this)/i, # suspicious javascript-type words /behaviou?r|expression|moz-binding|@import|@charset|(java|vb)?script|[]/"
oc_id >  Rails::Html::Sanitizer.full_sanitizer.new.sanitize("eval(User.delete_all('id = 1'))")
=> "eval(User.delete_all('id = 1'))"

stevendanna:
Defense against SQL injection should probably be in oc_erchef directly since there are multiple paths for information to get to that erchef API.  As far as I know, the information in question isn't being used in any SQL queries in the oc_id application.

I'm not against a few defense-in-depth protections here as well.  However, given the content of MSYS-55 I think we /also/ have to ensure information is being escaped/sanitized at the point of rendering OR that input is filtered on the erchef side.

stevendanna:
> Defense against SQL injection should probably be in oc_erchef directly since there are multiple paths for information to get to that erchef API.

I'll note that I don't think we have a SQL injection problem in erchef because of the way we use prepared statements to pass all user input explicitly as arguments; however, I haven't looked that the code in detail.

marcparadise:
Thanks for the details @piyushawasthi. As @stevendanna mentioned, the defense against sql injection is handled at the API layer in oc_erchef. A quick look through oc-id shows that the only DB access it needs is for oauth access grants, and that schema does not contain any user-editable fields, so HTML sanitation is the only concern here. 

> I'll note that I don't think we have a SQL injection problem in erchef because of the way we use prepared statements to pass all user input explicitly as arguments; however, I haven't looked that the code in detail.

It's a thing to verify, but I can't think of any cases where we run queries dynamically at all, never mind cases of concatenating user inputs to build a query.

piyushawasthi:
@marcparadise , @stevendanna - Thanks for you valuable comments, that's absolutely fine to put all the validations at API layer in oc_erchef, but please correct me if I am wrong `rails-html-sanitizer` only sanitize html tags. For example: If user inserts a name "John < O'Malley", then it's stored in DB as `"John &lt; O'Malley"` and it's actually not a valid name. Of-course we can't validate name 100% but we can reduce invalid data in our database. I think validating data before sending to API layer also is a good way to reduce number of re-render iteration from server so that it's better for performance tunning.

**I think validation should be on the following places:**
1: Client-side validation (for normal user, but easy to by-pass)
2: Validate data in (controller/model/helper) before sending to API (to reduce server interaction)
3: Server side validation (Always required to protect against the malicious users)

Please share your thoughts on this also let me know if I should modify the code to use Sanitize instead of Regex.

stevendanna:
@piyushawasthi If we are going to do input validation, then I think the regex is fine since on the erchef side I believe all of our input validation is regular-expression based anyway.  However, I think that in order to account for existing data, we'll still want to sanitize at the point of rendering the HTML, for those cases, the html sanitation helpers are likely way to go.

NimishaS:
@stevendanna , we can have a separate issue for using sanitize at HTML randering. Since this PR is for MSYS-555, it would be good to limit it's scope.
I guess for using sanitize at HTML randering, `chef-manage` and `oc-id` both will require changes.

stevendanna:
@NimishaS Feel free to carve up the work into tickets that make sense for y'all.  Overall, to close the core issue in MSYS-555 I think requires:

- Input validation in oc_erchef so all routes to adding user-data reject bad input
- Sanitization of displayed data in `oc-id` and `chef-manage` to account for existing bad data.
- (optionally) additional input validation in oc-id and chef-manage to avoid round-tripping bad data to the server when we know it will get reject. 

I think this lines up with the list @piyushawasthi created, just stated a bit differently.

stevendanna:
@btm @NimishaS What is the status here?

NimishaS:
@stevendanna , @btm has created POOL-589 and POOL-608 which needs to be done by another team. We need to wait for that.

stevendanna:
Here is the story so far:

- Since an input-filtering approach was going to be taken, we decided to do that filtering at the API level to ensure it was consistent across all ways to add users.

- This ran into all of the problems I expected us to hit with trying to run input validation on people's names (people's names that might also be coming from ill-managed fields in an LDAP schema)

- Because of the problems, we rolled back the input filtering.

Thus, I think we should still be doing something to address possible problems related to displaying these user-provided fields oc-id and in manage.  I think this should come in the form of sanitization at the time of output rendering.

Given all that, I'm going to close this PR.

PR #1262, opened 2017-05-08
[pedant] Enable compliance-proxy-tests, wait for listener

stevendanna:
We disabled these tests because occasionally the threaded compliance
stub server wouldn't start fast enough. This adds a simple function to
wait for the HTTP listener to be started before starting the tests.

### DIFF
--- a/oc-chef-pedant/spec/api/compliance_proxy_spec.rb
+++ b/oc-chef-pedant/spec/api/compliance_proxy_spec.rb
@@ -15,9 +15,9 @@ def wait_for_http_listener(uri_string, retries = 10)
     if retries <= 0
       raise e
     else
-      puts "GET #{uri_string} failed: #{e} retrying in 1 seconds (#{retries} retries remaining)"
+      puts "GET #{uri_string} failed: #{e} retrying in 0.1 seconds (#{retries} retries remaining)"
       retries -= 1
-      sleep 1
+      sleep 0.1
       retry
     end
   end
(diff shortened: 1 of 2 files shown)


oc-chef-pedant/spec/api/compliance_proxy_spec.rb
```diff
@@ -1,43 +1,73 @@
...
+  require 'webrick/https'
+  require 'openssl'
+  require 'thread'
+  Thread.new do
+    server = WEBrick::HTTPServer.new(:Port => port,
+                                     :SSLEnable => true,
+                                     :SSLVerifyClient => OpenSSL::SSL::VERIFY_NONE,
+                                     :SSLCertName => [%w[CN localhost]])
+    server.mount_proc '/' do |req, res|
+      if req.path =~ %r{/compliance/profiles/([^/]+)/?}
```
srenatus: [nit] the `if` and the `elsif` branch look like they could possibly be combined. But this is fine 👍

marcparadise: Would a `case` work better here?

oc-chef-pedant/spec/api/compliance_proxy_spec.rb
```diff
@@ -1,43 +1,73 @@
...
+      http.request request
+    end
+    puts "Succesfully made GET to #{uri_string}"
+  rescue StandardError => e
+    if retries <= 0
+      raise e
+    else
+      puts "GET #{uri_string} failed: #{e} retrying in 1 seconds (#{retries} retries remaining)"
+      retries -= 1
+      sleep 1
```
marcparadise: Minor, but we could use `sleep 0.1` here - if we poll a couple of extra times, no harm done.

stevendanna: 👍 I'm pretty sure that because of the default accept timeout is pretty large, we basically never loop anyway since we block trying to connect.

oc-chef-pedant/spec/api/compliance_proxy_spec.rb
stevendanna: I've left these as separate to make sure we are super explicit about what cases we expect this little stub to handle.

stevendanna:
Pushed some minor changes to address PR comments.

PR #1263, opened 2017-05-08
Set SAN on generated SSL certificate

### LINKED ISSUE #1260: Support 'SAN' when creating self-signed SSL certificate
## Description

As of Chrome 58, the web development team has removed/deprecated support for commonName matching in certificates, which is the logic used today (https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb#L59-L69) when Chef Server is installed and a self-signed certificate is created. 

In the past you could import / trust the self signed certificate in the local certificate store, but that is pointless now with the generated certificate still not supplying a SAN field in the cert. 

## Steps to reproduce
1. Install Chef Server on a new machine
1. In Chrome, Add self signed certificate to the OS' cert store (depends on platform)
1. Receive error that 'Subject Alternative Name' is missing

stevendanna:
This sets a Subject Alternative Name for newly-generated SSL
certificates. Newer browsers require that the SubjectAltName is set in
addition to the CommonName.

Fixes #1260

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
@@ -56,13 +56,21 @@ if (node['private_chef']['nginx']['ssl_certificate'].nil? &&
    ssl_keyfile = File.join(nginx_ca_dir, "#{node['private_chef']['nginx']['server_name']}.key")
    ssl_crtfile = File.join(nginx_ca_dir, "#{node['private_chef']['nginx']['server_name']}.crt")
 
+   server_name = node['private_chef']['nginx']['server_name']
+   server_name_type = if OmnibusHelper.is_ip?(server_name)
+                        "IP"
+                      else
+                        "DNS"
+                      end
+
    openssl_x509 ssl_crtfile do
-     common_name node['private_chef']['nginx']['server_name']
+     common_name server_name
      org node['private_chef']['nginx']['ssl_company_name']
      org_unit node['private_chef']['nginx']['ssl_organizational_unit_name']
      country node['private_chef']['nginx']['ssl_country_name']
      key_length node['private_chef']['nginx']['ssl_key_length']
      expire node['private_chef']['nginx']['ssl_duration']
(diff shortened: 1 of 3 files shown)


srenatus:
Interesting! [These docs](https://www.openssl.org/docs/manmaster/man5/x509v3_config.html#Subject-Alternative-Name) make it seem like also `IP:` would be a thing (both v4 and v6):

    subjectAltName=IP:192.168.7.1
    subjectAltName=IP:13::17

stevendanna:
@srenatus Right, I saw that but then wasn't sure if we actually wanted to go about parsing out hostnames from IP addresses; however, perhaps that wouldn't be too bad.  Looking at https://tools.ietf.org/html/rfc6125#section-6.4.4 it does seem important to populate the SAN data correctly since it appears that the CN isn't necessarily used as a fall back.

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
```diff
@@ -56,14 +56,21 @@
    ssl_keyfile = File.join(nginx_ca_dir, "#{node['private_chef']['nginx']['server_name']}.key")
    ssl_crtfile = File.join(nginx_ca_dir, "#{node['private_chef']['nginx']['server_name']}.crt")
 
+   server_name = node['private_chef']['nginx']['server_name']
+   server_name_type = if server_name =~ OmnibusHelper.is_ip?(server_name)
```
srenatus: I'm confused, doesn't `is_ip?` return a boolean?

stevendanna: *jedi hand wave* There was no rookie mistake here, what are you talking about?

Good catch, forgot to commit some local change.  Now with correct code and tests!

marcparadise:
How likely are we to determine the IP that nginx is listening on? Since it bind 0.0.0.0 by default, there might be some challenges in getting the right one?

marcparadise:
Never mind that - I just saw your impl for the additiion.

PR #1264, opened 2017-05-08
Fix issues in postgresql preflight validator

stevendanna:
- Not checking for db_superuser_password in user-config
- Querying the wrong attribute for the bookshelf attr.


stevendanna:
I'm going to go ahead and merge this, but I think we need to really take a careful look at the bootstrap preflight checks before the next release.

PR #1267, opened 2017-05-11
Mp/oc id missing all.svg

marcparadise:
enable compilation of missing assets, so that
requests to `all.svg` no longer 404.  This
dependency is defined and added in chef-web-core.
Building the assets in precompile generates everything including
all-FINGERPRINT.svg, however Rails is not using the fingerprint
in this case.

This is a simple fix after exploring the asset generation and
not finding a sane way to either a) get rails to use
the fingerprint for this resource ; or b) get it to load the
un-fingerprinted version out of the gem.

Since this is the only missing asset, no impact is expected by
enabling compilation of missing resources.

### DIFF
--- a/src/oc-id/config/environments/production.rb
+++ b/src/oc-id/config/environments/production.rb
@@ -27,7 +27,7 @@ OcId::Application.configure do
   # config.assets.css_compressor = :sass
 
   # Do not fallback to assets pipeline if a precompiled asset is missed.
-  config.assets.compile = false
+  config.assets.compile = true
 
   # Generate digests for assets URLs.
   config.assets.digest = true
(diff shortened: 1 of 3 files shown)


marcparadise:
As an added benefit this will stop ocid logs from capturing an exception on every hit.

src/oc-id/config/environments/production.rb
```diff
@@ -27,7 +27,7 @@
   # config.assets.css_compressor = :sass
 
   # Do not fallback to assets pipeline if a precompiled asset is missed.
-  config.assets.compile = false
+  config.assets.compile = true
```
stevendanna: Will this work if nodejs isn't installed on the machine?

marcparadise: I did run through with node uninstalled - but I'll verify that if it's never ther for the precompile phase, it still behaves just in case something got generated and retained.

marcparadise: I've confirmed that this works when there is never a node package installed.

PR #1270, opened 2017-05-12
WIP - DO NOT MERGE - streaming search results

sdelano:
**DO NOT MERGE**

This is very much a work in progress to determine the feasibility / performance of streaming search requests vs keeping the result set in memory.

Issues:
* one failing pedant test (you can see the databag case that I commented out)
* haven't even run / changed unit or integration tests
* if we get a database failure mid-stream the results are undefined
* request logging has been removed due to the new streaming callback

Performance results:
This unscientifically performs better than what we had before, but after digging into the `chef-load` results that I was seeing it's clear that the tests we're running are not returning results from the database. There's further investigation to be done, but the tests that were previously failing are now not failing and large search requests see quite an improvement in response time (e.g. decrease from 2.5 to 1.5 seconds in response time).

**TODO FOR FURTHER EXPERIMENTATION**:
* [ ] set magical nginx header to enable streaming / disable gzipping
* [ ] reintroduce logging at end of request

**TODO FOR POTENTIAL RELEASE**:
* [ ] fix databag 404 issue
* [ ] fix / update unit tests
* [ ] remove dead code
* [ ] investigate failure modes

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
@@ -140,28 +140,32 @@ to_json(Req, #base_state{chef_db_context = DbContext,
                                  filter_permitted_results(ReqId, RequestorId, OrgId, DbContext, IndexType, Ids);
                              false -> Ids
                          end,
-            {DbNumFound, Ans} = make_search_results(BulkGetFun, FilteredIds, BatchSize,
-                                                    Start1, SolrNumFound),
-            State1 = State#base_state{log_msg = search_log_msg(SolrNumFound,
-                                                               solr_ids_length(Ids), DbNumFound)},
-            case IndexType of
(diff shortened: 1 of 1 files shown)


stevendanna:
@sdelano Is this moving forward at all?  If not, perhaps we can close the PR but keep the branch around for if we ever pick it back up.

stevendanna:
@sdelano Does this work have a home anymore?  This seems like an obvious win from the erchef side in terms of memory management; however, if I recall correctly you were seeing some bad behaviour in nginx that we would have to sort out for this to be a win?

sdelano:
@stevendanna I'd say we can abort this work. It was in service of some potential performance issues in certain deployment scenarios in combination with Chef Automate, but those concerns have been addressed by tackling other more resource-hungry portions of the system.

IIRC, this change limited the amount of RAM consumed by large erchef searches but just pushed that resource consumption downstream to nginx.

PR #1273, opened 2017-05-15
Add a CODE_REVIEW_CHECKLIST.md

stevendanna:
The Chef Server 12.15 release had a number of regressions. As part of
the post-mortem for that release, we agreed that a PR checklist may be
helpful to have more fruitful discussions during code review.

This is a first attempt at such a checklist.

### DIFF
--- a/CODE_REVIEW_CHECKLIST.md
+++ b/CODE_REVIEW_CHECKLIST.md
@@ -1,0 +1,71 @@
+# Code Review Check List
+
+The goal of the code review checklist is to start productive questions
+during code review to help ensure that we are constantly improving the
+quality of the Chef Server.
+
+Change authors should review the checklist before submitting their
+pull request. Reviewers should consult the checklist during review in
+addition to reviewing the code for any language-specific standards we
+may have.
+
+- [ ] Do the commit messages and PR description give enough context on why
+      this change is being made? Would someone reading the commit message
+      6 months from now find it useful?
+
+- [ ] Does the code work? Does it compile without warnings?
+
+- [ ] Are the travis-based unit tests and the pedant test passing?
+
+- [ ] Are there integration tests for new user-facing features?
+
+- [ ] If this is a bug fix, is there a regression test for the bug?
+
+- [ ] Are the included tests likely to prevent future breakage of this
+      feature?
+
+- [ ] Has this PR changed the format of any on disk files that may be
+      used by add-ons? If so, has add-on compatibility been
+      specifically tested?
+
+- [ ] Has this PR changed an API response that may break API clients?
+      If so, does this change require an API version bump and has time
+      been set aside to update chef-client?
+
+- [ ] Are there constants that should be configurable?
+
+- [ ] Do any new configurables have a sensible default value?
+
+- [ ] Are any new user-visible settings/features documented? (via a
+      linked PR to [chef-web-docs](https://github.com/chef/chef-web-docs/))
+
+- [ ] Is there enough logging or other diagnostic information in the
+      code to aid future debugging?
+
+- [ ] Can any of the logging be removed or moved to a more detailed
+      log level?
+
+- [ ] Does this change introduce new dependencies? If so, do we
+      believe those dependencies are necessary and stable?
+
+- [ ] Does this change break the dev environment? If so, can we fix it
+      now or have we prioitized time to fix it in the near future?
+
+- [ ] Are failures from external systems or libraries handled? If not,
+      do we know what the behavior will be in the case of failure? Is
+      that the behavior we want?
+
+- [ ] Are variable and function names clear? Do functions/methods have
+      a clear purpose?
+
+- [ ] Is it clear how the code in the change addresses the stated goal
+      of the pull request?
+
+- [ ] Are there compromises this PR is making because of time or other
+      constraints? If so, are those compromises documented in `TODO`
+      or `NOTE` comments?
+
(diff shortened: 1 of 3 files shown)


CODE_REVIEW_CHECKLIST.md
```diff
@@ -0,0 +1,73 @@
...
+- [ ] Has this PR changed an API response that may break API clients?
+      If so, does this change require and API version bump and has
+      time been set aside to update chef-client?
+
+- [ ] Are there constants that should be configurable?
+
+- [ ] Do any new configurables have a sensible default value?
+
+- [ ] Are any new user-visible settings/features documented? (via a
+      linked PR to chef-docs)
```
srenatus: [pedant] Is "chef-docs" the collective noun for chef-{dev,web}-docs, or am I missing something?

CODE_REVIEW_CHECKLIST.md
```diff
@@ -0,0 +1,73 @@
...
+- [ ] If this is a bug fix, is there a regression test for the bug?
+
+- [ ] Are the included tests likely to prevent future breakage of this
+      feature?
+
+- [ ] Has this PR changed the format of any on disk files that may be
+      used by add-ons? If so, has add-on compatibility been
+      specifically tested?
+
+- [ ] Has this PR changed an API response that may break API clients?
```
srenatus: If we care about API clients (haha, of course we do), would it make sense to enumerate those _we support_? As in, you can't fix the world's Chef server API clients, but we can take care of chef-client. (_Or is_ {chef-client} that set?)

Also, we're not mentioning API versioning **or** versioning altogether -- would it make sense to add some item for that? I'd assume the contributed change itself won't bump the version, but **the necessity to bump the version** might be relevant for code review?

stevendanna:
> I do wonder if we have a set of first- and second-class deployment options? (I'm thinking about HA, tiered, external PG, external RMQ, and the like)

Yup, we have another task to define the list of our top-tier deployment scenarios.

CODE_REVIEW_CHECKLIST.md
srenatus: Just realised the slack channel is called "chef-docs"... not sure if that's a good thing, since that one's internal.

stevendanna: Let's make this a link to the chef-web-docs repo.

CODE_REVIEW_CHECKLIST.md
```diff
@@ -0,0 +1,73 @@
...
+      If so, does this change require and API version bump and has
```
stevendanna: s/and/an/

CODE_REVIEW_CHECKLIST.md
stevendanna: I think we mention API versioning in the sentence below this?  Overall, I think which clients to fix would be a discussion during the PR depending on which APIs were changing.

srenatus: Yep, I've read over that bit 😖

CODE_REVIEW_CHECKLIST.md
```diff
@@ -0,0 +1,73 @@
...
+- [ ] Are any new user-visible settings/features documented? (via a
+      linked PR to chef-docs)
+
+- [ ] Is there enough logging or other diagnostic information in the
+      code to aid future debugging?
+
+- [ ] Can any of the logging be removed or moved to a more detailed
+      log level?
+
+- [ ] Are any new logs properly rotated?
```
srenatus: Is it warranted to have this singled out like that? I mean, any new service could have various ways for filling up the filesystem. But I guess, producing logs is something every service does... 🤔

CODE_REVIEW_CHECKLIST.md
thommay: I think at a minimum we care about chef-client, ridley (for berkshelf), and possibly py-chef or chef-api as stretch goals. But the first two are critical.

CODE_REVIEW_CHECKLIST.md
stevendanna: I'll remove this one.  This list is kind of long and we were looking for way to shorten it a bit.  This question was on a list we started a bit ago about things to consider for a new service, but I'm not sure it is appropriate here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
...
+   ready to merge.
+
+5. Rebase your branch on master to pick up any changes that have
+   landed since your patch was opened and merge it!  (If you don't
+   have commit access, a maintainer will merge it for you after code
+   review is complete).
+
+## Reporting an Issue
+
+When reporting an issue, please try to include as much detail as
```
marcparadise: This might be better moved to an issue template.

CODE_REVIEW_CHECKLIST.md
stevendanna: @thommay I'm going to merge this without the explicit list, but I think we should continue to have a discussion about what clients we should be testing compat with.  Currently API compat is only tested with pedant, but perhaps we could add some "external integration" tests to ensure that we don't break any the core features of these products.  To be honest though, right now most bugs concern add-on compat and not REST API compat.

thommay: Yeah, I think that's fine. In general API clients should assume that chef-client (or pedant) is the reference implementation and be bug compatible with that, but it would be good to have some assurance.

PR #1276, opened 2017-05-16
12.15.7 release prep

marcparadise:
In addition to 12.15.7 release notes/version update, this PR also moves the product version out of `omnibus/projects/chef-server.rb` and into `omnibus-overrides.rb` to prevent a rebuild with every version bump. 

CHANGELOG will be updated before we close this out.

### DIFF
--- a/omnibus/config/projects/chef-server.rb
+++ b/omnibus/config/projects/chef-server.rb
@@ -24,16 +24,18 @@ package_name    "chef-server-core"
 replace         "private-chef"
 conflict        "private-chef"
 install_dir     "/opt/opscode"
-build_version   "12.15.6"
-build_iteration 1
-#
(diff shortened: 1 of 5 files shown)


omnibus/config/projects/chef-server.rb
```diff
@@ -24,16 +24,18 @@
 replace         "private-chef"
 conflict        "private-chef"
 install_dir     "/opt/opscode"
-build_version   "12.15.6"
-build_iteration 1
-#
-# Load dynamically updated overrides
+
+# In order to prevent unecessary cache expiration,
```
srenatus: :+1:

marcparadise:
Thanks for updating the docs too. :+1: on the additions. (Githun won't let me actually approve since I opened the PR. )

PR #1277, opened 2017-05-17
Pin berkshelf-no-depselector to avoid pulling in 6.0

stevendanna:


thommay:
wait why does chef-server have berkshelf in it?

stevendanna:
@thommay It is a build-time dep to compile the cookbooks required for reconfigure.

stevendanna:
Notably: https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/Berksfile

thommay:
Ah, i was concerned for a second there was some deep magic that relied on berkshelf. 👍

PR #1279, opened 2017-05-19
Add optional ulimit for erchef

danielsdeleo:
Add a chef-server.rb setting `opscode_erchef['memory_maxbytes']` which,
when set to a non-nil value, adds `-d $BYTES` to the `chpst` command
which sets the process environment for erchef. This setting limits the
data memory segment for the erchef process to the specified value, which
caps the amount of RAM erchef may use.

This setting is intended to be used for standalone topologies, such as
when Chef Automate and Chef Server are colocated on the same machine. In
these topologies, heavy search load can cause erchef to consume large
amounts of memory which then triggers memory related failures in other
components, such as Elasticsearch. The `memory_maxbytes` setting allows
the administrator to make erchef fail instead of the other components in
these overload situations, allowing the system to recover more rapidly.

This setting is inteded to act as an emergency "circuit breaker" for
excessive load situations and should be set at the highest value at
which it can be effective, i.e., the value at which the host system's
memory would be completely utilized. Future load testing efforts are
required to provide guidance around recommended values.

Signed-off-by: Daniel DeLeo <dan@chef.io>


omnibus/files/private-chef-cookbooks/private-chef/templates/default/sv-opscode-erchef-run.erb
```diff
@@ -6,4 +6,4 @@ exec 2>&1
...
+exec veil-env-helper --use-file -s chef-server.webui_pub_key -s opscode_erchef.sql_password -s bookshelf.access_key_id -s bookshelf.secret_access_key -s rabbitmq.password -s rabbitmq.management_password -s rabbitmq.actions_password -s oc_bifrost.superuser_id -o data_collector.token -o ldap.bind_password -- chpst -P <%= "-d #{node['private_chef']['opscode-erchef']['memory_maxbytes']}" if node['private_chef']['opscode-erchef']['memory_maxbytes'] %> -u <%= node['private_chef']['user']['username'] %> -U <%= node['private_chef']['user']['username'] %> env ERL_EPMD_ADDRESS=127.0.0.1 HOME=<%= node['private_chef']['opscode-erchef']['dir'] %> <%= "OPENSSL_FIPS=1" if node['private_chef']['fips_enabled'] %> /opt/opscode/embedded/service/opscode-erchef/bin/oc_erchef foreground
```
ryancragun: Totally pedantic, but I feel like it might make sense to make these conditional expressions tags include the line break dashes to trim the space when they're false, eg: `<%= foo %>` to `<%= foo -%>`

danielsdeleo: does erubis do that automatically? ISTR erubis not requiring the hyphen ending in places where regular Erb does.

stevendanna:
I'm OK with this since it is an opt-in feature; but I'm worried about how heavy of a hammer this is and how unpredictably it seems like it may cause us to fail.

I can imagine this thing would be hard to test in an automated fashion, but have we tried doing something like artificially creating memory pressure by setting a low max value and then watching it fail.  If so, how did it fail?

danielsdeleo:
I created a small OWCA instance (4GB RAM), created 600 node objects and then ran a load test of 100 simulated CCRs on a fairly tight interval (don't recall the exact params, but it was around 1m average interval), with all simulated runs doing a search for `*:*` on the nodes index. When enough of the simulated runs did their searches simultaneously, erchef hit a 1GB RAM limit I had set. The erlang VM's behavior in this case is to just exit without running any further erlang code. There are some mailing list discussions about changing this or adding some other option but there doesn't seem to be much enthusiasm for that. Anyway, runit was able to restart erchef in less than 10s in my configuration each time. In none of the cases was erchef unable to restart. I would suspect that some node saves were inconsistent, e.g., data wasn't correctly saved to both postgres and search, but I didn't test for that.

In any event, though this is a big hammer, it's a smaller hammer than the OOM killer wields, and is intended to be tuned such that you only hit this ulimit _just_ before the OOM killer would kick in and kill a process (like Elasticsearch) that has significantly worse MTTR (in my testing, I also saw Elasticsearch get a malloc failure, which has the same result because the JVM just exits in that case, so the outcome was the same in terms of visible behavior).

danielsdeleo:
@markan sadly I did not keep that in my investigation notes

danielsdeleo:
@markan got it off my demo box:

```
2017-06-01_23:04:51.34889 binary_alloc: Cannot reallocate 1143523 bytes of memory (of type "binary").
```

danielsdeleo:
guess it changes depending on what specifically erchef is doing when it gets the malloc failure:

```
2017-06-01_23:13:12.72098 eheap_alloc: Cannot allocate 6664208 bytes of memory (of type "old_heap").
```

PR #1283, opened 2017-05-30
[POOL-608] allow for case insensitive users search by email

srenatus:
So far, this is for `GET /users?email=foo%40bar.com`, and adding an index to make this not super slow (not verified if it's strictly necessary, so if tests in hosted reveal that we don't need this, we could drop b358900).

It doesn't yet forbid creating a second user with the same email (case sensitively).

### DIFF
--- a/src/oc_erchef/schema/deploy/users_email_functional_index.sql
+++ b/src/oc_erchef/schema/deploy/users_email_functional_index.sql
@@ -1,0 +1,10 @@
+-- Deploy enterprise_chef:users_email_functional_index to pg
+-- requires: users
+
+BEGIN;
+
+  -- TODO 2017/05/30 sr: check with ops if this should include CONCURRENTLY
+  -- https://www.postgresql.org/docs/9.2/static/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY
+  CREATE INDEX users_lower_email_idx ON users (lower(email));
+
+COMMIT;
--- a/src/oc_erchef/schema/deploy/users_email_functional_index.sql
+++ b/src/oc_erchef/schema/deploy/users_email_functional_index.sql
@@ -1,10 +1,4 @@
 -- Deploy enterprise_chef:users_email_functional_index to pg
 -- requires: users
 
-BEGIN;
-
-  -- TODO 2017/05/30 sr: check with ops if this should include CONCURRENTLY
-  -- https://www.postgresql.org/docs/9.2/static/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY
-  CREATE INDEX users_lower_email_idx ON users (lower(email));
-
-COMMIT;
+CREATE INDEX CONCURRENTLY users_lower_email_idx ON users (lower(email));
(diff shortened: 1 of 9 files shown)


stevendanna:
I believe we will need a partybus migration to deploy the new sqitch index on upgrades.

srenatus:
We've come to the conclusion that this is a cumbersome approach, and that it might be easier to get a migration in place for changing email's column type from `text` to `citext` (`c`ase `i`nsensitive). (That way, we've got stronger guarantees and less to worry about when it comes to inserting/updating data with properly downcased emails.)

I'll put this branch on hold and start another one for the migration.

srenatus:
After some more discussion, it might be the case that the best compromise is as follows:

1. query case insensitively
2. make manage react properly when more than one user with this email (or eMAIL or Email) is returned
3. ignore that the situation could be created again by someone adding a user record with a differently cased email

That way we don't need a migration. Also, we wave off any guarantees of duplicate entries not happening -- they may be there already (and we can't fix those), so they may be added as well. They just *may be there*.

srenatus:
This would be good to go. Just waiting for https://github.com/chef/chef-zero/pull/265 to merge so we can drop 8d50b22.

srenatus:
The chef-zero change this depends on has been merged ✔️ 

The only missing bit is asking ops if creating that index will be fine as-is for hosted. ❔

src/oc_erchef/schema/deploy/users_email_functional_index.sql
```diff
@@ -0,0 +1,10 @@
+-- Deploy enterprise_chef:users_email_functional_index to pg
+-- requires: users
+
+BEGIN;
+
+  -- TODO 2017/05/30 sr: check with ops if this should include CONCURRENTLY
+  -- https://www.postgresql.org/docs/9.2/static/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY
+  CREATE INDEX users_lower_email_idx ON users (lower(email));
```
stevendanna: Perhaps we should only create the index if it doesn't exist, that way ops can deploy it different if they choose to and this will still be safe to run.

srenatus: Uhm, yeah, that's a good idea. _CREATE INDEX IF NOT EXISTS_ it shall be.

stevendanna: Unfortunately, I think CREATE INDEX IF NOT EXISTS isn't supported in 9.2.  This stack-overflow answer presents alternatives for different versions:

stevendanna: https://dba.stackexchange.com/questions/35616/create-index-if-it-does-not-exist

srenatus: It would have been sooo nice. But thanks! :smiley:

srenatus:
I don't believe this has to be held back anymore. The index creation is done concurrently now, and the corresponding verification disabled. No user should be hurt in the process.

stevendanna:

PR #1287, opened 2017-06-01
CON-638 Experimental developer support for external ES5

lancewf:
This adds elastic search 5 support to chef_index but this support is
experimental.  It does not take care of upgrading a current chef-server
from ES 2 -> ES 5.  It also does not ensure that all other chef-server
functionality such as the reporting or the chef-backend plugin correctly
work when ES 5 is used.  Using ES5 as the search engine is completely
unsupported in a customer environment.

For the dev environment, a new optional VM has been added to
dev/Vagrantfile.  To enable it, check out defaults.yml for the commented
out 'elasticsearch' section.  It supports bring ES2 or ES5 currently.

@ssd, @sdelano, @marcparadise, @chef/chef-server-maintainers

### DIFF
--- a/dev/scripts/provision-elasticsearch.sh
+++ b/dev/scripts/provision-elasticsearch.sh
@@ -1,0 +1,17 @@
+add-apt-repository ppa:openjdk-r/ppa
+wget -qO - https://artifacts.elastic.co/GPG-KEY-elasticsearch | sudo apt-key add -
+
+if [ "${ELASTIC_VERSION}" = "2" ]; then
+  echo "deb https://packages.elastic.co/elasticsearch/2.x/debian stable main" | sudo tee -a /etc/apt/sources.list.d/elasticsearch-2.x.list
+else
+  echo "deb https://artifacts.elastic.co/packages/5.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-5.x.list
+fi
+
+apt-get update
+
+apt-get install openjdk-8-jdk -y
+export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
+
+apt-get install elasticsearch -y
+echo "http.host: 0.0.0.0" > /etc/elasticsearch/elasticsearch.yml
+service elasticsearch start
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
@@ -1,5 +1,6 @@
 require 'mixlib/shellout'
 require 'uri'
+require "net/http"
 
 class OmnibusHelper
   attr_reader :node
@@ -56,6 +57,40 @@ class OmnibusHelper
       host += ":" + url.port.to_s
     end
     host
+  end
+
+  def elastic_search_major_version
+    max_requests = 5
+    current_request = 1
+
+    if node['private_chef']['opscode-solr4']['external']
+      # This will throw an exception if you have an invalid URL. We want this behavior.
+      elastic_search_uri = URI.parse(node['private_chef']['opscode-solr4']['external_url'])
+      begin
+        req = Net::HTTP::Get.new(elastic_search_uri)
+        res = Net::HTTP.start(elastic_search_uri.hostname, elastic_search_uri.port, use_ssl: elastic_search_uri.scheme == "https") do |http|
+          http.request(req)
+        end
+      rescue => e
+        # Perform a blind rescue because Net:HTTP throws a variety of exceptions - some of which are platform specific.
+        Chef::Log.error "Failed to connect to elasticsearch service. Retrying.\n#{e}"
+        if current_request == max_requests
+          raise "Failed to connect to elasticsearch service. Ensure node['private_chef']['opscode-solr4']['external_url'] is correct.\n#{e}"
+        else
+          current_request += 1
+          sleep(current_request * 2)  # Exponential back-off.
+          retry
+        end
+      end
+      raise "Unable to interrogate elasticsearch server - HTTP#{res.code}:\n#{res.body}" if res.code.to_i >= 400
+      # This can raise exceptions if the response does not match the format we expect.
+      version = JSON.parse(res.body)['version']['number'].split('.').first.to_i
+      raise "Unsupported elasticsearch version of #{version}. There is current support for the major versions of 2 and 5." if version != 5 && version != 2
+      version
+    else
+      # Elasticsearch is disabled - this configuration setting should never be used in erlang.
+      0
+    end
   end
 
   def solr_url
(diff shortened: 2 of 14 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -58,6 +59,28 @@ def solr_root
...
+      elastic_search_url = node['private_chef']['opscode-solr4']['external_url']
+      begin
+        elastic_search_uri = URI.parse(elastic_search_url)
+        req = Net::HTTP::Get.new(elastic_search_uri)
+        res = Net::HTTP.start(elastic_search_uri.hostname, elastic_search_uri.port, use_ssl: elastic_search_uri.scheme == "https") do |http|
+          http.request(req)
+        end
+        version = JSON.parse(res.body)['version']['number'].split('.').first.to_i
+        version = default_version if version == 0
+        version
```
stevendanna: Should we check that the discovered version is one we support?

dev/scripts/provision-elasticsearch.sh
```diff
@@ -0,0 +1,17 @@
+add-apt-repository ppa:openjdk-r/ppa
```
ksubrama: Yeah makes sense.  I was just doing copy-pasta based on the surrounding code for postgres and chef-backend to make my life easier in the Vagrantfile.

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
ksubrama: I considered that - but then I didn't know what we did support.  There's only ES 1, 2 and 5 right now (6 is on the way).  I guess we don't support 1.  Are there any minor versions in 2 that we don't support?

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
```diff
@@ -33,6 +33,8 @@
 
 rabbitmq = OmnibusHelper.new(node).rabbitmq_configuration
 
+elastic_search_version = OmnibusHelper.new(node).elastic_search_major_version
```
ksubrama: It's being passed below in line 57.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/pedant_config.rb.erb
```diff
@@ -75,7 +75,7 @@ search_server "<%= @solr_url %>"
 
 <% if node['private_chef']['opscode-erchef']['search_provider'] == "elasticsearch" %>
 search_commit_url "/_refresh"
-search_url_fmt "/chef/_search?fq=X_CHEF_type_CHEF_X:%{type}&q=%{query}"
+search_url_fmt "/chef/_search?q=X_CHEF_type_CHEF_X:%{type}%%20%{query}"
```
ksubrama: Elastic search experts: Can you confirm if this is a legit change.  This is what @lancewf and I came up with to get the current tests passing but maybe there's an edge case we're missing.

src/oc_erchef/apps/chef_index/README.md
```diff
@@ -6,7 +6,7 @@ Chef is a system integration framework written in erlang and ruby and designed t
 
 The Chef Wiki is the definitive source of user documentation.
 
-    http://wiki.chef.io/display/chef/Home
```
ksubrama: The only line I think is important is the "put any changes that affect oc_reporting in chef_index".  I'll trim the rest.

Speaking of which...  I probably need to port some of this over don't I?

stevendanna: If and only if we care about reporting support on ES5 :D

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
stevendanna: In my diff, line 57 is literally a new empty line, is github trolling me?

ksubrama: Figured it out.  @lancewf 's editor had been trolling him the entire time we coded this up.  It put in 10 tabs or something instead of 20 spaces and I'm fairly sure github uses 4 or 8 space tabs.

src/oc_erchef/apps/chef_index/README.md
ksubrama: Oh good point.  Yeah f-that.  We should have automate notifications in a good enough place to ask anyone upgrading to ES5 to use that instead of reporting. #famouslastwords.

Heads up to @charlesjohnson.

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
lancewf: I am not sure how that happened. We even tested that change before we checked it in.

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
lancewf: I updated this part to log an error if the major version is not 2 or 5.

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
marcparadise: "Worked on my machine - it's an ~ops~ pipeline problem now!" :D

marcparadise:
A couple of things we'll need to verify for this: 

* Reporting uses a forked chef_index search interface. It does not currently support ES, but a sanity check to ensure that reporting pedant tests pass would be good - the path that uses search will be hit in those tests.  
* We're building this because we need it, but what do we want to do in terms of documenting the change for customers? This is a factor for cloud all-in-one installations , but also customers who are on the "backendless" path.

ksubrama:
We added some unit tests to the library in the omnibus cookbook.  I figure that the pedant tests exercising functionality should cover most of the other scenarios we care about and ensure the ES2/5 work like we expect solr to work.  It isn't automated at all - but I have run pedant tests locally for both ES2 and ES5.  There's one test that involves rabbitmq that seems to be flaking out but that test behaves the same way in master as well - so I'm not worrying about it.  All the other tests past.

I'll run the reporting pedant tests and see if that still works on just solr.

kmacgugan:
The query changes look good to me.

ksubrama:
Ran opscode-reporting-ctl test --all using solr on this branch!  Everything still passes.

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -58,6 +59,31 @@ def solr_root
     host
   end
 
+  def elastic_search_major_version
```
stevendanna: I'm a bit concerned that a transient ES issue during `reconfigure` could rather silently (one log message in hundreds of lines of output) reconfigure us to use version 2 instead of version 5 and produce very confusing errors for the user.

ksubrama: That's a good point...  Should we raise exceptions and fail hard for some of these?

ksubrama: Fixed to raise exceptions when faced with invalid data.  Also does retries on flaky connections.

PR #1288, opened 2017-06-02
oc-chef-pedant: fix override logic

srenatus:
Before, we'd end up with a payload that contained both
`"email" => "username@chef.io"` and `:email => "what-was-overridden"`.

### DIFF
--- a/oc-chef-pedant/lib/pedant/platform.rb
+++ b/oc-chef-pedant/lib/pedant/platform.rb
@@ -380,12 +380,12 @@ module Pedant
 
     def create_min_user(username, options = {})
       payload = {
-        "username" => username,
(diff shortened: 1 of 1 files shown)


stevendanna:
I think this is only a problem if you pass in symbols in your override array.  Won't this create problems for other call-sites that are using strings as symbols?  For instance, in the client_key_spec?

srenatus:
@stevendanna I believe the only call site _is indeed_ [this one](https://github.com/chef/chef-server/blob/608dbe94d15822a31849952e13549744fc40a702/oc-chef-pedant/spec/api/user_spec.rb#L91) and it's passing symbols for keys.

srenatus:
Sorry for the noise 😉

PR #1289, opened 2017-06-02
[POOL-606] oc-chef-pedant: improve oc_id API test coverage

srenatus:
This does not involve controllers/routes that:

a) send an email (we can't do that on prem)
b) require an existing oauth2 application
c) zendesk

Profile update does not include the email address because this, too,
would send an email.

### DIFF
--- a/oc-chef-pedant/lib/pedant/rspec/common.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/common.rb
@@ -387,6 +387,16 @@ module Pedant
           end
         end
 
+        # Used to easily documented endpoints that don't have tests
+        def self.untested_endpoint(endpoint_name, reason)
+          context endpoint_name do
+            it "returns success" do
+              pending(reason)
+              raise NotImplementedError
+            end
+          end
+        end
+
         # Run the given block once (using 'key' to identify
         # the block).  This allows us to do "before :each"
         # cleanup to clean up from previous test runs, since
--- a/oc-chef-pedant/spec/api/oc_id_spec.rb
+++ b/oc-chef-pedant/spec/api/oc_id_spec.rb
@@ -2,6 +2,38 @@ require 'pedant/rspec/common'
 require 'openssl'
 
 describe "oc_id API", :oc_id do
+
+  let(:request_headers) do
+    {
+      "Content-Type" => "application/x-www-form-urlencoded",
+      "Accept" => "text/html",
+      "Cookie" => csrf[:cookie]
+    }
+  end
+
+  let(:username) {  "pedant-oc-test-user-#{rand(10**7...10**8).to_s}" }
+  let(:password) { "foobar" } # hardcoded at the platform layer
+  let(:user_overrides) do
+    {
+      :overrides => {
+        "first_name" => "#{username}-first-name",
+        "middle_name" => "#{username}-middle-name",
+        "last_name" => "#{username}-last-name",
+        "display_name" => "#{username}-display-name",
+        "email" => "#{username}-email@host.com",
+      }
+    }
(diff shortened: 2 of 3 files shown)


oc-chef-pedant/spec/api/oc_id_spec.rb
```diff
@@ -2,6 +2,38 @@
...
+        "last_name" => "#{username}-last-name",
+        "display_name" => "#{username}-display-name",
+        "email" => "#{username}-email@host.com",
+      }
+    }
+  end
+
+  # create user before each test so that we don't regen the keys
+  # of the standard user used in other tests.
+  let(:oc_id_user) { platform.create_user(username, user_overrides) }
```
marcparadise: nit: Could we do this as `shared(:oc_id_user)` and clean it up in an `after(:all)`?  Disregard if the tests change user record state in a way that would affect other tests.  It's not a huge deal, I'm just hoping to avoid increasing overall pedant time where it can be done safely.

oc-chef-pedant/spec/api/oc_id_spec.rb
```diff
@@ -2,6 +2,38 @@
 require 'openssl'
```
marcparadise: Would it make sense to include a comment specifying endpoints that exist but are not covered, with a reference to the unit coverage? The possible down side is remembering to update that when it changes - but Id endpoints are relatively static.

oc-chef-pedant/spec/api/oc_id_spec.rb
srenatus: Cool. I didn't know how to do that. :+1:

oc-chef-pedant/spec/api/oc_id_spec.rb
srenatus: :+1: can do. It makes sense to not only have that information in the Jira ticket, but also where you might _want to see_ it. 😉

oc-chef-pedant/spec/api/oc_id_spec.rb
stevendanna: I've taken a stab at doing this but it gets a bit ugly to be honest since: (1) As far as I can tell `shared` isn't a thing, so you need to use instance variables in a `before` block and (2) we have at least 3 tests that change the user-object.  We can order those tests precisely so they don't conflict, but I think it is just best to eat the extra tests for now.

oc-chef-pedant/spec/api/oc_id_spec.rb
stevendanna: I added a new `untested_endpoint` method that we can use to document issues like this.

oc-chef-pedant/lib/pedant/rspec/common.rb
```diff
@@ -387,6 +387,16 @@ def self.isolate(message = nil, &examples)
           end
         end
 
+        # Used to easily documented endpoints that don't have tests
```
srenatus: :100: good idea

oc-chef-pedant/spec/api/oc_id_spec.rb
marcparadise: That works for now. `shared` is a pedant thing that's hiding somewhere in our libs.  Semantically similar to `let` but less robust for overriding iirc, and it persists across examples so that they can be accessed for setup in before/after :all blocks.

stevendanna: Hrm, perhaps we've broken it then, because it super-duper didn't work right for me.  BUT, it could be that I was actually hitting some of the data interactions.

PR #1290, opened 2017-06-02
Elvis for oc_erchef, oc_bifrost, and bookshelf

srenatus:
How about using this code _style_ checker for our code bases?

This PR introduces the latest version (053db5bded2b6b71894865472c484379876d7856) of [Elvis](https://github.com/inaka/elvis).

More info about the configurable rules [can be found here](https://github.com/inaka/elvis_core/wiki/Rules). The initial set is fairly conservative, but let's discuss them, and if we want to use this tool. 😃 

(Adding it for the other apps (excluding mover) could be done later...)


srenatus:
Btw, couldn't find any satisfying rebar3 plugin for this yet, fifo's is a bit behind (and I couldn't get it to work quickly), some other one (deadtrickster) is even more behind.

srenatus:
🏓 @chef/chef-server-maintainers

src/oc_erchef/elvis.config
```diff
@@ -0,0 +1,30 @@
+[
```
marcparadise: Let's add a comment to reference the config docs, https://github.com/inaka/elvis_core/wiki/Rules

src/oc_erchef/elvis.config
```diff
@@ -0,0 +1,30 @@
...
+          {elvis_style, no_if_expression},
+          {elvis_style, no_debug_call, #{ignore => [chef_test_db_helper,
+                                                    chef_test_suite_helper,
+                                                    oc_chef_wm_app]}},
+          {elvis_style, no_nested_try_catch},
+          {elvis_style, no_tabs},
+          {elvis_style, no_trailing_whitespace},
+          {elvis_style, operator_spaces},
+          {elvis_style, used_ignored_variable},
+          {elvis_style, variable_naming_convention}
```
marcparadise: A personal preference is to add `use_your_spacebar`, but I imagine that would have quite a bit of required cleanup associated.

srenatus: I don't quite understand, I'm afraid. Is that a rule? I don't find it. 😕

srenatus:
Note that with [this upstream fix](https://github.com/inaka/elvis/pull/448), elvis will default to being less noisy, too.

PR #1293, opened 2017-06-06
Fix `omnibus cache missing` etc

srenatus:
Seems like the _second parameter_ to `instance_eval` is not *only*

> used when reporting compilation errors

but also fixes the __FILE__ references. Without this change,

    omnibus cache missing

would end up trying to read omnibus_overrides.rb from a path relative to
the CWD, and therefore fail.

### DIFF
--- a/omnibus/.kitchen.yml
+++ b/omnibus/.kitchen.yml
@@ -6,8 +6,6 @@ driver:
     memory: 4096
   synced_folders:
     - ['..', '/home/vagrant/chef-server']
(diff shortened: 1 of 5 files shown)


srenatus:
:ping: @chef/chef-server-maintainers

srenatus:
@stevendanna I assume so, too. 😉 I'll schedule an adhoc build later.

srenatus:
_Now_ it started building locally. I've added commits updating the docs and adding `zip`.

PR #1294, opened 2017-06-06
[POOL-589] Add regexp for {first, middle, last, display}name

srenatus:
I've tried to be liberal. What this still forbids is anything html-y, or
SQL-y. It's probably not even close to capturing real-world names.

### DIFF
--- a/src/oc_erchef/apps/chef_objects/test/chef_user_tests.erl
+++ b/src/oc_erchef/apps/chef_objects/test/chef_user_tests.erl
@@ -26,7 +26,7 @@
 -include_lib("eunit/include/eunit.hrl").
 
 
--define(VD(D), chef_objects_test_utils:versioned_desc(Version,D)).
+-define(VD(D), chef_objects_test_utils:versioned_desc(Version, D)).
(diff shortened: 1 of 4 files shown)


src/oc_erchef/apps/chef_objects/test/chef_user_tests.erl
```diff
@@ -287,6 +287,33 @@ parse_binary_json_tests(Version) ->
...
+      end
+     }
+     || Field <- [<<"display_name">>, <<"firstname">>, <<"middlename">>, <<"lastname">>]
+    ]
+    ++
+    [
+     {?VD(lists:flatten(io_lib:format("Works with non-ASCII ~s", [Field]))),
+      fun() ->
+              %% "Maryam", #1 female name in the arab world as of 2015
+              Value = <<"مريم 1. O'Mara">>,
```
srenatus: Interesting. Must have to do with my editor not knowing right-to-left text, because it looks different there:
screen shot 2017-06-06 at 17 26 05

marcparadise:
I'll merge this after the travis re-run passes - looks like our authz connection pool ran out in CT on the last run.

PR #1299, opened 2017-06-12
Fix keys for validating name fields

srenatus:
It's got to be first_name, middle_name, and last_name.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -7,6 +7,19 @@ in the release. For a detailed list of changed components, refer to
 This document contains release notes for the current major release and all patches.
 For prior releases, see [PRIOR\_RELEASE\_NOTES.md](PRIOR_RELEASE_NOTES.md).
 
+## 12.15.X (TBD)
+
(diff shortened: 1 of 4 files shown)


RELEASE_NOTES.md
```diff
@@ -7,6 +7,18 @@ in the release. For a detailed list of changed components, refer to
 This document contains release notes for the current major release and all patches.
 For prior releases, see [PRIOR\_RELEASE\_NOTES.md](PRIOR_RELEASE_NOTES.md).
 
+## 12.15.X (TBD)
+
+* [Stricter validation of non-functional user record fields](https://github.com/chef/chef-server/pull/1294),
```
stevendanna: My only worry about this release note text is that it may imply that we currently have a problem with SQL injection even though we don't think that is the case.

srenatus: I'll rephrase it more defensively. :+1:

stevendanna: 👍  I like the new wording.

PR #1300, opened 2017-06-12
Lazily evaluate ES version detection code

stevendanna:
When running against chef-backend, the external elasticsearch node is proxies by HAProxy, which is started by Chef Server.  Thus, we need to make sure the ES version detection code runs at compile-time to avoid it getting ECONNREFUSED.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
@@ -4,6 +4,7 @@
 #
 # All Rights Reserved
 
+helper = OmnibusHelper.new(node)
 opscode_erchef_dir = node['private_chef']['opscode-erchef']['dir']
 opscode_erchef_log_dir = node['private_chef']['opscode-erchef']['log_directory']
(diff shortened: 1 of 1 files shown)


stevendanna:
@Sgtpluck Thanks for getting this started! You'll need to add the DCO Sign-off to your commit.  See here for details: https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-erchef.rb
```diff
@@ -55,7 +57,9 @@
                                                                  :actions_exchange => actions_exchange,
                                                                  :ldap_encryption_type => ldap_encryption_type,
                                                                  :solr_elasticsearch_major_version => elastic_search_version,
-                                                                 :helper => OmnibusHelper.new(node)))
+                                                                 :helper => OmnibusHelper.new(node))
+     }
+   }
```
phiggins: If the thing on line 51 is a hash, I think the curly brackets on lines 50 and 61 are unnecessary.

PR #1302, opened 2017-06-12
[No Merge] Logical Replication POC

marcparadise:
Opening this for comment/feedback.


marcparadise:
This is currently in an in-between state.  Friday the pieces were working through `dvm start oc_erchef` testing, manual calls, etc. Over the weekend I started moving in the direction of e direction of feeding the requests through, start to finish; that should be about 90% there at this point.  

Missing to get this minimally functional: 

* `migrator_encode:encode` is not validated in current form. Some pieces got moved from another module and modified for issues I caught reading through (namely, but I haven't re-validated after completing).
* migrator_target_db needs to connect to the target postgres instance and apply the statements. This is outlined in the 

A final note on the caching scheme: 
- that's what I _wanted_ to do if time allowed.  For short-term/PoC - no need to do this, but we will want that or something like it if we go forward with it. 
- if we're not doing it,  a better alternative might just be do to a prepare/execute/free sequence in `migrator_target_db` - it's just a couple of extra lines over executing raw sql, is probably less effort than changing the encoder to build the content with escaped inline values.

src/oc_erchef/apps/migrator/LICENSE
```diff
@@ -0,0 +1,29 @@
+Copyright (c) 2017, Marc A. Paradise <marc.paradise@gmail.com>.
```
stevendanna: :D Wrong license

src/oc_erchef/apps/migrator/src/migrator_change_listener.erl
```diff
@@ -0,0 +1,211 @@
...
+%  - I used epgsql directly:
+%       * sqerl only supports one DB connection - presently that's opscode_chef as the erchef user;
+%         as an alterntiave I could have used sqerl for the polling by giving the erchef user
+%         replication privileges, though initially the direction was a bit different and that made less sense.
+%       * we don't need pooling.  There's a dedicated connection to siphone dat from the logical repl slot
+%         and that's it.
+%       * When/if we move this to its own standalone service, that's a thing to revisit and check perf impact of.
+%
+% Open questions:
+%  - how do we handle updatesa that are done as part of triggers or stored procedures?
```
stevendanna: I think the way to do this is to set the session_replication_role on the session to replica: https://www.postgresql.org/docs/9.6/static/runtime-config-client.html

src/oc_erchef/apps/migrator/src/migrator_change_listener.erl
```diff
@@ -0,0 +1,211 @@
...
+%
+% Because this is PoC/spike work, you won't find new tests here yet.
+%
+% Some thoughts:
+%  - currently this does all parsing in the same proc. We may want to farm that out to
+%    one or more workers, which can sync and sequence on the trans id.  Particularly once we're
+%    doing this for multiple DBs.
+%  - This does attempt to prevent transaction loss by 'peek'ing one change at a time, then
+%    'get'ing it to clear it from the replication slot.  Errors that cause the supervisor to give up
+%    due to restart intensity limits would be considered fatal to the migration.
```
stevendanna: We'll have to see what this means for perf since we'll be getting the content twice, which I imagine could get a bit gnarly for cookbook content and node content.  One mitigation is that the double-transfer is over localhost so it isn't /awful/ but still something to keep in mind.

src/oc_erchef/apps/migrator/LICENSE
marcparadise: Damn those rebar templates.

src/oc_erchef/apps/migrator/src/migrator_change_listener.erl
marcparadise: Nice!

doc/scrap2.erl
```diff
@@ -0,0 +1,44 @@
+
+
+{ok, _Fields, Data} = epgsql:squery(C2, "select * from pg_logical_slot_get_changes('regression_slot',NULL, NULL, 'include-xids','1','force-binary','0', 'skip-empty-xacts', '1')").
```
marcparadise: We can probably nuke this file

src/oc_erchef/apps/migrator/test/migrator_tests.erl
```diff
@@ -0,0 +1,29 @@
+-module(migrator_tests).
+
+-compile([export_all]).
```
marcparadise: :+1: on this test - we'll also want something similar as a sanity check prior to accepting a command to cutover, I think.

codeadict: this `export_all` can be an issue if you plan to support OTP 20 in the future, even on 19 it may complain, can use:

```
-compile([nowarn_export_all, export_all]).
```

stevendanna:
I'm going to close this PR but leave the branch around.  Postgresql logical replication is definitely an interesting tool that we should keep in mind when solving future problems, but I think it is safe to say that this PR isn't going forward in the near term.

PR #1303, opened 2017-06-13
[POOL-589] Allow user creation with blank middle name.

stevendanna:


src/oc_erchef/apps/chef_objects/src/chef_regex.erl
```diff
@@ -74,6 +74,9 @@
 %% Validation regex for various names
 -define(HUMAN_NAME_REGEX, "[[:word:][:digit:]!'. -]+").
 
+%% Validation regex for various names that can be empty.
+-define(OPTIONAL_NAME_REGEX, "[[:word:][:digit:]!'. -]*").
```
stevendanna: Perhaps we should change `HUMAN_NAME_REGEX` since first_name and last_name are also optional?  The only non-optional field is display_name, but my guess is we have previously allowed empty strings there as well.

src/oc_erchef/apps/chef_objects/test/chef_user_tests.erl
```diff
@@ -180,6 +180,26 @@ parse_binary_json_tests(Version) ->
                 ?assertEqual(lists:sort(MinValid), lists:sort(GotData))
       end
      },
+     {?VD("Can create user with empty middle name"),
```
stevendanna: In addition to these unit tests, we will want to add oc-chef-pedant test cases for the /users API.

stevendanna:
👍 Github won't let me approve this since I opened it.

PR #1306, opened 2017-06-15
Update RELEASE_PROCESS.md and CODE_REVIEW_CHECKLIST.md

srenatus:
This is influenced by our internal ops' checklist for releases; however, I've not carried over stuff that is _nice to have_ but currently unrealistic.

Also updates references to DCC -> ACC, and the org name on that Automate instance.

For the code review checklist additions, I've added stuff that I thought reasonable, please scrutinise! 😉

### DIFF
--- a/CODE_REVIEW_CHECKLIST.md
+++ b/CODE_REVIEW_CHECKLIST.md
@@ -78,3 +78,6 @@ may have.
 
 - [ ] Should this change be reflected in the release notes? If so, has
       RELEASE_NOTES.md been updated?
+      (This doesn't include: dev VM changes, test-only changes,
+      code refactorings without behavioral changes, dependency
+      upgrades.)
--- a/CODE_REVIEW_CHECKLIST.md
+++ b/CODE_REVIEW_CHECKLIST.md
@@ -72,6 +72,16 @@ may have.
 - [ ] Will this change work on all of our supported platforms and
       browsers?
 
+      Supported platforms (architectures) are:
+      - el-5, el-6, el-7, sles-11, sles-12, ubuntu 12.04, ubuntu 14.04 (x86_64)
+      - el-7 (ppc64)
+      - el-7, ubuntu 14.04 (ppc64le)
+      - el-6, el-7, sles-11, sles-12 (s390x)
+
+      Note: For outside contributers, it is impossible to verify that.
+      A Chef employee with access to our internal CI infrastructure
+      will support them by executing test runs for outside PRs.
+
 - [ ] Do we expect significant changes in performance due to the
       changed code? Do we have any indication of the impact related to
       those performance changes?
--- a/CODE_REVIEW_CHECKLIST.md
+++ b/CODE_REVIEW_CHECKLIST.md
@@ -46,7 +46,9 @@ may have.
       log level?
 
 - [ ] Does this change introduce new dependencies? If so, do we
-      believe those dependencies are necessary and stable?
+      believe those dependencies are necessary and stable? Are all new
+      dependencies included in the omnibus package OR documented as
+      things we expect on the hosts (e.g. OpenSSH)?
 
 - [ ] Does this change break the dev environment? If so, can we fix it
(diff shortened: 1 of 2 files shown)


CODE_REVIEW_CHECKLIST.md
```diff
@@ -69,3 +71,10 @@ may have.
 
 - [ ] Will this change work on all of our supported platforms and
       browsers?
+
+- [ ] Do we expect significant changes in performance due to the
```
blakestier: This is very "scrutinizey": I think both this point and the one above are important-- is there a place that documents how to go about verifying supported platforms/performance impact? Should it be expected that if a code-writer has seen these that s/he/it note that on the PR in anticipation of the question?

CODE_REVIEW_CHECKLIST.md
```diff
@@ -69,3 +71,10 @@ may have.
 
 - [ ] Will this change work on all of our supported platforms and
```
blakestier: Where do we find what our supported platforms/browsers are?

CODE_REVIEW_CHECKLIST.md
```diff
@@ -69,3 +71,10 @@ may have.
...
       browsers?
+
+- [ ] Do we expect significant changes in performance due to the
+      changed code? Do we have any indication of the impact related to
+      those performance changes?
+
+- [ ] Should this change be reflected in the release notes? If so, has
```
blakestier: How do I know when a change should be reflected in release notes?

CODE_REVIEW_CHECKLIST.md
srenatus: You're right about this leading into long, dark alleys... My intention with including this was merely to have a reminder -- _"Is this going to be trouble for big deployments?"_ (where "big" again is very vague)...

Trying to come up with recent changes that would have been relevant for this rule, we've had the added functional PG index on `lower(email)` to allow for faster case-insensitive lookups. Adding that index feels instantaneous on my small users table, but could have been an issue for Hosted Chef. For this reason, we had changed the code so it'll create the index asynchronously.

Now this example is only for "upgrade-time", i.e. how long does it take to upgrade your chef-server, but I assume similar things could exist for others. (In the end, adding the functional index is a reaction to changing the search query -- only changing the query without adding the index would have turned a fast PG index lookup into a slow sequential scan.)

So, I don't believe this is a clear-cut criterium for accepting a change, but rather a reminder -- I could see value in splitting the list into two sections, but I also don't want to introduce too much possibly unnecessary structure -- what would you suggest, @blakestier? 😃

CODE_REVIEW_CHECKLIST.md
srenatus: Good catch -- While I'd say _"everything we build for in Jenkins"_, this is not something an outside contributor would be able to see. I'll update this.

srenatus: To be honest, I have no idea about the browsers! 😕 That makes this a bad checklist item. 

*But* since the web-accessible part of _Chef Server_ is quite minimal (oc_id), and should stay that way, this item doesn't say much anyways... 🤔

CODE_REVIEW_CHECKLIST.md
srenatus: Tried clarifying... But I've also noticed that it's not something external contributors would ever be able to check...

srenatus: Heh, that's not the markup I've intended, but I'll keep it like that. ✔️

CODE_REVIEW_CHECKLIST.md
blakestier: I think it's probably fine as is-- My concern is that this is an easy one to sort of gloss over if not asked to list or discuss "possible ramifications", but I don't think the overhead of requesting an answer is worthwhile, here (unless we see a trend of PRs that have no consideration of performance).

blakestier:
LGTM! Thanks for being thoughtful about my questions!

Issue #1308, opened 2017-06-15
Elasticsearch 5 Support Will Break External Solr


sdelano:
We still support the use of external search via Solr, but the changes with https://github.com/chef/chef-server/pull/1287 will break this.

The helper ([link](https://github.com/chef/chef-server/blob/fea4447c5ddcc31047576db6052b635e9316cbf7/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb#L66)) only checks whether we've configured external search, but not that the engine is Elasticsearch. If we're using Solr, this will break.

lancewf:
I will be looking into this problem today. I will run the pedant tests with an external solr service and see what breaks.

lancewf:
The only change that is needed is to add a check for node['private_chef']['opscode_erchef']['search_provider']` == `elasticsearch` [here](https://github.com/chef/chef-server/blob/fea4447c5ddcc31047576db6052b635e9316cbf7/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb#L66)

I am writing some tests and adding to the dev environment the option of an external solr.

PR #1309, opened 2017-06-16
[CON-689] issue 1308 Elasticsearch 5 Support Will Break External Solr

lancewf:

https://github.com/chef/chef-server/issues/1308

### DIFF
--- a/dev/Vagrantfile
+++ b/dev/Vagrantfile
@@ -13,7 +13,8 @@ IPS = {
   ldap: "192.168.33.152",
   custom: "192.168.33.153",
   reportingdb: "192.168.33.155",
-  elasticsearch: "192.168.33.156"
+  elasticsearch: "192.168.33.156",
+  solr: "192.168.33.157"
 }
 
 
@@ -104,6 +105,16 @@ Vagrant.configure("2") do |config|
     attributes['vm']['elasticsearch'] = nil
   end
 
+  if attributes['vm'].has_key? 'external_solr'
+    if attributes['vm']['external_solr']['start'] == true
+      config.vm.define('external_solr') do |c|
+        define_external_solr_server(c, attributes)
+      end
+    end
+  else
+    attributes['vm']['external_solr'] = nil
+  end
+
   if attributes['vm']['start-custom'] == true
     config.vm.define("custom") do |c|
       define_custom_server(c, attributes)
@@ -178,6 +189,15 @@ def define_chef_server(config, attributes)
                         "opscode_erchef['search_queue_mode']" => '"batch"'
       }
       json = simple_deep_merge(json, { "provisioning" => { "chef-server-config" => elasticsearch } })
+    end
+
+    if vmattr['external_solr'] && vmattr['external_solr']['start']
+      external_solr = { "opscode_solr4['external']" => "true",
+                        "opscode_solr4['external_url']" => '"http://solr.chef-server.dev:8983/solr"',
+                        "opscode_erchef['search_provider']" => '"solr"',
+                        "opscode_erchef['search_queue_mode']" => '"batch"'
+      }
+      json = simple_deep_merge(json, { "provisioning" => { "chef-server-config" => external_solr } })
     end
 
     dotfiles_path = vmattr["dotfile_path"] || "dotfiles"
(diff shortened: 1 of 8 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -84,9 +84,13 @@ def elastic_search_major_version
...
       raise "Unable to interrogate elasticsearch server - HTTP#{res.code}:\n#{res.body}" if res.code.to_i >= 400
       # This can raise exceptions if the response does not match the format we expect.
-      version = JSON.parse(res.body)['version']['number'].split('.').first.to_i
-      raise "Unsupported elasticsearch version of #{version}. There is current support for the major versions of 2 and 5." if version != 5 && version != 2
-      version
+      begin
+        version = JSON.parse(res.body)['version']['number'].split('.').first.to_i
+        raise "Unsupported elasticsearch version of #{version}. There is current support for the major versions of 2 and 5." if version != 5 && version != 2
+        version
+      rescue => e
```
srenatus: [nit] Can we say `StandardError` here?

sdelano: It also seems really odd that we raise inside of a `begin/rescue` block...

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -84,9 +84,13 @@ def elastic_search_major_version
       end
       raise "Unable to interrogate elasticsearch server - HTTP#{res.code}:\n#{res.body}" if res.code.to_i >= 400
       # This can raise exceptions if the response does not match the format we expect.
-      version = JSON.parse(res.body)['version']['number'].split('.').first.to_i
-      raise "Unsupported elasticsearch version of #{version}. There is current support for the major versions of 2 and 5." if version != 5 && version != 2
-      version
+      begin
+        version = JSON.parse(res.body)['version']['number'].split('.').first.to_i
```
sdelano: It doesn't appear that we're testing the raise here when we get no data, but maybe that's for another change?

sdelano:
Thank you so much for adding tests here around external services. This is awesome!

dev/Vagrantfile
```diff
@@ -104,6 +105,16 @@ Vagrant.configure("2") do |config|
     attributes['vm']['elasticsearch'] = nil
   end
 
+  if attributes['vm'].has_key? 'external_solr'
+    if attributes['vm']['external_solr']['start'] == true
```
phiggins: This would be a bit more idiomatic ruby written like:

```
if attributes['vm']['external_solr'] && attributes['vm']['external_solr']['start']
```

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
ksubrama: It was attempting to decorate the JSON exceptions with more context.  I restructured it to be more explicit.

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
ksubrama: Added

dev/Vagrantfile
ksubrama: It totally makes sense...  but for whatever reason all the other attributes['vm'] clauses use this strange pattern.  It might just be copy-pasta, but I it'd be better to just fix them all up in another pass.  Some of this (the chef_backend stuff assigning nil but then reading an attribute underneath it later) doesn't even seem quite right but those codepaths may never ever get used.  I might simplify it a bunch.

phiggins: This is definitely not a blocker. 👍

dev/Vagrantfile
```diff
@@ -180,6 +191,15 @@ def define_chef_server(config, attributes)
       json = simple_deep_merge(json, { "provisioning" => { "chef-server-config" => elasticsearch } })
     end
 
+    if vmattr['external_solr'] && vmattr['external_solr']['start']
+      external_solr = { "opscode_solr4['external']" => "true",
+                        "opscode_solr4['external_url']" => '"http://solr.chef-server.dev:8983/solr"',
+                        "opscode_erchef['search_provider']" => '"solr"',
+                        "opscode_erchef['search_queue_mode']" => '"batch"'
+      }
+      json = simple_deep_merge(json, { "provisioning" => { "chef-server-config" => external_solr } })
```
elliott-davis: I sort of hope there is a `complex_deep_merge` somewhere

PR #1310, opened 2017-06-16
Upgrade Postgres 9.2 to 9.6

srenatus:
I looks like the old machinery made for going from 9.1 to 9.2 still works just fine! 🎉 

Pending:
- [x] automatic upgrade tests (qa-chef-server-cluster)
- [x] check tunables one-by-one to see if any other ones are gone, clarify release notes
- [x] figure out an edge case or bug with the sentinel file creation
- [x] cleanup 9.2 stuff

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,6 +14,28 @@ For prior releases, see [PRIOR\_RELEASE\_NOTES.md](PRIOR_RELEASE_NOTES.md).
   the secret being changed. With the optional flag `--with-restart`,
   `chef-server-ctl set-secret` will attempt to automatically restart the
   dependent services.
+* [Upgrade to PostgreSQL 9.6](https://github.com/chef/chef-server/pull/1310),
+  Chef Server now uses the latest stable version of the 9.6 series (9.6.3). Upgrades of existing
+  installations are done automatically, but creating backups is advised.
+
+  The information below only applies if you have set a custom value for `checkpoint_segments`
+  in your `/etc/opscode/chef-server.rb`. If you have not set a custom value, there is nothing to
+  change:
+
+  The `checkpoint_segments` configuration setting is gone, so if you before have set
+
+      postgresql['checkpoint_segments'] = 10
+
+  you now have to specify (for example, please consult the documentation linked below)
+
+      postgresql['max_wal_size'] = '3G'
+
+  The default setting for `max_wal_size` is 1G. [The PostgreSQL release notes](https://www.postgresql.org/docs/9.6/static/release-9-5.html) mention a conversion
+  rule: `max_wal_size = (3 * checkpoint_segments) * 16MB`. They also say that the default value
+  for `max_wal_size` (1GB) should fit in most settings, so this conversion is not done automatically.
+
+  This update also adds two further new configurables in the ["Checkpoints" group](https://www.postgresql.org/docs/9.6/static/runtime-config-wal.html#RUNTIME-CONFIG-WAL-CHECKPOINTS), `min_wal_size` and
+  `checkpoint_flush_after`.
 
 ## 12.15.8 (2017-06-20)
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -37,6 +37,9 @@ For prior releases, see [PRIOR\_RELEASE\_NOTES.md](PRIOR_RELEASE_NOTES.md).
   This update also adds two further new configurables in the ["Checkpoints" group](https://www.postgresql.org/docs/9.6/static/runtime-config-wal.html#RUNTIME-CONFIG-WAL-CHECKPOINTS), `min_wal_size` and
   `checkpoint_flush_after`.
 
+  As part of the upgrade procdure, running `chef-server-ctl cleanup` will remove Postgres 9.2's data
+  and logs.
+
 ## 12.15.8 (2017-06-20)
 
 * [Stricter validation of non-functional user record fields](https://github.com/chef/chef-server/pull/1294),
--- a/omnibus/config/software/postgresql92-bin.rb
+++ b/omnibus/config/software/postgresql92-bin.rb
@@ -1,0 +1,57 @@
+#
+# Copyright 2017 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name "postgresql92-bin"
+
+default_version "9.2.21"
+
+license "PostgreSQL"
+license_file "COPYRIGHT"
+skip_transitive_dependency_licensing true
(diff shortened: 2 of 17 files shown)


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -618,7 +618,14 @@
 
 default['private_chef']['postgresql']['work_mem'] = "8MB"
 default['private_chef']['postgresql']['effective_cache_size'] = "#{(node['memory']['total'].to_i / 2) / (1024)}MB"
-default['private_chef']['postgresql']['checkpoint_segments'] = 3
+# Note: the checkpoint_segments setting was removed.
+# https://www.postgresql.org/docs/9.6/static/release-9-5.html says
+#   max_wal_size = (3 * checkpoint_segments) * 16MB
+# would be a usable conversion rule, but it also says the new setting's default
```
marcparadise: :+1: the defaults are sane according to several related docs I read (official, mailng list threads).

srenatus:
I _thought_ there was a problem when, for whatever reason, the upgrade process would fail; I think I saw a situation where we'd still get the sentinel file. However, I've tried reproducing this and couldn't ([recorded a video to convince myself](https://asciinema.org/a/Tv8g1gnly0J0eT9bnazD0GWeZ)).

srenatus:
want to give this another round of qa-chef-server-cluster tests, then I'll remove the `[hold]`. Reviews would be great, though 😀

srenatus:
🏓 @chef/chef-server-maintainers

omnibus/config/software/postgresql.rb
```diff
@@ -0,0 +1,62 @@
+#
```
marcparadise: At this point it looks like we might bre able to use the omnibus-software defition of postgres? Or asre we still stuck because of the version-specific prefix?
I'm wonder if we should just port that change over to the official definition?  Or make an alternative official definition (postgresql-versioned.rb?) that mirrors the pg one except for install path?

omnibus/config/software/postgresql92-bin.rb
```diff
@@ -14,7 +14,12 @@
 # limitations under the License.
 #
 
-name "postgresql92"
```
marcparadise: Similarly, a shared `postgres-versioned-bin-only` in omnibus-software?

omnibus/config/software/postgresql.rb
srenatus: 🎉 yes good idea. I'll do that.

omnibus/files/private-chef-cookbooks/private-chef/recipes/post_11_upgrade_cleanup.rb
```diff
@@ -65,16 +65,3 @@
 end
 
 private_chef_package_cleaner "varnish"
-
-# Remove old PostgreSQL data directory
-#
-# pg_upgrade will helpfully generate a cleanup script which removes
-# the previous database cluster.  This is nice, because we don't have
-# to worry about keeping track of the location of the old cluster.
```
marcparadise: Removing this  gets messy because of enterprise/private/old-oss chef server installations that could be getting upgraded from.
I'm concerned that the cleanup of the old data will no longer happen - because the postgres data was not always in a versioned directory IIRC - there may be older install that have the path as /var/opt/opscode/postgresql/data instead of postgresql/9.2/data.

Perhaps updating the new `postgres_upgrade_cleanup` to handle both would work?

This might also merit a modification of the upgrade pipeline to verify that old directories are removed.

RELEASE_NOTES.md
```diff
@@ -7,6 +7,29 @@ in the release. For a detailed list of changed components, refer to
 This document contains release notes for the current major release and all patches.
 For prior releases, see [PRIOR\_RELEASE\_NOTES.md](PRIOR_RELEASE_NOTES.md).
 
+## 12.15.X (TBD)
+
+* [Upgrade to PostgreSQL 9.6](https://github.com/chef/chef-server/pull/1310),
+  Chef Server now uses the latest stable version of the 9.6 series (9.6.3). Upgrades of existing
+  installations are done automatically, but creating backups is advised.
+  Note that the `checkpoint_segments` configuration setting is gone, so if you before have set
```
marcparadise: Maybe a prefix like: "The information below  only applies if you have set a custom value for `checkpoint_segments` in your `/etc/opscode/chef-server.rb`. If you have not set a custom value, there is nothing to change"
Maybe a preflight check against `checkpoint_segments` being present?

omnibus/files/private-chef-cookbooks/private-chef/recipes/post_11_upgrade_cleanup.rb
srenatus: It's merely moved to a generic location -- because it's now for the cleanup of *any* postgres upgrade. The comments [there](https://github.com/chef/chef-server/pull/1310/files#diff-1ada408659099d374677458edaaa8211R18) made me think that this script will contain the old data dir, irregardless of whether it is `/var/opt/opscode/postgresql/data` or versioned. (So it wouldn't care about going from 9.1 (unversioned) to 9.2, or 9.1 (unversioned) to 9.2, or 9.2 to 9.6.)

But I can look the left overs in the upgrade pipelines :+1:

RELEASE_NOTES.md
srenatus: Yeah you're right. 🙄 (Just kidding. You simple _are_ right there, and I'll add this 😉)

omnibus/config/software/postgresql.rb
srenatus: ⏩ https://github.com/chef/omnibus-software/pull/845 Any hints on how to make this less copy-pasty are very welcome!

srenatus:
@marcparadise I've checked off all the things from my list now ✔️ 

Well, a missing bit is checking the deleted old data dir in upgrade tests. But _if_ I find the right place for that test, it'll be in qa-chef-server-cluster anyways. So, could you throw some 👀 at those changes, please? I've intentionally not messed with the history yet, I'll squash and drop before merge.

omnibus_overrides.rb
```diff
@@ -10,3 +10,6 @@
 override :rubygems, version: "2.6.8"
 # This SHA is the last commit before the 6.0 release
 override :'berkshelf-no-depselector', version: '6016ca10b2f46508b1b107264228668776f505d9'
+
+override :'postgresql-bin-versioned', version: "9.2.21" # needed for upgrading
```
stevendanna: [minor] I wonder if we want to put the major version in the package name.  That is, this package would be called `postgresql92-bin`.

srenatus: That's actually what was there before -- but they commenters convinced me that there's a value in having `postgres-bin-versioned` in omnibus-software, which can be used for any version we'd like to _upgrade from_. [However, that bit hasn't passed review yet either](https://github.com/chef/omnibus-software/pull/845)

srenatus: ("commenters" = @marcparadise here, and @sdelano in slack)

ryancragun:
This looks good to me with the assumption that the omnibus/omnibus-software changes on your branches are merged upstream and we go back to following the master branch.

srenatus:
@ryancragun yeah I'll take care of cross-testing those things before merging both tomorrow 😃

srenatus:
Re: `[hold]`: This is probably ok, but needs one final verification with the packages from [here](https://packages.chef.io/files/unstable/chef-server/12.15.8+20170627132217/ubuntu/14.04/chef-server-core_12.15.8+20170627132217-1_amd64.deb).

Issue #1307, opened 2017-06-15
Elasticsearch 5 Support Breaks Reconfigure on Redirects


sdelano:
Preliminary Elasticsearch 5 support added in https://github.com/chef/chef-server/pull/1287 breaks reconfigure in OWCA-configured systems. This was discovered when testing the latest packages of Chef and Automate in OWCA. OWCA is configured to the use the Automate-provided Elasticsearch as external search for Chef Server.

It's currently configured as such:

```
opscode_solr4["external_url"] = "http://localhost:8080/elasticsearch"
```

In this case, `http://localhost:8080/elasticsearch` redirects to `http://localhost:8080/elasticsearch/`, causing the reconfigure to fail when parsing JSON because it's actually parsing the HTML of the redirect message.

Yes, configuring the external search endpoint to go to `http://localhost:8080/elasticsearch/` will fix this problem, and I've asked that Amazon do as much, but I believe that our tooling could behave better in the face of common configurations that otherwise results in a completely functional system (e.g. search still works when configured this way, so this helper should work as well).

sdelano:
I feel like there's an http client within Chef that we could be using that will follow redirects instead of using raw `Net::HTTP`.

sdelano:
Looks like `Chef::HTTP` does just this:

```
chef (12.19.36)> http = Chef::HTTP.new("http://localhost:8080/elasticsearch")
 => #<Chef::HTTP:0x00000006707c80 @url="http://localhost:8080/elasticsearch", @default_headers={}, @sign_on_redirect=true, @redirects_followed=0, @redirect_limit=10, @keepalives=false, @options={}, @middlewares=[]> 
chef (12.19.36)> http.get("")
 => "{\n  \"name\" : \"Skids\",\n  \"cluster_name\" : \"chef-insights\",\n  \"cluster_uuid\" : \"b0H5j-W9TGibALC5xqRAiQ\",\n  \"version\" : {\n    \"number\" : \"2.4.1\",\n    \"build_hash\" : \"c67dc32e24162035d18d6fe1e952c4cbcbe79d16\",\n    \"build_timestamp\" : \"2016-09-27T18:57:55Z\",\n    \"build_snapshot\" : false,\n    \"lucene_version\" : \"5.5.2\"\n  },\n  \"tagline\" : \"You Know, for Search\"\n}\n" 
chef (12.19.36)> 
```

ksubrama:
I knew I was missing an abstraction layer somewhere.

[...]

PR #1316, opened 2017-06-21
Adding tests for elasticsearch version request retries

lancewf:

Adding test for ticket #1307

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
@@ -113,7 +113,40 @@ describe OmnibusHelper, :es5 do
       end
     end
 
+    shared_context "elastic search version request fails" do
+      it 'should return 2 but some of the requests will fail' do
+        helper = described_class.new(node)
+        allow(Chef::HTTP).to receive(:new).with(external_url).and_return(client)
+
+        expect(client).to receive(:get).with('').exactly(times).times.and_raise('Bad connection')
+        expect(client).to receive(:get).with('').and_return(response)
+
+        allow(helper).to receive(:sleep).and_return(0)
+        expect(helper.elastic_search_major_version).to eq(2)
+      end
+    end
+
     context 'when elastic search is unavailable' do
+      context 'should return 2 but the request fails first time only' do
+        let(:times) {1}  
+        it_behaves_like "elastic search version request fails"
+      end
+
+      context 'should return 2 but the request fails the first two times' do
+        let(:times) {2}  
+        it_behaves_like "elastic search version request fails"
+      end
+
+      context 'should return 2 but the request fails the first three times' do
+        let(:times) {3}  
+        it_behaves_like "elastic search version request fails"
+      end
+      
+      context 'should return 2 but the request fails the first four times' do
+        let(:times) {4}  
+        it_behaves_like "elastic search version request fails"
+      end
+
       it 'should return raise an exception after 5 retries.' do
         helper = described_class.new(node)
         expect(Chef::HTTP).to receive(:new).with(external_url).exactly(5).times.and_return(client)


omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
```diff
@@ -114,6 +114,64 @@
     end
 
     context 'when elastic search is unavailable' do
+      it 'should return 2 but the request fails first time only' do
```
marcparadise: The tests look good, but I'm thinking they could be a shared example that exists locally this context, and uses `let` or similar  to set number of failures?

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
```diff
@@ -114,6 +114,64 @@
...
+        helper = described_class.new(node)
+        @times_called = 0
```
phiggins: Is `@times_called` used anywhere?

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
```diff
@@ -114,6 +114,64 @@
...
+        allow(Chef::HTTP).to receive(:new).with(external_url).and_return(client)
+
+        expect(client).to receive(:get).with('').exactly(3).times.and_raise('Bad connection')
+        expect(client).to receive(:get).with('').and_return(response)
+
+        allow(helper).to receive(:sleep).and_return(0)
+        expect(helper.elastic_search_major_version).to eq(2)
+      end
```
phiggins: It looks like there are two copies of this test?

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
phiggins: You could also use a loop for these tests that only differ by the count:

```
1.upto(4) do |n|
  it "eventually returns 2 after #{n} failures" do
    # ...
```

marcparadise: That would also work, though lately I have a preference for shared context in the same file vs dynamically created tests, for ease of understandability and to keep the amount of test-related "logic" to a minimum.

phiggins: Shared context has always seemed like wonky rspec voodoo to me, so I think it's just a matter of taste. :)

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
lancewf: No, `@times_called` is not used. I will remove it. Good catch.

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
lancewf: There is two. Removing the extra test.

lancewf:
I switched the code to use shared context and also removed the not used variable.

omnibus/files/private-chef-cookbooks/private-chef/spec/libraries/helper_spec.rb
```diff
@@ -113,7 +113,40 @@
       end
     end
 
+    shared_context "elastic search version request fails" do
```
phiggins: Does `shared_context` let you share specs? I thought it was only for sharing test setup.

phiggins: As long as the tests are actually running this seems good. :)

lancewf: I copied the idea from [here](https://github.com/chef/chef-server/blob/608dbe94d15822a31849952e13549744fc40a702/oc-chef-pedant/spec/api/versioned_behaviors/server_api_v1_spec.rb#L90). 

I did ensure that the test is working as intended by changing the `times` variable to 7 and watching it fail.

PR #1317, opened 2017-06-22
[dev] change update-changelog.sh to not leave data.tar.xz lying around

srenatus:
There's more to cleanup, but well, one piece at a time.


phiggins:
Makes sense to me. 👍

PR #1320, opened 2017-06-23
Update kitchen yaml Centos version(5.11) and include yum-centos cookbook

jaymalasinha:


omnibus/Berksfile
```diff
@@ -6,6 +6,7 @@ cookbook 'apt', '~> 2.0'
 
 cookbook 'omnibus'
 cookbook 'yum-epel', '~> 0.6'
+cookbook 'yum-centos'
```
srenatus: [nit] Do we need a pin of any sort? `~> 2.3` maybe, to not be too strict?

blakestier:
LGTM (also curious about pinning) 👍

PR #1323, opened 2017-06-26
upgrade ruby: 2.4.1

srenatus:
Still got "drop me" commits, but when the omnibus-software bits are merged, this should be ready.


src/oc_erchef/apps/chef_objects/priv/depselector_rb/Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'dep_selector', :git => 'https://github.com/chef/dep-selector', :branch => 'master'
-gem 'erlectricity', :git => 'https://github.com/sdelano/erlectricity', :branch => 'master'
+gem 'dep_selector', git: 'https://github.com/chef/dep-selector', branch: 'master'
+gem 'erlectricity', git: 'https://github.com/chef/erlectricity', branch: 'modernize'
```
phiggins: Should we merge the modernize branch into master before merging this PR? It seems weird to depend on some random git branch...

srenatus: @phiggins You're right, I think. Also, I don't believe anyone would be using that repo, so... if you'd like to, please feel free to adopt the PR 😃

srenatus:
The Net::HTTP in Ruby 2.4 definitely has [an issue with `Host` headers and IPv6](https://bugs.ruby-lang.org/issues/12642). But still I believe we should craft a better monkey patch than the one we've got in master currently. 🤔

srenatus:
How to reproduce the issue:

- in dev vm, have something listen on `::1`: `nc -6 -l -k :: 8080`
- `dvm load oc-chef-pedant`
- `cd /host/oc-chef-pedant; bundle exec irb`
- irb: `require 'pedant'`
- irb: `RestClient::Request.new(method: :get, url: "http://[::1]:8080/foo").execute` or
- irb: `Net::HTTP.get(URI("http://[::1]:8080"))`

those last two calls are both expected to *fail*, but you can watch the output of the `nc` call, and examine that it requests `Host: ::1:8080`

btm:
@srenatus is the `oc-chef-pedant/lib/pedant/core_ext/net_http.rb` file that is removed here the monkey patch you're referring to?

srenatus:
@btm that removed file is the _old monkeypatch_ that was there for Ruby 2.2.x. However, I couldn't find a realworld test case where this specific feature. From looking at the code, it did two things:

1. Fixing up the `Host` header for IPv6 **when using a proxy**
    - according to [a similar patch in chef/chef](https://github.com/chef/chef/blob/c8abc7988880adfd91f024ce02c80e56d569db2a/lib/chef/monkey_patches/net_http.rb#L26-L48) this has been resolved in 2.4.x (i.e., the test call from that snippet succeeds)
2. Quick-and-dirty SSL settings
    - I've threaded SSL settings through properly in this branch (it's the commits mentioning `ssl_version`), so there's no need for that

The new issue with rest-client and net/https [outlined in this commit](https://github.com/chef/chef-server/pull/1323#issuecomment-312307531) is, well, new:
* When accessing an IPv6 host _using the IPv6 address_, the `Host` header will be wrong, e.g. `::1:8080` instead of `[::1]:8080`. This could potentially be ignored by the server, however, it would only be a problem with virtual hosts using IPv6 addresses -- it's quite a stretch, I think.
* I haven't tested the proxy case -- there's a slight chance that while Net::HTTP has the issue resolved (see (1.) above), rest-client hasn't.
* As I said, I'm not sure under which circumstances this would really bite us. I believe we could also document, merge, wait and see.

srenatus:
Closing this as it's outdated, and the work continues in #1438.

PR #1327, opened 2017-06-30
[wip] upgrade Chef

srenatus:
Known TODOs:

* [ ] Upgrade Chef in oc_id
  <details>
    <summary>oc-id uses omniauth-chef, which also depends on `Chef ~> 12`; removing that requirements leads to a rack version incompatibility issue</summary>

  ```
  Bundler could not find compatible versions for gem "rack":
    In Gemfile:
      rails (= 4.2.8) was resolved to 4.2.8, which depends on
        actionpack (= 4.2.8) was resolved to 4.2.8, which depends on
          rack (~> 1.6)

      omniauth-chef was resolved to 0.3.0, which depends on
        chef (~> 13) was resolved to 13.0.113, which depends on
          chef-zero (>= 13.0) was resolved to 13.0.0, which depends on
            rack (~> 2.0)
  ```  
  :warning: this requires to upgrade to Rails 5.x.y; which bumping _maaany_ dependencies, so maybe it's better to
* [ ] Remove Chef dependency from oc-id -- it's only used in two places.


srenatus:
needs a rebase after #1323.

omnibus/files/private-chef-cookbooks/private-chef/resources/chef_run.rb
```diff
@@ -0,0 +1,60 @@
+provides :chef_run
+actions :run
```
tas50: you don't need actions or default_action

omnibus/files/private-chef-cookbooks/private-chef/resources/chef_run.rb
```diff
@@ -0,0 +1,60 @@
...
+default_action :run
+
+property :run_list, :kind_of => String, :name_attribute => true
+property :cookbook_path
+property :included_attrs, :kind_of => Array, :default => []
```
tas50: You don't need the kind_of part. Just the ruby type

omnibus/files/private-chef-cookbooks/private-chef/resources/chef_run.rb
```diff
@@ -0,0 +1,60 @@
+provides :chef_run
+actions :run
+default_action :run
+
+property :run_list, :kind_of => String, :name_attribute => true
```
tas50: This should be name_property not name_attribute

markan:
Closed by #1506

PR #1329, opened 2017-07-10
[cookbooks] Don't manage SHMMAX and SHMALL

stevendanna:
According to the postgresql documentation [1] a resonable value for
SHMMAX is now 1KB and that

> PostgreSQL requires a few bytes of System V shared memory (typically
> 48 bytes, on 64-bit platforms) for each copy of the server. On most
> modern operating systems, this amount can easily be
> allocated. However, if you are running many copies of the server, or
> if other applications are also using System V shared memory, it may
> be necessary to increase SHMMAX, the maximum size in bytes of a
> shared memory segment, or SHMALL, the total amount of System V
> shared memory system-wide. Note that SHMALL is measured in pages
> rather than bytes on many systems.

The previous values were required for Postgresql 9.2 which used
substantially more shared memory.

Removing the management of these system paramaters removes one of the
barriers to running the chef-server package directly in a container.

I've avoided doing any cleanup since this changed a core system
configuration which might also be needed by other software on the
system.

[1] https://www.postgresql.org/docs/9.6/static/kernel-resources.html

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -133,6 +133,18 @@ module PrivateChef
       end
     end
 
+    def deprecated_postgresql_settings
+      [ 'shmmax', 'shmall'].each do |setting|
+        if PrivateChef['postgresql'][setting]
+          ChefServer::Warnings.warn <<EOF
+The configuration parameter postgresql['#{setting}'] is no longer used
+by Chef Server. To limit the amount of shared memory used by
+postgresql use postgresql['shared_buffers'] instead.
+EOF
+        end
+      end
+    end
+
     # Mutate PrivateChef to account for common cases of user-provided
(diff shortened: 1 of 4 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -133,6 +133,18 @@ def import_legacy_service_config(old_service_key, new_service_key, keys)
       end
     end
 
+    def deprecated_postgresql_settings
```
srenatus: Could have also gone into [the postgres preflight validator](https://github.com/chef/chef-server/blob/366438de00fc294798441f0a5cba440dc52a2b42/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_postgres_validator.rb#L95-L103), I suppose? If you've avoided that for a reason, we might want to migrate the other check here, too? I can see that the exact message is different, but I wonder about the place where these kinds of checks should happen.

PR #1330, opened 2017-07-12
[omnibus] pin and bump for openresty

srenatus:
[see here](https://github.com/chef/omnibus-software/pull/863)


srenatus:
I still think we should get this sorted out. However, the code in this branch probably isn't the right way. So, the "Don't merge" is accurate. I propose we still keep it open for tracking purposes...

PR #1331, opened 2017-07-12
Enable FIPS mode in NGINX (and potentially other things)

jaym:
This patch writes out some openssl config that tells the library
that we want FIPS mode by default. NGINX calls `OPENSSL_config(null)`,
which first checks the environment variable OPENSSL_CONF for a config
file, and then falls back to the system default. In this case, we
have written a system default for chef server with `fips_mode = yes`.

Signed-off-by: Jay Mundrawala <jdmundrawala@gmail.com>


omnibus/files/private-chef-cookbooks/private-chef/templates/default/openssl.cnf.erb
```diff
@@ -0,0 +1,361 @@
+#
+# OpenSSL example configuration file.
```
srenatus: [nit] I'd be a little disappointed finding that comment in the rendered template :/ But that's just me

omnibus/files/private-chef-cookbooks/private-chef/templates/default/openssl.cnf.erb
```diff
@@ -0,0 +1,361 @@
...
+# This definition stops the following lines choking if HOME isn't
+# defined.
+HOME			= .
+RANDFILE		= $ENV::HOME/.rnd
+
+# Extra OBJECT IDENTIFIER info:
+#oid_file		= $ENV::HOME/.oid
+oid_section		= new_oids
+
+<% if node['private_chef']['fips_enabled'] -%>
```
stevendanna: I don't think this will work like we want in the case of turning fips mode on and back off.  Since once you turn fips_enabled to false, this template file won't get updated:

https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb#L40-L42

I think we may want a fips_disabled recipe that deletes this entire file when fips isn't on.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/openssl.cnf.erb
jaym: You are correct, the rest is the default file openssl puts there. Let me see if there is a way to include other files. Might be able to make it better

omnibus/files/private-chef-cookbooks/private-chef/templates/default/openssl.cnf.erb
jaym: Rookie mistake. Let me fix that up.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/openssl.cnf.erb
jaym: I dont see a way to do any sort of file nesting, however we can cull down this config file to just what we need. Most of these are either already defaults by the library or not useful to us.

srenatus: @jaym if it's not too much work, I'd think that would be nicer... but that's just an opinion, really...

jaym: Seems like a good option to me, we just need to take the fips stuff

omnibus/files/private-chef-cookbooks/private-chef/templates/default/sv-nginx-run.erb
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 exec 2>&1
-exec /opt/opscode/embedded/bin/veil-env-helper -o DATA_COLLECTOR_TOKEN=data_collector.token -s REDIS_PASSWORD=redis_lb.password -- chpst -P /opt/opscode/embedded/sbin/nginx -c <%= File.join(node['private_chef']['nginx']['dir'], "etc", "nginx.conf") %>
```
srenatus: [nit] I know it's been like that before, but I think this file could be helped by dropping some `\ \n` in there to make it less unreadable... 💭

jaym:
@stevendanna the issue you mentioned should now be resolved

jaym:
the more i think about it, i probably shouldn't be modifying `/opt/opscode`. Thoughts on putting the file in `/etc/opscode` or something else?

PR #1334, opened 2017-07-19
[ctl-commands] Restart haproxy on upgrade in chef-backend mode

stevendanna:
When configured to use Chef Backend, all postgresql requests are
routed through HAProxy. Thus, in order to ensure that we can access
postgresql during upgrades, we need to start haproxy rather than
postgresql.

NB: This assumes that the existing 15 second sleep is enough time to
allow haproxy to stabilize (i.e. mark non-leader nodes as DOWN). In
the reconfigure cookbook we do an active check that looks for the
steady-state rather than a sleep.


stevendanna:
@chef/chef-server-maintainers OK, this is ready for review.  I've decided if we want to put the more-robust haproxy status checking code in here as well, we should just finally do it the "possibly more correct" way and create a shared library that we can access from all of these minor ruby projects: https://github.com/chef/chef-server/pull/1336

PR #1336, opened 2017-07-20
Add LibCS library

stevendanna:
Short-term goals:

- Provide a place for shared code between chef-server-ctl, the
  reconfigure cookbooks and partybus.

- Centrally manage the ruby dependencies for those projects.

Longer-term goal:

- Move chef-server-ctl, partybus, and the reconfigure cookbooks into
  this new gem and manage them like normal ruby projects.


stevendanna:
Still a work in progress, but feel free to offer opinions :D

srenatus:
:+1: I'd like to have that 🎉

stevendanna:
I think that moving a lot around in the build would be a bad idea at this time until the habitat chef server stuff takes final form.  I will revisit this once that work has settled and we can find something that hopefully facilitates both setups.

PR #1339, opened 2017-07-24
[ctl-commands] Orphaned auth_actor cleanup tool

stevendanna:
Older versions of Chef Server produced orphaned auth_actor objects
easily. Some installations may have thousands of defunct entries
related to now-deleted (or never-created) clients. These defunct
entries can hurt performance and (in the worst case) make it
impossible to lis the membership of the group.

This command is based on a more straightfroward script created by
Josh Hudson:

https://gist.github.com/double-z/a062c9133719c2ac385a

It constructs a list of known actors (in the opscode_chef database)
and deletes any auth_actors in the bifrost database not contained in
the known actors list.

One issue with that script is that it requires the server to be
offline to run since otherwise newly-created clients might get
deleted.

This script is intended to run on-line. It uses a temporary new table
in bifrost to track newly created auth_actors. Any newly created
auth_actors are ignored when constructing batches of authz_id's to
ignore.

To avoid deleting clients whose creation spans the creation of the
tracking table and known actors list we use a one-time sleep.

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
+++ b/omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
@@ -206,30 +206,26 @@ def timed(description)
   res
 end
 
-# TODO(ssd) 2017-07-24: I know these functions need to use veil to get
-# the passwords. Right now has been written to potentialy run against
-# Chef Server 12.11.0
+def connection_params
+  running_config = JSON.parse(File.read("/etc/opscode/chef-server-running.json"))
+  pg_config = running_config['private_chef']['postgresql']
+
+  password = if respond_to?(:credentials) # veil helper method doesn't exist on older chef-servers
+               credentials.get('postgresql', 'db_superuser_password')
+             else
+               pg_config['db_superuser_password']
+             end
+
+  { 'host' => pg_config['vip'],
+    'port' => pg_config['port'],
+    'user' => pg_config['db_superuser'],
+    'password' => password }
+end
+
 def erchef_db
-  @erchef_db ||= begin
-                   running_config = JSON.parse(File.read("/etc/opscode/chef-server-running.json"))
-                   erchef_config = running_config['private_chef']['opscode-erchef']
-                   pg_config = running_config['private_chef']['postgresql']
-                   ::PGconn.open('user' => erchef_config['sql_user'],
-                                 'host' => pg_config['vip'],
-                                 'password' => erchef_config['sql_password'],
-                                 'port' => pg_config['port'],
-                                 'dbname' => 'opscode_chef')
-                 end
+  @erchef_db ||= ::PGconn.open(connection_params.merge('dbname' => 'opscode_chef'))
 end
 
 def bifrost_db
-  @bifrost_db ||= begin
-                    running_config = JSON.parse(File.read("/etc/opscode/chef-server-running.json"))
-                    pg_config = running_config['private_chef']['postgresql']
-                    ::PGconn.open('user' => pg_config['db_superuser'],
-                                  'host' => pg_config['vip'],
-                                  'password' => pg_config['db_superuser_password'],
-                                  'port' => pg_config['port'],
-                                  'dbname' => 'bifrost')
-                  end
+  @bifrost_db ||= ::PGconn.open(connection_params.merge('dbname' => 'bifrost'))
 end
--- a/omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
+++ b/omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
@@ -10,11 +10,32 @@ CREATE OR REPLACE RULE cleanup_auth_actor_creation_tracking AS ON INSERT TO auth
 COMMIT;
 SQL
 
+CREATE_SQL2 =<<SQL
+CREATE TABLE IF NOT EXISTS cleanup_known_auth_actors(
+    authz_id  CHAR(32)
+);
+SQL
+
 DELETE_SQL =<<SQL
 BEGIN;
 DROP RULE IF EXISTS cleanup_auth_actor_creation_tracking ON auth_actor;
 DROP TABLE IF EXISTS cleanup_tracking_auth_actors;
+DROP TABLE IF EXISTS cleanup_known_auth_actors;
 COMMIT;
+SQL
+
+CLEANUP_SQL = <<SQL
+WITH good_auth_actors AS (
+         SELECT authz_id
+         FROM cleanup_tracking_auth_actors
+              UNION
+              SELECT authz_id FROM cleanup_known_auth_actors),
+     orphaned_auth_actors AS (
+         SELECT authz_id
+         FROM auth_actor
+         WHERE authz_id NOT IN (SELECT authz_id FROM good_auth_actors)
+         LIMIT $1)
+DELETE FROM auth_actor WHERE authz_id IN (SELECT authz_id FROM orphaned_auth_actors)
 SQL
 
 add_command_under_category "cleanup-bifrost", "cleanup", "Cleanup orphaned bifrost objects.", 2 do
@@ -23,8 +44,12 @@ add_command_under_category "cleanup-bifrost", "cleanup", "Cleanup orphaned bifro
 
   OptionParser.new do |opts|
     opts.banner = "chef-server-ctl cleanup-bifrost [options]"
-    opts.on("-b SIZE", "--batch-size", "How many authz actors to delete at a time") do |b|
+    opts.on("-b SIZE", "--batch-size SIZE", "How many authz actors to delete at a time") do |b|
       options[:batch_size] = b.to_i
+    end
+
+    opts.on("-w SECONDS", "--wait-time SECONDS", "How many seconds to wait before stating scan (default: 50)") do |w|
+      options[:wait_time] = w.to_i
     end
 
     opts.on("--force-cleanup", "Clean up tracking tables (does not scan bifrost actors)") do |b|
@@ -43,22 +68,6 @@ add_command_under_category "cleanup-bifrost", "cleanup", "Cleanup orphaned bifro
     exit(1)
   end
 
-  if options[:cleanup_only]
-    remove_bifrost_tracking_table(bifrost_db)
-    exit(0)
-  end
-
-  if options[:estimate_only]
-    print_and_return_estimate(known_actor_list, bifrost_db)
-    exit(0)
-  end
-
-  run_cleanup(bifrost_db, options[:batch_size])
-end
-
-def run_cleanup(bifrost_db, batch_size)
-  safety_check(bifrost_db)
-  install_bifrost_tracking_table(bifrost_db)
   # NOTE(ssd) 2017-07-24:
   # For the most part, the tracking table manages in-flight requests for us.
   # However, it is still possible that the following sequence happens:
@@ -71,11 +80,38 @@ def run_cleanup(bifrost_db, batch_size)
   # clients, won't be in our tracking table, but will show up in our
   # search for orphaned authz_ids.
   #
-  # Thus, this sleep is a 1-time wait to avoid that scenario.
+  # The wait_time is a safety-timer we use to avoid this race.  The
+  # default of 50 seconds is:
   #
-  puts "One-time sleep to account for in-flight requests not captured by tracking table"
-  sleep 25
+  #   (# of db/upstream calls after auth_actor creation * 5 seconds) * 2
+  #
+  if !options[:wait_time]
+    options[:wait_time] = 50
+  elsif options[:wait_time] <= 0
+    puts "Invalid wait time: #{options[:wait_time]}"
+    exit(1)
+  end
+
+  if options[:cleanup_only]
+    remove_bifrost_tracking_table(bifrost_db)
+    exit(0)
+  end
+
+  if options[:estimate_only]
+    print_and_return_estimate(known_actor_list, bifrost_db)
+    exit(0)
+  end
+
+  run_cleanup(bifrost_db, options[:batch_size], options[:wait_time])
+end
+
+def run_cleanup(bifrost_db, batch_size, wait_time)
+  safety_check(bifrost_db)
+  install_bifrost_tracking_table(bifrost_db)
+
   begin
+    puts "Sleeping #{wait_time} seconds to account for in-flight requests not captured by tracking table"
+    sleep wait_time
     estimated_deletion_count = print_and_return_estimate(known_actor_list, bifrost_db)
 
     if estimated_deletion_count <= 0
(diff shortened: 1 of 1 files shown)


stevendanna:
@markan I was able to reduce the number of times we have to use an ad-hoc safety timer to one.  Given that, I think we could even bump it to the full nginx timeout (300 seconds) if we feel that is necessary.

The code is a bit of a mess but I wanted to get this out there to get your thoughts on the general strategy.

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,213 @@
...
+  #
+  # Thus, this sleep is a 1-time wait to avoid that scenario.
+  #
+  puts "One-time sleep to account for in-flight requests not captured by tracking table"
+  sleep 25
+  begin
+    estimated_deletion_count = print_and_return_estimate(known_actor_list, bifrost_db)
+
+    if estimated_deletion_count <= 0
+      puts "Estimated deletion count 0. Aborting"
```
marcparadise: Maybe something like `No orphaned authz records found.`?

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,213 @@
...
+  res = yield
+  printf ": #{Time.now - st}\n"
+  res
+end
+
+def to_pg_list(array)
+  "'" + array.join("','") + "'"
+end
+
+# TODO(ssd) 2017-07-24: I know these functions need to use veil to get
```
marcparadise: I thnk I remember that Veil could also read the old form of the secrets file?

marcparadise:
I am wondering if we could alter the actors table to include a default CURRENT_TIMESTAMP creation time, and use creation_time < current time instead of a tracking table? 
Something else that would simplify it is PG foreign data wrappers - but they weren't introduced until 9.3, unfortunately. .It'd let you just do a join across both DBs (or a right join to find missing entries).

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,213 @@
+require 'pg'
+
+CREATE_SQL =<<SQL
+BEGIN;
+CREATE TABLE IF NOT EXISTS cleanup_tracking_auth_actors(
```
marcparadise: Should we have `TEMPORARY` or possible `TEMPORARY UNLOGGED`(if we don't want the table to replicate)?

stevendanna: Interesting!  I definitely think that we should used `UNLOGGED`.  I'll need to look into how the PG gem manages the connection to know if TEMPORARY will actually work for us.

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
stevendanna: Yes, but either way I'll need a fallback for when the veil gem isn't installed so the user in question can install this by just dropping it into the correct directory.

stevendanna:
> I am wondering if we could alter the actors table to include a default CURRENT_TIMESTAMP creation time, and use creation_time < current time instead of a tracking table?

In the future, yes, but a goal of this is to also be able to drop into old installations which makes schema changes a no-go.  I am fine if we don't end up committing this.

> Something else that would simplify it is PG foreign data wrappers - but they weren't introduced until 9.3, unfortunately. .It'd let you just do a join across both DBs (or a right join to find missing entries).

Yup, if we could depend on 9.3 (which we can in the most recent version) we could you a fdw.  

Given this, perhaps I should move this to the `scripts/` directory for when we need it for older installs, and we can make one for *new* installs that could use one of these two ideas.

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,213 @@
...
+  # search for orphaned authz_ids.
+  #
+  # Thus, this sleep is a 1-time wait to avoid that scenario.
+  #
+  puts "One-time sleep to account for in-flight requests not captured by tracking table"
+  sleep 25
+  begin
+    estimated_deletion_count = print_and_return_estimate(known_actor_list, bifrost_db)
+
+    if estimated_deletion_count <= 0
```
srenatus: [nit] we've got `[0, x - y].max` below, so this will never be negative

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
srenatus: I think this line will follow [that output](https://github.com/chef/chef-server/pull/1339/files#diff-3b35eccc49cc7c77455e83f80f22ff8eR129):
```
 Total chef users+clients: 100
Total bifrost auth_actors: 100
Deletion Candidates (est): 0
```
So it should be understandable from context, I think?

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,213 @@
...
+def run_bifrost_scan(known_actor_list, batch_size, db)
+  total_deleted = 0
+  known_actors_str = to_pg_list(known_actor_list)
+  loop do
+    puts "Processing batch of #{batch_size}"
+
+    candidates = timed "    Retrieving batch from auth_actors" do
+      db.exec("SELECT authz_id FROM auth_actor WHERE authz_id NOT IN (#{known_actors_str}) LIMIT $1",
+              [batch_size])
+        .map {|r| r['authz_id'] }
```
srenatus: huh, I'd have expected a rubocop 💥 on the missing space here `{|r|` 🤔 (Not like I care too much, though)

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,213 @@
...
+end
+
+def run_bifrost_scan(known_actor_list, batch_size, db)
+  total_deleted = 0
+  known_actors_str = to_pg_list(known_actor_list)
+  loop do
+    puts "Processing batch of #{batch_size}"
+
+    candidates = timed "    Retrieving batch from auth_actors" do
+      db.exec("SELECT authz_id FROM auth_actor WHERE authz_id NOT IN (#{known_actors_str}) LIMIT $1",
```
srenatus: Just curious -- there's no way to template list args in SQL, I suppose? (We're using `$1` for `batch_size`, but string interpolation for `known_actors_str`, here, and `to_pg_list(candidates)` below)

(I'm not worried about SQLi there, really, just curious)

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
srenatus: > so the user in question can install this by just dropping it into the correct directory.

Oh! I see... 💡 (I wonder if we need a "properer" mechanism for this... like, _"dump your scripts here"_, have some metadata in them so we've got some basic version checks... *this* is a little wild west...)

stevendanna:
> For my understanding -- what drove you to do batches here? Could our customers' orphaned auth_objects exceed available memory? (...wouldn't PG handle that?)

The original code used batches so they are in part an artifact of iterative development. However, the place I'm looking at using this has 2.5 million entries that are likely to be cleaned up.  To get rid of the batches I think it would be good to come up with a single SQL query to run the correct delete rather than the current method of getting candidates, filtering them, and then deleting them.

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
stevendanna: @srenatus It should be possible but I gave up fighting with the old version of the pg gem we currently ship.

srenatus:
@stevendanna Ah, I see. Yeah, makes sense, I hadn't realised it's actually _us_ that get all the stuff into memory, and there's no PG builtin magic that could save us. Thanks!

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
stevendanna: @markan Great idea.  Creating a known table will also allow us to remove the inner loop to a single SQL query per iteration.

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,213 @@
...
+  #    - new record in oc-chef database
+  #
+  # In the this case, the authz_id won't be in our list of known
+  # clients, won't be in our tracking table, but will show up in our
+  # search for orphaned authz_ids.
+  #
+  # Thus, this sleep is a 1-time wait to avoid that scenario.
+  #
+  puts "One-time sleep to account for in-flight requests not captured by tracking table"
+  sleep 25
```
stevendanna: >  I'm assuming 25 secs came from something like (# of gen server calls between authz create and db write ) * gen_server timeout of 5 secs? 

Yes, more or less exactly that.  3 for the authz acl merging, 1 for possible key gen, and 1 for the final db write.  I'll document that and since this is a 1-time sleep now, I'll double it so we have some safety margin.

stevendanna:
@markan @marcparadise @srenatus Thanks for all the helpful comments.  I've made the following changes based on them:

1. We now insert the known actors list into a table rather than sending it over each time.

2. I've used UNLOGGED tables for the two tables we create.  It appears UNLOGGED and TEMPORARY are mutually exclusive.  I've gone with UNLOGGED for now so that if we need to debug a hard crash we can inspect the table in some cases.

3. Increased the safety timer to 50 seconds.

The code is still a bit messy since I'm trying to keep it "single file" for the moment to make it easy to use on a customer site.

TESTING

I've run this on a local chef server with 30,000 known clients and 1,000,000 orphaned objects.  It deleted entries at a rate of around 1 second per batch while 5 threads of concurrent client creation was generating load on the server.

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,213 @@
...
+        .map {|r| r['authz_id'] }
+    end
+
+    if candidates.length == 0
+      puts "No more candidates for deletion."
+      break
+    end
+
+    confirmed_candidates = timed "    Ignoring any recently created actors from candidates (n = #{candidates.length})" do
+      sql = "SELECT authz_id FROM cleanup_tracking_auth_actors WHERE authz_id IN (#{to_pg_list(candidates)})"
```
stevendanna: I've implemented more-or-less this in my recent update.  I'm still not thrilled about the string-interpolation on the INSERT, but I can live with it for now.

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,231 @@
...
+CREATE UNLOGGED TABLE IF NOT EXISTS cleanup_tracking_auth_actors(
+    authz_id  CHAR(32)
+);
+
+CREATE OR REPLACE RULE cleanup_auth_actor_creation_tracking AS ON INSERT TO auth_actor DO INSERT INTO cleanup_tracking_auth_actors VALUES (NEW.authz_id);
+COMMIT;
+SQL
+
+CREATE_SQL2 =<<SQL
+CREATE UNLOGGED TABLE IF NOT EXISTS cleanup_known_auth_actors(
```
stevendanna: It may be cleaner to re-use the cleanup_tracking_auth_actors table for tracking known actors as well.  That will also save a UNION in the delete query.

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,231 @@
...
+  @known_actor_list ||= timed "Fetching initial opscode_chef clients and user list" do
+    clients_and_users = erchef_db.exec("SELECT authz_id FROM clients UNION select authz_id FROM users")
+    clients_and_users.map do |real_actor|
+      real_actor['authz_id']
+    end
+  end
+end
+
+def safety_check(db)
+  res = db.exec("SELECT * FROM pg_tables
```
srenatus: Would it make sense to make _all SQL queries_ constants? There's not much to gain, though... It's surely fine to have the 1- and 2-liner queries inline, it's just... homogeneity?

srenatus: Nevermind, it's a dumb idea. 😉

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,231 @@
...
+  timed "Installing tracking tables into bifrost database" do
+    db.exec(CREATE_SQL)
+  end
+end
+
+def install_known_actor_table(db, list)
+  list_str = "('" + list.join("'), ('") + "')"
+  timed "Populating known actor table" do
+    db.exec(CREATE_SQL2)
+    db.exec("INSERT INTO cleanup_known_auth_actors VALUES #{list_str}")
```
srenatus: Is it possible that this becomes _too long_?

stevendanna: Yes. I can see two situations (1) the list is so large that holding it in memory and doing that join is too expensive or (2) the list gets so long that the INSERT query grinds to a halt.  

For some reason I was under the impression that our version of the PG gem didn't support COPY, but it appears that it does just fine from what I can see in git, so one option would be to use `COPY`. Another would be to chuck the list and do multiple INSERTS.

stevendanna:
> Also, it might be overengineering, but would there be a value in a "dry run" mode? As in, "show me the actors you would delete"?

It could be worth it.  I added the `--estimate-only` flag to be able to spot check whether the count of defunct auth_actors was in the right range; but this doesn't use the query that will actually be used for the delete.

stevendanna:
@srenatus I've moved to using `COPY` for the import of the known auth actors table. I haven't done performance testing on that change but the PostgreSQL documentation recommends COPY for importing and it gets rid of the last bit of string interpolation for queries in this script.

@chef/chef-server-maintainers I'd like to figure out what we think is required to make us comfortable merging this. This has worked at the customer-site it was intended for so there isn't any time pressure other than wanting to move it off the board if possible.

Here are a few changes/issues we've mentioned and my current thinking on them:

- We currently, inefficiently, pull the known actor list into memory and then ship it over the the bifrost database.

  An authz_id is a 32 character string. If we have 1M known actors and if we assume that via copying and other overhead we end up with 2.5 copies of the list, a "back-of-the-irb" memory  usage estimate is about 175MB. That is definitely bad if this were a long-running process but doesn't seem unreasonable for a rarely-used cleanup tool:

```
irb(main):001:0> require 'securerandom'
=> true
irb(main):002:0> require 'objspace'
=> true
irb(main):003:0> str = SecureRandom.hex(16)
=> "7d6e65a74732ac92e89e114e38590f46"
irb(main):004:0> str.length
=> 32
irb(main):005:0> ObjectSpace.memsize_of(str)
=> 73
irb(main):006:0> (((73 * 1_000_000)/1024.0)/1024.0)*2.5
=> 174.04556274414062
```
We could use PostgreSQL's foreign data wrappers to avoid copying the data into Ruby if this became an issue.

- Two ideas have been presented to avoid the tracking rules: (1) foreign data wrappers and (2) adding a created_at timestamp to the auth_actors table.

  I don't think FDWs will solve the problem that the tracking table solves.  The client creation is still split across multiple queries in erchef so even a cross database transaction in the cleanup code would have to account for half-created clients.

  A timestamp field would solve the problem and would drastically simplify this tool. A TIMESTAMP with or without timezone type in PostgreSQL is 8 bytes, which would be a 19.5% increase in the amount of data we are storing per row (bigint is 8 bytes, chararacter(32) is 32 + 1 bytes if I've read the PostgreSQL docs correctly) and introduce a simple schema migration [0].

  These costs might be worth it considering how simple the cleanup operation would become (i.e. we would avoid races by only considering bifrost actors created more than NGINX_TIMEOUT in the past). However, I'm fairly hesitant to make changes to a service that is otherwise performing well to facilitate a seldom-used cleanup tool.  I am happy to be convinced otherwise.

- We could experiment with better alternatives to the `LIMIT` clause in the delete query, but for now I like getting progress along the way even if it comes at the expense of longer overall runtims.

I see a couple of options:

1. Decide if this PR (or this PR with minor changes) is suitable to merge as a user-accessible command and merge it.

2. Decide that this PR isn't suitable to merge and create a new user-accessible command with a bit more careful engineering put into it.

3. Decide that we don't need this as a user-accessible command and merge it in the `dev/scripts` folder.

[0] https://www.postgresql.org/docs/9.6/static/datatype.html

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -150,11 +152,14 @@ def install_bifrost_tracking_table(db)
   end
 end
 
-def install_known_actor_table(db, list)
-  list_str = "('" + list.join("'), ('") + "')"
```
srenatus: oh I'm so glad to see this line go away 😉

omnibus/files/private-chef-ctl-commands/cleanup_bifrost.rb
```diff
@@ -0,0 +1,235 @@
...
+DROP TABLE IF EXISTS cleanup_tracking_auth_actors;
+DROP TABLE IF EXISTS cleanup_known_auth_actors;
+COMMIT;
+SQL
+
+CLEANUP_SQL = <<SQL
+WITH good_auth_actors AS (
+         SELECT authz_id
+         FROM cleanup_tracking_auth_actors
+              UNION
```
markan: Minor indent glitch here? Nice use of WITH here; it really helps readability for these kinds of multiclause queries.

markan:
+1 LGTM

PR #1346, opened 2017-07-31
[WIP] [CON-692] Add support for updating ES index definition.

marcparadise:
* add update supprot so that indices that have changed can be updated
* add ES version checking
* use ES version to determine field type mappings.

TODO: 
* [x] validate change detection and versoin behavior under ES 2.3
* [ ] validate update logic against ES5 - has been done by hand/manual PUTs so far: 
  * bring up chef-backend with ES 2.x
  * upgrade to 5.x package 
  * reconfigure chef server 
  * verify that the index changes are applied by querying the ES API and inspecting the new definition (all string types have been replaced with keyword or text. )

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/elasticsearch_index_provider.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/elasticsearch_index_provider.rb
@@ -5,32 +5,69 @@ class Chef
     class ElasticSearchIndex < Chef::Provider::LWRPBase
       use_inline_resources if defined?(:use_inlined_resources)
 
-      action :create do
-        if ! index_exists?
-          converge_by "Creating elasticsearch index #{new_resource.index_name}" do
-            solr_server.put(new_resource.index_name, Chef::JSONCompat.to_json(new_resource.index_definition))
+      action :create_or_update do
+        if index_exists?
+          if index_changed?
+            converge_block_for_update
           end
+        else
+          converge_block_for_create
         end
       end
 
       action :destroy do
-        if index_exists?
-          converge_by "Deleting elasticsearch index #{new_resource.index_name}" do
-            solr_server.delete(new_resource.index_name)
+        converge_block_for_destroy if index_exists?
+      end
+
+      def existing_index_mappings
+        unless @orig_mappings
+          index = JSON.parse(solr_server.get("#{new_resource.index_name}"))
+          @orig_mappings = index[new_resource.index_name]['mappings']
+            solr_server.put("#{url}?update_all_types",
+                            Chef::JSONCompat.to_json(mapping))
           end
+        end
+
+      end
+
+      def converge_block_for_create
+        converge_by "Creating elasticsearch index #{new_resource.index_name}" do
+          solr_server.put(new_resource.index_name, Chef::JSONCompat.to_json(new_resource.index_definition))
         end
       end
 
-      def index_exists?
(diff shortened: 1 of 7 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/elasticsearch_index_provider.rb
```diff
@@ -5,33 +5,70 @@ class Provider
     class ElasticSearchIndex < Chef::Provider::LWRPBase
       use_inline_resources if defined?(:use_inlined_resources)
```
stevendanna: While we are modifying this substantially, it may be a good time to move this to a new-style chef resource.

omnibus/files/private-chef-cookbooks/private-chef/libraries/elasticsearch_index_provider.rb
```diff
@@ -5,33 +5,70 @@ class Provider
...
+        converge_block_for_destroy if index_exists?
+      end
+
+      def existing_index_mappings
+        unless @orig_mappings
+          index = JSON.parse(solr_server.get("#{new_resource.index_name}"))
+          @orig_mappings = index[new_resource.index_name]['mappings']
+            solr_server.put("#{url}?update_all_types",
+                            Chef::JSONCompat.to_json(mapping))
           end
```
stevendanna: Something seems off with the indentation here, do we have an extra `end`?

marcparadise:
I'e pushed up my latest changes for reference but I am closing this PR. 

ES5 is backwards compatible with ES2 for our index. We will need to address the string type changes (split to text/keyword) , but the number of additional checks required for safely detecting when we need to change the index based on the resource declaration adds more complexity than we can meaningfully test . 

A better approach may be to make a specific command to apply ES5 index updates and a reindex  when the customer upgrades their external ES (backend or other) to ES5, possibly via a separate chef client run.  This would alllow us to not handle the special upgrade case inside the standard reconfigure behavior.

stevendanna:
@marcparadise OK, thanks for the update.  Does that mean for the purposes of the next release we aren't going to consider this blocking? Or do we still want to address the required changes for ES5 on new installs before the next release.

marcparadise:
Interestingly, the mappings using the old string definitions  - when applied to ES 5 on a new CS instance and examining the index we get back on query - they resolved exactly the way our attempted remapping did. (All non-analyzed fields -> keyword; the data field is text)

What this means is that: 

1) anyone installing on ES2 is fine
2) anyone installing on ES5 is fine
3) anyone upgrading backend to the ES5 version will be fine in the medium  term; but we will need to provide a path to safely drop/recreate/reindex before we move chef-backend to ES6 (and before we can support customers using ES6) 

In the PR I also tried to go the route of drop/recreate - but ran into a number of issues - the first of which was reliably detecting when the index def had changed.  Since we don't reindex in the normal case, an index recreate in a normal reconfigure is a thing to avoid. 

Here 
Note:       Mapper for [X_CHEF_id_CHEF_X] conflicts with existing mapping in
      other types:
        mapper [X_CHEF_id_CHEF_X] has different [index] values
        mapper [X_CHEF_id_CHEF_X] has different [doc_values] values
        mapper [X_CHEF_id_CHEF_X] has different [analyzer]"
}

Once I realized this was going pear-shaped I split off work into a new branch - final version (including some false attempts and workarounds I wouldn't want to actually ship) is in the branch ' mp/ES5-index-updating'.

PR #1347, opened 2017-08-01
[dev/scripts] Add start_depselector.sh

stevendanna:
This comes up every quarter or so and we always have to remember the
bit about file descriptors 3 and 4.

### DIFF
--- a/dev/scripts/start_depselector.sh
+++ b/dev/scripts/start_depselector.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+#
+# Start depselector in the forground in order to see any error
+# messages it is producing.  This is useful in some debugging
+# situations.
+#
+# The erlang runtime would open these FDs for us. If we don't open
+# them they get reserved by the ruby VM and then application fails to
+# start.
+ERLOUT=$(mktemp)
+echo "Depselector erlang port responses sent to $ERLOUT"
+exec 3<&0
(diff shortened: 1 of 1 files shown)


stevendanna:
You can send the initial get_pid message to it with:

```
printf "\x00\x00\x00\x0D\x83\x68\x01\x64\x00\x07\x67\x65\x74\x5F\x70\x69\x64" | ./start_depselector.sh
```

dev/scripts/start_depselector.sh
```diff
@@ -0,0 +1,15 @@
...
+#
+# Start depselector in the forground in order to see any error
+# messages it is producing.  This is useful in some debugging
+# situations.
+#
+# The erlang runtime would open these FDs for us. If we don't open
+# them they get reserved by the ruby VM and then application fails to
+# start.
+exec 3<&0
+exec 4>erlout
```
srenatus: [nit] Would this overwrite an existing `erlout` file? I'm currently not sure, since I've never seen `2>>output` or something similar...

stevendanna: `n>>word` does indeed work in Bash.  Happy to use that, however I'm not sure if appending random bytes is really better than overwriting. We could use a temporary file I suppose.

stevendanna:
@srenatus Now we use mktemp to create an output file.  Seems "good enough" for a debug script.

PR #1351, opened 2017-08-04
Only run smoke tests on s390x

stevendanna:
The s390x tests have become unstable.  This appears to be caused by a performance regression; however, we have not yet identified whether the problem is with the s390x testers themselves or an actual performance regression in the product.  To ensure that the pipeline can be used for other changes until this is solved, this PR modifies our CI test phase to only run the smoke tests on s390x.

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -evx
 
+echo "Sleeping even longer (120 seconds) to let the system settle"
+sleep 120
+
 sudo chef-server-ctl test -J $WORKSPACE/pedant.xml --all --compliance-proxy-tests
 
 if [ "$OMNIBUS_FIPS_MODE" == "true" ]
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -4,7 +4,14 @@ set -evx
 echo "Sleeping even longer (120 seconds) to let the system settle"
 sleep 120
 
-sudo chef-server-ctl test -J $WORKSPACE/pedant.xml --all --compliance-proxy-tests
+if [ "s390x" = "$(uname -m)" ]; then
+    # FIX ME FIX ME FIX ME
+    # This is to see if we can get the build passing at all on the s390x platform
+    # FIX ME FIX ME FIX ME
+    sudo chef-server-ctl test -J $WORKSPACE/pedant.xml --smoke --compliance-proxy-tests
+else
+    sudo chef-server-ctl test -J $WORKSPACE/pedant.xml --all --compliance-proxy-tests
+fi
 
 if [ "$OMNIBUS_FIPS_MODE" == "true" ]
 then


oc-chef-pedant/lib/pedant/utility.rb
```diff
@@ -81,5 +81,42 @@ def self.get_pedant_file(filename, &blk)
       end
       return nil # Cannot find the fixture. Raise an error?
     end
+
+    # queues are not readable (e.g. there's no rabbitmq) OR empty
+    def self.queues_empty?(opts = {quiet: true})
+      ENV['PATH'] = "/opt/opscode/embedded/bin:#{ENV['PATH']}"
+      output = `/opt/opscode/embedded/service/rabbitmq/sbin/rabbitmqctl list_queues -p /chef | awk '{sum += $2} END {print sum}'`
```
jaym: if only we had a metrics endpoint

oc-chef-pedant/lib/pedant/utility.rb
```diff
@@ -81,5 +81,42 @@ def self.get_pedant_file(filename, &blk)
...
+      if !status.success?
+        true
+      else
+        output.to_i == 0
+      end
+    end
+
+    def self.wait_until_queues_are_empty(i = 10, opts = {safety_sleep: true,
+                                                         quiet: true})
+      return if i == 0
```
jaym: `i <= 0` is probably the safer

oc-chef-pedant/lib/pedant/utility.rb
```diff
@@ -81,5 +81,42 @@ def self.get_pedant_file(filename, &blk)
...
+      ENV['PATH'] = "/opt/opscode/embedded/bin:#{ENV['PATH']}"
+      output = `/opt/opscode/embedded/service/rabbitmq/sbin/rabbitmqctl list_queues -p /chef | awk '{sum += $2} END {print sum}'`
+      status = $?
+      unless opts[:quiet]
+        STDERR.puts "Command exitstatus: #{status.exitstatus}"
+        STDERR.puts "Command output: #{output}"
+      end
+
+      if !status.success?
+        true
```
jaym: might be worth logging here

oc-chef-pedant/lib/pedant/rspec/search_util.rb
```diff
@@ -230,12 +231,20 @@ def setup_multiple_objects(object_type, number_to_create=5)
...
+            if objects.length > 0
+              Pedant::Utility.wait_until_queues_are_empty(i = 10, safety_sleep: false, quiet: true)
+            end
           end
 
           after :each do
             objects.each do |o|
               send(object_delete_method_symbol, admin_requestor, o['name'])
             end
+            # Technically we should call this here as well, but if the
```
jaym: would the names of the objects overlap in anyway? If so, is the queue drained and messages evaluated to completion in order?

stevendanna: Object updates are partitioned into 1 of 1024 queues based on object ID so all updates for a given object are ordered.

marcparadise:
Re the sleep - have we looked at increasing VM capacity? Last I saw, the testers were racking up a whole bunch of 'steal' time.

stevendanna:
> Re the sleep - have we looked at increasing VM capacity? Last I saw, the testers were racking up a whole bunch of 'steal' time.

Yes, the steal time is still the case (gets up into the 40% range when the tests start).  We've contacted the provider and have decided to up the memory assigned to our VMs.  It isn't clear to me why the memory would be the thing to bump given the steal time but perhaps the memory allocation means we won't be competing for some system level locks in the vm system on the hypervisor or something.

A full test run is going now since the change.  We'll see how it goes.

PR #1355, opened 2017-08-07
Fix partybus db schema upgrades when using External Postgres

irvingpop:
- Simple change to se the helper method `db_connection_uri` instead of hard-coding the connection string, this fixes upgrades with external Postgres

For discussion:   The helper method [db_connection_uri](https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb#L122) doesn't appear to be used anywhere else in the code.   Given the secrets changes, we can safely remove the credentials from the connection string, right?


irvingpop:
Apparently this is a non-issue,   all versions of Chef Server that support external Postgres don't depend on that partybus config file being right anyway.

PR #1356, opened 2017-08-08
[pedant] Wait for rabbitmq before every search test

stevendanna:
The search index is eventually consistent but the test code treats it
as if it were read-after-write consistent.

To mitigate this we wait for the indexing queue to be empty before
executing a search test.

This is going to slow these tests down, but we are trying to stabilize
the tests first.


jaym:
👍

PR #1358, opened 2017-08-11
Fixing reindex's use of elasticsearch scroll API

ksubrama:
Signed-off-by: Kartik Null Cating-Subramanian <ksubramanian@chef.io>

### DIFF
--- a/oc-chef-pedant/lib/pedant/rspec/search_util.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/search_util.rb
@@ -940,7 +940,7 @@ module Pedant
         should_find(temporary_data_bag_name, "test_item")
 
         # Now, drop all information from the search index
-        `#{executable} drop #{reindex_args.join(" ")} #{Pedant::Config.reindex_endpoint}`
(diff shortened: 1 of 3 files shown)


ksubrama:
@lancewf @elliott-davis @marcparadise @PrajaktaPurohit

oc-chef-pedant/lib/pedant/rspec/search_util.rb
```diff
@@ -950,7 +950,7 @@ def do_search(type, name, should_find=true)
         should_not_find_any(temporary_data_bag_name)
 
         # Now, send everything to be re-indexed
-        `#{executable} reindex #{reindex_args.join(" ")} #{Pedant::Config.reindex_endpoint}`
+        puts `#{executable} reindex #{reindex_args.join(" ")} #{Pedant::Config.reindex_endpoint}`
```
lancewf: Do we want to leave these puts in?

marcparadise: These tests fail often enough in the normal case that the addional output can only help.

PR #1359, opened 2017-08-11
Fix dvm Vagrantfile to launch external ES correctly

ksubrama:
Signed-off-by: Kartik Null Cating-Subramanian <ksubramanian@chef.io>

### DIFF
--- a/dev/Vagrantfile
+++ b/dev/Vagrantfile
@@ -186,7 +186,10 @@ def define_chef_server(config, attributes)
       elasticsearch = { "opscode_solr4['external']" => "true",
                         "opscode_solr4['external_url']" => '"http://elasticsearch:9200"',
                         "opscode_erchef['search_provider']" => '"elasticsearch"',
-                        "opscode_erchef['search_queue_mode']" => '"batch"'
+                        "opscode_erchef['search_queue_mode']" => '"batch"',
+                        "rabbitmq['enabled']" => "false",
+                        "rabbitmq['management_enabled']" => "false",
+                        "dark_launch['actions']" => "false"
       }
       json = simple_deep_merge(json, { "provisioning" => { "chef-server-config" => elasticsearch } })
     end


ksubrama:
@marcparadise @lancewf @PrajaktaPurohit

jaym:
I think the issue is that the config has changed and the old names don't map to the new names. The cookbook needs to be run unless a new build is used. It would be good to know what version of chef server was used to bootstrap with.

My thoughts on why I think this is because the error @ksubrama posted in the slacks was coming from https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_sup.erl#L77-L78
which I would expect to get if you're using to old cookbooks with the new code because we renamed `rabbitmq_management_service` to `rabbitmq_actions_management_service` in https://github.com/chef/chef-server/commit/e28760167e8fa8a255cff6f1a348957f0313eab6

stevendanna:
@jaym I think I might not have enough coffee at the moment but I'm not quite seeing how the name change in the rendered configuration would matter with respect to the user-facing configuration.

What we definitely want to be sure of is that no user-supplied configuration needs to be changed on upgrade.  Namely, people who currently have this:

```
rabbitmq['enable'] = false
rabbitmq['management_enabled'] = false
rabbitmq['queue_length_monitor_enabled'] = false
opscode_expander['enable'] = false
dark_launch['actions'] = false
```

Should continue to have rabbitmq completely disabled (including the new _status checks). I /think/ we are OK here, but just wanted to double check.

jaym:
If you have those parameters you specified, then yes, rabbitmq will be completely disabled and there weill be no health checks against it. My point was that the reason the dev environment was broken was because those parameters were not there by default, and the old omnibus package does not map to a usable package for bootstrapping the dvm.

And unless I'm missing something, actions and ES can coexist?

stevendanna:
@jaym I see, thanks.  Yes, actions and ES can coexist so I think you are correct that this PR shouldn't be necessary.

stevendanna:
We think we don't need this anymore but we can reopen if we are wrong.

PR #1364, opened 2017-08-11
[JEX-640] Remove dependency on chefops-dcc cookbook

schisamo:


cookbooks/chef-server-deploy/templates/default/chef-server.rb.erb
```diff
@@ -0,0 +1,27 @@
+topology 'standalone'
+api_fqdn '<%= @chef_server_deploy['chef_server_fqdn'] %>'
+notification_email "ops@chef.io"
```
tduffield: [nit] consistent between `"` and `'`

PR #1366, opened 2017-08-12
Use NGINX cert path attributes to configure RabbitMQ

schisamo:
These attributes work resolve correctly for a generated SSL cert OR a
full SSL cert path passed in via config.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
@@ -116,7 +116,6 @@ end
 
 # Configure Services
 [
-  "rabbitmq",
   "postgresql",
   "oc_bifrost",
   "oc_id",
@@ -128,6 +127,7 @@ end
   "opscode-chef-mover",
   "redis_lb",
   "nginx",
+  "rabbitmq",
   "keepalived"
 ].each do |service|
   if node["private_chef"][service]["external"]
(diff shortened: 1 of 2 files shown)


schisamo:
We found this bug when deploying our new internal Chef Server acceptance environment in which we configure the path to an SSL cert/key (vs using the generated one).

jaym:
I think https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb#L79-L80 needs to happen before rabbitmq

jaym:
this does indeed break things. Here's `rabbitmq.config`:
```erlang
root@vagrant:/var/opt/opscode/rabbitmq/etc# cat rabbitmq.config
[
 {kernel, [{inet_dist_use_interface, {127,0,0,1}}]},
 {ssl, [{versions, ['tlsv1.2','tlsv1.1']}]},
 {rabbit, [{tcp_listeners, [5672]},
               {ssl_options,
                 [{certfile, ""},
                  {keyfile,  ""},
                    {versions, ['tlsv1.2','tlsv1.1']}]}
  ]},
 {rabbitmq_management,
     [{listener, [
               {ip, "127.0.0.1"},
               {port,     15672 },
               {ssl, true }
               % The Rabbit Management Plugin will use the global Rabbit SSL Config
              ]}
  ]}
].
```

jaym:
I think everything will work correctly if you swap https://github.com/chef/chef-server/blob/master/inspec/chef-server-smoke/controls/base.rb#L15-L16

schisamo:
@jaym I moved configuration of `rabbitmq` after `nginx`.

jaym:
👍

schisamo:
Successful Ad-hoc build: http://wilson.ci.chef.co/job/chef-server-trigger-ad_hoc/2/downstreambuildview

PR #1370, opened 2017-08-14
Set the correct owner and permissions for SSL certificate and key

jaym:
This is a follow up to https://github.com/chef/chef-server/pull/1366 and
https://github.com/chef/chef-server/pull/1368.

This patch will set the owner of the ssl certificate and key to the
user that is running rabbitmq/nginx. In #1366, we discovered that
rabbitmq was expecting certs to be in the nginx directory, however
that wasn't necessarily the case when the ssl certificate was not
generated by private-chef-cookbooks. In these cases, the rabbitmq
management interface would be completely broken as the certs we
asked it to use were not there. After fixing that, #1368 was required
because the custom certificate that was being set for the accpetance
environment was owned by root and had the permissions 600, which meant
the opscode user could not read it. Rabbitmq does not start as root like
nginx, so rabitmq was unable to read the certificate. While #1368 got
our environment working, I suspect others may run into a similar issue
unless we explicitly set the owner and permissions on the files to what
we expect.

### DIFF
--- a/cookbooks/chef-server-deploy/metadata.rb
+++ b/cookbooks/chef-server-deploy/metadata.rb
@@ -4,7 +4,7 @@ maintainer_email 'engineering@chef.io'
 license 'Apache 2.0'
 description 'Installs/Configures Chef Server in ACC'
 long_description 'Installs/Configures Chef Server in ACC'
-version '0.1.7'
+version '0.1.8'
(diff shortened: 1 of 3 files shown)


jaym:
i think that makes sense, everything else is owned by root

PR #1373, opened 2017-08-15
Update Release Notes for 12.16

stevendanna:


schisamo:
I _think_ you want to apply the `Changelog: Skip Update` label to this PR since you modified things by hand.

stevendanna:
@schisamo I modified the RELEASE_NOTES rather than CHANGELOG.  We'll need to sort out whether we want to keep the RELEASE_NOTES in the long run.

schisamo:
@stevendanna I still don't think you need this type of "housekeeping" PR to be part of the `CHANGELOG`.

PR #1377, opened 2017-08-15
Render supermarket oc-id file contents into NGINX config

schisamo:
We previously just touched the file as the `opscode` user so NGINX could
serve the file up. That is not a long term solution as `chef-server-ctl
reconfigure` explicitly sets the owner to `root`:
https://git.io/v7F4z

We'll just read the contents of the file during our deployment CCR (when
privileges are elevated).


cookbooks/chef-server-deploy/libraries/helpers.rb
```diff
@@ -5,6 +5,12 @@ def liveness_agent_recipe_url
       release = JSON.parse(github_api.get('/repos/chef/automate-liveness-agent/releases/latest'))
       release['assets'].find { |a| a['name'] == 'automate-liveness-recipe.rb' }['browser_download_url']
     end
+
+    def oc_id_applciation_config(app_name)
```
robbkidd: Misspelled ... but misspelled the same way in both places!

PR #1383, opened 2017-08-24
[SUSTAIN-632] Do not update the users table to contain the public key. The keys table should be the only source of truth

PrajaktaPurohit:
TODO:
 - [x] Add sqitch tag
 - [x] Add omnitruck changes
 - [x] If `knife create client` is the only way to create clients, this bug does not affect clients.
 - [x] Update function returns [[<<"update_user">>, <<>>]] that does not get handled in our code
```
==> /var/log/opscode/opscode-erchef/current <==
2017-08-23_22:48:46.28882 [error] {<<"method=PUT; path=/users/test; status=500; ">>,{error,{error,{case_clause,[[{<<"update_user">>,<<>>}]]},[{chef_db,update,3,[{file,"/var/cache/omnibus/src/oc_erchef/_build/default/lib/chef_db/src/chef_db.erl"},{line,259}]},{oc_chef_wm_base,update_from_json,4,[{file,"/var/cache/omnibus/src/oc_erchef/_build/default/lib/oc_chef_wm/src/oc_chef_wm_base.erl"},{line,1121}]},{webmachine_resource,resource_call,3,[{file,"/var/cache/omnibus/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_resource.erl"},{line,203}]},{webmachine_resource,do,3,[{file,"/var/cache/omnibus/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_resource.erl"},{line,148}]},{webmachine_decision_core,resource_call,1,[{file,"/var/cache/omnibus/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},{line,47}]},{webmachine_decision_core,accept_helper,1,[{file,"/var/cache/omnibus/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},{line,613}]},{webmachine_decision_core,decision,1,[{file,"/var/cache/omnibus/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},{line,518}]},{webmachine_decision_core,handle_request,2,[{file,"/var/cache/omnibus/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},{line,32}]}]}}}
```
- [x] Add user function returns [[<<"add_user">>,<<>>]] that does not get handled in sqerl
```
(oc_erchef@api)1> [error] {<<"method=POST; path=/users; status=500; ">>,{error,{error,{case_clause,{ok,[[{<<"add_user">>,<<>>}]]}},[{sqerl,statement,4,[{file,"/host/src/oc_erchef/_build/default/lib/sqerl/src/sqerl.erl"},{line,111}]},{timer,tc,1,[{file,"timer.erl"},{line,166}]},{stats_hero,ctime,3,[{file,"/host/src/oc_erchef/_build/default/lib/stats_hero/src/stats_hero.erl"},{line,131}]},{chef_db,create,3,[{file,"/host/src/oc_erchef/_build/dev/lib/chef_db/src/chef_db.erl"},{line,168}]},{oc_chef_wm_base,create_from_json,5,[{file,"/host/src/oc_erchef/_build/dev/lib/oc_chef_wm/src/oc_chef_wm_base.erl"},{line,1061}]},{webmachine_resource,resource_call,3,[{file,"/host/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_resource.erl"},{line,203}]},{webmachine_resource,do,3,[{file,"/host/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_resource.erl"},{line,148}]},{webmachine_decision_core,resource_call,1,[{file,"/host/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},{line,47}]}]}}}
```

http://wilson.ci.chef.co/job/chef-server-trigger-ad_hoc/10/downstreambuildview/

### DIFF
--- a/src/oc_erchef/apps/chef_db/priv/pgsql_statements.config
+++ b/src/oc_erchef/apps/chef_db/priv/pgsql_statements.config
@@ -484,17 +484,10 @@
 %% User-related
 %%
 {insert_user,
-  <<"INSERT INTO users (id, authz_id, username, email, hashed_password, salt,
-     hash_type, last_updated_by, created_at, updated_at, external_authentication_uid,
-      recovery_authentication_enabled, serialized_object, admin, pubkey_version) VALUES
-      ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, false, 0)">>}.
+  <<"SELECT add_user($1, $2, $3, $4, 'this_is_not_a_key', 0, $5, $6, $7, $8, $9, $10, $11, $12, $13, false)">>}.
 
 {insert_user_v0,
-  <<"INSERT INTO users (id, authz_id, username, email, public_key, pubkey_version,
-      hashed_password, salt, hash_type, last_updated_by, created_at,
-      updated_at, external_authentication_uid,
-      recovery_authentication_enabled, serialized_object, admin) VALUES
-      ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, false)">>}.
+  <<"SELECT add_user($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, false)">>}.
 
 {find_user_by_username_v0,
   <<"SELECT u.id, authz_id, username, email, k.public_key, k.key_version pubkey_version,
@@ -560,22 +553,7 @@
 {list_users_by_ext_auth_uid, <<"SELECT username FROM users WHERE external_authentication_uid = $1">>}.
 
 {update_user_by_id_v0,
-<<"UPDATE users
-     SET pubkey_version = $1,
-         public_key = $2,
-         hashed_password = $3,
-         salt = $4,
-         hash_type = $5,
-         serialized_object = $6,
-         external_authentication_uid = $7,
-         recovery_authentication_enabled = $8,
-         email = $9,
-         username = $10,
-         last_updated_by = $11,
-         updated_at = $12,
-         admin = false
-  WHERE id = $13">>
-}.
+<<"SELECT update_user($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, false, $13)">>}.
 
 {list_common_orgs_for_users,
  <<"SELECT o.name as name, o.full_name as full_name, o.id as guid"
@@ -587,18 +565,4 @@
 }.
 
 {update_user_by_id,
-<<"UPDATE users
-     SET hashed_password = $1,
-         salt = $2,
-         hash_type = $3,
-         serialized_object = $4,
-         external_authentication_uid = $5,
-         recovery_authentication_enabled = $6,
-         email = $7,
-         username = $8,
-         last_updated_by = $9,
-         updated_at = $10,
-         admin = false,
-         pubkey_version = 0
-  WHERE  id = $11">>
-}.
+<<"SELECT update_user(0, 'this_is_not_a_key', $1, $2, $3, $4, $5, $6, $7, $8, $9, $10, false, $11)">>}.
--- a/src/oc_erchef/apps/chef_db/src/chef_db.erl
+++ b/src/oc_erchef/apps/chef_db/src/chef_db.erl
@@ -155,7 +155,6 @@ make_context(ApiVersion, ReqId) ->
 make_context(ApiVersion, ReqId, Darklaunch) ->
     #context{server_api_version = ApiVersion, reqid = ReqId, darklaunch = Darklaunch}.
 
-
 -spec create(object_rec(), #context{}, object_id()) -> ok | {conflict, term()} | {error, term()}.
 create(#chef_cookbook_version{} = Record, DbContext, ActorId) ->
     create_object(DbContext, create_cookbook_version, Record, ActorId);
--- a/src/oc_erchef/apps/chef_db/src/chef_sql.erl
+++ b/src/oc_erchef/apps/chef_db/src/chef_sql.erl
@@ -990,7 +990,10 @@ create_object(#chef_sandbox{id=SandboxId,
             %% the failure back up
             Error
     end.
--spec create_object(atom(), tuple() | list()) -> {ok, non_neg_integer()} | {error, term()} | {conflict, term()}.
+-spec create_object(atom(), tuple() | list()) -> {ok, non_neg_integer()} |
+                                                 {ok,[[{binary(),<<>>}]]} |
+                                                 {error, term()} |
+                                                 {conflict, term()}.
 %% okay, that's pretty ugly, but no more so than all the hacks in here and
 %% chef_db for plain old cookbooks, their versions, and their checksums
 %% and at least, it's in a transaction
@@ -1002,6 +1005,14 @@ create_object(insert_cookbook_artifact_version = QueryName, Args) when is_list(A
         {error, {<<"CS001">>, _}} -> {error, invalid_checksum};
         {error, _Why} = Error -> Error;
         {conflict, _Why} = Conflict -> Conflict
+    end;
+create_object(QueryName, Args) when QueryName =:= insert_user;
+                                    QueryName =:= insert_user_v0,
+                                    is_list(Args)  ->
+   case sqerl:select(QueryName, Args, first_as_scalar, [add_user]) of
+        {ok, 1} -> {ok, 1};
+        {ok, 0} -> {error, <<"Record not created '", (term_to_binary(QueryName))/binary, "'.">>};
+        Error -> Error
     end;
 create_object(QueryName, Args) when is_atom(QueryName), is_list(Args) ->
     sqerl:statement(QueryName, Args, count);
@@ -1081,10 +1092,20 @@ delete_object(delete_cookbook_by_orgid_name = Query, OrgId, Name) ->
 update(ObjectRec, ActorId) ->
     chef_object:update(ObjectRec, ActorId, fun select_rows/1).
 
+do_update(QueryName, UpdateFields) when QueryName =:= update_user_by_id;
+                                        QueryName =:= update_user_by_id_v0,
+                                        is_list(UpdateFields)  ->
+    case sqerl:select(QueryName, UpdateFields, first_as_scalar, [update_user]) of
+        {ok, 1} -> {ok, 1};
+        {ok, 0} -> {error, <<"Record not updated '", (term_to_binary(QueryName))/binary, "'.">>};
+        {ok, none} -> {ok, not_found};
+        Error ->
+            Error
+    end;
 do_update(QueryName, UpdateFields) ->
     case sqerl:statement(QueryName, UpdateFields) of
         {ok, 1} -> {ok, 1};
-        {ok, none} -> {ok, not_found};
+        {ok, none} -> not_found;
         Error ->
             Error
     end.
--- a/src/oc_erchef/schema/deploy/create_and_update_users.sql
+++ b/src/oc_erchef/schema/deploy/create_and_update_users.sql
@@ -1,0 +1,114 @@
+-- Deploy enterprise_chef:create_and_update_users to pg
+-- Inserts a sentinel value into the public_key field of a user table.
+-- Ensures keys table is the only source of truth for public_key
+
+BEGIN;
+
+DROP FUNCTION If EXISTS add_user(character,character,text,text,text,integer,text,text,password_hash_type,character,timestamp without time zone,timestamp without time zone,text,boolean,text,boolean);
+
+CREATE OR REPLACE FUNCTION add_user(p_id users.id%TYPE,
+                                    p_authz_id users.authz_id%TYPE,
+                                    p_username users.username%TYPE,
+                                    p_email users.email%TYPE,
+                                    p_public_key users.public_key%TYPE,
+                                    p_pubkey_version users.pubkey_version%TYPE,
+                                    p_hashed_password users.hashed_password%TYPE,
+                                    p_salt users.salt%TYPE,
+                                    p_hash_type users.hash_type%TYPE,
+                                    p_last_updated_by users.last_updated_by%TYPE,
+                                    p_created_at users.created_at%TYPE,
+                                    p_updated_at users.updated_at%TYPE,
+                                    p_external_authentication_uid users.external_authentication_uid%TYPE,
+                                    p_recovery_authentication_enabled users.recovery_authentication_enabled%TYPE,
+                                    p_serialized_object users.serialized_object%TYPE,
+                                    p_admin users.admin%TYPE) RETURNS integer AS $$
+  DECLARE
+      inserteduser integer;
+  BEGIN
+   -- Need this insert despite of the trigger since the trigger updates the keys table
+   -- from the value inserted in the users table. But we insert sentinel in the users table.
+    IF p_public_key != 'this_is_not_a_key' THEN
+      INSERT INTO keys
+        (id, key_name, public_key, key_version, created_at,  expires_at)
+      VALUES
+         (p_id, 'default', p_public_key, p_pubkey_version, now(), 'infinity'::timestamp);
+    END IF;
+    WITH createduser AS
+      (INSERT INTO users
+        (id, authz_id, username, email, public_key, hashed_password, salt, hash_type,
+         last_updated_by, created_at, updated_at, external_authentication_uid,
+         recovery_authentication_enabled, serialized_object, admin,
+         pubkey_version)
+       VALUES (p_id, p_authz_id, p_username, p_email, 'this_is_not_a_key',
+         p_hashed_password, p_salt, p_hash_type, p_last_updated_by, p_created_at, p_updated_at,
+         p_external_authentication_uid, p_recovery_authentication_enabled,
+         p_serialized_object, p_admin, p_pubkey_version) RETURNING 1)
+      SELECT count(*) FROM createduser INTO inserteduser;
+      RETURN inserteduser;
+  END;
+$$ LANGUAGE plpgsql;
+
+DROP FUNCTION IF EXISTS update_user(integer,text,text,text,password_hash_type,text,text,boolean,text,text,character,timestamp without time zone,boolean,character);
+
+CREATE OR REPLACE FUNCTION update_user(p_pubkey_version users.pubkey_version%TYPE,
+                                       p_public_key users.public_key%TYPE,
+                                       p_hashed_password users.hashed_password%TYPE,
+                                       p_salt users.salt%TYPE,
+                                       p_hash_type users.hash_type%TYPE,
+                                       p_serialized_object users.serialized_object%TYPE,
+                                       p_external_authentication_uid users.external_authentication_uid%TYPE,
+                                       p_recovery_authentication_enabled users.recovery_authentication_enabled%TYPE,
+                                       p_email users.email%TYPE,
+                                       p_username users.username%TYPE,
+                                       p_last_updated_by users.last_updated_by%TYPE,
+                                       p_updated_at users.updated_at%TYPE,
+                                       p_admin users.admin%TYPE,
+                                       p_id character(32)) RETURNS integer AS $$
+    DECLARE
+      updateduser integer;
+    BEGIN
+    -- If user sets public_key = null in api v0, default public_key is expected to be
+    -- deleted from the keys table. Not supported for v1.
+    IF p_public_key IS NULL THEN
+      DELETE FROM keys WHERE id = p_id AND key_name = 'default';
+    END IF;
+    -- Need this insert despite of the trigger since the trigger updates the keys table
+    -- from the value inserted in the users table. But we insert sentinel in the users table.
+    IF p_public_key != 'this_is_not_a_key' THEN
+      UPDATE keys SET public_key = p_public_key,
+                      key_version = p_pubkey_version,
+                      expires_at = 'infinity'::timestamp
+      WHERE id = p_id AND key_name = 'default';
+        INSERT INTO keys (id, key_name, public_key, key_version, created_at, expires_at)
+        SELECT p_id, 'default', p_public_key, p_pubkey_version, now(), 'infinity'::timestamp
+        WHERE NOT EXISTS (SELECT 1 FROM keys WHERE id = p_id AND key_name = 'default');
+      -- TODO: Use the On Conflict to do Upsert once we drop support for 9.2.
+      -- INSERT INTO keys (id, key_name, public_key, key_version, created_at, expires_at)
+      --   VALUES (p_id, 'default', p_public_key, p_pubkey_version, now(), 'infinity'::timestamp)
+      --   ON CONFLICT (id, key_name)
+      -- DO UPDATE SET public_key = p_public_key,
+      --               key_version = p_pubkey_version,
+      --               expires_at = 'infinity'::timestamp;
+    END IF;
+    WITH changeduser AS
+      (UPDATE users SET
+        username = p_username,
+        email = p_email,
+        public_key = 'this_is_not_a_key',
+        hashed_password = p_hashed_password,
+        salt = p_salt,
+        hash_type = p_hash_type,
+        last_updated_by = p_last_updated_by,
+        updated_at = p_updated_at,
+        external_authentication_uid = p_external_authentication_uid,
+        recovery_authentication_enabled = p_recovery_authentication_enabled,
+        serialized_object = p_serialized_object,
+        admin = p_admin,
+        pubkey_version = p_pubkey_version
+      WHERE id = p_id RETURNING 1)
+    SELECT count(*) FROM changeduser INTO updateduser;
+    RETURN updateduser;
+   END
+$$ LANGUAGE plpgsql;
+
+COMMIT;
--- a/src/oc_erchef/schema/deploy/keys_update_trigger.sql
+++ b/src/oc_erchef/schema/deploy/keys_update_trigger.sql
@@ -1,33 +1,43 @@
 -- Deploy keys_update_trigger
 --
 -- This keeps the keys table consistent when the clients and user tables update their keys
---
+-- We need these triggers to maintain compatibility with older versions of chef-server.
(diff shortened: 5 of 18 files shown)


src/oc_erchef/apps/chef_db/src/chef_sql.erl
```diff
@@ -1004,7 +1004,7 @@ create_object(insert_cookbook_artifact_version = QueryName, Args) when is_list(A
         {conflict, _Why} = Conflict -> Conflict
     end;
 create_object(QueryName, Args) when is_atom(QueryName), is_list(Args) ->
-    sqerl:statement(QueryName, Args, count);
+    sqerl:select(QueryName, Args, count);
```
marcparadise: Why did you need to change this to `select`?  Context for concern: creation of most other object types is done via insert statements, and this would affect them as well.

src/oc_erchef/schema/sqitch.plan
```diff
@@ -80,3 +80,7 @@ node_policy_name_policy_group 2015-09-04T01:15:03Z Daniel DeLeo <ddeleo@lorentz.
 
 users_email_functional_index [users] 2017-05-30T12:10:32Z Stephan Renatus <srenatus@chef.io> # Add functional index for index-based lower(email) lookups
 @users_email_functional_index 2017-05-30T13:27:15Z Stephan Renatus <srenatus@chef.io> # Add functional index for index-based lower(email) lookups
+
+create_and_update_users 2017-08-23T22:53:21Z Prajakta Purohit <prajakta@chef.io> # Adding functions to create and update users by inserting a sentinel value in the public_key_columns
+keys_update_trigger [keys_update_trigger@users_email_functional_index] 2017-08-23T23:12:29Z Prajakta Purohit <prajakta@chef.io> # The insert and update triggers ignore rows with sentinel value for public_key.
```
marcparadise: The creation of this change as `@users_email_functional_index` seems off to me since we want it to go in sequence after the previous one - and if sqitch tracks this as dependent on users_email_functional_index, it may not go in that order.  Could you verify that this won't be a thing?

src/oc_erchef/apps/chef_db/src/chef_db.erl
```diff
@@ -167,6 +166,7 @@ create(ObjectRec0, #context{server_api_version = ApiVersion, reqid = ReqId}, Act
     Fields = chef_object:fields_for_insert(ObjectRec2),
     case stats_hero:ctime(ReqId, {chef_sql, create_object},
                           fun() -> chef_sql:create_object(QueryName, Fields) end) of
+        {ok,[[{<<"add_user">>,<<>>}]]} -> ok;
```
marcparadise: This is very specific to the add_user query/function.  If we are not able to get a return value consistent with other inserts (number of rows inserted), it's worth a closer look at epgsql+sqerl to understand the  returned `[[{binary(), binary()}]]` from stored proc calls. Would it be sufficient to match similar to the below w/ guards:

src/oc_erchef/schema/deploy/create_and_update_users.sql
```diff
@@ -0,0 +1,88 @@
...
+                                       p_external_authentication_uid users.external_authentication_uid%TYPE,
+                                       p_recovery_authentication_enabled users.recovery_authentication_enabled%TYPE,
+                                       p_email users.email%TYPE,
+                                       p_username users.username%TYPE,
+                                       p_last_updated_by users.last_updated_by%TYPE,
+                                       p_updated_at users.updated_at%TYPE,
+                                       p_admin users.admin%TYPE,
+                                       p_id character(32)) RETURNS integer AS $update_user$
+    BEGIN
+    IF p_public_key != 'thisisnotakey' THEN
```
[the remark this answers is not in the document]
PrajaktaPurohit: The reason we are not using null is - null in v0 api implies that the user wants to delete the key, hence the use of a sentinel value.

src/oc_erchef/apps/chef_db/src/chef_sql.erl
PrajaktaPurohit: We changed the insert statements to be a function in plpgsql, and those return in the form : {ok,[[{<<"add_user">>,<<>>}]]}
but select expects return in the form:  https://github.com/chef/sqerl/blob/master/src/sqerl.erl#L94-L101
which is why the change from statement to select.

src/oc_erchef/schema/sqitch.plan
PrajaktaPurohit: That is a direct change due to `sqitch rework keys_update_trigger` but will make sure.

src/oc_erchef/schema/revert/keys_update_trigger.sql
```diff
@@ -1,14 +1,45 @@
...
+     IF NEW.public_key IS NOT NULL THEN
+       INSERT INTO keys
+         (id, key_name, public_key, key_version, created_at,  expires_at) VALUES
+         (NEW.id, 'default', NEW.public_key, NEW.pubkey_version, now(), 'infinity'::timestamp);
+     END IF;
+     RETURN NULL; -- result is ignored since this is an AFTER trigger
+   END;
+$add_key$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE FUNCTION update_key() RETURNS TRIGGER AS $update_key$
```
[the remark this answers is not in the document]
marcparadise: Disregard, it was misnamed from the start and changing it now will complicate the sqitch bits unnecessarily.

src/oc_erchef/schema/revert/keys_update_trigger@users_email_functional_index.sql
```diff
@@ -0,0 +1,14 @@
+-- Revert keys_update_trigger
```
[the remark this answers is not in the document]
PrajaktaPurohit: right the current keys_update_trigger.sql does that, this is the revert for the file before making my changes.

src/oc_erchef/apps/chef_db/src/chef_sql.erl
PrajaktaPurohit: updated as discussed!

src/oc_erchef/schema/deploy/create_and_update_users.sql
```diff
@@ -0,0 +1,92 @@
...
+                        key_version = p_pubkey_version,
+                        expires_at = 'infinity'::timestamp
+        WHERE id = p_id AND key_name = 'default';
+       INSERT INTO keys (id, key_name, public_key, key_version, created_at, expires_at)
+         SELECT p_id, 'default', p_public_key, p_pubkey_version, now(), 'infinity'::timestamp
+         WHERE NOT EXISTS (SELECT 1 FROM keys WHERE id = p_id AND key_name = 'default');
+      ELSE
+       DELETE FROM keys WHERE id = p_id AND key_name = 'default';
+      END IF;
+    END IF;
```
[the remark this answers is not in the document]
PrajaktaPurohit: we still need this because we will not be inserting the key into the users table in all the new requests (after this release) so just leaving the trigger in will update 'thisisnotakey' into the keys table. This function does the right things and the trigger will only be useful for older versions of chef-server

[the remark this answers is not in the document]
markan: In SQL NULL returns NULL for all comparisons. So the DELETE will never fire above I think, but instead NULL != to 'thisisnotakey' will return NULL, which is falsy, and the inner if will never be reached.

```
CREATE OR REPLACE FUNCTION test(val text)
RETURNS text
AS $$
DECLARE
  result text;
BEGIN
   IF val != 'A' THEN
     IF val IS NOT NULL THEN
      result := 'NOT A, NOT NULL';
     ELSE 
      result := 'NOT A, BUT NULL';
     END IF;
   ELSE
      result := 'A';
   END IF;
   RETURN result;
END;
$$;
```

```
opscode_chef=# select test(NULL);
 test 
------
 A
```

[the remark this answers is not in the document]
markan: We probably should have unit test coverage of some of these functions to insure they work as expected.

src/oc_erchef/schema/deploy/keys_update_trigger.sql
```diff
@@ -5,29 +5,35 @@
...
-   BEGIN
-     IF NEW.public_key IS NOT NULL THEN
-       INSERT INTO keys
-         (id, key_name, public_key, key_version, created_at,  expires_at) VALUES
-         (NEW.id, 'default', NEW.public_key, NEW.pubkey_version, now(), 'infinity'::timestamp);
-     END IF;
-     RETURN NULL; -- result is ignored since this is an AFTER trigger
-   END;
+  BEGIN
+    IF NEW.public_key != 'thisisnotakey' THEN
```
markan: See the NULL comparision comment for create_and_update_users.sql above

src/oc_erchef/schema/deploy/keys_update_trigger.sql
```diff
@@ -5,29 +5,35 @@
...
-         SELECT NEW.id, 'default', NEW.public_key, NEW.pubkey_version, now(), 'infinity'::timestamp
-         WHERE NOT EXISTS (SELECT 1 FROM keys WHERE id = NEW.id AND key_name = 'default');
-     ELSE
-       DELETE FROM keys WHERE id = NEW.id AND key_name = 'default';
-     END IF;
-     RETURN NULL; -- result is ignored since this is an AFTER trigger
-   END;
+  BEGIN
+    IF NEW.public_key != 'thisisnotakey' THEN
+      IF NEW.public_key IS NOT NULL THEN
```
markan: See the NULL comparision comment for create_and_update_users.sql above

src/oc_erchef/apps/chef_db/priv/pgsql_statements.config
```diff
@@ -484,17 +484,10 @@
 %% User-related
 %%
 {insert_user,
-  <<"INSERT INTO users (id, authz_id, username, email, hashed_password, salt,
-     hash_type, last_updated_by, created_at, updated_at, external_authentication_uid,
-      recovery_authentication_enabled, serialized_object, admin, pubkey_version) VALUES
-      ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, false, 0)">>}.
+  <<"SELECT add_user($1, $2, $3, $4, 'thisisnotakey', 0, $5, $6, $7, $8, $9, $10, $11, $12, $13, false)">>}.
```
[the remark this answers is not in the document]
PrajaktaPurohit: done!

src/oc_erchef/schema/deploy/create_and_update_users.sql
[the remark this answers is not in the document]
PrajaktaPurohit: That is an awesome catch! updated the cases to fall through to the delete operation when expected.

I will check the unit tests in bifrost and add something to test this functionality.

src/oc_erchef/schema/deploy/keys_update_trigger.sql
PrajaktaPurohit: updated.

src/oc_erchef/apps/chef_db/src/chef_db.erl
```diff
@@ -168,6 +167,7 @@ create(ObjectRec0, #context{server_api_version = ApiVersion, reqid = ReqId}, Act
     case stats_hero:ctime(ReqId, {chef_sql, create_object},
                           fun() -> chef_sql:create_object(QueryName, Fields) end) of
         {ok, 1} -> ok;
+        {ok, 0} -> <<"Record not created '", ObjectRec2/binary, "'.">>;
```
stevendanna: The dialyzer spec on this function needs to be updated.  Perhaps it would be good to use `{error, <<"Record..._`  here as you did below. I think this case would be pretty rare as I imagine if we can't create what we are trying to create, we would get an error back rather than a successful query?

src/oc_erchef/schema/deploy/create_and_update_users.sql
```diff
@@ -0,0 +1,101 @@
...
+      DELETE FROM keys WHERE id = p_id AND key_name = 'default';
+    END IF;
+    IF p_public_key != 'this_is_not_a_key' THEN
+      UPDATE keys SET public_key = p_public_key,
+                      key_version = p_pubkey_version,
+                      expires_at = 'infinity'::timestamp
+      WHERE id = p_id AND key_name = 'default';
+        INSERT INTO keys (id, key_name, public_key, key_version, created_at, expires_at)
+        SELECT p_id, 'default', p_public_key, p_pubkey_version, now(), 'infinity'::timestamp
+        WHERE NOT EXISTS (SELECT 1 FROM keys WHERE id = p_id AND key_name = 'default');
```
stevendanna: I know this is how we are doing "upsert" in the original code, but I think this construction is racy.  My guess is that concurrent user updates are rare enough that we can stay with this though.

src/oc_erchef/schema/deploy/create_and_update_users.sql
```diff
@@ -0,0 +1,105 @@
...
+        (id, authz_id, username, email, public_key, hashed_password, salt, hash_type,
+         last_updated_by, created_at, updated_at, external_authentication_uid,
+         recovery_authentication_enabled, serialized_object, admin,
+         pubkey_version)
+       VALUES (p_id, p_authz_id, p_username, p_email, 'this_is_not_a_key',
+         p_hashed_password, p_salt, p_hash_type, p_last_updated_by, p_created_at, p_updated_at,
+         p_external_authentication_uid, p_recovery_authentication_enabled,
+         p_serialized_object, p_admin, p_pubkey_version) RETURNING 1)
+      SELECT count(*) FROM createduser INTO inserteduser;
+      RETURN inserteduser;
```
christophermaier: You don't need this indirection of using CTEs to always insert the literal value `1` into a variable.

Why does this function need to return anything if it's always returning `1`? Is the desire to indicate whether something was inserted?

src/oc_erchef/schema/revert/create_and_update_users.sql
```diff
@@ -0,0 +1,8 @@
+-- Revert enterprise_chef:create_and_update_users from pg
+
+BEGIN;
+
+DROP FUNCTION IF EXISTS add_user();
+DROP FUNCTION IF EXISTS update_user();
```
christophermaier: The deploy script implies that these functions existed in the system before, but this revert will not replace those older implementations.

src/oc_erchef/schema/revert/keys_update_trigger.sql
```diff
@@ -1,14 +1,45 @@
...
+   BEGIN
+     IF NEW.public_key IS NOT NULL THEN
+       UPDATE keys SET public_key = NEW.public_key, key_version = NEW.pubkey_version, expires_at = 'infinity'::timestamp
+         WHERE id = NEW.id AND key_name = 'default';
+       INSERT INTO  keys (id, key_name, public_key, key_version, created_at, expires_at)
+         SELECT NEW.id, 'default', NEW.public_key, NEW.pubkey_version, now(), 'infinity'::timestamp
+         WHERE NOT EXISTS (SELECT 1 FROM keys WHERE id = NEW.id AND key_name = 'default');
```
christophermaier: Looks like you're trying to do an upsert? Can you use the `ON CONFLICT` clause instead (see https://www.postgresql.org/docs/current/static/sql-insert.html)... not sure what version of PG y'all are on these days).

I saw this was being done above, as well.

src/oc_erchef/schema/t/test_create_and_update_users.sql
```diff
@@ -0,0 +1,216 @@
+CREATE OR REPLACE FUNCTION test_zzzzcreate_and_update_users()
```
christophermaier: Why "zzzz"?

src/oc_erchef/schema/t/test_create_and_update_users.sql
```diff
@@ -0,0 +1,216 @@
...
+RETURNS SETOF TEXT
+LANGUAGE plpgsql
+AS $$
+BEGIN
+
+-- v0 add_user tests --
+RETURN QUERY SELECT is(add_user('11111111111111111111111111111111',
+                                '22222222222222222222222222222222',
+                                'aaaa',
```
christophermaier: For these dummy values, I'd recommend having them describe what they are as much as you can given any constraints that exist on the data.

So if "aaaa" is a username, use "test_user" instead, etc.

src/oc_erchef/schema/t/test_create_and_update_users.sql
```diff
@@ -0,0 +1,216 @@
...
+    ARRAY['this_is_not_a_key'],
+    'unchanged public_key in users table for v0'
+);
+
+PREPARE "after update email v0" AS SELECT email from users
+                                     WHERE id='11111111111111111111111111111111';
+RETURN QUERY SELECT results_eq(
+    '"after update email v0"',
+    ARRAY['changed_email_address_v0@chef.io'],
+    'after update email address for v0'
```
christophermaier: I'd recommend stating these messages in terms of the desired behavior you're going after, instead of the mechanics of what is happening. That makes it easier to figure out what the overall intention of the code is.

src/oc_erchef/schema/deploy/create_and_update_users.sql
christophermaier: Ah, I commented on this down below... if it's available on the Postgres version you're using, an `ON CONFLICT` clause would be better for this.

Alternatively, using CTEs for this might be preferable, as you could contain the operations within a single statement.

You're also in PL/pgSQL, so you could also use exception handling, but that's more expensive.

src/oc_erchef/schema/deploy/create_and_update_users.sql
```diff
@@ -0,0 +1,105 @@
...
+      VALUES
+         (p_id, 'default', p_public_key, p_pubkey_version, now(), 'infinity'::timestamp);
+    END IF;
+    WITH createduser AS
+      (INSERT INTO users
+        (id, authz_id, username, email, public_key, hashed_password, salt, hash_type,
+         last_updated_by, created_at, updated_at, external_authentication_uid,
+         recovery_authentication_enabled, serialized_object, admin,
+         pubkey_version)
+       VALUES (p_id, p_authz_id, p_username, p_email, 'this_is_not_a_key',
```
christophermaier: If the intention is to have the `keys` table be the source of truth for keys, why keep key-related information in the `users` table at all? Is this step 1 of a broader refactoring?

src/oc_erchef/schema/t/test_create_and_update_users.sql
PrajaktaPurohit: wanted that test to run last - will edit.

src/oc_erchef/schema/deploy/create_and_update_users.sql
PrajaktaPurohit: yep - sort of a phase1 since on env like hosted we need to support older versions of chef-server.

src/oc_erchef/schema/deploy/create_and_update_users.sql
PrajaktaPurohit: yea - to make sure that the insert did happen.

src/oc_erchef/schema/deploy/create_and_update_users.sql
stevendanna: Unfortunately, I'm not sure we can drop support for 9.2 yet (we should check with product) which means no `ON CONFLICT`.  I think overall using `CTE`s would also make some of this a bit more readable; Perhaps we can do that as a more easily reviewable follow-up improvement. (Getting a bit worried that there is a lot going on here as this is already code not many know well.)

src/oc_erchef/schema/t/test_create_and_update_users.sql
PrajaktaPurohit: updated!

src/oc_erchef/schema/t/test_create_and_update_users.sql
PrajaktaPurohit: updated!

src/oc_erchef/schema/deploy/create_and_update_users.sql
PrajaktaPurohit: I have added the new syntax as a comment in there - a TODO for the future!

src/oc_erchef/schema/deploy/create_and_update_users.sql
PrajaktaPurohit: RE: You don't need this indirection of using CTEs to always insert the literal value 1 into a variable. - will update that but maybe as a second pass, if that is ok?

src/oc_erchef/schema/revert/create_and_update_users.sql
PrajaktaPurohit: Yes - this is just safeguarding for the future, we don't expect the functions to actually be present.

src/oc_erchef/schema/revert/keys_update_trigger.sql
PrajaktaPurohit: added a comment can be changes once we drop support for postgres 9.2/9.3

src/oc_erchef/schema/t/test_create_and_update_users.sql
PrajaktaPurohit: done!

src/oc_erchef/schema/t/test_create_and_update_users.sql
PrajaktaPurohit: done!

src/oc_erchef/schema/revert/keys_update_trigger.sql
PrajaktaPurohit: Actually maybe I will leave this one as is - since it is the revert of the update trigger.

src/oc_erchef/apps/chef_db/src/chef_db.erl
```diff
@@ -258,6 +258,7 @@ update(#chef_cookbook_version{org_id =OrgId} = Record, #context{reqid = ReqId} =
 update(ObjectRec, #context{reqid = ReqId}, ActorId) ->
     case stats_hero:ctime(ReqId, {chef_sql, do_update},
                           fun() -> chef_sql:update(ObjectRec, ActorId) end) of
+        0 -> {error, <<"Record not updated '", ObjectRec/binary, "'.">>};
```
marcparadise: It looks like the changes [do_update](https://github.com/chef/chef-server/pull/1383/files#diff-43908ac1f0b634347760bb3d12a9cba8R1100) will return `{error, Reason}` - can we remove this branch from here? 

Same question for `create`, looks like the {ok. 0} return is handled there.

src/oc_erchef/apps/chef_db/src/chef_sql.erl
```diff
@@ -1081,6 +1092,16 @@ delete_object(delete_cookbook_by_orgid_name = Query, OrgId, Name) ->
 update(ObjectRec, ActorId) ->
     chef_object:update(ObjectRec, ActorId, fun select_rows/1).
 
+do_update(QueryName, UpdateFields) when QueryName =:= update_user_by_id;
+                                        QueryName =:= update_user_by_id_v0,
+                                        is_list(UpdateFields)  ->
+    case sqerl:select(QueryName, UpdateFields, first_as_scalar, [update_user]) of
+        {ok, 1} -> {ok, 1};
+        {ok, 0} -> {error, <<"Record not updated '", (term_to_binary(QueryName))/binary, "'.">>};
+        {ok, none} -> {ok, not_found};
```
marcparadise: the caller seems is expecting `not_found` instead of `{ok, not_found}`.

src/oc_erchef/schema/deploy/keys_update_trigger.sql
```diff
@@ -1,33 +1,43 @@
 -- Deploy keys_update_trigger
 --
 -- This keeps the keys table consistent when the clients and user tables update their keys
---
+-- We need these triggers to maintain compatibility with older versions ok chef-server.
```
marcparadise: s/ok/of

src/oc_erchef/schema/deploy/keys_update_trigger@users_email_functional_index.sql
```diff
@@ -0,0 +1,45 @@
...
+
+CREATE OR REPLACE FUNCTION update_key() RETURNS TRIGGER AS $update_key$
+   BEGIN
+     IF NEW.public_key IS NOT NULL THEN
+       UPDATE keys SET public_key = NEW.public_key, key_version = NEW.pubkey_version, expires_at = 'infinity'::timestamp
+         WHERE id = NEW.id AND key_name = 'default';
+       INSERT INTO  keys (id, key_name, public_key, key_version, created_at, expires_at)
+         SELECT NEW.id, 'default', NEW.public_key, NEW.pubkey_version, now(), 'infinity'::timestamp
+         WHERE NOT EXISTS (SELECT 1 FROM keys WHERE id = NEW.id AND key_name = 'default');
+     ELSE
```
marcparadise: This is a port of the existing logic --  but it at always makes me say, "What what?" when I read it.  Could you add a comment explaining why a null key causes a deletion? (v0 api behavior compatibility)

src/oc_erchef/apps/chef_db/src/chef_db.erl
PrajaktaPurohit: there are 2 different code paths that call into either the do_update or update functions, so we need this in both places.

PrajaktaPurohit: yes!

src/oc_erchef/schema/deploy/keys_update_trigger@users_email_functional_index.sql
PrajaktaPurohit: done!

PR #1384, opened 2017-08-24
Stats endpoint

jaym:


src/oc_erchef/apps/oc_chef_wm/src/chef_wm_pooler_collector.erl
```diff
@@ -0,0 +1,76 @@
...
+    [gauge_metric([{pool_name, PoolName}], PoolData#pool.queue_max) || {PoolName, PoolData} <- Pools].
+
+
+
+create_gauge(Name, Help, Data) ->
+    create_mf(Name, Help, gauge, ?MODULE, Data).
+
+monitored_pools() ->
+    % TODO(jaym) 08-23-17: Move this out to configuration
+    [sqerl, oc_chef_authz_http, chef_index_http, chef_depsolver].
```
irvingpop: I'm not 100% sure, but I suspect that keygen, chef_objects and data_collector may also have pools to monitor

jaym: Tried those, didn't work :(

jaym: You may have been thinking of this:
```
(oc_erchef@api)1> chef_keygen_cache:status().
[{keys,10},
 {max,10},
 {max_workers,2},
 {cur_max_workers,2},
 {inflight,[]},
 {avail_workers,2},
 {start_size,0}]
```

omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
```diff
@@ -56,6 +56,12 @@
         {error_logger_hwm, <%= @log_rotation['max_messages_per_second'] %>}
     ]},
 
+    {prometheus, [{collectors, [default,
+                                <% if node['private_chef']['postgresql']['enable'] && !node['private_chef']['postgresql']['external'] -%>
```
stevendanna: Is there anything about the stats we collect that would make it hard to collect for external DBs?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
```diff
@@ -0,0 +1,59 @@
...
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+
+-module(chef_wm_stats).
+
+-ifdef(TEST).
+-compile(export_all).
```
srenatus: [nit] This doesn't export more than `-export()` already exports, does it?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
```diff
@@ -0,0 +1,59 @@
...
+     [
+      {"application/json", to_json},
+      % I think there is a bug in webmachine where it wont allow us to use
+      % 'text/plain; version=0.0.4'.
+      % TODO: Understand https://github.com/basho/webmachine/blob/develop/src/webmachine_util.erl#L140-L158
+      {{"text/plain",[{"version","0.0.4"}]}, to_text},
+      {"text/plain", to_text}
+     ], Req, State}.
+
+to_json(Req, State) ->
```
srenatus: Is this meant for later inclusion? I wonder if this should _provide application/json_ if it doesn't actually provide it.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
```diff
@@ -0,0 +1,59 @@
...
+         allowed_methods/2,
+         content_types_provided/2,
+         to_json/2,
+         to_text/2]).
+
+-include_lib("webmachine/include/webmachine.hrl").
+-define(A2B(X), erlang:atom_to_binary(X, utf8)).
+
+init(_Any) ->
+    {ok, <<"{}">>}.
```
srenatus: What's the use of passing `<<"{}">>` around as `State`? (Am I missing something?)

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
```diff
@@ -0,0 +1,59 @@
...
+-endif.
+
+-export([init/1,
+         allowed_methods/2,
+         content_types_provided/2,
+         to_json/2,
+         to_text/2]).
+
+-include_lib("webmachine/include/webmachine.hrl").
+-define(A2B(X), erlang:atom_to_binary(X, utf8)).
```
srenatus: [nit] Is this used?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
jaym: Yep, I'll be working on this today. Prometheus is a secondary format and json will be default

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
jaym: Not sure, I just copy pasted from elsewhere. Will take a closer look here

omnibus/files/private-chef-cookbooks/private-chef/templates/default/oc_erchef.config.erb
jaym: Not hard, but I was more thinking about who owns the service. The idea was, if the service is not on the box, it's not our problem to monitor. I could be convinced otherwise, however there might be weirdness if say multiple frontends were to report a service that they didn't own.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_prometheus_json_format.erl
```diff
@@ -0,0 +1,100 @@
...
+            %TODO(jaym) 08/27/17: It might be worthwhile making this defensive to failure... always cleanup
+            erase(?PROCESS_DICT_STORAGE),
+            jiffy:encode(Data)
+    end.
+
+registry_collect_callback(Registry, Collector) ->
+  Callback = fun (MF) ->
+                 Data = get(?PROCESS_DICT_STORAGE),
+                 put(?PROCESS_DICT_STORAGE, [mf_to_erl(MF) | Data])
+             end,
```
stevendanna: It is a bit of a bummer that the API here doesn't let us pass around an accumulator instead.

jaym: yea :(
Would have been a lot more cleaner

src/oc_erchef/apps/chef_db/priv/pgsql_statements.config
```diff
@@ -19,6 +19,14 @@
 
 {ping, <<"SELECT 'pong' as ping LIMIT 1">>}.
 
+{stats, <<"SELECT SUM(seq_scan) as seq_scan, SUM(seq_tup_read) as seq_tup_read, SUM(idx_scan) as idx_scan,"
+          " SUM(idx_tup_fetch) as idx_tup_fetch, SUM(n_tup_ins) as n_tup_ins, SUM(n_tup_upd) as n_tup_upd,"
+          " SUM(n_tup_del) as n_tup_del, SUM(n_live_tup) as n_live_tup, SUM(n_dead_tup) as n_dead_tup,"
+          " SUM(heap_blks_read) as heap_blks_read, SUM(heap_blks_hit) as heap_blks_hit,"
+          " SUM(idx_blks_read) as idx_blks_read, SUM(idx_blks_hit) as idx_blks_hit, SUM(toast_blks_read) as toast_blks_read,"
+          " SUM(toast_blks_hit) as toast_blks_hit, SUM(tidx_blks_read) as tidx_blks_read,"
+          " SUM(tidx_blks_hit) as tidx_blks_hit FROM pg_stat_all_tables, pg_statio_all_tables">>}.
```
ksubrama: I think this is inaccurate.  "," does a cartesian join and inflates the number of final rows.  I'm still trying to figure out if there was a join here we wanted or if we just wanted to get a composite query and how to do that (because reading postgresql to me is like spaniard trying to read Brazilian Portuguese...  so close but yet so far).

jaym: there's no join. just a composite of 2 queries, perhaps i was trying to be too clever

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
```diff
@@ -169,6 +169,25 @@
...
 
+if node['private_chef']['opscode-erchef']['stats_user']
+  stats_api_passwd = PrivateChef.credentials.get('opscode_erchef', 'stats_password')
+  stats_passwd_file = node['private_chef']['opscode-erchef']['stats_password_file']
+
+  execute 'stats_api_append_password' do
+    command "openssl passwd -apr1 '#{stats_api_passwd}' >> #{stats_passwd_file}"
+    action :nothing
+    sensitive true
+  end
```
stevendanna: We need to store this password back into the secrets store somewhere.  Perhaps rather than doing this as an execute we can move this into a library and then generate the password when we generate the other passwords in the initial configuration parsing.

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
```diff
@@ -169,6 +169,25 @@
...
+
+  file stats_passwd_file do
+    content "#{node['private_chef']['opscode-erchef']['stats_user']}:"
+    mode '0400'
+    owner OmnibusHelper.new(node).ownership['owner']
+    group OmnibusHelper.new(node).ownership['group']
+    notifies :run, 'execute[stats_api_append_password]', :immediately
+  end
+end
```
stevendanna: As of the veil work, we try not to put any passwords outside of the veil store when the user has `insecure_addon_compat false`.  We've never really talked about how strict we want to be about that moving forward since that work was done to address some particular customer concerns.

Unfortunately, from the docs it looks like it might be hard/impossible to use the basic_auth module with out sticking the credential in that file.  I suppose we could change over to using access_by_lua, but that seems a bit complex for this.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_prometheus_json_format.erl
```diff
@@ -0,0 +1,100 @@
...
+content_type() ->
+    <<"application/json">>.
+
+-spec format() -> binary().
+format() ->
+    format(default).
+
+-spec format(Registry :: prometheus_registry:registry()) -> binary().
+format(Registry) ->
+    put(?PROCESS_DICT_STORAGE, []),
```
stevendanna: My only concern with using the process dictionary here is out of ignorance.  I don't know if putting what could be a lot of data in the process dictionary will cause us any problems down the road.  The prometheus code uses the ram_disk module for all of it's formatters; however I have no idea if that is better or worse.  Perhaps for now this is just something we live with but keep an eye on.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -647,6 +648,14 @@ def gen_redundant(node_name, topology)
       end
     end
 
+    def ensure_stats_password
```
stevendanna: The RFC for this feature says:

> A username and password will be generated if it not provided.

So, I think we should probably be treating this like the other non-optional passwords where we generate the username and password if they are not provided.

src/oc_erchef/apps/chef_db/src/chef_pgsql_collector.erl
```diff
@@ -0,0 +1,92 @@
...
+                                   counter_metric/2]).
+
+-behaviour(prometheus_collector).
+
+-define(METRICS, [
+                  {<<"seq_scan">>, {pg_stat_seq_scan, counter, "Number of sequential scans initiated", fun erlang:binary_to_integer/1}},
+                  {<<"seq_tup_read">>, {pg_stat_seq_tup_read, counter, "Number of live rows fetched by sequential scans", fun erlang:binary_to_integer/1}},
+                  {<<"idx_scan">>, {pg_stat_idx_scan, counter, "Number of index scans initiated", fun erlang:binary_to_integer/1}},
+                  {<<"idx_tup_fetch">>, {pg_stat_tup_fetch, counter, "Number of live rows fetched by index scans", fun erlang:binary_to_integer/1}},
+                  {<<"n_tup_ins">>, {pg_stat_n_tup_ins, counter, "Number of rows inserted", fun erlang:binary_to_integer/1}},
```
stevendanna: We should make it clear in these descriptions that these are sums across all tables perhaps.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
ksubrama: It should export everything in the file (for any tests to use them).

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
ksubrama: Cleaned up.

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
ksubrama: Wait I get this password from the secret store (unless I don't understand how this works).  I do chef-server-ctl add-secret opscode_erchef stats_password "12345".  This derived password is just an scrypt salted hash of that password.

omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb
ksubrama: The password is salted and hashed from my understanding.

```
cat /var/opt/opscode/nginx/stats_htpasswd
foo:$apr1$S.a9NOEq$YiWDrTTJb68HjyFjm9CmH/```

So it might be ok?

Comment (ksubrama) [reply omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb]: http://nginx.org/en/docs/http/ngx_http_auth_basic_module.html does seem to mention that the actual algorithms used might be...  weak?  I get suspicious when I see MD5 in there.  But there seems to be no other nice way to do this.

File: omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb (hunk 12 above)
Comment (ksubrama) [reply omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb]: Ah I missed that.  Ok - I need to figure out how to save a generated password.

File: omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb (hunk 10 above)
Comment (stevendanna) [reply omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb]: @ksubrama Ah, yeah, somehow I overlooked the fact that this is in fact NOT a "plain text password" in that file.  I think that means we are OK.

File: src/oc_erchef/apps/oc_chef_wm/src/chef_wm_prometheus_json_format.erl (hunk 11 above)
Comment (jaym) [reply src/oc_erchef/apps/oc_chef_wm/src/chef_wm_prometheus_json_format.erl]: Yea, I wanted to use ram file, but I couldn't figure out how to do it without rolling my own json serializer. That being said, there shouldn't be a lot of data. The other option I could think of was creating a gen server to store the state.

If it makes you feel better, webmachine makes fairly heavy use of the process dictionary: https://github.com/webmachine/webmachine/blob/82472528d8a735d28fa92a7c6e03fc97cb4d1913/src/webmachine_decision_core.erl#L28

Comment (stevendanna) [reply src/oc_erchef/apps/oc_chef_wm/src/chef_wm_prometheus_json_format.erl]: I'm +1 on leaving this as is for now.  Thanks for following up.

File: src/oc_erchef/apps/chef_db/src/chef_pgsql_collector.erl (hunk 13 above)
Comment (ksubrama) [reply src/oc_erchef/apps/chef_db/src/chef_pgsql_collector.erl]: :+1:

File: omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb (hunk 12 above)
Comment (ksubrama) [reply omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb]: :+1:

File: omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb (hunk 9 above)
Comment (ksubrama) [reply omnibus/files/private-chef-cookbooks/private-chef/recipes/nginx.rb]: I refactored this so that it autogenerates a password if not provided and also just generates the file in one go (instead of 2 steps).  This way the nginx server can get restarted properly only when the password file actually changes.

**Comment by jaym** [comment]
@ksubrama you mentioned you had some pedant tests. If so, could you push them up.

File: oc-chef-pedant/spec/api/stats_spec.rb [hunk 14]
```diff
@@ -0,0 +1,258 @@
...
+  it "returns metrics of the correct type", :smoke do
+    response = JSON.parse(get(request_url, platform.non_admin_user))
+    response_type_map.each do |name, type|
+      stat = response.find { |s| s["name"] == name }
+      expect(stat["metrics"]).not_to be_empty
+      stat["metrics"].each do |metric|
+        expect(metric).to have_key("value")
+        if type == "GAUGE" || type == "COUNTER"
+          expect(Integer(metric["value"])).to be_a(Integer)
+        elsif type == "UNTYPED" && boolean_stats.include?(name)
```
jaym: I think this is incorrect for untyped. The only thing that you can say about untyped is that it provides a value.

oc-chef-pedant/spec/api/stats_spec.rb
```diff
@@ -0,0 +1,258 @@
...
+
+  it "returns metrics of the correct type", :smoke do
+    response = JSON.parse(get(request_url, platform.non_admin_user))
+    response_type_map.each do |name, type|
+      stat = response.find { |s| s["name"] == name }
+      expect(stat["metrics"]).not_to be_empty
+      stat["metrics"].each do |metric|
+        expect(metric).to have_key("value")
+        if type == "GAUGE" || type == "COUNTER"
+          expect(Integer(metric["value"])).to be_a(Integer)
```
jaym: I think guages can be float as well

oc-chef-pedant/spec/api/stats_spec.rb
```diff
@@ -0,0 +1,258 @@
...
+      :body => response_body
+    })
+  end
+
+  it "returns metrics of the correct type", :smoke do
+    response = JSON.parse(get(request_url, platform.non_admin_user))
+    response_type_map.each do |name, type|
+      stat = response.find { |s| s["name"] == name }
+      expect(stat["metrics"]).not_to be_empty
+      stat["metrics"].each do |metric|
```
jaym: instead of iterating inside the `it`, is it possible to create an `it` for each so that if there is a failure, you know what is broken?

jaym: Something like https://coderwall.com/p/5kfxhg/dynamic-rspec-tests

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
```diff
@@ -0,0 +1,58 @@
...
+    {['GET'], Req, State}.
+
+content_types_provided(Req, State) ->
+    {
+     [
+      {"application/json", to_json},
+      % I think there is a bug in webmachine where it wont allow us to use
+      % 'text/plain; version=0.0.4'.
+      % TODO: Understand https://github.com/basho/webmachine/blob/develop/src/webmachine_util.erl#L140-L158
+      {{"text/plain",[{"version","0.0.4"}]}, to_text},
```
jaym: So to follow the rfc, we're supposed to accept `&format=`.
Seeing as we dont do that, we should probably remove `text/plain` for now. Thoughts?

stevendanna: I'd rather us just quickly implement a format parameter, but I understand that there is probably some desire to get this PR wrapped up.

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -17,6 +17,15 @@ def ownership
     {"owner" => owner, "group" => group}
   end
 
+  def apr1_password(password)
+    cmd = Mixlib::ShellOut.new("openssl passwd -apr1 '#{password}'")
```
stevendanna: Should we use a full path to our embedded openssl install here?

@jaym Do we need to do anything here for FIPS?

jaym: `OPENSSL_FIPS=1`

jaym: i highly doubt that will work with fips. It might be safer to use `crypt()` instead. Not sure you can have a working system where `crypt()` doesn't work

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
ksubrama: Done :)

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_stats.erl
```diff
@@ -41,15 +41,24 @@ allowed_methods(Req, State) ->
...
+    JsonProvider = {"application/json", to_json},
+    TextProviders = [
       % I think there is a bug in webmachine where it wont allow us to use
       % 'text/plain; version=0.0.4'.
       % TODO: Understand https://github.com/basho/webmachine/blob/develop/src/webmachine_util.erl#L140-L158
       {{"text/plain",[{"version","0.0.4"}]}, to_text},
-      {"text/plain", to_text}
-     ], Req, State}.
+      {"text/plain", to_text}],
+    case wrq:get_qs_value("format", Req) of
```
jaym: cool

oc-chef-pedant/spec/api/stats_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    "pg_stat_tup_fetch" => "COUNTER",
+    "pg_stat_idx_scan" => "COUNTER",
+    "pg_stat_seq_tup_read" => "COUNTER",
+    "pg_stat_seq_scan" => "COUNTER"
+  }
+
+  # Run command once and cache the value.
+  STATS_PASSWORD =
+    begin
+      cmd = Mixlib::ShellOut.new("chef-server-ctl show-secret opscode_erchef stats_password")
```
stevendanna: Alternatively to this, we can assume the password is in the environment:

https://github.com/chef/chef-server/blob/master/oc-chef-pedant/lib/pedant.rb#L79-L80

And then do something like this to feed it into the env at runtime:

https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-ctl-commands/test.rb#L8-L9

Perhaps also putting a guard around the tests to disable them if the secret doesn't exist.

stevendanna: You'd also need to add the stats password here: https://github.com/chef/chef-server/blob/master/dev/defaults.yml#L250-L255

jaym:
👍

PR #1397, opened 2017-10-09
Mitigate Host Header cache-poisoning attacks

stevendanna:
Previously, a request such as:

    curl -H Host:evil.com http://chef-server.dev

would create a response where the Location was set to the user
provided Host header.

    Location: https://evil.com/

If this 301 was erroneously cached, it may then be served to another
user, whose traffic might be redirected to an attacker-controlled
server.

This PR changes the nginx configuration so that we use the server_name
rather than the user-provided Host header when generating redirects.


stevendanna:
Alternative to https://github.com/chef/chef-server/pull/1398 for discussion

stevendanna:
Gonna close this one.

PR #1398, opened 2017-10-09
Mitigate Host Header cache-poisoning attacks

stevendanna:
Previously, a request such as:

    curl -H Host:evil.com http://chef-server.dev

would create a response where the Location was set to the user
provided Host header.

    Location: https://evil.com/

If this 301 was erroneously cached, it may then be served to another
user, whose traffic might be redirected to an attacker-controlled
server.

This PR changes the nginx configuration so that we only respond to
requests with Host headers that match the configured FQDNs for the
machine. Any requests for an unknown host will get a 404.

To ease the transition burden, we also add any IPs on the machine and
localhost to the list of known server names.

The main dis-advantage is that it we can't account for IPs that route
to the machine but aren't accessible via Ohai.

This adds two new configuration options

    delivery['nginx']['strict_host_header'] = true
    delivery['nginx']['use_implicit_hosts'] = true

The old behavior can be restored with

    delivery['nginx']['strict_host_header'] = false
    delivery['nginx']['use_implicit_hosts'] = false

This is similar to the approach taken in chef/automate#1118 (apologies
private). A much more simple solution would be to replace $host with
$server_name in the rewrite/return rule. The concern with this
approach in Automate was that users hitting the site with an IP
address would get redirected to a hostname, causing consternation.
This might not be as much of a concern here because we only use $host
for the HTTP->HTTPS rewrite, so the user could avoid such
consternation by using HTTPS from the outset.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
@@ -105,6 +105,33 @@ http {
     proxy_cache_path  <%= File.join(@dir, "cache", "cookbooks") %> levels=1:2 keys_zone=cookbooks:50m max_size=<%= @cache_max_size %> inactive=600m;
     proxy_temp_path <%= File.join(@dir, "cache-tmp") %>;
 
+    <%- if @strict_host_header %>
+    # Reject requests with unknown Host header
+      <%- if @non_ssl_port %>
+    server {
+      listen <%= @non_ssl_port %> default_server;
+      server_tokens off;
+      server_name _;
+      return 404;
+    }
+      <%- end %>
+
+    server {
+      listen <%= @ssl_port %> ssl;
+      server_tokens off;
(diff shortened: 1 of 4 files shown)


stevendanna:
Alternative: https://github.com/chef/chef-server/pull/1397

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
```diff
@@ -105,6 +105,33 @@ http {
...
+      ssl_certificate <%= @ssl_certificate %>;
+      ssl_certificate_key <%= @ssl_certificate_key %>;
+      ssl_session_timeout 5m;
+      ssl_protocols <%= @ssl_protocols %>;
+      ssl_ciphers <%= @ssl_ciphers %>;
+      ssl_prefer_server_ciphers on;
+
+      server_name _;
+      return 404;
+  }
```
markan: Minor indentation glitches here.

PR #1399, opened 2017-10-10
Update Release_Notes for version 12.17.0

PrajaktaPurohit:


RELEASE_NOTES.md
```diff
@@ -8,6 +8,10 @@ This document contains release notes for the current major release and all patch
 For prior releases,
 see [PRIOR\_RELEASE\_NOTES.md](PRIOR_RELEASE_NOTES.md).
 
+## 12.17.0 (2017-10-10)
```
stevendanna: Minor note: This will actually go out as 12.17.1 or greater because of expeditor.  We can potentially avoid the version bump with the "Version: Skip Bump" tag, but I'm unsure if that will end up skipping any other steps.  Might be worth trying just to find out.

I think in the long run, we might want to reconsider how we manage this file.

RELEASE_NOTES.md
```diff
@@ -8,6 +8,10 @@ This document contains release notes for the current major release and all patch
...
+
+This release updates the version of java to 8u144 to address CVE-2017-3526.
```
srenatus: I might be misreading the CHANGELOG (again!) -- but it looks like the stats endpoint has gone unmentioned so far: [see here](https://github.com/chef/chef-server/blob/master/CHANGELOG.md#merged-pull-requests-1)

stevendanna: @srenatus Good catch!  We DEFINITELY want to talk about the stats feature in the release notes.  It is great.

PrajaktaPurohit: 👍

RELEASE_NOTES.md
```diff
@@ -8,6 +8,16 @@ This document contains release notes for the current major release and all patch
 For prior releases,
 see [PRIOR\_RELEASE\_NOTES.md](PRIOR_RELEASE_NOTES.md).
 
+## 12.17.0 (2017-10-10)
+
+This release has security updates and enhancements.
+ - Update the version of java to 8u144 to address CVE-2017-3526.
+ - Add a `/_stats` endpoint to Chef Server that responds with 
+   stastistics in the json format about
```
stevendanna: Might want to mention it does the prometheus text format as well.

RELEASE_NOTES.md
```diff
@@ -8,6 +8,16 @@ This document contains release notes for the current major release and all patch
...
+     1. The Chef Server instance
+     2. Services required by Chef Server 
+     3. Services colocated with that instance
```
stevendanna: Maybe rather than including this list, we could say something like.  "The new endpoint conforms to Chef RFC 93 (https://github.com/chef/chef-rfc/blob/master/rfc093-server-stats-endpoint.md).  Currently, it exposes statistics about connection pool usage inside erchef, Postgresql, and the Erlang VM.  Further statistics will be added as needed."

PR #1409, opened 2017-10-24
Update the Release Process document

PrajaktaPurohit:


RELEASE_PROCESS.md
```diff
@@ -32,6 +32,13 @@ In order to release, you will need the following accounts/permissions:
 
 - [ ] Announce your intention to drive the release to #cft-announce and #chef-server on slack.
 
+### Getting the build to be released into current
+
+- Update the version file to the version to be released.
```
stevendanna: We only need to do this if we want to bump the minor version rather than just the patch.

RELEASE_PROCESS.md
```diff
@@ -32,6 +32,13 @@ In order to release, you will need the following accounts/permissions:
...
+- Update the release notes with the version, date and context.
```
stevendanna: This duplicates a step down in `Preparing for the release`. I think that we can delete that section completely now that we have this new section.

stevendanna:
@PrajaktaPurohit Thanks for these updates, it would be great to get these in before the next release so we can further refine them from here.  One thing that might be good to do in general is make it more clear that we now promote binary artifacts and don't need "release builds" unless we have last minute changes.

PR #1411, opened 2017-10-24
WIP: Be defensive against cloud plugins that don't report what we expect.

stevendanna:
Looking at the Ohai source code, it appears that cloud_v2 might be the
more consistent interface to use; however, in either case, I'd like to
do some more research to be sure.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/nginx_erb.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/nginx_erb.rb
@@ -1,3 +1,5 @@
+require 'ipaddr'
+
 class NginxErb
 
   attr_reader :node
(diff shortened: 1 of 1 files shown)


tas50:
Just a heads up if we move to Chef 13 the improvements to Ohai 13 help us out a bit here.

stevendanna:
@tas50 Yeah, I think we never would have had this bug in Chef 13.

omnibus/files/private-chef-cookbooks/private-chef/libraries/nginx_erb.rb
```diff
@@ -57,6 +55,29 @@ def ipv6?
     node['private_chef']['nginx']['enable_ipv6']
   end
 
+  def ip_or_nil
```
srenatus: Isn't this missing an `(a)`? Or am I missing something?
```ruby
def ip_or_nil(a)
```

markan:
Now that we've updated Chef and Ohai to 13 is this still needed?

tas50:
This shouldn’t be an issue with modern Ohai.

markan:
Closing because I think it's fixed by the ohai 13 upgrade. Feel free to reopen if I'm wrong.

PR #1413, opened 2017-10-25
Add expeditor subscription to trigger uncached build

jaymalasinha:
Merges to master branch in omnibus-software with updated software definitions will  have expeditor trigger an uncached chef-server build

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -41,6 +41,16 @@ merge_actions:
       only_if_labels:
         - "Expeditor: ACC Only"
 
+# Subscription to workload action of pull request merge to master branch on
+# omnibus-software project will allow us to trigger an uncached omnibus build
+# for chef-server when the omnibus-software has modified software config files.
+subscriptions:
+  # chef/omnibus-software
+  - workload: pull_request_merged:chef/omnibus-software:master:*
+    actions:
+      - built_in:trigger_omnibus_expcache_build:
+          only_if_modified:
+      	    - config/software/*
 # These actions are taken, in the order specified, when an Omnibus artifact is promoted
 # within Chef's internal artifact storage system.
 #


stevendanna:
tenor-203999282

.expeditor/config.yml
```diff
@@ -41,6 +41,16 @@ merge_actions:
...
+# Subscription to workload action of pull request merge to master branch on
+# omnibus-software project will allow us to trigger an uncached omnibus build
+# for chef-server when the omnibus-software has modified software config files.
+subscriptions:
+  # chef/omnibus-software
+  - workload: pull_request_merged:chef/omnibus-software:master:*
+    actions:
+      - built_in:trigger_omnibus_expcache_build
+          only_if_modified:
+      	    - config/software/*
```
stevendanna: INTERESTING! So only_if_modified works on subscribes and the test applies to the subscribed to project?

.expeditor/config.yml
```diff
@@ -41,6 +41,16 @@ merge_actions:
...
         - "Expeditor: ACC Only"
 
+# Subscription to workload action of pull request merge to master branch on
+# omnibus-software project will allow us to trigger an uncached omnibus build
+# for chef-server when the omnibus-software has modified software config files.
+subscriptions:
+  # chef/omnibus-software
+  - workload: pull_request_merged:chef/omnibus-software:master:*
+    actions:
+      - built_in:trigger_omnibus_expcache_build
```
tduffield: You need a `:` at the end of this line to make it valid YAML.

.expeditor/config.yml
tduffield: Everything is actually a subscribes. Merge and Artifact Actions get evaluated down to subscribes. action filters (i.e. only_if_modified) are evaluated against the workload (in this case, the PR merge from omnibus-software). It's still a bit confusing, which is why subscribes isn't documented yet. It's possible the syntax may evolve.

jaymalasinha: @stevendanna Yes.  We have successfully tested the trigger action on any PR merge to omnibus-software with harmony but this is the first update with only_if_modified condition which we thought would be of value to chef-server.   Will keep an eye for the next few merges to verify.

.expeditor/config.yml
jaymalasinha: Updated.

PR #1414, opened 2017-10-26
[SUSTAIN-728] Add /_acl for cookbook_artifacts

nsdavidson:
Currently, knife-ec-backup (via ChefFS) returns a 404 when attempting
to retrieve ACLs on `cookbook_artifacts`.  This change adds the `/_acl`
endpoint to `cookbook_artifacts` so that they can be retrieved and
written.

Signed-off-by: Nolan Davidson <ndavidson@chef.io>


stevendanna:
Note, that approval is on the erlang code :D.  We'll clearly need to sort out what to do on the chef-zero size to get travis happy.

PR #1420, opened 2017-11-03
[rabbitmq] Remove guest user by default

PrajaktaPurohit:


omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
```diff
@@ -115,6 +115,13 @@
     retries 10
   end
 
+  # Remove default guest users
+  execute "#{rmq_ctl} delete_user guest" do
+    environment (rabbitmq_env)
+    user opc_username
+    only_if "#{rmq_ctl} list_users| grep guest", :environment => rabbitmq_env
```
srenatus: What would happen if we executed this unconditionally? Delete might be idempotent? Just wondering if we can save us a call to `rabbitmqctl`, since these are quite expensive 🤔

PrajaktaPurohit: Looks like we can save that call, changes made! thank you :)

srenatus: Hmm have you pushed that? My expectation -- if we can save a call -- would be to have this block as
```ruby
execute "#{rmq_ctl} delete_user guest" do
  environment (rabbitmq_env)
  user opc_username
end
```
So, no `only_if` at all. Does that make sense? 😉

PR #1422, opened 2017-11-07
Bump oc-bifrost-pedant Gemfile.lock

markan:
Testing bumping gemfile.lock helps travis out.


omnibus_overrides.rb
```diff
@@ -4,6 +4,7 @@
 override :chef, version: "v12.19.36"
 override :ohai, version: "v8.23.0"
 override :ruby, version: "2.2.8"
-override :rubygems, version: "2.6.13"
+override :rubygems, version: "2.6.14"
+override :bundler, version: "1.16.0"
```
stevendanna: If we are going to add a new pin here, let's add a comment explaining why.  For better or worse we've been aggressively walking back the pinning in in our builds.

PR #1425, opened 2017-11-13
[reconfigure] Rescue all IPAddr errors

stevendanna:
This bug was discovered in Chef Automate when a similar feature was
added there.  It isn't clear what values of `server_name` lead to
this (possibly `nil` or an Integer).


stevendanna:
cc @ryancragun Do you remember what input caused you to find this?  I only ask because it feels like most values that would cause this problem would also cause problems in other code later on.

ryancragun:
Yohai, IIRC sometimes some hostnames would cause `IPAddr.new()` to throw an `AddressFamilyError`, so I just made it generic to handle all the cases.

Issue #972, opened 2016-10-17
chef server welcome page cannot be disabled


theckman:
Hello,

I'm running an open source version of the Chef server, with nodes that connect to the Chef server over the WAN. Current versions (`12.x+`?) of the Chef server have a web page that's displayed to the world indicating it's a Chef server. I've included the contents of the page below.

I've done some digging, and I haven't found a way to disable it, without the equivalent of `echo '' >` in to the `index.html` file. I hope I haven't missed the docs, if so I apologize for the waste of time.

Would it be possible to add an option to disable this landing page? Although I suppose one could argue I'm asking for a layer of security by obscurity, I'd prefer not to announce to the world that this endpoint is a Chef server.

My initial idea is that blank `index.html` page being rendered during the `chef-server-ctl reconfigre` run, based on an attribute in the Chef server config file, would be one way to solve this.

Thank you for your time!

```
Are You Looking For the Chef Server?

Hello! It looks like you were trying to browse to your Chef Server but you haven't installed the Management Console.

If you prefer, you can access the server programmatically with the Chef Server API. To learn how to do this, head over to our API Documentation pages.

If you do want to use your browser to manage your server, then install the Management Console. It's free for up to 25 nodes.

One way to install the Management Console is to log in as root and type this command:

 # chef-server-ctl install chef-manage 

Alternatively, you can download the Management Console as an installable package from the Management Console downloads page.

You may also be interested in our other add-ons. These are available as packages from our Chef Server downloads page or you can type chef-server-ctl install. Without arguments, the command lists the packages available for installation.
```

stevendanna:
👍 on being able to turn this page off.  I think it is helpful to have on by default for people who get confused by the lack of a web interface, but we should be able to toggle it off.

ketank-new:
one way to also to obtain the chef-server landing page is as follows

1) install chef-server rpm
2) run the chef-server on specific port number <port no>
3) re check if the chef-server process is running in process manager using 'ps' command
4) to obtain the landing page use ' wget localhost:<port no>' .. you should get indx.html


Let me know if this helps

theckman:
@ketank-new I'm not sure your comment is relevant. This ticket is asking for a way to *disable* the page from being displayed by the Chef server. How would downloading it from a Chef server help you not display it on the public Internet?

PR #1428, opened 2017-11-28
Disable Welcome Page

lancewf:
This allows a user to disable chef server's welcome page.
When the welcome page is disabled, a 404 will be returned when a request
is made that does not match any needed pattern.

To disable the welcome page add the below line to the chef-server.rb
file.

`nginx['show_welcome_page'] = false`

The change is only a cookbook change that modifies nginx's config file

This would close https://github.com/chef/chef-server/issues/972

I did not find a problem when installing chef manage with the welcome page disabled. Are there other addons or configurations that I would need to test? 

I am looking into adding a few tests.  

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
@@ -44,19 +44,10 @@
     # Whitelist the docs necessary to serve up error pages and friendly
     # html to non-chef clients hitting this host.
     location ~ "^/[0-9]{3,3}\.(json|html)|favicon.ico|index.html$" {
-      <% unless node['private_chef']['nginx']['show_welcome_page'] %>
-        return 404;
-      <% end -%>
     }
     location "/css/" {
-      <% unless node['private_chef']['nginx']['show_welcome_page'] %>
-        return 404;
-      <% end -%>
     }
     location "/images/" {
-      <% unless node['private_chef']['nginx']['show_welcome_page'] %>
-        return 404;
-      <% end -%>
     }
 
     location /version {
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
@@ -44,10 +44,19 @@
     # Whitelist the docs necessary to serve up error pages and friendly
     # html to non-chef clients hitting this host.
     location ~ "^/[0-9]{3,3}\.(json|html)|favicon.ico|index.html$" {
+      <% unless node['private_chef']['nginx']['show_welcome_page'] %>
+        return 404;
+      <% end -%>
     }
     location "/css/" {
+      <% unless node['private_chef']['nginx']['show_welcome_page'] %>
+        return 404;
+      <% end -%>
     }
     location "/images/" {
+      <% unless node['private_chef']['nginx']['show_welcome_page'] %>
+        return 404;
+      <% end -%>
     }
 
     location /version {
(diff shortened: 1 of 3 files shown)


omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -44,10 +44,19 @@
     # Whitelist the docs necessary to serve up error pages and friendly
     # html to non-chef clients hitting this host.
     location ~ "^/[0-9]{3,3}\.(json|html)|favicon.ico|index.html$" {
```
srenatus: [nit] Looking at this, I wonder if we can't combine those last two blocks into the RE-block here 🤔

stevendanna: [NOTE]  Note that this location block is doing all the heavy lifting of disabling the welcome page.  Even though this location block won't actually match most initial requests that would result in the welcome page, those requests eventually get here: https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/dispatch.lua.erb#L49  The ngx.exec does an internal redirect to index.html which then hits the 404.

lancewf:
Do I need to increase the version number on the metadata.rb for this update? 
https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/metadata.rb#L7

lancewf:
@stevendanna Thank you for pointing out that lua file. I moved the code to disable the welcome page there. 

I could not figure out a good way to write tests for this change.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/dispatch.lua.erb
```diff
@@ -44,9 +44,15 @@ if mode == "api" then
...
-      -- more friendly to web browsers.
-      return ngx.exec("/index.html")
+      <% if node['private_chef']['nginx']['show_welcome_page'] %>
+        -- the request did not originate with a chef client, we'll give something
+        -- more friendly to web browsers.
+        return ngx.exec("/index.html")
+      <% else %>
+        -- return 404 because the welcome page has been disabled
+        ngx.status = ngx.HTTP_NOT_FOUND
+        ngx.exit(ngx.HTTP_OK)
```
srenatus: I don't know much about `ngx_lua`, but [these docs](https://github.com/openresty/lua-nginx-module#ngxexit) make me think we might be able to go with a single `ngx.exit(ngx.HTTP_NOT_FOUND)` call. But you've potentially tried that already, if so please ignore `:)`

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/dispatch.lua.erb
```diff
@@ -44,9 +44,15 @@ if mode == "api" then
         ngx.exit(ngx.HTTP_OK)
       end
     else
-      -- the request did not originate with a chef client, we'll give something
-      -- more friendly to web browsers.
-      return ngx.exec("/index.html")
+      <% if node['private_chef']['nginx']['show_welcome_page'] %>
```
srenatus: [nit] we could use `-%>` here for less newlines but also it doesn't really matter

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/scripts/dispatch.lua.erb
lancewf: @srenatus Yeah, I did try that. When you use `ngx.exit(ngx.HTTP_NOT_FOUND)` only you get back the below page
screen shot 2017-11-29 at 4 33 53 pm

If you leave it the way it is you get back nothing in the response body with the status of 404. I think this would be better to not return anything in the response body. This would better hide the fact that this is a chef-server endpoint.

stevendanna:
>Do I need to increase the version number on the metadata.rb for this update?
https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/metadata.rb#L7

Nope, don't worry about it.

PR #1429, opened 2017-11-28
[partybus] Add `infer` command to fix missing migration-level file

stevendanna:
Some users may be restoring a backup of external data sources and no
longer have a migration-level file. In this case, a `chef-server-ctl
upgrade` will fail because it will attempt to run all migrations,
starting from the beginning.

The `partybus infer` command attempts to reconstruct the
migration-level based on what it sees in the database.

For migrations that deploy a sqitch tag, we infer the migration has
run from the presence of the sqitch tag in the sqitch table.

Non-sqitch migrations are less reliable to detect and in most cases we
do not attempt to detect them since most should be idempotent.

Further, we assume the following:

- Migrations from versions prior to Enterprise Chef 11.1 are not
  supported and thus the lowest possible migration level we should
  ever infer is 1.13

- The Chef Server 12.0.0 upgrade deployed migrations 1.14 through
  1.19.  We will assume that our system has either completely applied
  this set of upgrades (and is thus on 1.19 or greater) or hasn't (and
  is thus on 1.13)

For the migrations we will currently infer, the following identifies
how we infer them or why we think it is safe to rerun.

# Migration-level Detection Methods

- `1.34`: Detect sqitch tag

- `1.33`: Detect sqitch tag

- `1.32`: Safe to rerun: simple service restart

- `1.31`: Safe to rerun: rabbitmqctl change_password to the /existing/ password

- `1.30`: Custom check function since migration appears unsafe to rerun.

- `1.29`: Likely safe to rerun: ACL addition should be idempotent in
       bifrost. We could be overwriting a user's intention to /remove/
       the server-admins from having READ on their user

- `1.28`: Likely safe to rerun: server-admins group isn't recreated if it
       exists:

     https://github.com/chef/chef-server/blob/master/src/chef-mover/src/mover_server_admins_global_group_callback.erl#L74-L76

     ACL modifications should be idempotent:
     https://github.com/chef/chef-server/blob/master/src/chef-mover/src/mover_server_admins_global_group_callback.erl#L79-L96

- `1.27`: Detect sqitch tag

- `1.26`: Safe to rerun: Only un-renamed groups are selected for
       migration:

     https://github.com/chef/chef-server/blob/master/src/chef-mover/src/mover_global_admins_rename_callback.erl#L48-L56

- `1.25`: Safe to rerun: Group member addition should be
       idempotent. Since global groups aren't user-modifiable, we
       can't be modifying user-altered data

- `1.24`: Safe to rerun: rm -rf on a cache directory

- `1.23`: Safe to rerun: Only orgs without policy containers are
       selected:

     https://github.com/chef/chef-server/blob/master/src/chef-mover/src/mover_policies_containers_creation_callback.erl#L76-L81

- `1.22`: Detect sqitch tag

- `1.21`: Detect sqitch tag

- `1.20`: Detect sqitch tag

- `1.19`: Detect sqitch tag of migration 1.16, assume all of 1.14-1.19
       have completed.  See note on this above.

- `1.13`: Detect sqitch tag

### DIFF
--- a/omnibus/files/private-chef-upgrades/001/030_actor_keys_access_group.rb
+++ b/omnibus/files/private-chef-upgrades/001/030_actor_keys_access_group.rb
@@ -17,3 +17,38 @@ define_upgrade do
     stop_services(["opscode-chef-mover"])
   end
 end
+
+define_check do
+  # If there is at least one group with public_key_read_access group,
+  # assume the migration has run. This might not be true, but if there are
+  # no existing read_access_groups, we believe the migration should be
+  # safe to rerun.
+  #
+  # We can erroneously mark this migration as "run" in the following cases:
+  #
+  # (1) The user has created a group called public_read_access
+  #     unrelated to this migration.
+  #
+  # (2) If the user upgraded, failed to apply this upgrade, but then
+  #     somehow was able to create *new* organization using the
+  #     newer version of erchef.
+  #
+  # We can erroneously mark this migration as "not run" in the following cases:
+  #
+  # (1) The migration ran but this installation has no organizations.
+  #     Re-running the migration will have no ill effect as it
+  #     operates on a per-organization basis.
+  #
+  # (2) The migration ran but then the user deleted all of the
+  #     public_read_access groups. Re-running the migration should not
+  #     cause an error but the user might be surprised to find those
+  #     groups re-appear.  We think this case is rare.
+  #
+  begin
+    c = pg_conn_for('oc_erchef')
+    r = c.exec("SELECT COUNT(*) FROM groups WHERE name = 'public_key_read_access'")
+    r.first['count'].to_i > 0
+  ensure
+    c.close if c
+  end
+end
--- a/omnibus/partybus/bin/partybus
+++ b/omnibus/partybus/bin/partybus
@@ -24,6 +24,7 @@ Usage: partybus ACTION
 
 Actions:
   init        Set the initial migration level
+  infer       Infer the current migration level
   upgrade     Run through the pending upgrades
   help        Print this help message
 EOU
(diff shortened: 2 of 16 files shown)


omnibus/partybus/lib/partybus/dsl_runner.rb
```diff
@@ -10,19 +10,53 @@ def initialize(migration_file)
...
   end
 
+  def run_check
+    if @check_block
+      Partybus::UpgradeAPI::V1.new(&@check_block).eval_result
+    else
+      # NOTE(ssd) 2017-11-28: The assumption here is that a migration
+      # is safe to re-run unless otherwise specififed.
+      # Counter-intuitively, we define checks for the safest
+      # migrations to re-run (sqitch) becuase they are easy to check
```
srenatus: [typo] `s/uase/ause/`

omnibus/partybus/lib/partybus/migration_api/v1.rb
```diff
@@ -114,6 +116,27 @@ def run_sqitch(target, service, opts = {})
         run_command(command, env: {"PGPASSWORD" => options[:password]})
       end
 
+      CHECK_SQITCH_SQL="SELECT 1 FROM sqitch.tags WHERE tag=$1"
+      def check_sqitch(target, service, opts = {})
+        options = default_opts_for_service(service).merge(opts)
+        c = pg_conn_for(service, opts)
+        result = c.exec(CHECK_SQITCH_SQL, [target])
+        puts result.cmd_tuples
```
srenatus: [nit] probably due to this PR's WIP-iness? ;)

stevendanna:
> Would you think that this is could be reliable enough to get rid of the migration level file altogether?

That is an interesting question.  I'm not sure.  For anyone who successfully upgrades to the latest version, we could make it reliable going forward by requiring that every migration has a well-defined check function.

Currently, the approach I am taking is to move that migration state into a database table to make it less likely that it gets lost, but it is interesting to think about whether we could remove the state tracking completely.

omnibus/files/private-chef-upgrades/001/030_actor_keys_access_group.rb
```diff
@@ -17,3 +17,35 @@
...
+  # (1) The user has created a group called public_read_access unrelated to this migration.
+  #
+  # (2) If the user upgraded, failed to apply this upgrade, but then
+  #     somehow was able to create *new* organization using the
+  #     newer version of erchef.
+  #
+  # We can erroneously mark this migration as "not run" in the following cases:
+  #
+  # (1) The migration ran but this installation has no organizations.
+  #     Re-running the migration will have no ill effect as the
```
danielsdeleo: you accidentally the rest of the sentence 😝

omnibus/partybus/bin/partybus
```diff
@@ -43,9 +44,97 @@ def load_migrations
...
+    migration_16 = migrations.find {|m| m.major == 1 && m.minor == 16}
+    if migration_16.run_check
+      log "Inferring migration level of Chef Server 12.0.0 (1.19) from presence of migration 16 and absence of migration 20"
+      migration_19 = migrations.find {|m| m.major == 1 && m.minor == 19}
+      write_migration_state(migration_19)
+      exit(0)
+    end
+
+    # Since migrations 14-19 shipped as a set in Chef Server 12.0.0
+    # and since we don't suppot upgrades from versions before
```
danielsdeleo: *support*

omnibus/partybus/lib/partybus/dsl_runner.rb
```diff
@@ -10,19 +10,53 @@ def initialize(migration_file)
     @file_path     = migration_file.path
   end
 
-  def run
+  def load
```
ryancragun: [nit] I'm generally not a fan of overloading methods from `Kernel`, but this seems okay.

stevendanna: I can change it to `load_migration`.  That'll also fix the syntax highlighting :D

omnibus/partybus/bin/partybus
```diff
@@ -43,9 +44,97 @@ def load_migrations
...
+      applied_version = m
+      break
+    else
+      log "#{m}: NOT APPLIED"
+    end
+
+    # A number of migrations before 1.20 (the Chef Server 12.0) are
+    # mover migrations that are harder to check for.  If we make it
+    # all the way to here, we stop checking and hope one of the two
+    # special cases below apply.
```
marcparadise: Looking to see if couch is still running/installed might give us a more definitive answer for minimum level that has or hasn't been applied.

omnibus/partybus/bin/partybus
```diff
@@ -73,7 +162,18 @@ File.open(Partybus.config.migration_state_file, File::RDWR | File::CREAT) do |f|
     when "init"
       migrations = load_migrations
       set_initial_migration_state(migrations)
+    when "infer"
+      force = ARGV[1] == "force"
```
marcparadise: Should this be `ARGV.include?"force"`to account for other possible flags?

omnibus/partybus/bin/partybus
stevendanna: Unfortunately, I'm not sure it does because maybe they didn't run cleanup.  I suppose if we /don't/ see couchdb then we have more evidence that we go through migrations 14-19 (right now we check if migration 16 is completed and use that to assume that they successfully upgraded to 12.0.0).  If we /do/ see couchdb and we assume they would have run cleanup, then we have evidence that they haven't upgraded to 12, but I am not sure if I trust that more or less than just looking at migration 16.

I think I'm OK with this for now because it makes mostly conservative judgements. We are taking the following leaps of faith: (1) My reading of the idempotency of some of the mover migrations and (2) that most likely we aren't dealing with a half-upgraded Chef Server 12.0.0 installation.  I'm not sure looking at couchdb changes that substantially unlesss I'm missing something.

PR #1430, opened 2017-11-30
[partybus] Store migration-level in postgresql database

stevendanna:
The migration-level was previously stored in a flat-file on disk.
This file then had to be specifically accounted at multiple points in
a Chef Server installations life cycle.

By placing the migration-state in the database, we hope to decrease
problems related to incorrectly synced migration-level files.

Both the /infer/ and /upgrade/ commands will import existing
migration-level files into the database if the database is not already
initialized.

TODO:

- [ ] For now I've treated PostgreSQL as a key-value store and didn't
  change what metadata we store. We should decide now if we think we
  want to do something fancier.

### DIFF
--- a/omnibus/partybus/bin/partybus
+++ b/omnibus/partybus/bin/partybus
@@ -6,6 +6,7 @@ $:.unshift File.expand_path('../../lib', __FILE__)
 require 'partybus'
 require 'partybus/dsl_runner'
 require 'partybus/migrations'
+require 'partybus/migration_state'
 
 # load the config from a file
 require File.expand_path('../../config.rb', __FILE__)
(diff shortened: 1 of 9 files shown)


stevendanna:
@chef/chef-server-maintainers We've talked about doing this a few times.  I worked on it when working on the migration-level init PR but pulled it out to make that change safe to apply.  This doesn't feel particularly urgent at the moment, so I would definitely love thoughts here.

omnibus/partybus/bin/partybus
```diff
@@ -148,41 +126,38 @@ def do_upgrade(migration_state, migrations)
...
-      migration_state = load_current_migration_state
-      migrations = load_migrations
-      do_upgrade(migration_state, migrations)
-    when "help"
-      print_usage_and_exit(0)
-    else
-      print_usage_and_exit
-    end
+# Obtain the file lock on the migration state file
+migration_state = Partybus::MigrationState.new
```
stevendanna: We probably don't want to do this just to print help!

rhass:
@PrajaktaPurohit This may or may not make your life better with chef-backend.... but in my experience with chef-server it's a good thing for us to test and land.

@btm Any objections to me adding a card to test this out?

PR #1432, opened 2017-12-05
Change elasticsearch standard analyzer to keyword analyzer

piyushawasthi:
Added `keyword analyzer` to prevent escaping forward slashes in elasticsearch.

**Description:**
By default elasticsearch using `standard analyzer` divides text into terms by ` Unicode Text Segmentation algorithm` that basically escaping forward slashes so I am changing that to `keyword analyzer` that's basically a “noop” analyzer that accepts whatever text it is given and outputs the exact same text as a single term.

Reference URL - https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-analyzers.html

**Local Testing Output**
Pass:
```
$ bundle exec knife search node 'stack_id:arn\:aws\:cloudformation\:eu-west-1\:569478613970\:stack/*'
$ bundle exec knife search node 'stack_id:arn\:aws\:cloudformation\:eu-west-1\:569478613970\:stack\\/*'
$ bundle exec knife search node 'stack_id:arn\:aws\:cloudformation\:eu-west-1\:569478613970\:stack\\//*'
```
Fail:
```
$ bundle exec knife search node 'stack_id:arn\:aws\:cloudformation\:eu-west-1\:569478613970\:stack\//*'
$ bundle exec knife search node 'stack_id:arn\:aws\:cloudformation\:eu-west-1\:569478613970\:stack\/*'
```
Testing still In-progress will update after some more test.

Signed-off-by: piyushawasthi <piyush.awasthi@msystechnologies.com>


PrajaktaPurohit:
Not sure if we would need this for our case - but could be something to consider: https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html

PrajaktaPurohit:
This looks good but what I do not understand is why are we not able to escape '/' for elasticsearch.

@stevendanna and I looked some at the search code and here are a few things that we think might offer a solution that might be closer to fixing this issue.

Some background: The way our search code is structured is that chef_index handles dealing with the search backend (default solr and also support elasticsearch). We use chef_lucene grammar to parse out the query and send it to the respective provider. We see an issue with the way our grammar is currently structured i.e: https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_index/src/chef_lucene.peg#L62 says that a valid_letter is any combination of start_letter followed by zero or more occurrences of alpha-numeric value, some special characters (*?@_+./-) OR  zero or more occurrences of \ followed by a special character.
But if we look at the definition of special_characters (https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_index/src/chef_lucene.peg#L68) it does not include a '/'

TLDR: https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_index/src/chef_lucene.peg#L62 and https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_index/src/chef_lucene.peg#L68 make it such that we cannot escape a '/'

So a way to go forward would be:

include '/' in special characters like so:
```
special_char <- '[' / ']' / '\\' / '"' / [!(){}^~*?:/] ~;
```

That way to do that would be:
1. Change the existing chef_lucene.peg file. We would have to check with that is to make sure that escaping / ('\/' in the search query) does not break solr.

2. If 1. breaks solr, then the path forward might be to create 2 peg files. The existing (chef_lucene.peg) one for chef_lucene_solr.peg and the other including '/' in the special_char as mentioned above as chef_lucene_elasticsearch.peg. 
https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_index/src/chef_index_query.erl#L27 seems to indicate that we have knowledge of the search provider, based on that https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_index/src/chef_index_query.erl#L95 would parse either the chef_lucene_solr or chef_lucene_elasticsearch grammar.

I have not read enough about elastic search to know if in addition to this we need the keyword analyzer, but this change should take us beyond the current "ERROR: knife search failed: invalid index name or query"

piyushawasthi:
Hi @prajakta, as per your comment I have updated code for `special_char <- '[' / ']' / '\\' / '"' / [!(){}^~*?:/] ~;` in `https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_index/src/chef_lucene.peg#L68` after that I have index data as
```
curl -XPUT 'elasticsearch:9200/_settings' -H "Content-Type:application/json" -d '{
    "index.search.slowlog.threshold.query.debug" : "1ms",
    "index.search.slowlog.threshold.fetch.debug": "1ms",
    "index.indexing.slowlog.threshold.index.info": "1ms"
}'
```
getting `{"acknowledged":true}` but I am getting same error.
```
$ bundle exec knife search node 'stack_id:arn\:aws\:cloudformation\:eu-west-1\:569478613970\:stack/*'
ERROR: knife search failed: invalid indexed name or query
```
*Chef Server Log*
```
==> /var/log/opscode/nginx/access.log <==
192.168.33.1 - - [08/Dec/2017:16:46:56 +0530]  "GET /organizations/test/search/node?q=stack_id:arn%5C:aws%5C:cloudformation%5C:eu-west-1%5C:569478613970%5C:stack/*&start=0&rows=1000 HTTP/1.1" 400 "0.021" 41 "-" "Chef Knife/13.6.10 (ruby-2.4.0-p0; ohai-13.6.0; x86_64-linux; +https://chef.io)" "127.0.0.1:8000" "400" "0.021" "13.6.10" "algorithm=sha1;version=1.1;" "admin" "2017-12-08T11:16:56Z" "2jmj7l5rSw0yVb/vlWAYkK/YBwk=" 1131

==> /var/log/opscode/opscode-erchef/erchef.log <==
2017-12-08 16:46:56.757 [error] {<<"method=GET; path=/organizations/test/search/node; status=400; ">>,"Bad Request"}

==> /var/log/opscode/opscode-erchef/crash.log <==
2017-12-08 16:46:56 =ERROR REPORT====
{<<"method=GET; path=/organizations/test/search/node; status=400; ">>,"Bad Request"}

==> /var/log/opscode/opscode-erchef/current <==
2017-12-08_11:16:56.76729 [error] {<<"method=GET; path=/organizations/test/search/node; status=400; ">>,"Bad Request"}

==> /var/log/opscode/opscode-erchef/requests.log.1 <==
2017-12-08T11:16:56Z erchef@127.0.0.1 method=GET; path=/organizations/test/search/node?q=stack_id:arn%5C:aws%5C:cloudformation%5C:eu-west-1%5C:569478613970%5C:stack/*&start=0&rows=1000; status=400; req_id=g3IAA2QAEGVyY2hlZkAxMjcuMC4wLjEBAAHjDAAAAAEAAAAA; org_name=test; msg={solr_400,"/chef/_search"}; couchdb_groups=false; couchdb_organizations=false; couchdb_containers=false; couchdb_acls=false; 503_mode=false; couchdb_associations=false; couchdb_association_requests=false; req_time=19; rdbms_time=0; rdbms_count=3; solr_time=14; solr_count=1; user=admin; req_api_version=1; 
```

PR #1433, opened 2017-12-06
[SUSTAIN-600] Make request logging configurable

PrajaktaPurohit:
This makes request logging configurable for oc_erchef, oc_bifrost and bookshelf with individual switches.

```root@api:/vagrant# cat /etc/opscode/chef-server.rb 
opscode_erchef['enable_request_logging'] = false
```

ensure that there are no request logs generated:
```
root@api:/vagrant# ls -l /var/log/opscode/opscode-erchef/
total 840
-rw-r--r-- 1 root    root        38 Dec  6 20:32 config
-rw-r--r-- 1 opscode opscode 411921 Dec  6 21:57 crash.log
-rw-r--r-- 1 opscode opscode 220151 Dec  6 21:57 current
-rw-r--r-- 1 opscode opscode 209261 Dec  6 21:57 erchef.log
-rw------- 1 opscode opscode      0 Dec  6 20:32 lock
drwxr-x--- 2 opscode opscode   4096 Dec  6 21:30 sasl
-rw-r--r-- 1 opscode opscode      0 Dec  6 21:30 sasl-error.log
```

All tests pass.

Green build: http://wilson.ci.chef.co/job/chef-server-trigger-ad_hoc/37/downstreambuildview/

TODO: 
[x] verify with knife
[x] upload cookbooks
[x] bootstrap a node

Signed-off-by: Prajakta Purohit <prajakta@opscode.com>

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/bookshelf.config.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/bookshelf.config.erb
@@ -122,15 +122,15 @@
 <% end -%>
  {webmachine, [
           {log_handlers, [
-%%% Log handling is disabled, because it is redundant (nginx also logs requests)
-%%% If debug logging is needed, this can be uncommented to start logging somewhat verbose logs
-%%%              {oc_wm_request_logger, [
-%%%                       {file, "<%= File.join(@log_directory, 'requests.log') %>"},
-%%%                       {file_size, 100},  %% Size in MB
-%%%                       {files, 5},
-%%%                       {annotations, [user, req_id]}
-%%%                       ]
-%%%               }
+          <% if node['private_chef']['bookshelf']['enable_request_logging'] %>
+              {oc_wm_request_logger, [
+                       {file, "<%= File.join(@log_directory, 'requests.log') %>"},
+                       {file_size, 100},  %% Size in MB
+                       {files, 5},
+                       {annotations, [user, req_id]}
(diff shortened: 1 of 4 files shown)


PrajaktaPurohit:
Would it be useful to have like a master[webmachine][enable_request_logging] attribute that can superseed the existing individual ones?

I kept going both ways about it and have not added one in.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/bookshelf.config.erb
```diff
@@ -124,13 +124,15 @@
           {log_handlers, [
 %%% Log handling is disabled, because it is redundant (nginx also logs requests)
 %%% If debug logging is needed, this can be uncommented to start logging somewhat verbose logs
+%%%          <% if node['private_chef']['webmachine']['bookshelf']['enable_request_logging'] %>
```
stevendanna: Rather than keeping this commented, what do you think about uncommenting it, changing the default config value to false, an moving the explanation over to the attribute file.

PrajaktaPurohit: yesss!!!

PrajaktaPurohit:
That was another thing I was pondering over. but for example with bookshelf['enable_request_logging'], I would expect the attribute to be used under bookshelf rather than webmachine; hence the inclusion of the new key there.

PR #1440, opened 2017-12-19
Upgrade to RAML 1.0

thommay:
This lets us at least build with a fresh install of RAML2html


thommay:
I also wired in /universe since I missed that…

PR #1442, opened 2017-12-28
[SUSTAIN-807] Clean up oc_id PID on restart

nsdavidson:
When a Chef Server host is taken down hard, the PID file for the oc_id Rails
app does not cleaned up. Upon restart of the system, the PID stored in that file is
often assigned to another service. When this happens, oc_id fails to
restart, because Rails/Rack sees there's a process with the PID stored
in the oc_id PID file that it does not own. This sends oc_id into a
restart loop with this log message:

```
A server is already running. Check
/opt/opscode/embedded/service/oc_id/tmp/pids/server.pid.
```

Cleaning up the pidfile in the runit script before launching the rails
app prevents this.

Signed-off-by: Nolan Davidson <ndavidson@chef.io>


PrajaktaPurohit:
Looks good to me!

PR #1469, opened 2018-02-12
removing forcing of the doc formatter

lamont-granquist:
the output is so long that it blows out travis's line limits and the way
it gets forced here is difficult to override.

it can go into .rspec files in projects to bring it back if people
really like it by default, but its very wrong to have it here.

Signed-off-by: Lamont Granquist <lamont@scriptkiddie.org>

### DIFF
--- a/oc-chef-pedant/lib/pedant/config.rb
+++ b/oc-chef-pedant/lib/pedant/config.rb
@@ -1,4 +1,4 @@
-# Copyright: Copyright (c) 2012 Opscode, Inc.
+# Copyright: Copyright (c) 2012-2018, Chef Software Inc.
 # License: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


lamont-granquist:
looks like master isn't actually running pedant tests correctly at all even though they're green.

in playing around with this it looks like the tests seem to require the retry formatter which seems to force the doc formatter somehow and any way i beat on it, i just make the tests explode in red.

if this actually worked it'd be nice to do this here, since doing it downstream seems to be vaguely impossible, at least i've hit my limit of trying to sort out what is going wrong.

rhass:
@lamont-granquist Your comment about "it looks like the tests seem to require the retry formatter which seems to force the doc formatter somehow and any way i beat on it" leads me to believe this PR doesn't do anything. Am I understanding you correctly?

lamont-granquist:
no, its that the forcing of the doc formatter seems to break the retry stuff.  i could not sort out what hackery the retry formatter was doing and if it could be made to be compatible with the doc formatter.

lamont-granquist:
(and didn't look into if this was now backed into rspec or if there was a better config to use to get retries)

PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/542#1fdef55b-256a-4358-a0c5-41dcf1b5a104

PR #1472, opened 2018-02-14
Adding Habitization - this PR supersedes PR#1416

jeremymv2:
teamwork

This is the result of a tremendous team effort spent over the last couple of months. I'm just trying to drive it home with this PR. It is a branch copy of #1416; rebased and with all commits squashed and authors
attributed below. #1416 can be left intact for posterity and reference of original work with all the commits.

Habitat pkgs in this branch have been tested and are functional.

The vast majority of changes in this PR are for Habitat _only_ and will not affect any current functionality.  The few exceptions to this are some minor changes to files in:
 * omnibus/files/private-chef-ctl-commands/
 * oc-chef-pedant/lib/

Co-authored-by: Elliott Davis <edavis@chef.io>
Co-authored-by: Irving Popovetsky <irving@chef.io>
Co-authored-by: Mark Anderson <mark@chef.io>
Co-authored-by: Nolan Davidson <ndavidson@chef.io>
Co-authored-by: Thomas Cate <thomascate@gmail.com>

Signed-off-by: Jeremy J. Miller <jm@chef.io>

PR #1475, opened 2018-02-21
habitat mode for chef-server-ctl command

jeremymv2:
This PR depends on #1472 

Here we're adapting the chef-server-ctl commands and files to allow for a new Habitat mode.

It should just work in either mode - like magic!

magic


jeremymv2:
👍

PR #1478, opened 2018-02-28
[upgrades] Don't early return in upgrade 1.31

stevendanna:
Calling return from a migration callback results in:

   LocalJumpError: unexpected return


stevendanna:
This is untested, but was the root problem in a recent support ticket.

PR #1479, opened 2018-03-16
WIP Fix erlang dependency loading.

markan:
Work in progress rework of the dependency loading. This breaks the
contract around the library loading, and I'm still sorting that
out. Putting this out there for commentary on the general approach.

Rebar3 now provides a feature for overriding the loading of
dependencies. If a dep is provided in the _checkouts directory, rebar3
will build it in preference to the version loaded with get-deps.

This rewrite uses a soft link from _checkouts to /host/external-deps;
the dev at this time will have to have a link on the host side to
their checkout of the dep.

I'm not quite sure how the old code was supposed to work with loading
things into /host/*, as sync was pretty aggressive about stomping on
things there.

### DIFF
--- a/dev/defaults.yml
+++ b/dev/defaults.yml
@@ -77,6 +77,7 @@ vm:
        - rel/
        - _rel/
        - _build/
+       - _checkouts/
        - chef-mover/rel/mover/
        - ebin/
        - .eunit/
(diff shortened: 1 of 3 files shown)


markan:
I need to figure out if this is still needed/useful.

stevendanna:
I'm going to close this for now.  My guess is dev-vm is going to need some re-vamping in the near future and we can pick it back up as part of that work.  I'll leave the branch around as it may be helpful.

PR #1482, opened 2018-04-05
Ma/update all the things

markan:
This has two major parts:
+ Bundle update Gemfile locks
+ Update to use Chef 13, including a bunch of cookbook and tool fixes.


omnibus/config/software/cleanup.rb
```diff
@@ -7,6 +7,10 @@
   # Delete cached .gem files and git checkouts
   delete "#{install_dir}/embedded/lib/ruby/gems/2.2.0/cache/*.gem"
   delete "#{install_dir}/embedded/service/gem/ruby/2.2.0/cache/*.gem"
+  delete "#{install_dir}/embedded/lib/ruby/gems/2.3.0/cache/*.gem"
+  delete "#{install_dir}/embedded/service/gem/ruby/2.3.0/cache/*.gem"
+  delete "#{install_dir}/embedded/lib/ruby/gems/2.4.0/cache/*.gem"
+  delete "#{install_dir}/embedded/service/gem/ruby/2.4.0/cache/*.gem"
```
thommay: could this just be `.../ruby/2.*/cache/*.gem` so we don't have to update every ruby version bump?

PR #1483, opened 2018-04-08
lock to chef 13.x for habitat pkgs

jeremymv2:
Additionally, some build issues are fixed and the docker-compose
simplified.


jeremymv2:
@markan awesome! I would like to get this in so that we can build the hab pkgs and keep their releases/tags in sync.

irvingpop:
👍 ship it!

PR #1485, opened 2018-04-10
Added chef-server-ctl version functionality

thomascate:
Added basic functionality for chef-server-ctl to report the version from the version-manifest.txt file. If it detects it's running in Habitat it will derive the version from the IDENT file in the hab package.

Signed-off-by: Thomas Cate <tcate@chef.io>

### DIFF
--- a/omnibus/files/private-chef-ctl-commands/version.rb
+++ b/omnibus/files/private-chef-ctl-commands/version.rb
@@ -1,0 +1,35 @@
+#
+# Copyright:: Copyright (c) 2018 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+require 'json'
+
+add_command_under_category "version", "general", "Display current version of Chef Server.", 2 do
+
+  begin
+    # detect if running as a habitat service
+    if File.exist?('/hab/svc/chef-server-ctl/PID')
+      ident_file = File.read('../IDENT')
+      version = "chef-server #{ident_file.split('/')[2]}"
+    else
+      version = JSON.parse(File.read('/opt/opscode/version-manifest.json'))['build_version']
+    end
+
+    puts version
+  rescue Errno::ENOENT => e
+    puts "Error determining version!"
+    puts "#{e.message}"
+    exit(1)
+  end
+
+end


omnibus/files/private-chef-ctl-commands/version.rb
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+add_command_under_category "version", "general", "Display current version of Chef Server.", 2 do
+
+  # detect if running as a habitat service
+  if File.exist?('/hab/svc/chef-server-ctl/PID')
+    #code for populating version through /version endpoint
```
jeremymv2: left over comment artifact perhaps?

omnibus/files/private-chef-ctl-commands/version.rb
```diff
@@ -0,0 +1,30 @@
...
+    puts "running in hab"
```
jeremymv2: nit: I don't really think this is necessary - seems like a left over debugging statement perhaps?

omnibus/files/private-chef-ctl-commands/version.rb
```diff
@@ -0,0 +1,30 @@
...
+    ident_file = File.read('../IDENT')
+    version = "chef-server #{ident_file.split('/')[2]}"
+  else
+    version = File.readlines('/opt/opscode/version-manifest.txt')[0]
+  end
+
+  puts version
+
```
jeremymv2: nit: perhaps put an error handler after line 29 to better handle any `Errno::ENOENT`?

```
rescue Errno::ENOENT => ex
  puts "Error determining version! #{ex.message}"
```

The `rescue` would line up with the `add_command_under_category` and `end`

jeremymv2:
Having this simple yet important feature would be nice!

omnibus/files/private-chef-ctl-commands/version.rb
thomascate: good catch

omnibus/files/private-chef-ctl-commands/version.rb
thomascate: good catch

omnibus/files/private-chef-ctl-commands/version.rb
thomascate: adding error checking, I'm going to use e as the var and log instead of puts, to match how errors are handled in other parts of chef-server-ctl. For example rotate_credentials.rb

omnibus/files/private-chef-ctl-commands/version.rb
```diff
@@ -0,0 +1,34 @@
...
+add_command_under_category "version", "general", "Display current version of Chef Server.", 2 do
+
+  begin
+    # detect if running as a habitat service
+    if File.exist?('/hab/svc/chef-server-ctl/PID')
+      ident_file = File.read('../IDENT')
+      version = "chef-server #{ident_file.split('/')[2]}"
+    else
+      version = File.readlines('/opt/opscode/version-manifest.txt')[0]
```
markan: I might do ```JSON.parse(File.read('/opt/opscode/version-manifest.json'))['build_version']``` instead. I suspect the json file might be more future proof that relying on the first  line of the txt file.

markan:
Once you have all the changes, please squash/rebase and make sure you have signed off on the commit (git commit -s --amend) will fixup missing signoffs.

omnibus/files/private-chef-ctl-commands/version.rb
thomascate: I was on the fence as to which file to use, I'll switch over to the json.

PR #1488, opened 2018-04-20
Add sysvinitrc tag

markan:
Without this tag we generate a bogus /etc/inittab line and break things.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -42,6 +42,11 @@ default['private_chef']['fips_enabled'] = ChefConfig.fips?
 default['private_chef']['required_recipe']['enable'] = false
 default['private_chef']['required_recipe']['path'] = nil
 
+### For runit on sysvinit systems; without that we generate a empty tag value in inittab and break sysvinit.
+### Note if you happen to manage the chef-server with chef, and install your own runit, you will need to choose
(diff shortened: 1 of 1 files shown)


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -42,6 +42,9 @@
 default['private_chef']['required_recipe']['enable'] = false
 default['private_chef']['required_recipe']['path'] = nil
 
+### For runit on sysvinit systems; Without that we generate a bad tag value and break sysvinit.
+default['private_chef']['sysvinit_id'] = 'CS'
```
stevendanna: We used to be using the default here:

https://github.com/chef-cookbooks/enterprise-chef-common/blob/master/libraries/component_runin_supervisor_resource_mixin.rb#L10

so I'm guessing something changed on our upgrade of Chef that made this /not/ the default anymore or something.

If we want to avoid unnecessary modification to the inittab I suppose we could make this `SV`.

markan: I was going back and forth on that. My original thinking was that we kind of step on the key for anyone managing their chef server with chef and using the runit cookbook, which wasn't cool. On the other hand, we've historically been using the SV key, and changing that would mess with upgrades. In the light of days, I think going with the upgrade safe approach is better now.

markan:
Merging to stabilize build. The travis tests are failing due to an unrelated issue with chef-zero

PR #1489, opened 2018-04-20
fix builds in privilege restricted environments like builder

jeremymv2:
Attempts to fix Builder builds currently failing because they occur in a privilege restricted environment and cannot change owner or group based on the archive metadata during a tar extract.

This should introduce a new behavior to the tar command where owner/group archive metadata is ignored.

Additionally, it adds `core/libxml2` and `core/libxslt` to the build deps of `oc_id` so that we can then pass custom paths during the `nokogiri` gem build. 

Otherwise, when we bundle install, `nokogiri` fails to build because of a `tar` extraction failure of libxml like [this](https://bldr.habitat.sh/#/pkgs/chef-server/oc_id/builds/959414338453241856)

Further Discussion on the `tar` extraction issue: https://github.com/habitat-sh/builder/issues/365

### DIFF
--- a/src/oc-id/habitat/plan.sh
+++ b/src/oc-id/habitat/plan.sh
@@ -22,6 +22,8 @@ pkg_build_deps=(
   core/tar
   core/pkg-config
   core/coreutils
+  core/libxml2
+  core/libxslt
 )
 pkg_binds_optional=(
   [database]="port"
@@ -70,8 +72,6 @@ _tar_pipe_app_cp_to() {
   tar="$(pkg_path_for tar)/bin/tar"
 
   "$tar" -cp \
-      --owner=root:0 \
-      --group=root:0 \
       --no-xattrs \
       --exclude-backups \
       --exclude-vcs \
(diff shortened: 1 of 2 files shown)


jeremymv2:
@irvingpop I agree, I think your idea will move us forward so I've gone ahead and added the customization of the nokogiri gem build.

src/oc-id/habitat/plan.sh
```diff
@@ -93,6 +94,10 @@ do_install() {
       | _tar_pipe_app_cp_to "$HOME"
   bundle config path ${HOME}/vendor/bundle
   bundle config build.sqlite3 --with-sqlite3-lib=$(pkg_path_for core/sqlite)/lib
+  bundle config build.nokogiri --with-xml2-include=$(pkg_path_for core/libxml2)/include/libxml2 \
```
irvingpop: ah whoops, appears this is still missing `--use-system-libraries` @jeremymv2

PR #1495, opened 2018-04-26
Ma/update pc and rebar

markan:
Lock port compiler; it's the only thing we don't lock.

Update rebar to 3.5.0


src/sky/_build/default/lib/cf/rebar.config
```diff
@@ -0,0 +1,2 @@
+{erl_opts, [debug_info]}.
```
srenatus: Did you intend to commit this folder? 😉

PR #1497, opened 2018-05-04
Expeditor cleanups after review with releng, plus an nginx logging bugfix for dockerized environments

irvingpop:
- Change pkg_origin to chef for all packages
- switch from hard-coded origins to flexible ones for dbdpg and
openresty-noroot
- remove unnecessary `pkg_{lib,bin,include}_dirs` statements from hab
plans in order to stop confusing binlink and build-time linking
- remove unnecessary `pkg_source="nosuchfile.tar.gz"` and corresponding
`do_download()` and `do_verify()` steps
- remove commented out things, echos and printenvs
- remove `openresty-lpeg` plan as it wasn't being used
- Making nginx access logging configurable so it works in dockerized and non-dockerized setups
- Pull in nginx mime.types rather than trying to guess the hab origin of
the package it resides in

Signed-off-by: Irving Popovetsky <irving@chef.io>


src/chef-server-ctl/habitat/plan.sh
```diff
@@ -53,25 +43,17 @@ do_before() {
...
-
-do_prepare() {
-  return 0
-}
-
+#
+# do_prepare() {
+#   return 0
+# }
+#
```
stevendanna: Let's delete rather than comment to keep in the spirit of the rest of the PR.

irvingpop:
Great catches @stevendanna 

The rebar3 change is unintentional, and looks like the result of the hab plans patching those files in-place:
https://github.com/chef/chef-server/blob/master/src/oc_erchef/habitat/plan.sh#L85

I'm going to investigate solutions today that don't impact those files directly.  (heads up @markan @jeremymv2 )

irvingpop:
okay so it appears that running ` fix_interpreter ./rebar3` is no longer necessary and now functions correctly inside the build.  I'm going to remove it.

irvingpop:
Just did a full rebuild and test, ready for re-review

docker-compose.yml
```diff
@@ -11,7 +11,7 @@
 version: '2.1'
 services:
   postgresql:
-    image: chefdemo/postgresql:stable
+    image: chefdemo/postgresql:${VERSION:-stable}
```
jeremymv2: I believe the omission of VERSION was intentional before, since the VERSION of postgresql and elasticsearch doesn't really match up the rest of the services.  Perhaps leave it off, or make it a variable name specific for postgresql?

I think we want to account for two scenarios: 
* running this compose with docker images build and published locally, from running `./habitat_pkgs_build.sh` (in which case VERSION won't be 'stable' and VERSION won't match what's local)
* running the compose to pull images from docker hub

docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
         token = "${AUTOMATE_TOKEN:-93a49a4f2482c64126f7b6015e6b0f30284287ee4054ff8807fb63d9cbd1c506}"
 
   elasticsearch:
-    image: chefdemo/elasticsearch5:stable
+    image: chefdemo/elasticsearch5:${VERSION:-stable}
```
jeremymv2: Same small concern here as I had for postgresql

src/nginx/habitat/config/chef_http_lb_common
```diff
@@ -16,7 +16,7 @@
     set_by_lua $data_collector_token 'return os.getenv("DATA_COLLECTOR_TOKEN")';
     {{~/if}}
 
-    access_log /dev/stdout opscode;
+    access_log {{cfg.access_log}} opscode;
```
jeremymv2: I built the pkg, exported to docker, started it via compose and noticed that this wasn't set.

```
/ # echo $HAB_CHEF_SERVER_NGINX
access_log = "/dev/tty"
/ # grep access_log /hab/svc/chef-server-nginx/config/chef_https_lb.conf
    access_log  opscode;
/ #
```

After glancing at the reset of the file, in this context, it should be `{{../cfg.access_log}}`

irvingpop:
thanks for the feedback @jeremymv2 , I addressed those comments

PR #1505, opened 2018-06-01
Update os versions in omnibus kitchen.yml

markan:
We want more modern versions of Ubuntu and Centos when building via omnibus.

### DIFF
--- a/omnibus/.kitchen.yml
+++ b/omnibus/.kitchen.yml
@@ -9,7 +9,8 @@ driver:
 
 provisioner:
   name: chef_zero
-  require_chef_omnibus: 14.1.12
+  product_name: chef
+  product_version: 14.1.12
 
 platforms:
   - name: ubuntu-18.04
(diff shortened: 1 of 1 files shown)


omnibus/.kitchen.yml
```diff
@@ -12,19 +12,24 @@ provisioner:
...
+  - name: ubuntu-14.04
     run_list: apt::default
-  - name: ubuntu-10.04
+  - name: ubuntu-12.04
     run_list: apt::default
-  - name: centos-5.11
-    run_list:  yum-centos::vault
-  - name: centos-6.7
-  - name: centos-7.2
+  - name: centos-6.9
```
tas50: If you make this centos-6 you'll always get the latest one that Seth publishes.

omnibus/.kitchen.yml
```diff
@@ -12,19 +12,24 @@ provisioner:
...
+  - name: centos-7.3
```
tas50: Same here. Use centos-7

Issue #265, opened 2015-05-19
Dozens of Chef warnings on reconfigure


stevendanna:
On the latest nighties, we are seeing dozens of warnings again during reconfigure.  Here is a breakdown:

```
root@api:~# chef-server-ctl reconfigure 2>&1 | grep WARN | awk '{ $1=""; $2=""; print}' | sort | uniq -c | sort -nr
     42   This will no longer work in Chef 13: you must use 'provides' to provide DSL.
     17   Class Chef::Resource::RunitService does not declare 'provides :runit_service'.
      8   In Chef 13, this will break: you must call provides to mark the names you provide, even if you also override provides? yourself.
      6   Using an LWRP by its name (PrivateChefPgUser) directly is no longer supported in Chef 12 and will be removed. Use Chef::Resource.resource_for_node(node, name) instead.
      6   Class Chef::Resource::PrivateChefPgUser does not declare 'provides :private_chef_pg_user'.
      5   Using an LWRP by its name (PrivateChefPgUserTableAccess) directly is no longer supported in Chef 12 and will be removed. Use Chef::Resource.resource_for_node(node, name) instead.
      5   Using an LWRP by its name (PrivateChefPgDatabase) directly is no longer supported in Chef 12 and will be removed. Use Chef::Resource.resource_for_node(node, name) instead.
      5   Class Chef::Resource::PrivateChefPgUserTableAccess does not declare 'provides :private_chef_pg_user_table_access'.
      5   Class Chef::Resource::PrivateChefPgDatabase does not declare 'provides :private_chef_pg_database'.
      3   No analytics oc-id-application present. Skipping
      2   Using an LWRP by its name (PrivateChefPgUpgrade) directly is no longer supported in Chef 12 and will be removed. Use Chef::Resource.resource_for_node(node, name) instead.
      2   Using an LWRP by its name (PrivateChefPgCluster) directly is no longer supported in Chef 12 and will be removed. Use Chef::Resource.resource_for_node(node, name) instead.
      2   Using an LWRP by its name (PrivateChefKeepalivedSafemode) directly is no longer supported in Chef 12 and will be removed. Use Chef::Resource.resource_for_node(node, name) instead.
      2   Class Chef::Resource::PrivateChefPgUpgrade does not declare 'provides :private_chef_pg_upgrade'.
      2   Class Chef::Resource::PrivateChefPgCluster does not declare 'provides :private_chef_pg_cluster'.
[...]

PR #1506, opened 2018-06-01
Upgrade to Chef 14

markan:
Upgrade to chef 14, and clean up some residual issues left from the original Chef 13 work.

This picks up some work that was done in PR #1327 but didn't get included when we actually upgraded to chef 13.


omnibus/files/private-chef-cookbooks/private-chef/resources/chef_run.rb
```diff
@@ -0,0 +1,58 @@
+provides :chef_run
+
+property :run_list, :kind_of => String, :name_attribute => true
```
tas50: Get rid of :kind_of =>
Also name_attribute should be name_property

omnibus/files/private-chef-cookbooks/private-chef/resources/chef_run.rb
```diff
@@ -0,0 +1,58 @@
+provides :chef_run
```
stevendanna: We might want to explicitly test if this thing still works manually.  It is pretty tightly coupled to chef-client internals....

markan: Testing a plugin install should be sufficient, right?

markan:
This closes #1327 (much of it was directly lifted from there, in fact)

markan:
It also closes #265

Issue #196, opened 2015-05-01
404 page is generic rails


stevendanna:
In chef/oc-id#49 @mmzyk said:

> If a user hits a 404 page that exists within oc-id they will get a generic rails 404 page. This is a puzzling experience to users and the 404 page should be made to match the rest of the oc-id styling and give better messaging.

PR #1507, opened 2018-06-01
Fix enterprise chef common typo

markan:
We switched from x.y.z to vx.y.z tag format and I didn't catch that, breaking things.


markan:
Merging without review because its a trivial change, and un-breaks the builds. (see internal jenkins chef-server build #196)

PR #1514, opened 2018-06-28
MSYS-848 : Prevent Chef Server headers from exposing openresty name and version

piyushawasthi:
When connecting to the Chef Server, it currently exposes in a `Server` header that the openresty package is used and it's current version. For security reasons we would prefer to not expose that information.

Reproduction steps:
```
openssl s_client -connect my_chef_server:443
[snip ssl related output]
HEAD / HTTP/1.0
```

depending on the Chef server version you'll see output containing a line like:
```
Server: openresty/1.11.2.1
```

This can also be seen over HTTP:
```
root@chef-server:/var/opt/opscode/nginx/etc# nc localhost 80
HEAD / HTTP/1.0

HTTP/1.1 301 Moved Permanently
Server: openresty/1.11.2.1
Date: Mon, 25 Jun 2018 20:59:57 GMT
Content-Type: text/html
Content-Length: 191
Connection: close
Location: https://chef-server.ninjr.org/
```

When installed on a server the nginx configs are in /var/opt/opscode/nginx/etc. It appears that adding this configuration will prevent the line from being displayed:
```
more_clear_headers Server;
```

Acceptance Criteria:
Chef Server does not print information about openresty via http or https.

Signed-off-by: piyushawasthi <piyush.awasthi@msystechnologies.com>

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
@@ -141,6 +141,7 @@ http {
           server {
             <%= @helper.listen_port('http') %>
             server_name <%= @helper.server_name %>;
+            more_clear_headers Server;
             access_log /var/log/opscode/nginx/rewrite-port-80.log;
             return 301 https://$host$request_uri;
           }
@@ -156,6 +157,7 @@ http {
   server {
     <%= @helper.listen_port(node['private_chef']['lb_internal']['chef_port']) %>
     server_name <%= @helper.server_name %>;
+    more_clear_headers Server;
 
     client_max_body_size <%= @client_max_body_size %>;
     proxy_set_header        Host            $host;
@@ -188,6 +190,7 @@ http {
   server {
     <%= @helper.listen_port(node['private_chef']['lb_internal']['oc_bifrost_port']) %>
     server_name <%= @helper.server_name %>;
+    more_clear_headers Server;
 
     client_max_body_size <%= @client_max_body_size %>;
     proxy_set_header        Host            $host;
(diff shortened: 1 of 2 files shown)


omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx.conf.erb
```diff
@@ -141,6 +141,7 @@ http {
           server {
             <%= @helper.listen_port('http') %>
             server_name <%= @helper.server_name %>;
+            more_clear_headers <%= @helper.server_name %>;
```
srenatus: ❓ From your commit message, I would expect this to be `more_clear_headers Server;`, but this renders to something else. How come?

piyushawasthi: @srenatus : Sorry I have update that actually on dev I am looking something and that was push please have a look.

srenatus:
@piyushawasthi Sorry I hadn't done that earlier today; but I've just checked another product. (I had vaguely remember that we solved the same issue once before over there.)

Back then, we'd used
```
server_tokens off;
```
[See docs on `server_tokens`](http://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens) and [this is the commit in question](https://github.com/chef/automate/commit/a54c2d91a7c2360379dfc272d58ea8a766952d26).

It seems to me that this distinction is relevant:
>  Enables or disables emitting nginx version **on error pages** and in the “Server” response header field. 

I know this isn't strictly about error pages, but what do you think, could you try if `server_tokens off;` does the trick, too? 😃

stevendanna:
The `server_tokens` setting will only omit the version but a Server header that identifies the server as "openresty" or "nginx" will still be sent in the response.  Which one to use really depends on whether we want complete silence or just want to supress the version.

srenatus:
@stevendanna @piyushawasthi Wouldn't it make sense to do _both_ then?

PR #1516, opened 2018-07-03
Ma/appbundle and move

markan:
Rework chef-server-ctl. 
This makes it a gem, appbundles it, and moves it to the src directory alongside the habitat stuff. Previously we were letting the dependencies float free, and that breaks when we upgrade other gems.

This is built on #1513, and will replace it if it goes smoothly.

Will squash this up for final merge.


omnibus/config/software/private-chef-ctl.rb
```diff
@@ -14,13 +14,16 @@
...
+source path: "#{project.files_path}/../../src/chef-server-ctl"
 
 license :project_license
 skip_transitive_dependency_licensing true
 
+dependency "postgresql96" # for libpq
+
+dependency "appbundler"
 dependency "highline-gem"
 dependency "sequel-gem"
```
lamont-granquist: these gems should get moved into the Gemfile.lock to get accurately appbundle'd.

if there are c-libs in the omnibus sources those deps should get moved here.

src/chef-server-ctl/lib/chef_server_ctl/helpers/key_ctl_helper.rb
```diff
@@ -0,0 +1,192 @@
+require "chef/config"
```
marcparadise: This looks like a straight move w/ no changes?

src/chef-server-ctl/plugins/upgrade.rb
```diff
@@ -3,11 +3,8 @@
 # All Rights Reserved
 #
 
-begin
-  require_relative 'open_source_chef12_upgrade'
-rescue LoadError
-  require '/opt/opscode/embedded/service/omnibus-ctl/open_source_chef12_upgrade'
-end
+# TODO refactor this for gemified chef-server-ctl
```
marcparadise: Is this an intentionally left behind TODO?

src/chef-server-ctl/lib/chef_server_ctl/helpers/key_ctl_helper.rb
markan: Yes; looks like the mv got lost somehow and turned into an add + delete

markan:
oc_erchef is failing in ct. I'm still merging this since master (a5b55c864411fcaea21b7da3fd5646e6c8db306a)  has the same issue, and is broken in other ways that this fixes.

PR #1526, opened 2018-07-20
Remove a few more couchdb configurables

stevendanna:
I imagine sometime in the 2020s we'll be able to grep the code base
and not see `couch` in it. I found these when reviewing available
erchef configuration. I have not tested this at all. If something is
using this I really want to find out.

### DIFF
--- a/src/oc_erchef/config/vars.config
+++ b/src/oc_erchef/config/vars.config
@@ -1,7 +1,6 @@
 %% -*- mode: erlang -*-
 {couchdb_server, "localhost"}.
 {couchdb_port, 5984}.
-{couchdb_max_conn, 50}.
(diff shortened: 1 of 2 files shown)


src/oc_erchef/config/vars.config
```diff
@@ -1,7 +1,6 @@
 %% -*- mode: erlang -*-
-{couchdb_server, "localhost"}.
+{couchdb_server, "localhost"}
```
stevendanna: Yikes, interseting that this typo snuck in and that it doesn't break anything

markan: There should be a period there. That ends the erlang expression. I suspect that file is entirely replaced by the file generated by the chef-server cookbooks, and so this error never blows things up.

PR #1528, opened 2018-07-24
Prepare 12.18 Chef Server release.

rhass:
The patch version is not specified in the RELEASE_NOTES.md since expeditor controls the patch version.


cp713a:
Please help.

Anyone by chance has a copied of the chef-server 12.18.0 somewhere?

rhass:
@cp713a You can get `12.18.14` from Chef Downloads: https://downloads.chef.io/chef-server/stable/12.18.14

If you need `12.18.0` you will have to build it from the tag. However, there probably isn't a reason to use `12.18.0` over `12.18.14`. The change history between the two looks big, but it's not too bad. https://github.com/chef/chef-server/compare/12.18.0...12.18.14

Good luck!

PR #1530, opened 2018-08-01
Remove old github changelog generator

tas50:
This should all be unused at this point since we bump the changelog with expeditor.


tas50:
@stevendanna We could probably rig up a similar functionality in the promotion step if we really wanted, but I'd argue that sort of thing should just go in the release notes, which are still human managed.

markan:
You might need to kick the travis tests to rerun the failing test. Otherwise you should be able to rebase this and merge. (Please rebase before the merge, it makes things in git visualizers more legible.

tas50:
Rebase. We'll see now

PR #1533, opened 2018-08-02
[chef-server-ctl] Install only appbundled version

markan:
### Description

We don't want to gem-install any versions of chef-server ctl in the user path, only the appbundled version.

### Issues Resolved

This addresses some potential breakage where we pick up the non-appbundled version in /opt/opscode/embedded/bin. 


PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/354#045b131a-486a-475c-a2d2-bd2c69877025

PR #1538, opened 2018-08-09
WIP [chef-server-ctl] Make chef-server-ctl configurable

stevendanna:
This PR adds the ability to inject Chef Server configuration items
required by chef-server-ctl subcommands. A Config class has been added
that provides access to this configuration globally.

The short-term goal is to make it straightforward to get more of the
chef-server-ctl commands working from Habitat packages.  We now have 2
sets of Habitat packages:

- the docker-focused Habitat packages in this repository
- the A2-focused Habitat packages in the A2 repository

The A2 packages consume the packages in this repository.

Previously, the server-admins, key-rotation, reindex, bifrost-cleanup,
filtered-dump, and password commands were unlikely to work with either
set of packages. Now, these commands can work in our Habitat packages
if the wrappers set the appropriate configuration.

The long-term goal of this is two-fold. First, we want to make
chef-server-ctl are more standalone application. The first step of
this was the work that moved the command into its own gem. In addition
to supporting Habitat, making these dependencies configurable means we
can do things like:

- Inject stub versions of commands to test failure conditions Point

- chef-sever-ctl and chef-zero or similar API stubs for testing

- Decouple sub-command logic from the particulars of omnibus
  packaging, paving the way for future packaging changes.

The second long-term goal is to make the Habitat packages in Chef
Server more generically consumable for downstream consumers such as
A2.

## Ugh, why environment variables?

In the long-run I think we might want a custom configuration file
based on mixlib-config or similar. However, environment variables
have the following benefits:

- Easy to set from Habitat wrappers
- Easy to set at the command line for testing
- Doesn't require immediately fixing all of chef-server-ctl's command
  line argument processing bugs
- Avoids pre-baking a config-file format until we know what we need.

I think we should consider these environment variables UNSUPPORTED at
the moment and thus developers are free to change them

## Should we get rid of CHEF_SECRETS_DATA in favor of this?

No, at least not yet. Getting rid of CHEF_SECRETS_DATA in chef-server
itself will require being very careful to not add any credentials to
any new on-disk files.

## Why didn't you fix all of the commands?

I've limited this initial change to the commands supported in the A2
Habitat packages. This also has the benefit of not touching some of
the more complex commands just yet. The un-modified commands include:

  - reconfigure
  - oc-id commands
  - secrets rotation commands
  - All runit-based commands
  - HA commands
  - upgrade/chef-11-upgrade
  - gather-logs

Fixing all of chef-server-ctl will be a lengthy process since we have
a number of environments to test it in.

### Check List

- [ ] All tests pass

### DIFF
--- a/src/chef-server-ctl/lib/chef_server_ctl/config.rb
+++ b/src/chef-server-ctl/lib/chef_server_ctl/config.rb
@@ -1,0 +1,162 @@
+require 'chef_server_ctl/log'
+
+# ChefServerCtl::Config is a global configuration class for
+# ChefServerCtl subcommands.
+#
+# We use a global at the moment to avoid too much upheaval in the
+# various subcommands.
+#
+# Configuration is based on environment variables to make it easy to
+# implement wrappers in our Habitat packaged versions of chef-server.
+#
+# If the environment variables become too unwieldy, we can change them
+# as long as we remember to go fix the Habitat wrappers.
+#
+# TODO(ssd) 2018-08-08: Maybe use a configuration file instead?  I've
+# opted against it for now to avoid having to write yet-another-file
+# out during reconfiugration.
+module ChefServerCtl
+  module Config
+    DEFAULT_KNIFE_CONFIG_FILE = "/etc/opscode/pivotal.rb".freeze
+    DEFAULT_KNIFE_BIN = "/opt/opscode/embedded/bin/knife".freeze
+    DEFAULT_LB_URL = "https://127.0.0.1".freeze
+    DEFAULT_FIPS_LB_URL = "http://127.0.0.1".freeze
+    DEFAULT_RABBITMQCTL_BIN = "/opt/opscode/embedded/service/rabbitmq/sbin/rabbitmqctl".freeze
+    DEFAULT_ERCHEF_REINDEX_SCRIPT = "/opt/opscode/embedded/service/opscode-erchef/bin/reindex-opc-organization".freeze
+
+    def self.init(ctl)
+      @@ctl = ctl
+      Log.debug("Using KNIFE_CONFIG_FILE=#{self.knife_config_file}")
+      Log.debug("Using KNIFE_BIN=#{self.knife_bin}")
+      Log.debug("Using BIFROST_URL=#{self.bifrost_url}")
+      Log.debug("Using LB_URL=#{self.lb_url}")
+      Log.debug("Using HABITAT_MODE=#{self.habitat_mode}")
+    end
+
+    # knife_config should be the path to a configuration file that
+    # allows the `knife` executable to run with pivotal permissions.
+    def self.knife_config_file
+      if ENV['CSC_KNIFE_CONFIG_FILE']
+        ENV['CSC_KNIFE_CONFIG_FILE']
+      else
+        DEFAULT_KNIFE_CONFIG_FILE
+      end
+    end
+
+    # knife_bin is the command used to execute knife.
+    def self.knife_bin
+      if ENV['CSC_KNIFE_BIN']
+        ENV['CSC_KNIFE_BIN']
+      else
+        DEFAULT_KNIFE_BIN
+      end
+    end
+
+    # rabbitmqctl_bin is the command used to execute rabbitmqctl. This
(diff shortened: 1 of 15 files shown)


src/chef-server-ctl/plugins/password.rb
```diff
@@ -67,5 +67,11 @@ def run_knife_opc_cmd(cmd, message)
 end
 
 def ldap_authentication_enabled?
-  running_config['private_chef']['ldap'] && running_config['private_chef']['ldap']['enabled']
+  # NOTE(ssd) 2018-08-09: Current neither of the Habitat-ized Chef
```
stevendanna: s/Current/Currently/

stevendanna:
For chef-server employee's, here is an example of the kind of wrapper command this facilitates: https://github.com/chef/a2/blob/4853fa014eafb07d6a4aec780c0f248e0bbead63/components/automate-cs-nginx/cmd/chef-server-ctl/ctl.go

src/chef-server-ctl/lib/chef_server_ctl/config.rb
```diff
@@ -0,0 +1,170 @@
...
+#
+# If the environment variables become too unwieldy, we can change them
+# as long as we remember to go fix the Habitat wrappers.
+#
+# TODO(ssd) 2018-08-08: Maybe use a configuration file instead?  I've
+# opted against it for now to avoid having to write yet-another-file
+# out during reconfiugration.
+module ChefServerCtl
+  module Config
+    DEFAULT_KNIFE_CONFIG_FILE = "/etc/opscode/pivotal.rb".freeze
```
danielsdeleo: frozen string literals would probably also do the trick but NBD

src/chef-server-ctl/lib/chef_server_ctl/config.rb
```diff
@@ -0,0 +1,170 @@
...
+      if ENV['CSC_RABBITMQCTL_BIN']
+        ENV['CSC_RABBITMQCTL_BIN']
+      else
+        DEFAULT_RABBITMQCTL_BIN
+      end
+    end
+
+    # fips_enabled indicates whether the chef-server is running in
+    # fips mode.
+    def self.fips_enabled
```
markan: Jenkins omnibus tests fail near here. That build was for the a339ca4 change, but I don't see anything obvious to fix it in later commits. I'll dig into things once I finish debugging the build failures on master.
```
    chef-server: /opt/opscode/embedded/lib/ruby/gems/2.5.0/gems/chef-server-ctl-1.0.0/lib/chef_server_ctl/config.rb:72:in `fips_enabled': undefined method `[]' for nil:NilClass (NoMethodError)
    chef-server: 	from /opt/opscode/embedded/lib/ruby/gems/2.5.0/gems/chef-server-ctl-1.0.0/lib/chef_server_ctl/config.rb:81:in `lb_url'
    chef-server: 	from /opt/opscode/embedded/lib/ruby/gems/2.5.0/gems/chef-server-ctl-1.0.0/lib/chef_server_ctl/config.rb:33:in `init'
    chef-server: 	from /opt/opscode/embedded/lib/ruby/gems/2.5.0/gems/chef-server-ctl-1.0.0/bin/chef-server-ctl:267:in `<top (required)>'
    chef-server: 	from /opt/opscode/embedded/bin/chef-server-ctl:23:in `load'
    chef-server: 	from /opt/opscode/embedded/bin/chef-server-ctl:23:in `<main>'
```

markan: Look like running_config isn't set up at this point in the code (we've just created a new Omnibus::ChefServerCtl in line 264 of /opt/opscode/embedded/lib/ruby/gems/2.5.0/gems/chef-server-ctl-1.0.0/bin/chef-server-ctl

stevendanna: @markan this should be fixed since we no longer try to print lb_url in init

stevendanna: Confirmed that the tests are now passing for at least x86: http://wilson.ci.chef.co/job/chef-server-test/253/

PR #1540, opened 2018-08-11
Postgres listens on loopbacks, not localhost

markan:
### Description

We previously set postgresql to use `listen_addresses = 'localhost'`
for standalone installs. But that can be problematic, depending on the
/etc/hosts file; on some platforms that resolves to the ipv6 loopback,
and postgres seems to take this as ’listen only to ipv6 loopback (::1)
’.  A lot of the services use hardcoded ivp4 loopback to connect, and
fail.

This should not affect tiered or HA systems, since in those cases we
listen on a specific IP address, not localhost.


stevendanna:
Will this fail if ::1 isn't available?

irvingpop:
as an alternative, we could use Ohai to determine the available loopback addresses and listen on all of them:
```
chef (14.3.37)> node['network']['interfaces']['lo0']['addresses'].keys.join(',')
 => "127.0.0.1,::1,fe80::1"
```

markan:
Updated with @irvingpop's suggestion, and confirmed this fixes the issue with s390x on centos.

PR #1542, opened 2018-08-14
Modernize Expeditor config; Promote harts/containers

schisamo:
This PR makes a number of pipeline related improvements:

- Move away from `artifact_actions` and begin using `subscriptions` to
  control all promotion actions
- Declare `post_commit` config in `merge_actions`
- Automatically delete feature branches following a successful merge
- Create a `manifest.json` file at `*.hart` build time. This becomes the
  release artifact that maps a set of Habitat pkg idents to a particular
  Chef Server version
- Execute Habitat channel and Docker container tag promotions at the
  appropriate time

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -11,6 +11,13 @@ automate:
   project: chef-server
 
 github:
+  delete_branch_on_merge: true
+  # The file where the MAJOR.MINOR.PATCH version is kept. The version in this file
+  # is bumped automatically via the `built_in:bump_version` merge_action.
+  version_file: "VERSION"
+  # The file where our CHANGELOG is kept. This file is updated automatically with
+  # details from the Pull Request via the `built_in:update_changelog` merge_action.
+  changelog_file: "CHANGELOG.md"
   release_branch:
     - master:
         # we'll have to bump this later, as globs won't let us do > 18
@@ -18,19 +25,13 @@ github:
     - 12-18-stable:
         # this branch should only be active
         version_constraint: "12.18.*"
-  # The file where the MAJOR.MINOR.PATCH version is kept. The version in this file
-  # is bumped automatically via the `built_in:bump_version` merge_action.
-  version_file: "VERSION"
-  # The file where our CHANGELOG is kept. This file is updated automatically with
-  # details from the Pull Request via the `built_in:update_changelog` merge_action.
-  changelog_file: "CHANGELOG.md"
 
 # Habitat + Docker exporting
 habitat_packages:
   - openresty-noroot:
       source: src/openresty-noroot
       bldr_paths: src/openresty-noroot/*
-  - oc-id:
+  - oc_id:
       source: src/oc-id
       bldr_paths: src/oc-id/*
       export:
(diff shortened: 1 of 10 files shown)


.expeditor/config.yml
```diff
@@ -92,14 +107,22 @@ subscriptions:
...
-    - built_in:update_acc
-  promoted_to_current:
-  promoted_to_stable:
-    - built_in:rollover_changelog
-    - built_in:notify_chefio_slack_channels
+  # TODO: add action to update changelog with modified omnibus components
+  - workload: artifact_published:unstable:chef-server:*
+    actions:
+      - built_in:update_acc
+  - workload: artifact_published:current:chef-server:*
```
tduffield: I'm not sure what specifying this without any actions will do, but probably it will break things.

.expeditor/config.yml
```diff
@@ -92,14 +107,22 @@ subscriptions:
...
+  - workload: artifact_published:stable:chef-server:*
+    actions:
+      - built_in:rollover_changelog
+      - built_in:notify_chefio_slack_channels
+
+promote:
+  action:
```
tduffield: please use `actions`

.expeditor/config.yml
```diff
@@ -92,14 +107,22 @@ subscriptions:
...
+    - built_in:promote_artifactory_artifact
+    - bash:.expeditor/promote_harts_and_containers.sh
```
tduffield: these two bash scripts need to be `post_commit: true`

.expeditor/license_scout.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
tduffield: Does `chef-server` have these components? Do we need to update this or remove it?

.expeditor/create_manifest.rb
```diff
@@ -0,0 +1,129 @@
...
+  # releases every 2 weeks, often giving us very little time to
+  # validate that the new release works as expected for us.
+  #
+  # IF YOU UPDATE THESE PINS YOU MUST ALSO UPDATE THE core/hab PIN IN components/automate-deployment/habitat/plan.sh
+  #
+  pinned_hab_components = {
+    "hab"          => { "origin" => "core", "name" => "hab",          "version" => "0.59.0", "release" => "20180712155441"},
+    "hab-sup"      => { "origin" => "core", "name" => "hab-sup",      "version" => "0.59.0", "release" => "20180712161546"},
+    "hab-launcher" => { "origin" => "core", "name" => "hab-launcher", "version" => "7797",   "release" => "20180625172404"}
+  }
```
jeremymv2: 🤔 we're definitely going to want these updated to 0.64.1 for the Chef Server packages.
0.59.0 does not have the features we need.

schisamo: Those items were just cargo culted from the A2 version of this manifest. We don't really need them in this Chef Server iteration. In A2 the deployment service (which consumes this manifest) uses that data to know what version of the sup and launcher to install.

jeremymv2: In that case, could we remove these pins and just pick up the latest version of each component?

jeremymv2:
@schisamo Is there any way to tell Expeditor in this config to publish all docker containers with the same Habitat components (hab, hab-sup, launcher)?

In order to be operationally consumable in a docker environment, the containers that comprise the Chef Server stack all really need to be running the same Hab versions. At present, the only way to make this happen is to "touch" a file for each habitat package (oc_erchef, oc_id, chef-server-ctl etc) in a PR in order to for them all to rebuild and publish to Dockerhub with matching Habitat Supervisors. It might be nice to enhance this workflow somehow.

What sort of tag will the containers end up with in Dockerhub? 
Ideally, we'd be able to tell a customer, go `docker pull` tag `12.17.50-20180504221721` or `stable` for all the Chef Server hab pkg containers and the tags make sense in that they container the same semver of the Omnibus pkgs for that same release.

tduffield:
@jeremymv2 at the moment that is not something Expeditor supports.

PR #1544, opened 2018-08-15
[omnibus] Rabbitmq uses loopback for its name

markan:
WIP; This doesn't work in all configurations, refining still.

### Description

localhost is unsafe in some configurations; not all systems resolve it
consistiently. Trying ipv4 loopback (127.0.0.1) as an alternative.

NOTE: Erlang distribution can be either a shortname (myhost) or a
longname (myhost.example.com). IPv4 addresses are treated as
longnames, and we have to switch to using those.

### Issues Resolved

This fixes issues with machines that use ipv6 loopback for localhost.


markan:
Well, even with the restart logic, there's some issues here, as the DB name for rabbit uses the name, e.g. /var/opt/opscode/rabbitmq/db/rabbit@127.0.0.1

markan:
Closing, as we have another, cleaner approach.

PR #1566, opened 2018-09-05
Add an X-Orgname header to data_collector requests

mivok:
This allows easy identification of the chef organization making the request, and allows sending data collector requests to different automate instances based on organization name, by means of an intermediate proxy. This is intended to be used with hosted chef.

My erlang skills are practically non-existent, so if there's a better way to do the erchef side of things, let me know.


stevendanna:
It looks like we never really built any test scaffolding for data collector calls unfortunately, so adding end-to-end tests might be better for a follow up.  As discussed in chat, one way forward with testing might be something similar to the compliance_proxy_spec.rb tests in pedant.

PrajaktaPurohit:
Closing after a chat with the Ops team. This was required for managed automate project with is not on the cards.

PR #1567, opened 2018-09-05
Update chef-client to 14.4.56

tas50:
This eliminates the need for the openssl cookbook, which slims down chef-server a small bit.


tas50:
Thanks @markan. Travis green really doesn't mean much here. Thanks for kicking off a Jenkins job.

PR #1578, opened 2018-10-02
Experiment around reindexing fixes (WIP)

markan:
This is an attempt to address an issue (#1586) we've seen when the indexer gets backlogged. We put a work item on the queue via a gen_server , and it is waiting to be indexed. But because it's backlogged, we don't get to it, and fail to reply to the caller before the 5 sec timeout fires. And so the server returns 500, and the client retries. However the indexer will eventually get to the item, and index it successfully. Indexing is basically the last thing we do in the request, and so we've successfully done all of the work, but the client still retries because we've told it has failed. This ends up causing load amplification on an already backlogged server. 

This increases the timeout and attempts to do better error logging.

This PR as is is incomplete, as it doesn't check the call chain to see if there's something else that will timeout once this gen server timeout is increased.

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/chef_reindex.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/chef_reindex.erl
@@ -35,6 +35,7 @@
 -export([
          make_context/0,
          reindex/2,
+         reindex/3,
          reindex_by_id/4,
          reindex_by_name/4
         ]).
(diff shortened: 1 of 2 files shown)


src/oc_erchef/apps/oc_chef_wm/src/chef_reindex.erl
```diff
@@ -255,8 +256,14 @@ log_failures(OrgName, [Failure|Rest]) ->
...
     Acc;
 humanize_failures([H|T], Acc) ->
-    {Id, Reason} = H,
-    humanize_failures(T, [{Id, pretty_reason(Reason)} | Acc]).
+    case H of
+        {Id, Reason} ->
+            humanize_failures(T, [{Id, pretty_reason(Reason)} | Acc]);
+        X ->
+            lager:error("humanize failures couldn't make sense of failure ~s~p", [X]),
+            humanize_failures(T, [{Id, pretty_reason(Reason)} | Acc]);
```
stevendanna: Id and Reason are unbound here, maybe just make X our reason and put some dummy ID in

markan:
Resolved in #1945

PR #1579, opened 2018-10-02
increase authn:keygen_timeout for oc_erchef hab pkg

jeremymv2:
### Description
This is a very simple change to the `oc_erchef` Hab package default configuration.
We noticed that the `keygen_timeout` value was too small out of the box and would continually spam these warnings:

```
oc-erchef.default(O): [warning] {chef_keygen_cache,keygen_timeout}
```

Increasing to this proposed value yielded no adverse affects in our testing over the months and so I decided  to just make it the new default 😄 

### DIFF
--- a/src/oc_erchef/habitat/default.toml
+++ b/src/oc_erchef/habitat/default.toml
@@ -23,7 +23,7 @@ strict_search_result_acls=false
 keygen_cache_workers=10
 keygen_cache_size=10
 keygen_start_size=0
-keygen_timeout=1000
+keygen_timeout=5000
 
 [oc_chef_authz]
 


stevendanna:
> Increasing to this proposed value yielded no adverse affects in our testing over the months and so I decided to just make it the new default 😄

Bumping this is probablly fine, but 20 seconds seems very long.  Is this testing something we can put in the pipeline and use it to choose a smaller value?  Happy to just merge this as is though.

jeremymv2:
@stevendanna I consistently got timeouts with the timeout at `2000` on an quiet system, therefore I'm thinking doubling that and rounding up to 5s is reasonable. I have squashed and pushed up.

stevendanna:
5 seconds is exactly in the realm of what I would expect.  I'm actually pretty surprised we don't see this fail more often at 1 seconds since we are forking a command that does crypto potentially at early boot.

stevendanna:
I just noticed we are using 20s over in A2: https://github.com/chef/a2/blob/master/components/automate-cs-oc-erchef/pkg/api/config_request.go#L52

PR #1582, opened 2018-10-18
Use our standard ruby-cleanup definition

tas50:
This looks like it will shrink our install by ~5% on disk by cleaning up
more files. We'll need a real build before standing by that number
though.


markan:
Testing build in ad-hoc # 168

Issue #176, opened 2015-05-01
post-install messages still show up if the package fails to install


stevendanna:
From @smith in chef/opscode-omnibus#427:

> Tried this:
> 
> ```
> root@private-chef:/opt/opscode/sv/opscode-webui# dpkg -i /tmp/cache/private-chef_11.2.99+20140830193320.git.2.dd76a61-1_amd64.deb
> (Reading database ... 136994 files and directories currently installed.)
> Preparing to replace private-chef 11.1.5+20140725034804.git.249.b22037f-1 (using .../private-chef_11.2.99+20140830193320.git.2.dd76a61
> -1_amd64.deb) ...
> Unpacking replacement private-chef ...
> dpkg: error processing /tmp/cache/private-chef_11.2.99+20140830193320.git.2.dd76a61-1_amd64.deb (--install):
>  unable to make backup link of `./opt/opscode/embedded/cookbooks/private-chef/providers/pg_cluster.rb' before installing new version:
> Operation not permitted
> dpkg-deb: subprocess paste killed by signal (Broken pipe)
> You have upgraded Private Chef!
> 
> The next step in the upgrade process is to run:
> 
> private-chef-ctl upgrade
> 
> Then, to remove configuration files, logs, directories,
> users, etc. that were used by internal services that
> have been removed from this version of Private Chef,
> please run:
> 
> private-chef-ctl cleanup
> 
> (Add the '--no-op' option to see what would be removed by
> this command)
> Errors were encountered while processing:
>  /tmp/cache/private-chef_11.2.99+20140830193320.git.2.dd76a61-1_amd64.deb
> ```
> 
> Notice that the install failed due to some permissions issues with dev-vm mounting, but the messages telling me to run upgrade/cleanup were still printed. These should only show up on a successful package install. On Ubuntu 10.04.

stevendanna:
I've reproduced this bug in the following way:

```
# chattr +ia /opt/opscode/embedded/cookbooks/private-chef/providers/pg_cluster.rb
#  ... 100617 files and directories currently installed.)
Preparing to unpack .../chef-server-core_12.0.8-1_amd64.deb ...
Unpacking chef-server-core (12.0.8-1) over (12.0.8-1) ...
dpkg: error processing archive /vagrant/chef-server-core_12.0.8-1_amd64.deb (--install):
 unable to make backup link of `./opt/opscode/embedded/cookbooks/private-chef/providers/pg_cluster.rb' before installing new version: Operation not permitted
[...]

PR #1583, opened 2018-10-18
Update Chef 14.4 -> 14.5

tas50:
We might as well stay current here. I fixed a license string to be SPDX compliant while I was in the gemspec and nuked a double dep definition between the gemspec and gemfile.


markan:
Glitch building rerunning as #176

tas50:
@markan Did this end up passing in jenkins?

markan:
@tas50 Nope, another glitch. I rebased on master and I'm rerunning.

PR #1589, opened 2018-11-07
Use the embedded env when running escript

stevendanna:
This patch has been running regularly on a branch but had not yet
been ported over to here.

Signed-off-by: Ryan Cragun <ryan@chef.io>


stevendanna:
I went ahead and merged this.  Since none of the CI in this project tests this code, it is unlikely that CI issues were related to this change.

PR #1590, opened 2018-11-07
Add SSL configuration parameters to pedant

stevendanna:
This adds SSL-related configurables to pedant.  

This fix has been running regularly as part of a Habitatized Chef server
built from a branch containing a collection of patches.  Moving this patch over
will help us get rid of that branch.


stevendanna:
I rebased this and ran it through wilson.  It seems to be failign on RHEL but I'm not sure it is related to this PR, still investigating: http://wilson.cd.chef.co/job/chef-server-test/375/

stevendanna:
Issue cleared up on retry: http://wilson.cd.chef.co/job/chef-server-test/376/

Issue #183, opened 2015-05-01
Wrong configuration file referenced in error message for missing server section in chef-client.rb


stevendanna:
From @kevindickerson in chef/opscode-omnibus#723:

> This raise [1] produces a RuntimeError message. The message is incorrect. The message has not been updated to reflect the new location for chef-server.rb for Chef 12.

> *Incorrect*
> The RuntimeError message directs users to look at /etc/opscode/private-chef.rb.

> *Correct*
> The RuntimeError message directs users to look at /etc/opscode/chef-server.rb.

> References
> 1. https://github.com/chef/opscode-omnibus/blob/12.0.5/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb#L390

PR #1591, opened 2018-11-07
Server admin mutual TLS support

stevendanna:
### Description

This is built on top of `ma/chef-server-ctl-config` and should be 
merged after. The relevant commit for review is https://github.com/chef/chef-server/commit/00c7458172f91515abbb9c32162a8e7cb2201070

This fix has been running regularly as part of a Habitatized Chef server
built from a branch containing a collection of patches.  Moving this patch over
will help us get rid of that branch.

Our Habitatized Chef Server users mutual TLS authentication between all
components.  This fixes various calls inside the chef-server-ctl commands
to pass relevant TLS options during bifrost API calls.

This is a separate PR from ma/chef-server-ctl-config only to avoid invalidating the testing that has been done on that branch.


markan:
Running tests as ad-hoc #183

PR #1593, opened 2018-11-09
Trigger nightly build with Expeditor

schisamo:
Now that we removed the nightly trigger from the Jenkins pipeline (per
chef-cookbooks/opscode-ci#936) we need to update the Expeditor config to
begin triggering a nightly build of the project. This build also expires
the Omnibus cache so we can be sure Chef Server builds correctly even if
some transitive deps have been udpated.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -62,6 +62,13 @@ habitat_packages:
       export:
         - docker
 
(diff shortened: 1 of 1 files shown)


markan:
Rebased, should be good to merge.

PR #1594, opened 2018-11-13
Update loofah and rubyzip gems

tas50:
Address CVEs in both of these deps.


markan:
Replaced by #1660

PR #1601, opened 2018-12-18
Add note about out of date dep to trigger a rebuild

btm:
openresty is built against core/openssl/1.0.2n/20180608102213, latest is 1.0.2q


btm:
let's see what this gets us built.

PR #1621, opened 2019-01-31
Remove duplicate rubygems pin and roll back to 2.7.X

tas50:
This version was being pinned twice. No need for that.

### DIFF
--- a/omnibus_overrides.rb
+++ b/omnibus_overrides.rb
@@ -4,8 +4,8 @@
 #
 override :erlang, version: "18.3.4.9"
 override :lua, version: "5.1.5"
-override :rubygems, version: "3.0.2"
-override :bundler, version: '~> 1.17'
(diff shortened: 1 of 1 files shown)


tas50:
@markan It looks like we weren't actually running rubygems 3.x then because that code is not 3.x ready. Let me update this to just pin to 2.7.8 and we can get things over to 3.x later.

markan:
This is ready to merge once we fix the failing tests in travis. Don't think its related to this PR but haven't had a chance yet to confirm.

markan:
After rebasing it looks like the only thing left is the pin back to rubygems 2.7.X; I assume we still want to do that, right?

tas50:
I've entirely forgotten why we had to roll back to 2.x.

markan:
I think we don't need this anymore, closing.

PR #1626, opened 2019-02-11
Replace Jenkins omnibus build/test pipeline with Buildkite

jeremiahsnapp:
Signed-off-by: Jeremiah Snapp <jeremiah@chef.io>


jeremiahsnapp:
@tduffield the buildkite builders are ephemeral so there isn't a cache in the first place so these adhoc builds are actually similar to the "expire cache adhoc builds".

@schisamo thanks for catching that "promote to current" code in the delivery cookbook. it's removed now.

PR #1633, opened 2019-03-01
Use expeditor's default promote config

jeremiahsnapp:
There's no need to explicitly define Expeditor's default promote config so let's remove it.


jeremiahsnapp:
Closing this because we prefer to keep it explicitly defined.

PR #1637, opened 2019-03-07
Force a habitat rebuild to pull in new openssl

stevendanna:
I forgot to set the Build All tag on the last attempt to do this.


cookbooks/chef-server-deploy/recipes/omnibus_standalone.rb
```diff
@@ -147,7 +147,7 @@
   default_type application/json;
   return 200 '#{oc_id_applciation_config('supermarket')}';
 }
-EOF
+       EOF
```
stevendanna: The linter cannot be more wrong about this, but OK.

PR #1645, opened 2019-03-15
Update gitignore

markan:


markan:
Merged as trivial change.

PR #1649, opened 2019-03-26
Double the erchef max_request_size

irvingpop:
### Description

Nearly every Chef customer needs to increase this erchef `max_request_size` if they perform inspec scanning via the audit cookbook.  Apparently you can't use many of our CIS profiles out of the box without encountering the dreaded `413 Request entity too large` error.

An original motivation for capping this value was anemic Solr performance, which is no longer an issue since we've switched to Elasticsearch.

### Issues Resolved

We have more than enough support ticket evidence to support doubling this value by default, as support team starts by doubling the value and often going up to 5MB.

### Check List

- [ ] All tests pass


irvingpop:
well the good news? is that [our docs](https://docs.chef.io/install_server_pre.html) don't even mention CPU requirements anymore.  which is probably good because CPU power has significantly increased since the original requirements were written 😄

irvingpop:
Crazy idea:   How hard would it be in erchef to modify the error message "Request Entity Too Large" to be more informative?    It could be as simple as instructing the user which value to adjust, or as complex as telling the user the body size and recommending a value.

markan:
It looks like we already create a message here: https://github.com/chef/chef-server/blob/ac11df3d7c88b164a16beccdd18c12d403e87967/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_enforce.erl#L59 and pass it through here: https://github.com/chef/chef-server/blob/651ef2aa96cee20619c82e80236b350513bf64e0/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl#L888

Not sure if it's being dropped somewhere, but we do at least mention the size. We certainly could enhance that to mention the config values requiring adjustment.

irvingpop:
I think the message might be getting dropped/replaced by nginx when it proxies.   so maybe this is a line of work we can undertake at a different time.

irvingpop:
thanks @markan !   I'll PR this over to docs as well

PR #1651, opened 2019-03-30
Allow configuration of local rabbitmq acctions_password

Annih:
### Description

Since the code has been refactored (#1128) to allow configuration of external rabbitmq credentials, there was no more way to configure the ones for a local rabbitmq.

This code leverage the [add_secret helper](https://github.com/chef/chef-server/blob/5c7c988cba529c4794b21faa9c6bd135f37f47c8/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb#L553) that read the chef-server.rb config first, and in case of missing secret generate a new one.

One can now define the rabbitmq actions_password for the local rabbitmq by setting below line in their /etc/opscode/chef-server.rb

```
  rabbitmq['actions_password'] 'very_secure_local_rabbitmq_password'
```
### Check List

- [ ] All tests pass


markan:
Thanks for your submission! We'll review this and get back to you in the next few days.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -637,7 +637,7 @@ def generate_rabbit_actions_password
                         value: PrivateChef["external_rabbitmq"]["actions_password"],
                         frozen: true, force: true)
       else
-        credentials.add("rabbitmq", "actions_password", length: 100)
+        add_secret({group: "rabbitmq", name: "actions_password", length: 100}, true)
```
lamont-granquist: i think you can use `add_secret(group: "rabbitmq", name: "actions_password", length: 100)` here

markan:
So could you provide a little more details about your use case? I take it that letting chef-server manage that secret is not ok. Are you using the chef-server rabbitmq for some other app, and need to manage the password so that the app can access? Or are you running a non chef-server managed rabbitmq that you need to access, but it doesn't fit well with how we modeled the external rabbitmq work? 

In today's review meeting we had a couple of thoughts. 
1) Is is there a way to either treat your local rabbitmq as an external one in chef-secrets? That would work if the local rabbitmq was completely outside of chef-server's control. 
2) Otherwise does it make sense to extend the command here (https://github.com/chef/chef-server/blob/master/src/chef-server-ctl/plugins/secrets.rb#L9)  to better support your use case?

The goal is to support your use case while using the chef-secrets infrastructure and keeping chef-server.rb free of any secrets. 

Thanks again!

Annih:
> So could you provide a little more details about your use case? I take it that letting chef-server manage that secret is not ok. Are you using the chef-server rabbitmq for some other app, and need to manage the password so that the app can access? Or are you running a non chef-server managed rabbitmq that you need to access, but it doesn't fit well with how we modeled the external rabbitmq work?

I'm currently refactoring - using Chef - the way we setup our Tiered chef servers. So the rabbitmq is local on my backend server, and I need to configure the same action password on the frontend. I know the documentation says "you should sync the whole content of /etc/opscode", but IMHO this is not the Chef way to do it.

> Is is there a way to either treat your local rabbitmq as an external one in chef-secrets? That would work if the local rabbitmq was completely outside of chef-server's control.

Yes I think I could configure the backend's RabbitMQ as an external one.

> Otherwise does it make sense to extend the command here (https://github.com/chef/chef-server/blob/master/src/chef-server-ctl/plugins/secrets.rb#L9)  to better support your use case?

Why not.

> The goal is to support your use case while using the chef-secrets infrastructure and keeping chef-server.rb free of any secrets.

I was not aware of your plan to remove all secrets from `chef-server.rb`, so I can now understand your concern.
Do you plan to also remove the `private-chef-secrets.json` file? To be honest I could configure this file directly, but I don't see why it is better to configure a *json* file instead of *rb* one, in either way the passwords are readable.

In the other hand, I don't understand why all other secrets are configurable via this file, but the rabbitmq action password is not.
Even if you don't recommend using the `chef-server.rb` file to store password, could it be possible to have all passwords handled the same way?

Annih:
So @markan any thought? I know it may not be the best patch matching the direction Chef is going, but don't you think it is OK to have all settings managed the same way at least?

Annih:
@lamont-granquist or @markan sorry to ping you again on this, but is your "final" opinion on that review?
I would prefer a "no" as an answer than no answer, don't hesitate to close this PR.

lamont-granquist:
(i have zero expertise on this issue, best I could do was that ruby style tip, i can't give a yes or a no)

PrajaktaPurohit:
@annih We apologize for the delay in response to your Pull Request. We need to do some code archeology to figure if that queue is still in use, in which case I would be more than happy to merge your PR. Thank you for contribution to Chef Infra Server.

PrajaktaPurohit:
Followup: chef-actions queue on rabbitmq might be used only in chef actions which is currently deprecated.
More investigation needed to confirm this.

PR #1656, opened 2019-04-03
Force rebuild to update ruby

jaym:
Signed-off-by: Jay Mundrawala <jmundrawala@chef.io>


jaym:
@tduffield #1654 seems to need the change the license_scout.sh

tduffield:
@jaym not when it is rebased on #1655.

PR #1658, opened 2019-04-03
Get habitat build working with changes from #1654

jaym:
The pwd for building changed with #1654. Hopefully
this makes it not matter

Signed-off-by: Jay Mundrawala <jmundrawala@chef.io>

### DIFF
--- a/src/bookshelf/habitat/plan.sh
+++ b/src/bookshelf/habitat/plan.sh
@@ -23,11 +23,19 @@ pkg_binds_optional=(
   [chef-server-ctl]="secrets"
 )
 
+# set_pwd gives a consistent current working directory when building
+# foo/bar/habitat/ vs foo/bar
+set_pwd() {
+    cd "$PLAN_CONTEXT/../"
+}
+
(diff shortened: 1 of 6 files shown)


src/bookshelf/habitat/plan.sh
```diff
@@ -23,11 +23,19 @@ pkg_binds_optional=(
   [chef-server-ctl]="secrets"
 )
 
+# set_pwd gives a consistent current working directory when building
```
markan: I admit I'm a little unclear about the impact of this; why is it necessary? How does it affect local package builds (non-expeditor)

jaym: I think tom is going to do something slightly different. Basically, with the new build, it's doing the equivalent of `build src/bifrost/habitat` instead of what we were doing before `build src/bifrost`. Hab sets different working directories for these, and our plans relied on the working directory to work correctly, so they don't work with the new build changes

jaym:
@tduffield is doing stuff to fix this

PR #1663, opened 2019-04-25
[nginx] Don't log 404s to the error log

stevendanna:
In the Chef API, 404s are typically not errors. Rather they are often
the expected response about an object that doesn't exist.

The log_not_found directive stops nginx from logging 404s as errors;
however, they should still be logged to the request log
(i.e. access.log).

References:

- http://nginx.org/en/docs/http/ngx_http_core_module.html#log_not_found


stevendanna:
Note: This is untested.  Opening this to continue the discussion around 404 logging in the Chef Server more centrally.

markan:
This looks pretty reasonable. Just kicked off a build in CI: 
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/125#6c27fc3f-20ea-4353-a57b-cc07c378b5bb

PR #1665, opened 2019-05-06
render profiles and data-collector upstreams correctly

sdelano:
when setting the configuration value `insecure_addon_compat`, the
chef-server library code does not render the secrets from the
secrets helper into the PrivateChef hash, which results specifically
in `data_collector['token']` not showing up there. when configured
to proxy compliance profiles and the data-collector endpoint, however,
the nginx recipe was relying on this field being present for the
upstreams to be rendered into the config. this resulted in us never
rendering the correct upstreams in `insecure_addon_compat` mode.

this change makes it so that the profiles and data-collector upstreams
will always be rendered if their root_url is configured. if, by
chance, the user doesn't have the `data_collector token` secret set,
we'll now see a 401 response code and an error message instead of a
confusing and misleading 404.

Signed-off-by: Stephen Delano <stephen@chef.io>

### Issues Resolved

Internal support issue: https://getchef.zendesk.com/agent/tickets/21598

### Check List

- [ ] All tests pass


markan:
Running CI as https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/141#617d6407-2bdf-434f-8b66-5d2eb7ecc2d1. I'll merge if it's good.

PR #1671, opened 2019-05-31
Ma/use specific path

markan:
### Description

Pedant uses two methods to find it's chef version
1) is to require chef/version
2) is to run knife and pattern match the version string

Closes #1669 

### Issues Resolved

### Check List

- [x] New functionality includes tests
- [x] All tests pass

### DIFF
--- a/oc-chef-pedant/Gemfile
+++ b/oc-chef-pedant/Gemfile
@@ -7,6 +7,11 @@ gem 'pry-byebug'
 gem 'pry-stack_explorer'
 gem 'rake'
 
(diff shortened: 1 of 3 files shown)


markan:
@lamont-granquist will including chef in the Gemfile cause troubles?

markan:
Had some discussion with @lamont-granquist about whether the knife command and the path fixing is necessary. It appears it's needed for some of the dev-vm magic (at least things failed there when I did a local build). Not clear why, but I'm going to merge this now to fix builds and revisit later.

PR #1672, opened 2019-06-01
Ma/license check

markan:
### Description
Note, based on ma/update_runit, because chef-15 requires that.

This updates to chef-15, which requires license acceptance, and adds a license acceptance requirement to chef-server-ctl reconfigure.

### Issues Resolved

[List any existing issues this PR resolves, or any Discourse or
StackOverflow discussion that's relevant]

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [x] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #1674, opened 2019-06-05
Ma/license check

markan:
### Description

This updates to chef-15, which requires license acceptance, and adds a license acceptance requirement to chef-server-ctl reconfigure.

This reprises #1672 which was closed because of a github oops. (Don't delete the base branch of a live PR.)

### Issues Resolved

### Check List

- [ ] All tests pass


omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -288,7 +288,7 @@ def generate_hash
       results["private_chef"]["logs"]["log_retention"] = PrivateChef["log_retention"]
       results["private_chef"]["logs"]["log_rotation"] = PrivateChef["log_rotation"]
       results["private_chef"]["dark_launch"] = PrivateChef["dark_launch"]
-      results["private_chef"]["opscode-erchef"]["max_request_size"] = PrivateChef["opscode_erchef"]["max_request_size"]
+#      results["private_chef"]["opscode-erchef"]["max_request_size"] = PrivateChef["opscode_erchef"]["max_request_size"]
```
marcparadise: This doesn't seem related?

omnibus/omnibus-test.sh
```diff
@@ -61,7 +61,7 @@ y/8SReCpC71R+Vl6d4+Dw6GFdL+6k6W558dPfq3UeV8HPWQEaM7/jXDUKJZ0tB6a
 -----END DH PARAMETERS-----
 " | sudo tee /etc/opscode/dhparam.pem
 
-sudo chef-server-ctl reconfigure || true
+sudo chef-server-ctl reconfigure --chef-license accept || true
```
marcparadise: some cases we're doing `--chef-license accept` and some `--chef-license=accept`; could you make it consistently one way or the other?

tyler-ball: Here I would do `accept-no-persist` because I'm thinking this test runs on a CI environment. I would rather not leave license acceptance marker files around on CI nodes.

omnibus_overrides.rb
```diff
@@ -7,9 +7,11 @@
 override :rubygems, version: "3.0.2"
 override :bundler, version: '~> 1.17'
 override :'omnibus-ctl', version: "master"
-override :chef, version: "v14.11.21"
-override :ohai, version: "v14.8.11"
+override :chef, version: "v15.0.269"
```
tyler-ball: If we're using Chef 15 I think we should use ruby 2.6.3, rubygems and bundler 1.17.2 3.0.3 based on the [deps that ship with our chef omnibus packages](https://github.com/chef/chef/blob/d810b363da8ccdbed92ce92419d6c6e7f2f0835e/omnibus_overrides.rb)

src/chef-server-ctl/bin/chef-server-ctl
```diff
@@ -233,6 +234,9 @@ EOM
 
     # Override reconfigure to skip license checking
     def reconfigure(*args)
+
+      LicenseAcceptance::Acceptor.check_and_persist!("infra-server", ChefServerCtl::VERSION.to_s)
```
tyler-ball: Do we have access to the config from the user specified `omnibus.rb` here? If so I think we should provide that to `check_and_persist!` [like we do in chef-client](https://github.com/chef/chef/blob/d810b363da8ccdbed92ce92419d6c6e7f2f0835e/lib/chef/application.rb#L258). This gives users another way to accept the license - by adding `chef_license "accept"` into their `omnibus.rb`.

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
markan: This is needed for chef 15; issue fixed a different way in the end.

src/chef-server-ctl/bin/chef-server-ctl
markan: We don't have an omnibus.rb visible; but maybe we should look at adding it to the /etc/opscode/chef-server.rb config file.

omnibus/omnibus-test.sh
markan: Fixed

omnibus_overrides.rb
markan: Fixed

src/chef-server-ctl/bin/chef-server-ctl
markan: Unfortunately, we're in a bit of a chicken and egg situation with the chef-server.rb config; the parsing of that file happens inside the private-chef cookbook, so it's not accessible here until after the first successful reconfigure.

markan:
Latest version is running in CI as https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/226

src/chef-server-ctl/bin/chef-server-ctl
tyler-ball: I think `chef-server.rb` was what I was thinking of. If it is not available here that is fine by me. Potentially a future improvement we could make, although it sounds like it would take a lot of refactoring.

src/chef-server-ctl/Gemfile.lock
```diff
@@ -46,100 +47,122 @@ GEM
   specs:
     addressable (2.6.0)
       public_suffix (>= 2.0.2, < 4.0)
-    appbundler (0.12.0)
+    appbundler (0.13.0)
```
tyler-ball: Looks like the current build failures are because 0.13.0 got yanked from Rubygems. @lamont-granquist Do you know if this is expected?

PR #1691, opened 2019-06-26
[chef-server-ctl] Fix rabbitmq fail on first reconfigure

### LINKED ISSUE #1690: Rabbitmq fails to restart on first reconfigure
We have an intermittent failure on the first run of chef-server-ctl. Rerunning always seems to fix the problem. The failure is late enough in the first run that we're just running triggered restarts, and all configuration is done. 

### Chef Server Version
Chef Server 13.0.14

### Platform Details
Ubuntu 16.04, in vagrant/virtualbox (also seems to intermittently repro on AWS m4.large)

### Configuration
Standalone,  & Tiered, HA. New install or upgrade.

### Scenario:
Install chef server

### Steps to Reproduce:
dpkg -i chef-server-core_13.0.14-1_amd64.deb 
chef-server-ctl reconfigure --verbose  --chef-license accept

### Expected Result:
chef-server-ctl to succeed

### Actual Result:

      
---- Begin output of /opt/opscode/embedded/bin/sv restart /opt/opscode/service/rabbitmq ----

      
STDOUT: timeout: run: /opt/opscode/service/rabbitmq: (pid 7674) 68s

    
Full dump here (this is from a dev-vm but we can reproduce elsewhere)
https://gist.github.com/markan/ac0dbcf8ed9a9fab968868fbbff9bfcd

markan:
### Description

As described in issue #1690, the first run of `chef-server-ctl
reconfigure` fails with a timeout at the very end of run, when executing delayed restarts:

    STDOUT: timeout: run: /opt/opscode/service/rabbitmq: (pid 7674) 68s

Successive runs of reconfigure work fine.

This appears to be a problem buried deep inside the rabbitmq restart logic; we weren't able to make much progress with the root cause.

As an experiment were replacing the restart action on the rabbitmq service with stop, then start actions.

If this doesn't work, a few other options suggest themselves.

We can add a execute block with default action nothing that aggressively shuts down rabbitmq (normal shutdown followed by pkill if that doesn't work), then waits, then starts up rabbitmq again. Instead of sending a restart action (or the stop/start here) we send the execute action to this block.

The sv_timeout parameter might also be used to lengthen the timeout before we fail. However that seems less than ideal if we're really stuck; we could just hang longer before failing.

See also:

https://wiki.openstack.org/wiki/OpsGuide-Maintenance-RabbitMQ#RabbitMQ_service_hangs

### Issues Resolved

#1690 

### Check List

- [ ] All tests pass

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
@@ -97,6 +97,28 @@ component_runit_service "rabbitmq" do
   control ['t']
 end
 
+# Does this belong inside the is_data_master block below?
+# Walk of shame moment. We're not sure why the graceful stop of
+# rabbitmq is failing. Things seemed to work in runit 4.1.1, but it's
+# not working in 5.1.1. Instead we are timing out during stop/restart
+# actions This is a workaround to ensure we are actually restarting
+# the process in a clean-ish way. Hopefully we will either figure out
+# the root cause or remove rabbitmq soon.
+#
+script 'hard_kill_rabbitmq' do
+  interpreter "bash"
+  code <<-EOH
+timeout 60 /opt/opscode/embedded/bin/sv stop /opt/opscode/service/rabbitmq
+if [ $? -ne 0 ] 
+then
+   kill --term `/opt/opscode/sv/rabbitmq/supervise/pid`
+fi
+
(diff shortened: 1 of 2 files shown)


markan:
This seems to fix the issue when run locally, but needs wider testing

PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/255#7385705e-e2b5-4912-a326-0e7b2d5c0e3d

PR #1694, opened 2019-07-01
Remove the old delivery directory

tas50:
We're not using delivery anymore so we shouldn't need this.


stevendanna:
This was added by releng and is explained here:

https://github.com/chef/chef-server#acc-pipeline--chef-expeditor

and I believe the goal was to replace the qa-chef-server stuff with it: https://github.com/chef/chef-server/blob/master/RELEASE_PROCESS.md#testing-the-release

I'm personally fine removing this as I think that any process that depends on ACC should probably not be long for this world and that effort is better spent elsewhere.

That said, if we are going to remove this then we should probably also remove `terraform/`, ` cookbooks/chef-server-deploy/`, and `inspec/`.

PrajaktaPurohit:
Captured the issue in : https://github.com/chef/chef-server/issues/1713

PR #1696, opened 2019-07-02
Update yard to current

tas50:
There's a CVE that doesn't impact us, but is showing up in GitHub.


markan:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/265

PR #1701, opened 2019-07-09
Turn off smart builds

jaym:
While the manifest generated ties together a release of automate, it is
still confusing to have components at different versions for chef
server. Also, automate depends on openresty from chef server, which
doesn't get changed often thus never picks up changes upstream.


stevendanna:
A full build takes 13 minutes: https://buildkite.com/chef/chef-chef-server-master-habitat-build/builds/32#49c9309d-4aa6-4fbc-ae88-5cdbc26dd8a4

Seems worth it for the predictability of having everything build.

PR #1705, opened 2019-07-11
Bump bundler to 1.17.3

markan:
### Description

We now enforce there being only one version of bundler in
ruby-cleanup, but the omnibus def differers from what we pull in
elsewhere.

### Issues Resolved

Fixes broken builds (https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/279#344a6234-8a0e-468f-8317-10e30cef8dcf)

### Check List

- [ ] All tests pass

### DIFF
--- a/omnibus_overrides.rb
+++ b/omnibus_overrides.rb
@@ -5,7 +5,7 @@
 override :erlang, version: "18.3.4.9"
 override :lua, version: "5.1.5"
 override :rubygems, version: "3.0.3"
-override :bundler, version: "1.17.2" # currently pinned to what ships in Ruby to prevent double bundler
+override :bundler, version: "1.17.3"
 override :'omnibus-ctl', version: "master"
 override :chef, version: "v15.0.300"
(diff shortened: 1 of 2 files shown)


omnibus_overrides.rb
```diff
@@ -5,7 +5,7 @@
 override :erlang, version: "18.3.4.9"
 override :lua, version: "5.1.5"
 override :rubygems, version: "3.0.3"
-override :bundler, version: "1.17.2" # currently pinned to what ships in Ruby to prevent double bundler
+override :bundler, version: "1.17.3" # currently pinned to what ships in Ruby to prevent double bundler
```
tas50:
```suggestion
override :bundler, version: "1.17.3"
```

PrajaktaPurohit: I like having the comment in there, just so that our future selves know easily what that change was made, but that can be put as the commit message perhaps?

PR #1725, opened 2019-08-20
WIP Lbaker/ssl internal postgres

lbakerchef:
### Description

Enable SSL configuration between chef-server and an internal PostgreSQL instance.

### Issues Resolved

### Check List

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -691,7 +691,10 @@ default['private_chef']['postgresql']['archive_timeout'] = 0 # 0 is disabled.
 # allowable choices come from the PGSSLMODE environment var for libpq.
 # supported sslmode values: disable | require
 default['private_chef']['postgresql']['sslmode'] = 'disable'
-
+# for now, copy these by hand into the directory, `chmod 0600 *.key; chown root *.key`
+# later, this will be automated and these comments will be removed
+default['private_chef']['postgresql']['ssl_cert_file'] = "#{default['private_chef']['postgresql']['data_dir']}/api.chef-server.dev.crt"
+default['private_chef']['postgresql']['ssl_key_file']  = "#{default['private_chef']['postgresql']['data_dir']}/api.chef-server.dev.key"
 # This is based on the tuning parameters here:
 #
 #  https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -693,8 +693,8 @@ default['private_chef']['postgresql']['archive_timeout'] = 0 # 0 is disabled.
 default['private_chef']['postgresql']['sslmode'] = 'disable'
 # for now, copy the nginx certs by hand into the postgres data directory, chmod and chown
 # later, this will be automated and these comments will be removed
-default['private_chef']['postgresql']['ssl_cert_file'] = "#{default['private_chef']['postgresql']['data_dir']}/#{node['fqdn']}.crt"
-default['private_chef']['postgresql']['ssl_key_file']  = "#{default['private_chef']['postgresql']['data_dir']}/#{node['fqdn']}.key"
+default['private_chef']['postgresql']['ssl_cert_file'] = "#{node['private_chef']['postgresql']['data_dir']}/#{node['fqdn']}.crt"
+default['private_chef']['postgresql']['ssl_key_file']  = "#{node['private_chef']['postgresql']['data_dir']}/#{node['fqdn']}.key"
 # This is based on the tuning parameters here:
 #
 #  https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -663,8 +663,7 @@ default['private_chef']['postgresql']['version'] = "9.6"
 default['private_chef']['postgresql']['enable'] = true
 default['private_chef']['postgresql']['external'] = false
 default['private_chef']['postgresql']['ha'] = false
-default['private_chef']['postgresql']['dir'] = "/var/opt/opscode/postgresql/#{node['private_chef']['postgresql']['version']}"
-default['private_chef']['postgresql']['data_dir'] = "/var/opt/opscode/postgresql/#{node['private_chef']['postgresql']['version']}/data"
+default['private_chef']['postgresql']['dir'] =           "/var/opt/opscode/postgresql/#{node['private_chef']['postgresql']['version']}"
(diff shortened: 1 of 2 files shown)


omnibus/files/private-chef-cookbooks/private-chef/templates/default/postgresql.conf.erb
```diff
@@ -77,7 +77,9 @@ max_connections = <%= node['private_chef']['postgresql']['max_connections'] %>
 # - Security and Authentication -
 
 #authentication_timeout = 1min		# 1s-600s
-#ssl = off				# (change requires restart)
+ssl = <%= (node['private_chef']['postgresql']['sslmode'] == 'disable' ? 'off' : 'on') + ' # (change requires restart)' %>
+ssl_cert_file = <%= "'#{node['private_chef']['postgresql']['ssl_cert_file']}' # (change requires restart)" %> %>
+ssl_key_file  = <%= "'#{node['private_chef']['postgresql']['ssl_key_file']}'  # (change requires restart)" %> %>
```
PrajaktaPurohit: might be useful to add conditionals around these to be set only if the sslmode is set to require.

PrajaktaPurohit: #(change requires restart) seems to be a comment on the line. Not needed to be included in the value of the keys for ssl, ssl_cert_file or ssl_key_file

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -691,7 +691,10 @@
 # allowable choices come from the PGSSLMODE environment var for libpq.
 # supported sslmode values: disable | require
 default['private_chef']['postgresql']['sslmode'] = 'disable'
-
+# for now, copy these by hand into the directory, `chmod 0600 *.key; chown root *.key`
+# later, this will be automated and these comments will be removed
+default['private_chef']['postgresql']['ssl_cert_file'] = "#{default['private_chef']['postgresql']['data_dir']}/api.chef-server.dev.crt"
+default['private_chef']['postgresql']['ssl_key_file']  = "#{default['private_chef']['postgresql']['data_dir']}/api.chef-server.dev.key"
```
PrajaktaPurohit: would the `api.chef-server.dev.key` and `api.chef-server.dev.crt` exist by default?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/postgresql.conf.erb
lbakerchef: Originally I had conditionals, but removed them because it has no effect anyway if ssl=off, and the lines simply point to cert files which will already exist (after recipe is created to place them).

I can add the conditionals back if you like.  Let me know your thoughts.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
lbakerchef: Yes, my idea was to have a recipe place those files by default (similar to nginx certificates existing by default).

omnibus/files/private-chef-cookbooks/private-chef/templates/default/postgresql.conf.erb
PrajaktaPurohit: I think the pattern is to put #(change requires restart) as a comment, do we need it to be a part of that expression?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -663,8 +663,7 @@
 default['private_chef']['postgresql']['enable'] = true
 default['private_chef']['postgresql']['external'] = false
 default['private_chef']['postgresql']['ha'] = false
-default['private_chef']['postgresql']['dir'] = "/var/opt/opscode/postgresql/#{node['private_chef']['postgresql']['version']}"
-default['private_chef']['postgresql']['data_dir'] = "/var/opt/opscode/postgresql/#{node['private_chef']['postgresql']['version']}/data"
```
PrajaktaPurohit: I think this needs to be left in. This dir is an important attribute for postgres and gets used on multiple occassions.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -691,7 +690,10 @@
 # allowable choices come from the PGSSLMODE environment var for libpq.
 # supported sslmode values: disable | require
 default['private_chef']['postgresql']['sslmode'] = 'disable'
-
+# for now, copy the nginx certs by hand into the postgres data directory, chmod and chown
+# later, this will be automated and these comments will be removed
+default['private_chef']['postgresql']['ssl_cert_file'] = "#{default['private_chef']['postgresql']['dir']}/data/#{node['fqdn']}.crt"
```
PrajaktaPurohit: you might have to use the node attributes instead.
`default['private_chef']['postgresql']['ssl_cert_file'] = node['private_chef']['postgresql']['data_dir']/#{node['fqdn']}.crt` (assuming that is the name of the cert)

omnibus/files/private-chef-cookbooks/private-chef/templates/default/postgresql.conf.erb
PrajaktaPurohit: until the recipe to add the default certs exists, I would prefer adding conditionals.
a) to check if sslmode is require and
b) to confirm that those files exist.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
lbakerchef: It was accidentally removed.  I already added these back in yesterday, and pushed :)

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
lbakerchef: Done.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/postgresql.conf.erb
lbakerchef: The comment only needs to be a part of the expression if we want the comment rendered to the output file.  Anyway, I removed it.

Conditionals added.  Will add code to check if files exist.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
lbakerchef: By the way, maybe I've forgotten the rationale.  Why do we want or need custom-nameable and custom-locateable certs (ssl_cert_file and ssl_key_file) for internal postgresql?  What's the use case?

lbakerchef: To answer my own question above, I found this from @markan lurking in my notes:
```
I think there’s two use cases to cover
1) Self signed certs of some sort; by default I think we generate one for nginx, we should reuse it for postgres.
2) Externally generated certs (the user has a cert for their host generated however they like)
In both cases we probably should follow the path that we use for nginx, and the same certs should be used for both nginx and postgres.
```

PrajaktaPurohit:
Closing this for now. Can reopen when needed.

PR #1748, opened 2019-09-19
upgrade rebar3 to 3.6.2

lbakerchef:
### Description

We would like to update the rebar3 dependency to the latest. However 3.12.0 fails in the oc_erchef unit tests die to issue mentioned below.

Updating to 3.6.2 until it is fixed.

### Task Outline

- [x] investigating build error
seems to be suffering from: https://github.com/erlang/rebar3/issues/1985
dropped back to 3.11.1 - still fails
trying 3.6.2
- [x] rebase onto master


markan:
The errchef_ct test failure in buildkite might be relevant.

This isn't really a reviewable PR obviously, but if it passes the verify tests and you can build an omnibus package with it, it's good to merge.

lbakerchef:
Yup, I saw that build failure and plan to investigate it.  I'm guessing it's something small and easy to fix.

lbakerchef:
Investigating...

PrajaktaPurohit:
looks like it is a bug since Dec 2018 (from the tag 3.7.0) Until there is a fix, https://github.com/erlang/rebar3/releases/tag/3.6.2 might be the most we can update to.

lbakerchef:
trying now

PrajaktaPurohit:
Passing adhoc build: https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/584

PR #1750, opened 2019-09-19
Add copy/paste execution strings for convenient builds

lbakerchef:
Paste this into your shell for convenient builds!

```cd && . load-omnibus-toolchain.sh && cd chef-server/omnibus && sudo chown -R vagrant ~/.bundle/  && bundle install --binstubs && bin/omnibus build chef-server -l internal```

### DIFF
--- a/omnibus/README.md
+++ b/omnibus/README.md
@@ -182,7 +182,14 @@ $ kitchen login ubuntu-1604
 ...
 [vagrant@ubuntu...] $ bin/omnibus build chef-server -l internal
 ```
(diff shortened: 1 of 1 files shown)


PrajaktaPurohit:
@lbakerchef - this is helpful only first time the build is run. For all the subsequent builds we would run only the last command. It would help to get some clarification around that.

lbakerchef:
For the lazy approach, you can just paste the same thing and it still works (it's what I do).  I think it's 'idempotent' so runs through the first part real quick.  "One execution string to rule them all and bind them."  But I will add some clarification.

PR #1767, opened 2019-09-24
Update Chef Infra Client to 15.3

tas50:
Update Chef Infra Client to 15.3 in all the places I could find it. This also switches from chef_backup supplied via git to chef_backup supplied via gem. I updated chef_backup to allow for mixlib-shellout 3.x which was blocking new Chef Infra Client releases and I set it up in Expeditor / Buildkite. We should just ship it if there's additional blockers in the future. Sourcing it via git increases the size of the installer as we pull in all the git repo files vs. the smaller subset of files we ship in the gem.


PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/468

PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/470#c9ae27ca-d225-40e0-b1e2-2c084a1bdc1a <-passing adhoc build.

PR #1774, opened 2019-10-01
Bump omnibus chef-infra to 15.4.20

christopher-snapp:
chef-server builds began breaking when chef/omnibus-software#1099 introduced an appbundle of the inspec binary.

The issue is that chef-server is pinned to the latest stable release of chef-infra (15.3.14) which is dependent on `inspec-core-bin` (4.16.0) that causes the following error:

```
There was an error parsing `Gemfile`: There are no gemspecs at /opt/opscode/embedded/lib/ruby/gems/2.6.0/gems/inspec-core-bin-4.16.0.  Bundler cannot continue.
```

This fix should be considered temporary and we should ensure it gets updated to the next stable version of chef infra when possible.

### DIFF
--- a/omnibus_overrides.rb
+++ b/omnibus_overrides.rb
@@ -7,7 +7,7 @@ override :lua, version: "5.1.5"
 override :rubygems, version: "3.0.3"
 override :bundler, version: "1.17.2"  # pin to avoid double bundle error
 override :'omnibus-ctl', version: "master"
-override :chef, version: "v15.3.14"
+override :chef, version: "v15.4.20"
 override :ohai, version: "v15.3.1"
 override :ruby, version: "2.6.3"
 


christopher-snapp:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/487

christopher-snapp:
> We should bump it a few other places as well. see: https://github.com/chef/chef-server/blob/master/doc/FrequentTasks.md#updating-chef-client

Chef Infra 15.4.20 isn't available on rubygems.

The thinking behind leaving the other places unmodified in Chef Server is that there isn't a stable release available that works (current stable is 15.3.14) with current Omnibus Software.

This PR was intended to fix the currently broken Omnibus builds with the intention being that I'd provide another PR that updates *all* the necessary dependencies in Chef Server once Chef Infra gets a new stable version.

PrajaktaPurohit:
@christopher-snapp could you create an issue to update to the latest chef-infra client as it becomes available and link it here? The change looks good. Thank you!

christopher-snapp:
Issue #1776 is tracking the necessary changes once a new stable release of chef infra becomes available.

Issue #1797, opened 2019-10-21
Testing ldap scenarios within the test matrix


PrajaktaPurohit:
This usecase will setup a scenario in the test infrastructure to test ldap setup.
1. With standalone Chef Infra Server
2. With tiered Chef Infra Server

REF: https://docs.chef.io/server_ldap.html

PR #1812, opened 2019-11-04
Add external openldap terraform scenario

christopher-snapp:
This PR adds an omnibus external openldap terraform scenario as requested in #1797 

### Description

This scenario instantiates an OpenLDAP Server followed by a Chef Infra Server utilizing an Omnibus built artifact downloaded from `$upgrade_version_url` as the install package.

The Chef Infra Server will receive a `/etc/opscode/chef-server.rb` configuration file that is setup to use the external LDAP server over TLS.

Once both servers are installed and configured, the pedant tests are run against the Chef Infra Server.

**_Note that smoke and pedant tests currently fail in this scenario._**

### Check List


terraform/scenarios/omnibus-external-openldap/files/chef-server.rb
```diff
@@ -0,0 +1,16 @@
...
+data_collector['token'] = 'foobar'
+
+profiles['root_url'] = 'http://localhost:9998'
+
+ldap['base_dn'] = 'ou=chefs,dc=chef-server,dc=dev'
+ldap['bind_dn'] = 'cn=admin,dc=chef-server,dc=dev'
+ldap['bind_password'] = 'H0\/\/!|\/|3tY0ur|\/|0th3r'
+ldap['host'] = 'ldap.chef-server.dev'
+ldap['login_attribute'] = 'uid'
+ldap['tls_enabled'] = true
```
PrajaktaPurohit: it might be worth making a note what happens / does not happen if ldap['ssl_enabled'] is set to true.

christopher-snapp: Ok, I'll toss a comment in the code.

I'm still confused by what the original intent of `ldap['ssl_enabled'] = true` was and whether it was only intended for Active Directory / Global Catalog.

Whenever I've set `ldap['ssl_enabled'] = true` against an OpenLDAP instance I only see the following error on the Chef Manage login screen:
```
The AD/LDAP server is down. Please contact an administrator. 
```

A TCP dump captured during the login attempt shows the connection being reset:
TCP dump

With `ldap['ssl_enabled'] = false` and `ldap['tls_enabled'] = true` we are able to connect to the OpenLDAP instance without issue:
tcpdump

PR #1816, opened 2019-11-05
Split up terraform scenarios' remote-exec scripts

christopher-snapp:
By splitting up the monolithic `remote-exec` scripts in the terraform
scenarios we make it easier to debug failures as each script is run
separately in its own SSH session.

### Check List


christopher-snapp:
> Very much like this change; any investment in debugability is always welcome. Out of curiosity in this approach is there a clear demarkation between the different provisioner remote-exec
> runs?

`Provisioning with 'remote-exec'` is a search string that will let someone walk through the different SSH sessions.

PR #1846, opened 2019-11-23
Remove Ruby 2.5.6 Pinning

christopher-snapp:
### Description

The build environment no longer manages multiple versions of Ruby.
   
By removing the pinning to a specific version of Ruby, we will always use the latest version supplied by the build environment.

Unfortunately, `oc-id` is built on Rails `4.2.11` which fails its verification tests with `ThreadError: already initialized`.

To allow Rails `4.2.11` to work with Ruby `>2.6.0` we need a small patch added via `src/oc-id/spec/support/monkeypatch.rb` to prevent rails from reusing `ActionController::TestResponse` object in tests.
    
### Check List

### DIFF
--- a/.expeditor/license_scout.sh
+++ b/.expeditor/license_scout.sh
@@ -5,11 +5,6 @@ set -eou pipefail
 if [[ "${EXPEDITOR:-false}" == "true" ]]; then
   apt-get update
   apt-get install -y libpq-dev libsqlite3-dev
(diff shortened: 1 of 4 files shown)


christopher-snapp:
An adhoc build with these changes can be found here:

https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/738

PR #1874, opened 2019-12-31
Add buildkite steps to the `integration_test` pipeline

### LINKED ISSUE #1742: Extend Chef Infra Server's build pipeline to include integration tests of multiple installation topologies.
## Current State

The Chef Infra Server product currently relies on manual testing of different installation topologies which may result in unstable release artifacts reaching customers. 

## Desired State

Buildkite steps should be added to the Chef Infra Server release pipeline following the standalone testers of the build artifact.  These Buildkite steps would leverage multiple Terraform scenarios each responsible for setting up a different topology for integration testing.

christopher-snapp:
### Description

This PR adds buildkite steps to the `integration_test` pipeline that execute specific terraform scenarios.

### Issues Resolved
#1742 

### Check List

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -110,6 +110,12 @@ subscriptions:
           post_commit: true
       - bash:.expeditor/purge_cdn.sh:
           post_commit: true
+  - workload: buildkite_build_canceled:{{agent_id}}:integration_test:*
+    actions:
+      - bash:.expeditor/integration_test.pipeline.sh
+  - workload: buildkite_build_failed:{{agent_id}}:integration_test:*
+    actions:
+      - bash:.expeditor/integration_test.pipeline.sh
 
 promote:
   actions:
--- a/.expeditor/integration_test.pipeline.sh
+++ b/.expeditor/integration_test.pipeline.sh
@@ -1,0 +1,292 @@
+#!/usr/bin/env bash
+
+prog='integration_test.pipeline.sh'
+action=
+
+# verify command dependencies
+[[ "$(command -v terraform)" ]] || error 'terraform command is not available'
+
+# verify environment variables
+[[ -z "$BUILDKITE_BUILD_NUMBER" ]] && error 'BUILDKITE_BUILD_NUMBER environment variable is required!'
+[[ -z "$BUILDKITE_LABEL" ]] && error 'BUILDKITE_LABEL environment variable is required!'
+
+# allow for environment override of build number
+[[ -z "$BUILD_NUMBER" ]] && BUILD_NUMBER="${EXPEDITOR_BUILD_NUMBER}"
+[[ -z "$BUILD_NUMBER" ]] && BUILD_NUMBER="${BUILDKITE_BUILD_NUMBER}"
+export BUILD_NUMBER
+
+# apply defaults if necessary
+[[ -z "$TF_VAR_build_prefix" ]] && TF_VAR_build_prefix="${BUILD_NUMBER}-"
+export TF_VAR_build_prefix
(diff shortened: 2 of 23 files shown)


.expeditor/config.yml
```diff
@@ -110,6 +110,12 @@ subscriptions:
           post_commit: true
       - bash:.expeditor/purge_cdn.sh:
           post_commit: true
+  - workload: buildkite_build_canceled:{{agent_id}}:integration_test:*
+    actions:
+      - bash:.expeditor/integration_test.pipeline.sh destroy
+  - workload: buildkite_build_failed:{{agent_id}}:integration_test:*
+    actions:
+      - bash:.expeditor/integration_test.pipeline.sh destroy
```
tduffield: Unfortunately, you cannot pass parameters into `bash` actions. The script will need to detect that it needs to destroy based on the environment variables passed in. I would suggest the `EXPEDITOR_BUILD_STATE` environment variable, which would equal either `canceled` or `failed`.

christopher-snapp: ah, ok sorry I didn't catch that!

I can work around that in the script using `EXPEDITOR_BUILD_STATE` as suggested.

On a side note can you verify that `EXPEDITOR_BUILD_NUMBER` would contain just the integer for the *failed | canceled* job?

christopher-snapp: I've updated the shell script to override the action based on `EXPEDITOR_BUILD_STATE`:

https://github.com/chef/chef-server/pull/1874/files#diff-03877f02b41fcff024599b3a33aff44bR158-R159

.expeditor/integration_test.pipeline.sh
```diff
@@ -0,0 +1,294 @@
...
+
+# allow for environment override of build number
+[[ -z "$BUILD_NUMBER" ]] && BUILD_NUMBER="${EXPEDITOR_BUILD_NUMBER}"
+[[ -z "$BUILD_NUMBER" ]] && BUILD_NUMBER="${BUILDKITE_BUILD_NUMBER}"
+export BUILD_NUMBER
+
+# apply defaults if necessary
+[[ -z "$TF_VAR_build_prefix" ]] && TF_VAR_build_prefix="${BUILD_NUMBER}-"
+export TF_VAR_build_prefix
+[[ -z "$TF_VAR_aws_contact" ]] && TF_VAR_aws_contact="${AWS_CONTACT:-releng}"
```
tduffield: The default contact shouldn't be releng -- it should be someone on the Chef Server team.

christopher-snapp: @tduffield I used `releng` as the default value as that is what would show up as the contact for the resources created during a buildkite run.  This seemed appropriate since the resources are being created in a VPC in the `chef-cd` account that a chef-server developer wouldn't have access to clean up if something went wrong.  My attempt was to follow a similar pattern to the instances that are normally created by other buildkite runs in the `delivered-default` VPC.
Would it be preferred to use `buildkite` as the contact tag rather than `releng`?  Or should I really just do `csnapp`?

tduffield: So the "owner" are the individuals who can provide context on how/why certain infrastructure is being used. While the infra may run in an account managed by releng, we don't "own" it per say. If anything in that situation we'd just be a middleman.

christopher-snapp: I've updated the contact to `chef-eng-team` as recommended by @OBrienCommaJosh

christopher-snapp:
@PrajaktaPurohit I've added a commit to this PR to reflect your request that the full test matrix be configured in buildkite.

PR #1876, opened 2020-01-15
Praj/solr to es

### LINKED ISSUE #1732: Chef Infra Server search(solr) uses supported/secure software(es), resolving open CVEs
Replaces solr with Elasticsearch to remove unsupported solr package that contains security vulnerabilities.

Operationally brings Chef Infra Server users toward migrating to the Automate Infra API (automate uses elasticsearch)

Aha! Link: https://chef.aha.io/epics/CHEF-E-2

PrajaktaPurohit:
### Description

This change adds capability to replace internal solr(for searching) with elasticsearch.
The solr, rabbitmq and opscode-expander services are shutdown.

By default the internal search will be with elasticsearch.

```
default['private_chef']['opscode-solr4']['external'] = true
default['private_chef']['opscode-solr4']['external_url'] = 'http://localhost:9200'

default['private_chef']['elasticsearch']['enable'] = true

default['private_chef']['opscode-erchef']['search_provider'] = 'elasticsearch' # solr, elasticsearch
default['private_chef']['opscode-erchef']['search_queue_mode'] = 'batch' # rabbitmq, batch, or inline
```

### Issues Resolved

#1732 

### Check List

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -130,7 +130,7 @@ default['private_chef']['haproxy']['etcd_port'] = 2379
 ####
 # RabbitMQ
 ####
-default['private_chef']['rabbitmq']['enable'] = true
+default['private_chef']['rabbitmq']['enable'] = false
 default['private_chef']['rabbitmq']['ha'] = false
 default['private_chef']['rabbitmq']['dir'] = '/var/opt/opscode/rabbitmq'
 default['private_chef']['rabbitmq']['data_dir'] = '/var/opt/opscode/rabbitmq/db'
@@ -226,13 +226,13 @@ default['private_chef']['jetty']['log_directory'] = '/var/opt/opscode/opscode-so
 ####
 # Chef Solr 4
 ####
-default['private_chef']['opscode-solr4']['enable'] = true
+default['private_chef']['opscode-solr4']['enable'] = false
 #
 # Set this to point at a solr/cloudsearch installation
 # not controlled by chef-server
 #
-default['private_chef']['opscode-solr4']['external'] = false
-default['private_chef']['opscode-solr4']['external_url'] = nil
+default['private_chef']['opscode-solr4']['external'] = true
+default['private_chef']['opscode-solr4']['external_url'] = "http://localhost:9200"
 default['private_chef']['opscode-solr4']['ha'] = false
 default['private_chef']['opscode-solr4']['dir'] = '/var/opt/opscode/opscode-solr4'
 default['private_chef']['opscode-solr4']['data_dir'] = '/var/opt/opscode/opscode-solr4/data'
@@ -268,7 +268,7 @@ default['private_chef']['opscode-solr4']['elasticsearch_replica_count'] = 1
 ####
 # Chef Expander
 ####
-default['private_chef']['opscode-expander']['enable'] = true
+default['private_chef']['opscode-expander']['enable'] = false
 default['private_chef']['opscode-expander']['ha'] = false
 default['private_chef']['opscode-expander']['dir'] = '/var/opt/opscode/opscode-expander'
 default['private_chef']['opscode-expander']['log_directory'] = '/var/log/opscode/opscode-expander'
@@ -285,8 +285,7 @@ default['private_chef']['opscode-expander']['retry_wait'] = 1
 var_base = '/var/opt/opscode'
 log_base = '/var/log/opscode'
 
-default['private_chef']['elasticsearch']['enable'] = false
-default['private_chef']['elasticsearch']['first_internal_install'] = false
+default['private_chef']['elasticsearch']['enable'] = true
 elasticsearch = default['private_chef']['elasticsearch']
 
 # These attributes cannot be overridden in chef-server.rb
@@ -444,8 +443,8 @@ default['private_chef']['opscode-erchef']['depsolver_timeout'] = 5000
 default['private_chef']['opscode-erchef']['ibrowse_max_sessions'] = 256
 default['private_chef']['opscode-erchef']['ibrowse_max_pipeline_size'] = 1
 # general search settings used to set up chef_index
-default['private_chef']['opscode-erchef']['search_provider'] = 'solr' # solr, elasticsearch
-default['private_chef']['opscode-erchef']['search_queue_mode'] = 'rabbitmq' # rabbitmq, batch, or inline
+default['private_chef']['opscode-erchef']['search_provider'] = 'elasticsearch' # solr, elasticsearch
+default['private_chef']['opscode-erchef']['search_queue_mode'] = 'batch' # rabbitmq, batch, or inline
 default['private_chef']['opscode-erchef']['search_batch_max_size'] = '5000000'
 default['private_chef']['opscode-erchef']['search_batch_max_wait'] = '10'
 # solr_service configuration for erchef. These are used to configure an opscoderl_httpc pool
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -299,6 +299,7 @@ elasticsearch['temp_directory'] = "#{var_base}/elasticsearch/tmp"
 elasticsearch['log_directory'] = "#{log_base}/elasticsearch"
 elasticsearch['log_rotation']['file_maxbytes'] = 104857600
 elasticsearch['log_rotation']['num_to_keep'] = 10
+elasticsearch['listen'] = '127.0.0.1'
 elasticsearch['port'] = 9200
 elasticsearch['enable_gc_log'] = false
 elasticsearch['initial_cluster_join_timeout'] = 90
@@ -858,7 +859,7 @@ default['private_chef']['dark_launch']['private-chef'] = true
 default['private_chef']['dark_launch']['sql_users'] = true
 default['private_chef']['dark_launch']['add_type_and_bag_to_items'] = true
 default['private_chef']['dark_launch']['reporting'] = true
-# It appears that actions rabbitmq was used for oc_actions and analytics. 
+# It appears that actions rabbitmq was used for oc_actions and analytics.
 default['private_chef']['dark_launch']['actions'] = false
 
 ###
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
@@ -112,17 +112,11 @@ class OmnibusHelper
   end
 
   def es_index_definition
-    if node['private_chef']['elasticsearch']['first_internal_install'] == true
+    es_version = elastic_search_major_version
+    if elastic_search_major_version == 6
       es_6_index
-    else
-      # For external elasticsearch or chef-backend elasticsearch will be running
-      # before we try to create the index
-      es_version = elastic_search_major_version
-      if es_version == 6
-        es_6_index
-      elsif [2, 5].include?(es_version)
-        es_5_or_2_index
-      end
+    elsif [2, 5].include?(es_version)
+      es_5_or_2_index
     end
   end
 
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
@@ -383,6 +383,8 @@ module PrivateChef
       PrivateChef['bookshelf']['listen'] ||= PrivateChef['default_listen_address']
       PrivateChef['rabbitmq']['node_ip_address'] ||= PrivateChef['default_listen_address']
       PrivateChef['redis_lb']['listen'] ||= PrivateChef['default_listen_address']
+      # Should there be value for es here?
+      #PrivateChef['elasticsearch']['ip_address'] ||= PrivateChef['default_listen_address']
       PrivateChef['opscode_solr4']['ip_address'] ||= PrivateChef['default_listen_address']
       PrivateChef['postgresql']['listen_address'] ||= '*' # PrivateChef["default_listen_address"]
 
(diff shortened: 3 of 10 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -103,15 +103,106 @@ def elastic_search_major_version
         # Decorate any JSON parsing exception or numeric exceptions when accessing and parsing the version field.
         raise "Unable to parse elasticsearch response #{e}"
       end
-      raise "Unsupported elasticsearch version of #{version}. There is current support for the major versions of 2 and 5." if version != 5 && version != 2
```
marcparadise: Informational: Anything older than 6.3 is EOL

omnibus/package-scripts/chef-server/postinst
```diff
@@ -11,6 +11,45 @@ function error_exit
...
+            ;;
+        *)
+            echo "Please reinstall the older version again and reconfigure."
+            exit
+            ;;
+  esac
+}
+
+first_elasticsearch_internal_install () {
+  IFS=. read -r major minor patch <<< "$version"
```
marcparadise: Can we capture this in preinst instead, so that we don't need to operator to do a rollback/reinstall of the older version? 
One concern is that they may no longer have the exact version number they previously had installed available.

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -49,6 +49,33 @@
...
+  only_if { File.exist?('/tmp/first_time_internal_elasticsearch_install') }
+end
+
+directory '/opt/opscode/sv/rabbitmq' do
+  recursive true
+  action :delete
+  only_if { File.exist?('/tmp/first_time_internal_elasticsearch_install') }
+end
+
+directory '/opt/opscode/sv/opscode-expander' do
```
marcparadise: We'll also want to clean up the rabbit/solr log services. 

Should we also be deleting the data directory for solr4/rabbitmq? 

Instead of doing this per directory, is it possible to use the runit service resource w/ action :delete and have it do handle all the directories?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -846,7 +847,7 @@
 default['private_chef']['dark_launch']['sql_users'] = true
 default['private_chef']['dark_launch']['add_type_and_bag_to_items'] = true
 default['private_chef']['dark_launch']['reporting'] = true
-default['private_chef']['dark_launch']['actions'] = true
+default['private_chef']['dark_launch']['actions'] = false
```
marcparadise: This is minor but potentially - customers will start getting 404 from that endpoint IIRC, is that going to be a surprise for anyone? Will it cause any problems with data feeds into automate? (I'm not sure if these come through `actions` or a different path.)

omnibus/package-scripts/chef-server/postinst
```diff
@@ -11,6 +11,45 @@ function error_exit
 	exit 1
 }
 
+version=$(awk '/chef-server/{ split($2,a,"+");print a[1];exit }' /opt/opscode/version-manifest.txt)
```
PrajaktaPurohit: Try to see if we can decide if a migration is needed based on info in var/opt/opscode/upgrades/migration-level from the api.

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -49,6 +49,33 @@
...
+  recursive true
+  action :delete
+  only_if { File.exist?('/tmp/first_time_internal_elasticsearch_install') }
+end
+
+file '/tmp/first_time_internal_elasticsearch_install' do
+  node.override['private_chef']['opscode-solr4']['enable'] = false
+  node.override['private_chef']['rabbitmq']['enable'] = false
+  node.override['private_chef']['elasticsearch']['first_internal_install'] = true
```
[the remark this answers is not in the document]
PrajaktaPurohit: That was mostly to be able to identify the version of elasticsearch installed. Updated the the PR to support older versions of es. That also uses this attribute.

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-solr4-external.rb
```diff
@@ -32,38 +33,35 @@
...
+                                                    "norms" => {
+                                                      "enabled" => false
+                                                    }
                                                   },
-                            'data_bag' => { 'type' => 'string',
-                                            'index' => 'not_analyzed',
-                                            'norms' => {
-                                              'enabled' => false,
-                                            },
+                            "data_bag" => { "type" => "keyword",
```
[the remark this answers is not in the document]
PrajaktaPurohit: you are absolutely right - added a test case to test that.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
PrajaktaPurohit: As far as I am aware, the /actions endpoint was not being used for anyting other than actions and analytics. But I will try to look for more evidence that, that is actually true.

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
PrajaktaPurohit: Looks like there is no action :delete for the runit cookbook, but the recipe now does delete the log dirs as well. Thank you!

omnibus/files/private-chef-upgrades/001/035_solr_to_elasticsearch_migration.rb
```diff
@@ -0,0 +1,35 @@
+define_upgrade do
+  if Partybus.config.bootstrap_server
+
+    must_be_data_master
+
+    # Make sure API is down
+    stop_services(["nginx", "opscode-erchef"])
+    # Try to call chef-server-ctl reindex
+    start_services(["opscode-chef-mover",
```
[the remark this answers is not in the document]
PrajaktaPurohit: yeah.

omnibus/package-scripts/chef-server/postinst
PrajaktaPurohit: Deciding to not use the logic in preinst and postinst script since, the migration api is not available to us before the install.

omnibus/package-scripts/chef-server/postinst
PrajaktaPurohit: yep! done!

omnibus/package-scripts/chef-server/postinst
```diff
@@ -11,6 +11,45 @@ function error_exit
...
+  # change minor to 1 when doing the actual promotion
+  if [ $major -ge 13 -a $minor -ge 0 ]; then
+    echo "Installing internal elasticsearch."
+    if [ -e /opt/opscode/sv/opscode-solr4 ]; then
+      echo "This install will replace solr with elasticsearch."
+      echo "The migration is a downtime operation."
+      echo "Please take a full backup before proceeding."
+      confirm
+    else
+      echo "sv solr4 does not exist you were already on elasticsearch for search. no reindex."
```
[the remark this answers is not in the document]
PrajaktaPurohit: yes - will get rid of this before merging, it is to mostly make sure the code works as expected.

src/oc_erchef/apps/chef_index/src/chef_elasticsearch.erl
```diff
@@ -87,8 +87,9 @@ query_body(#chef_solr_query{
 
 fields_tag() ->
     case envy:get(chef_index, solr_elasticsearch_major_version, 2, non_neg_integer) of
-        5 -> <<"stored_fields">>;
```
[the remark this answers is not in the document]
PrajaktaPurohit: yep - done!

omnibus/package-scripts/chef-server/postinst
```diff
@@ -11,6 +11,10 @@ function error_exit
 	exit 1
 }
 
+# If we have progressed from the preinst script
+# It implies that the user has agreed to proceed with the reindexing
+touch /tmp/first_time_internal_elasticsearch_install
```
stevendanna: I think we should use a path in a folder we otherwise control, such as `/var/opt/opscode`

omnibus/files/private-chef-upgrades/001/035_solr_to_elasticsearch_migration.rb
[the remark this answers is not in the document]
stevendanna: How is mover involved in 503 mode? I think the reindex tool can do that via redis redirectly?

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -150,6 +178,11 @@
     component_runit_service service do
       action :disable
     end
+  elsif node['private_chef']['opscode-solr4']['external'] &&
```
marcparadise, line 181: Should we also disable this in any case we have an indexing mode that doesn't use it?

PrajaktaPurohit: which would be elasticsearch only right?

marcparadise: I was looking at this attribute 
```
default['private_chef']['opscode-erchef']['search_queue_mode'] = "batch" # rabbitmq, batch, or inline
```
I think only `rabbitmq` mode requires opscode-expander.

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -49,6 +49,33 @@
...
+end
+
+directory '/opt/opscode/sv/opscode-expander' do
+  recursive true
+  action :delete
+  only_if { File.exist?('/tmp/first_time_internal_elasticsearch_install') }
+end
+
+file '/tmp/first_time_internal_elasticsearch_install' do
+  node.override['private_chef']['opscode-solr4']['enable'] = false
```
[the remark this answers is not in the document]
PrajaktaPurohit: done!

omnibus/files/private-chef-cookbooks/private-chef/recipes/opscode-chef-mover.rb
```diff
@@ -70,5 +70,5 @@
 
 # We want the service defined, but dead
 component_runit_service 'opscode-chef-mover' do
-  action :down
+  action :create
```
PrajaktaPurohit, line 73: Check if we can completely leave out the mover service.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -285,8 +285,8 @@
 var_base = '/var/opt/opscode'
 log_base = '/var/log/opscode'
 
-default['private_chef']['elasticsearch']['enable'] = false
-default['private_chef']['elasticsearch']['first_internal_install'] = false
+default['private_chef']['elasticsearch']['enable'] = true
```
stevendanna, line 288: We need to make a change to libraries/private_chef.rb to disable elasticsearch on the frontend.

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -112,18 +112,11 @@ def elastic_search_major_version
...
-    if node['private_chef']['elasticsearch']['first_internal_install'] == true
-      es_6_index
-    else
-      # For external elasticsearch or chef-backend elasticsearch will be running
-      # before we try to create the index
-      es_version = elastic_search_major_version
-      if es_version == 6
+      if elastic_search_major_version == 6
         es_6_index
       elsif [2, 5].include?(es_version)
```
stevendanna, line 117: `es_version` will be undefined here. I think this should be `elatic_search_major_version`

omnibus/files/private-chef-cookbooks/private-chef/libraries/private_chef.rb
```diff
@@ -383,6 +383,8 @@ def gen_backend_default(bootstrap)
       PrivateChef['bookshelf']['listen'] ||= PrivateChef['default_listen_address']
       PrivateChef['rabbitmq']['node_ip_address'] ||= PrivateChef['default_listen_address']
       PrivateChef['redis_lb']['listen'] ||= PrivateChef['default_listen_address']
+      # Should there be value for es here?
+      #PrivateChef['elasticsearch']['ip_address'] ||= PrivateChef['default_listen_address']
```
stevendanna, line 387: Right now our we have the following in the elasticsearch config template:

```
network.host: <%= node['private_chef']['opscode-erchef']['vip']  %>
```

This likely needs to be changed to something ES specific.

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
PrajaktaPurohit: I am sure I have pushed this change twice - not sure whay it did not reflect. Will check.

omnibus/files/private-chef-cookbooks/private-chef/recipes/elasticsearch.rb
```diff
@@ -137,5 +137,10 @@
   to elasticsearch_conf_dir
 end
 
+directory '/opt/opscode/sv/opscode-solr4' do
```
stevendanna, line 140: 💭 We think we might not need this now that we have the prefligth check that makes sure only 1 is enabled.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -226,13 +226,13 @@
...
-default['private_chef']['opscode-solr4']['enable'] = true
+default['private_chef']['opscode-solr4']['enable'] = false
 #
 # Set this to point at a solr/cloudsearch installation
 # not controlled by chef-server
 #
-default['private_chef']['opscode-solr4']['external'] = false
-default['private_chef']['opscode-solr4']['external_url'] = nil
+default['private_chef']['opscode-solr4']['external'] = true
+default['private_chef']['opscode-solr4']['external_url'] = "http://localhost:9200"
```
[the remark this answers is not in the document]
markan: After talking to @PrajaktaPurohit yes we are, and this makes sense to me now.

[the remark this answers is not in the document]
stevendanna: Yeah, we are gonna look into _not_ overloading it as a follow up, time permitting

PR #1898, opened 2020-01-29
Fix erlang version and db config for habitat build

uwej711:
### Description

- Update the habitat build to use erlang 20 as building with erlang 18 no longer works
- Fix the habitat database configuration for external database

### Check List

### DIFF
--- a/src/bookshelf/habitat/plan.sh
+++ b/src/bookshelf/habitat/plan.sh
@@ -3,7 +3,7 @@ pkg_origin=chef
 pkg_license=('Apache-2.0')
 pkg_maintainer="The Chef Server Maintainers <support@chef.io>"
 pkg_deps=(
-  core/erlang18
+  core/erlang20
   core/cacerts
   core/coreutils
   core/gcc-libs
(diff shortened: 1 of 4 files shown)


PrajaktaPurohit:
Hey @uwej711. Thank you for looking at the habitat packaging of chef-server. Habitat packaging is definitely our future direction, but we’re not there yet, and it’s not supported at this time. In particular we don’t (yet) have automated test pipelines around the habitat packages; our current usage is limited to periodically taking releases and integrating them into Chef-Automate. So any new changes (including yours) won’t be tested and integrated immediately, and we won’t have great confidence about their functionality in the habitat environment. Improving that is a near term goal, but probably won’t surface for a few months.
That being said, your changes are certainly more correct than what we have, and on that basis we want to get them merged as soon as we can. Just didn’t want to mislead you as to the current state of affairs. 
That being said, if possible it would be nice to split the PR into 2.

uwej711:
OK, see 
- https://github.com/chef/chef-server/pull/1902
- https://github.com/chef/chef-server/pull/1903

PR #1919, opened 2020-02-13
Update pointers to erlware_commons

lbakerchef:
The erlware_commons branch we are pointing to was rebased and a commit
was squashed.  This change points us to the updated erlware_commons commit.


PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/927

PR #1946, opened 2020-04-04
Praj/enable internal es

PrajaktaPurohit:
### Description
This is not ready for deploy for production.

This change will let the user manually enable elastic search.
Make sure to take ample backups before making the following changes.

Add the following to /etc/opscode/chef-server.rb

```
opscode_solr4['external'] = true
opscode_solr4['external_url'] = 'http://localhost:9200'
elasticsearch['enable'] = true
elasticsearch['first_internal_install'] = true
opscode_erchef['search_provider'] = 'elasticsearch'
opscode_erchef['search_queue_mode'] = 'batch'
```

Manually run reindex command `chef-server-ctl reindex` after upgrade.

https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1075#fb50d184-6c1a-40b2-b389-47afdd637333

### Check List

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
@@ -50,6 +50,12 @@ include_recipe 'private-chef::plugin_discovery'
 include_recipe 'private-chef::plugin_config_extensions'
 include_recipe 'private-chef::config'
 
+if node['private_chef']['elasticsearch']['first_internal_install']
+  node.override['private_chef']['opscode-solr4']['enable'] = false
+  node.override['private_chef']['rabbitmq']['enable'] = false
+  node.override['private_chef']['opscode-expander']['enable'] = false
+end
+
 if node['private_chef']['fips_enabled']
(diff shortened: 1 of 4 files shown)


omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -50,6 +50,12 @@
 include_recipe 'private-chef::plugin_config_extensions'
 include_recipe 'private-chef::config'
 
+if node['private_chef']['elasticsearch']['first_internal_install']
```
markan, line 53: So there's no pre-defined default for this, we're relying on nil being 'falsey', right?

PrajaktaPurohit:
Create this as a TODO to not miss out on adding the check: https://github.com/chef/chef-server/issues/1950

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
PrajaktaPurohit: added to the default attributes, defaults to false.

PR #1948, opened 2020-04-06
Fix pedant to not be strict about cookbook artifact metadata

lamont-granquist:
This broke when I added the eager_load_libraries metadata.  The server
doesn't actually care about additional metadata (which we've done before
for years, including ohai_version, chef_version, privacy and the gem
metadata).


PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1073

PR #1984, opened 2020-05-13
Add preflight checks

PrajaktaPurohit:
Check for some of the conditions that might disrupt
 expected behaviour of internal elasticsearch install

### Description

preflight check verify_es_disabled_if_user_set_external_solr
checks that if user has selected external search; 
the internal es service is turned off

preflight check verify_unused_services_are_disabled_if_using_internal_es
opscode_solr4, rabbitmq and opscode-expander services
should be turned off if using internal elasticsearch.

Question:
```
opscode_solr4['external'] = false
elasticsearch['enable'] = false
no error - is that ok?
```

```
Config:
opscode_solr4['enable'] = true
opscode_solr4['external'] = true
rabbitmq['enable'] = true
opscode_expander['enable'] = true

Reconfigure:
-----------------------------------------------------------------------
SOLR002: The Chef Infra Server is configured to enable both
         Elasticsearch and Solr. This is an unsupported configuration.
         Please update /etc/opscode/chef-server.rb such that only one
         search provider is enabled.
-----------------------------------------------------------------------

Config:
#opscode_solr4['enable'] = true
opscode_solr4['external'] = true
rabbitmq['enable'] = true
opscode_expander['enable'] = true

Reconfigure:
-----------------------------------------------------------------------
SOLR003: The Chef Infra Server is configured to use an external search
         index but the internal Elasticsearch is still enabled. This is
         an unsupported configuration.
         To disable the internal Elasticsearch, add the following to /etc/opscode/chef-server.rb:
             elasticsearch['enable'] = false
         To use the internal Elasticsearch, consider removing the configuration
         entry for opscode_solr4['external'].
-----------------------------------------------------------------------

Config:
opscode_solr4['enable'] = true
opscode_solr4['external'] = true
#rabbitmq['enable'] = true
#opscode_expander['enable'] = true
elasticsearch['enable'] = false

Reconfigure:
-----------------------------------------------------------------------
SOLR004: The Chef Infra Server is configured to use an external search
         provider but the internal Solr is still enabled. This is an
         unsupported configuration.
         To disable the internal Solr, add the following to /etc/opscode/chef-server.rb:
             opscode_solr4['enable'] = false
         Alternatively, if you are attempting to use the internal Solr,
         add the following to /etc/opscode/chef-server.rb:
             opscode_solr4['enable'] = true
             opscode_solr4['external'] = false
-----------------------------------------------------------------------

Config:
#opscode_solr4['enable'] = true
opscode_solr4['external'] = false
rabbitmq['enable'] = true
opscode_expander['enable'] = true
elasticsearch['enable'] = true

Reconfigure:
-----------------------------------------------------------------------
SOLR006: The Chef Infra Server is configured to use its internal
         Elasticsearch installation, but RabbitMQ has been manually
         enabled. This is an unsupported configuration.
         Please update /etc/opscode/chef-server.rb to ensure that
         rabbitmq is disabled:
             rabbitmq['enable'] = false

SOLR007: The Chef Infra Server is configured to use its internal
         Elasticsearch installation, but opscode-expander has been
         manually enabled. This is an unsupported configuration.
         Please update /etc/opscode/chef-server.rb to ensure that
         opscode-expander is disabled:
             opscode_expander['enable'] = false
-----------------------------------------------------------------------

Config:
#opscode_solr4['enable'] = true
opscode_solr4['external'] = false
#rabbitmq['enable'] = true
#opscode_expander['enable'] = true
elasticsearch['enable'] = true

Reconfigure:
-----------------------------------------------------------------------
SOLR008: The Chef Infra Server is configured with
             opscode_solr4['external'] = false
         but this is incompatible with elasticsearch being enabled.
         Please remove this line from /etc/opscode/chef-server.rb.
-----------------------------------------------------------------------

Config:
#opscode_solr4['enable'] = true
opscode_solr4['external'] = true
rabbitmq['enable'] = true
opscode_expander['enable'] = true
elasticsearch['enable'] = false

Reconfigure:
-----------------------------------------------------------------------
SOLR010: No external url specified for Solr depsite opscode_solr4['external']
         being set to true. To use an external solr instance, please set
         opscode_solr4['external_url'] to the external solr endpoint.
-----------------------------------------------------------------------
```

Todo:
- Do we want to add any chef-backend preflight scenarios?

Notes:
Created https://github.com/chef/chef-server/issues/1986 with a gist from all the comments in the code.

### Check List

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
@@ -16,48 +16,140 @@
 require_relative './preflight_checks.rb'
 
 class SolrPreflightValidator < PreflightValidator
-  attr_reader :cs_solr_attr, :node_solr_attr, :cs_erchef_attr, :node_erchef_attr
+  # The cs_*attr variables hold the user-defined configuration
+  attr_reader :cs_solr_attr, :cs_erchef_attr,
+              :cs_elasticsearch_attr, :cs_rabbitmq_attr,
+              :cs_opscode_expander_attr
+
+  # The node_*attr variables hold the default configuration
+  attr_reader :node_solr_attr, :node_elasticsearch_attr, :node_erchef_attr
 
   def initialize(node)
     super
+
     @cs_solr_attr = PrivateChef['opscode_solr4']
     @cs_erchef_attr = PrivateChef['opscode_erchef']
+    @cs_elasticsearch_attr = PrivateChef['elasticsearch']
+    @cs_rabbitmq_attr = PrivateChef['rabbitmq']
+    @cs_opscode_expander_attr = PrivateChef['opscode_expander']
+
     @node_erchef_attr = node['private_chef']['opscode-erchef']
     @node_solr_attr = node['private_chef']['opscode-solr4']
+    @node_elasticsearch_attr = node['private_chef']['elasticsearch']
   end
 
   def run!
-    verify_sane_reindex_sleep_times
+    verify_consistent_reindex_sleep_times
+
+    verify_one_search_provider
+    verify_es_disabled_if_user_set_external_solr
+    verify_unused_services_are_disabled_if_using_internal_es
+    verify_no_explicit_external_disable_when_es_enabled
+
     warn_unchanged_external_flag
     verify_external_url
     verify_erchef_config
   end
 
-  def verify_sane_reindex_sleep_times
+  def verify_consistent_reindex_sleep_times
     final_min = cs_erchef_attr['reindex_sleep_min_ms'] || node_erchef_attr['reindex_sleep_min_ms']
     final_max = cs_erchef_attr['reindex_sleep_max_ms'] || node_erchef_attr['reindex_sleep_max_ms']
     if final_min > final_max
-      fail_with <<~EOM
+      fail_with err_SOLR001_failed_validation(final_min, final_max)
+    end
+  end
 
-        opscode_erchef['reindex_sleep_min_ms'] (#{final_min}) is greater than
-        opscode_erchef['reindex_sleep_max_ms'] (#{final_max})
+  def verify_one_search_provider
+    if elasticsearch_enabled? && solr_enabled?
+      fail_with err_SOLR002_failed_validation
+    end
+  end
 
-        The maximum sleep time should be greater or equal to the minimum sleep
-        time.
-      EOM
+  # If an external search provider (ES or Solr) was explicitly enabled
+  # by the user, then we expect that both internal search services
+  # should be disabled.
+  def verify_es_disabled_if_user_set_external_solr
+    # NOTE(ssd) 2020-05-13: This might impact the people who we gave
+    # pre-release access to.
+    if cs_solr_attr['external'] && elasticsearch_enabled?
+      fail_with err_SOLR003_failed_validation
+    end
+
+    if solr_enabled? && external?
+      fail_with err_SOLR004_failed_validation
+    end
+  end
+
+  # The user might have a combination of solr, rabbitmq or
+  # opscode-expander explicitly enabled in their config.
+  #
+  # If we are using the new internal elasticsearch, this is an
+  # unsupported configuration.
+  #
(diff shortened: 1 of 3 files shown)


stevendanna:
@PrajaktaPurohit I like the idea of using the preflight checks for this 👍 

I've added a few more checks and updated the logic just a bit.  I left some inline todo comments as well.

I also changed all the warnings to failures. I think we are catching cases that we are 90% confident are going to be errors.  I would rather fail early before making any system changes in those cases.  Even if it generates a support ticket, it won't be one in which we've introduced unexpected downtime.

We might want to get an opinion from @sean-horn or @moutons here as well.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
```diff
@@ -48,6 +66,172 @@ def verify_sane_reindex_sleep_times
...
+    end
+  end
+
+  # If an external search provider (ES or Solr) was explicitly enabled
+  # by the user, then we expect that both internal search services
+  # should be disabled.
+  def verify_es_disabled_if_user_set_external_solr
+    # NOTE(ssd) 2020-05-13: This might impact the people who we gave
+    # pre-release access to.
+    if cs_solr_attr['external'] && elasticsearch_enabled?
```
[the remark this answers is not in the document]
stevendanna: @OBrienCommaJosh What do you think from the product side here? is it reasonable for us to ask customers using external search indexing to add new configuration in response to this upgrade?

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
```diff
@@ -48,6 +66,172 @@ def verify_sane_reindex_sleep_times
...
+  #
+  # To test: setup standalone cs with 13.2.0 add
+  # opscode_solr4['enable'] = true
+  # rabbitmq['enable'] = true
+  # opscode_expander['enable'] = true to chef-server.rb
+  # then upgrade to 13.3
+  def verify_unused_services_are_disabled_if_using_internal_es
+    if elasticsearch_enabled?
+      if cs_solr_attr['enable']
+        fail_with <<~EOM
```
stevendanna, line 140: Instead of immediately calling fail_with, we could track the failures in some array or string and then fail once at the bottom to avoid too much iteration.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
```diff
@@ -48,6 +66,172 @@ def verify_sane_reindex_sleep_times
...
+      EOM
+      end
+    end
+  end
+
+  def verify_no_explicit_external_disable_when_es_enabled
+    return if cs_solr_attr['external'].nil?
+
+    # TODO(ssd) 2020-05-13: Rather than this check, we could just
+    # ignore that configuration when parsing the configuration.
```
stevendanna, line 190: Another alternative is to not overload the `external` config item like we are currently. But I'm not sure on the level of effort there.

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
PrajaktaPurohit: Done

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_solr_validator.rb
```diff
@@ -16,48 +16,140 @@
...
+  # opscode-expander explicitly enabled in their config.
+  #
+  # If we are using the new internal elasticsearch, this is an
+  # unsupported configuration.
+  #
+  def verify_unused_services_are_disabled_if_using_internal_es
+    if elasticsearch_enabled?
+      msg = []
+      if cs_solr_attr['enable']
+        msg << err_SOLR005_failed_validation
```
PrajaktaPurohit, line 93: err_SOLR002_failed_validation will always catch this case; But would like to leave this here till the code stabilizes. (the migration PR is merged)

PR #1985, opened 2020-05-13
[chef-server-ctl] Add new check-config command

stevendanna:
This new command runs just the configuration portion of the
reconfigure cookbooks, including the preflight checks.

### DIFF
--- a/omnibus/config/software/private-chef-cookbooks.rb
+++ b/omnibus/config/software/private-chef-cookbooks.rb
@@ -48,6 +48,15 @@ build do
         ]
       )
     end
+    File.open("#{install_dir}/embedded/cookbooks/check-config.json", "w") do |f|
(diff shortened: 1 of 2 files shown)


omnibus/config/software/private-chef-cookbooks.rb
```diff
@@ -48,6 +48,15 @@
         ]
       )
     end
+    File.open("#{install_dir}/embedded/cookbooks/check-config.json", "w") do |f|
+      f.write FFI_Yajl::Encoder.encode(
+        run_list: [
+          'recipe[private-chef::plugin_discovery]'
+          'recipe[private-chef::plugin_config_extensions]'
+          'recipe[private-chef::config_config]'
```
stevendanna, line 56:
```suggestion
          'recipe[private-chef::config]'
```

PR #2005, opened 2020-05-18
Automatically announce release on promotion to stable

schisamo:
This commit fully brings the `chef/chef-server` repo onto our streamlined release notes process:
- Pending release notes are added to https://github.com/chef/chef-server/wiki/Pending-Release-Notes
- On the promotion to stable:
  - Release notes are published to packages.chef.io
  - A release announcement is published to https://discourse.chef.io/c/chef-release/9

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -90,15 +90,13 @@ subscriptions:
     actions:
       - bash:.expeditor/promote_harts_and_containers.sh:
           post_commit: true
-      - bash:.expeditor/purge_cdn.sh:
-          post_commit: true
+      - purge_packages_chef_io_fastly:{{target_channel}}/chef-server/latest
   - workload: artifact_published:stable:chef-server:*
     actions:
       - built_in:rollover_changelog
       - bash:.expeditor/promote_harts_and_containers.sh:
           post_commit: true
-      - bash:.expeditor/purge_cdn.sh:
-          post_commit: true
+      - purge_packages_chef_io_fastly:{{target_channel}}/chef-server/latest
       - built_in:notify_chefio_slack_channels
   - workload: schedule_triggered:chef/chef-server:master:nightly_build:*
     actions:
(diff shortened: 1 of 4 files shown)


.expeditor/config.yml
```diff
@@ -96,7 +96,11 @@ subscriptions:
       - built_in:rollover_changelog
       - bash:.expeditor/promote_harts_and_containers.sh:
           post_commit: true
+      - bash:.expeditor/publish-release-notes.sh:
+          post_commit: true
```
PrajaktaPurohit, line 100: Is every step in the expeditor config executed?
Or would this fail if the previous step failed?

.expeditor/publish-release-notes.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -eou pipefail
+
+git clone https://x-access-token:${GITHUB_TOKEN}@github.com/chef/chef-server.wiki.git
+
+pushd ./chef-workstation.wiki
```
PrajaktaPurohit, line 7: should this be chef-server.wiki?

schisamo: Good catch!

.expeditor/config.yml
schisamo: I believe when the first step fails the whole actionset fails. Is this correct @tduffield ?

PrajaktaPurohit: Ideally we would want to release a hab build every time we release an omnibus builds, but if that cannot happen due to some reasonand we absolutely need to make an omnibus release, we would have to comment out the section to promote_harts_and_containers in the config.yml and then the release would go through as expected. Is that correct?

tduffield: @schisamo is correct. If an action fails, the action set stops. There is a way to get around this using the `always_run` parameter. 

From a Release Engineering perspective, if we cannot release the Habitat Builds we should not be releasing any builds. They are all a single Release Artifact. However, you are correct @PrajaktaPurohit that in this scenario you would need to comment out the `promote_harts_and_containers` script, effectively removing Habitat from the release pipeline. 

I would strongly caution against treating the Chef Habitat packages as second class artifact. If a fix is absolutely needed for an omnibus release, chances are that it is absolutely needed for the Chef Habitat release as well.

PR #2011, opened 2020-05-20
Praj/limit integration test cases

PrajaktaPurohit:
### Description

Limit the scenarios that are run per commit.
Add a default config to run all scenarios before release.

https://buildkite.com/chef/chef-chef-server-master-integration-test/builds/110#0a0bc308-0d9e-4cae-8864-d817c61c6d32

TODO: update the release process docs to run the full pipeline of the integration tests.


stevendanna:
This looks good to me. @christopher-snapp wondering what you think about this?  The idea here is that it would be useful for us to have a smaller subset of tests to iterate on.  We can then run the _full suite a bit less often.

PR #2014, opened 2020-05-20
[terraform] Don't use quotes in the makefile

stevendanna:
With the quotes in place, the underlying terraform vars would end up
with quotes in them, breaking a lot of the logic.  After we fix this,
we should probably undo

https://github.com/chef/chef-server/commit/e59d9cfc1141a42fc250359d1de92b195fa66812

which was me not realizing what was actually happening.


PrajaktaPurohit:
good catch!

Issue #2019, opened 2020-05-21
Clarify Push Jobs IPv6 support


stevendanna:
Currently, the push-jobs add-on does not appear to work out of the box with IPv6 because we generate nginx upstream configuration that doesn't correctly put the IP address in `[]`s, resulting in the following error:

```
2020-05-21_08:17:23.97520 nginx: [emerg] invalid port in upstream "2a05:d01c:13:cb01:2e94:2b02:9f4f:8732:10003" in /var/opt/opscode/nginx/etc/addon.d/10-push_jobs_upstreams.conf:5
```

While we could likely work-around this in our tests via more explicitly setting some configuration value, I think we should probably clarify whether we expect pushy to work in IPv6 mode and if so make sure that whatever workaround we put in our tests is also put into the documentation.

Note, that I may have missed something here as we have a lot of issues we are working through at the moment.

@markan Do you have any further context here?

markan:
At this point I would say we shouldn't expect it to work with IPv6 out of the box. Probably wouldn't be too hard to fix; but would require code changes. We would have to check that the client manages ports correctly in ruby as well. When we wrote the client Ruby had some similar bugs with port specifications in URLs, and I don't know if we worked around them, or if the workarounds stayed correct after multiple Ruby upgrades.

stevendanna:
@markan Awesome, thanks for the confirmation.  Our IPv6 tests will likely require some follow up work for a number of reasons, so we can look at this again in the future, but I think given our current set of priorities, deferring that work for a bit makes sense.

tas50:
Closing this out since we're no longer supporting push jobs

PR #2020, opened 2020-05-21
Disable push jobs tests in tiered ipv6 pipeline

PrajaktaPurohit:

REF: https://github.com/chef/chef-server/issues/2019

### DIFF
--- a/.expeditor/integration_test.pipeline.yml
+++ b/.expeditor/integration_test.pipeline.yml
@@ -195,6 +195,7 @@ steps:
       SCENARIO: omnibus-tiered-fresh-install
       PLATFORM: ubuntu-18.04
       ENABLE_IPV6: true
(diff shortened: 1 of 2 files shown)


.expeditor/integration_test.pipeline.yml
```diff
@@ -212,6 +213,7 @@ steps:
       SCENARIO: omnibus-tiered-upgrade
       PLATFORM: rhel-7
       ENABLE_IPV6: true
+      ENABLE_ADDON_PUSH_JOBS: false
```
stevendanna, line 216: Thanks! We probably want to do the same thing in integration_test_full I think

PrajaktaPurohit: Yeah I was contemplating that, but I see the value in disabling it, since we have the issue to track that created. 👍

Issue #1713, opened 2019-08-02
Purge or use some of the older unused directories.


PrajaktaPurohit:
### Describe the Enhancement:
 I would like to keep only working usable code in the chef-server repo. Since the delivery code is cleaned up, there might be some surrounding support cleanup that needs to be done.

### Describe the Need:
Tech Debt.

### Can We Help You Implement This?:
Creating an issue from : https://github.com/chef/chef-server/pull/1694

It would be useful to delete/reuse the following directories:
 terraform/, cookbooks/chef-server-deploy/, and inspec/.

PrajaktaPurohit:
https://github.com/chef/chef-server/pull/2026

PR #2026, opened 2020-05-23
Remove legacy ACC configuration

stevendanna:
I don't think the ACC pipeline is used anymore. If that is the case,
let's remove these old cookbooks from the repository.


PrajaktaPurohit:
https://github.com/chef/chef-server/issues/1713

PR #2029, opened 2020-05-28
[habitat] Explicitly install ruby 1.17

stevendanna:
A couple of changes converged to break the current build

- Certain gems in our ecosystem now require Ruby 2.6+
- The core-plans refresh upgraded bundler to 2.1

I originally attempted to upgrade us to bundler 2.1, but ran into
multiple problems. They are likely all fixable, but the 45 minute
cycle time per omnibus build means I'll likely need more time to
iterate on the solution there.

This change:

1) Installs bundler 1.17 during the Habitat build via the `gem`
   commands. It installs it to a GEM_HOME in the package's pkg_prefix
   so it should be available at runtime.

2) Updates various hooks and binaries to set GEM_HOME

3) Upgrades us from core/ruby to core/ruby26.

4) Removes the dependency on core/bundler and instead depend on the
   bundler binary shipped in core/ruby26. When calling `bundle` from a
   directory with a Gemfile, this should detect the required version
   of bundler and use it.

5) Removes some runtime uses of `hab pkg path` in hooks and bin
   stubs. Calling `hab pkg path` in a hook or binstub is almost always
   a bug because it is not locked to the version that the package
   installed.

6) Fixes a few missing dependency problems I found along the way.


src/oc_erchef/habitat/hooks/run
```diff
@@ -6,10 +6,10 @@ export RELX_CONFIG_PATH="{{pkg.svc_config_path}}/sys.config"
 export VMARGS_PATH="{{pkg.svc_config_path}}/vm.args"
 export CHEF_SECRETS_DATA=$(cat {{pkg.svc_config_path}}/veil-secrets.json)
 # these are needed in order to make depselector start up
-export PATH={{pkgPathFor "core/ruby"}}/bin:{{pkgPathFor "core/bundler"}}/bin:${PATH}
+export PATH={{pkgPathFor "core/ruby26"}}/bin:${PATH}
 export LD_LIBRARY_PATH={{pkgPathFor "core/libffi"}}/lib:${LD_LIBRARY_PATH}
-export GEM_HOME={{pkgPathFor "core/bundler"}}
-export GEM_PATH={{pkgPathFor "core/bundler"}}
+export GEM_HOME={{pkgPathFor "core/ruby26"}}
```
stevendanna, line 11: This really should be pointing at the bundle/vendor dir inside the erchef package?  As in, I think this line was wrong before and continues to be wrong now, but I'm trying to limit how much I change at once.

stevendanna:
I'm having trouble actually getting this whole system of packages up to test, but this should get the build passing again.

Issue #2033, opened 2020-06-06
Failing FIPS tests on upgrade to chef 16.1.16 via chef-cli


PrajaktaPurohit:
### Chef Server Version
13.2.44 onwards

### Platform Details
Tried on rhel-8

### Configuration
FIPS mode

### Scenario:
https://buildkite.com/chef/chef-chef-server-master-integration-test/builds/139 fails in FIPS tests.

### Steps to Reproduce:
Create an instance with omnibus-fips-ipv4-rhel-8. That will cause cookbook upload tests to fail.

Tracing this further down:
Chef Infra server uses different version fo ruby:
omnibus_overrides: chef-15.8.23
oc-id: chef-12.22.5
chefspec -> chef-cli : chef-16.1.16

Chefspec pulls chef-cli: https://github.com/chefspec/chefspec/blob/master/chefspec.gemspec#L27
chef-cli pulls the latest chef: https://github.com/chef/chef-cli/blob/master/chef-cli.gemspec#L46
(16.1.16 in this case) 

It fails with:
```
[root@ip-10-0-19-207 ec2-user]# /opt/opscode/embedded/service/gem/ruby/2.6.0/bin/knife cookbook upload packagecloud -VV
INFO: Using configuration from /home/ec2-user/.chef/knife.rb
DEBUG: FIPS mode is enabled.
DEBUG: No chefignore file found. No files will be ignored!
DEBUG: No chefignore file found. No files will be ignored!
DEBUG: No chefignore file found. No files will be ignored!
Traceback (most recent call last):
	31: from /opt/opscode/embedded/service/gem/ruby/2.6.0/bin/knife:23:in `<main>'
	30: from /opt/opscode/embedded/service/gem/ruby/2.6.0/bin/knife:23:in `load'
	29: from /opt/opscode/embedded/lib/ruby/gems/2.6.0/gems/chef-16.1.16/bin/knife:24:in `<top (required)>'
	28: from /opt/opscode/embedded/lib/ruby/gems/2.6.0/gems/chef-16.1.16/lib/chef/application/knife.rb:163:in `run'
	27: from /opt/opscode/embedded/lib/ruby/gems/2.6.0/gems/chef-16.1.16/lib/chef/knife.rb:229:in `run'
	26: from /opt/opscode/embedded/lib/ruby/gems/2.6.0/gems/chef-16.1.16/lib/chef/knife.rb:489:in `run_with_pretty_exceptions'
[...]

PR #2035, opened 2020-06-06
Pin to a single version of chef client consistently in omnibus

PrajaktaPurohit:
A lot of software dependencies pull a floating latest version of the chef-client as a result of which we have multiple versions of clients installed on the chef-server.

This will make it so that a single version of the chef-client is installed on the chef-server.

A better way to do this would be to remove the omnibus-software dependencies for gems and install them from the omnibus/Gemfile (or the respective Gemfiles) #2060 

#2033 

https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1609
https://buildkite.com/chef/chef-chef-server-master-habitat-build/builds/268
https://buildkite.com/chef/chef-chef-server-master-integration-test/builds/322


omnibus/Gemfile
```diff
@@ -1,11 +1,12 @@
 source 'https://rubygems.org'
 
 gem 'rake'
-
+# chef-cli pulls chef >= 15.0 which is 16.1.16 at the time of failure.
+# https://github.com/chef/chef/commit/003fbc132935961e93667f11c3f45ce4914b83ac
+gem 'chef', '=15.8.23'
```
tas50, line 6: You can probably get away with 15.11.whatever here. 16.something is where the change was made

PrajaktaPurohit: Right - I wanted to pin to the same version pinned in omnibus_overrides. We can upgrade the version overall once we know this change works.

omnibus/Gemfile
```diff
@@ -1,13 +1,14 @@
...
+# chef-cli pulls chef >= 15.0 which is 16.1.16 at the time of failure.
+# https://github.com/chef/chef/commit/003fbc132935961e93667f11c3f45ce4914b83ac
+gem 'chef', '=15.12.22'
 gem 'chefspec'
 # chef-cli >=3.0.4 requires Ruby version >= 2.7 which chef-server isn't ready for
 gem 'chef-cli', '=3.0.1'
-
 gem 'berkshelf'
-gem 'bundler', '>1.10'
+gem 'bundler', '=1.17.2'
```
stevendanna, line 11: Note for other reviewers, our current plan is to try so sort out the bundler version shenanigans once oc-id is updated, as that has been the main pain-point in attempting to upgrade to bundler 2 in the past.

omnibus/config/software/fixie-gem.rb
```diff
@@ -15,7 +15,7 @@
 #
 
 name "chef_fixie"
-default_version "0.4.0"
+default_version "master"
```
markan, line 18: Yeah, this is probably for the best. It's not like the update rate of fixie is high.

stevendanna (approved):

PR #2038, opened 2020-06-17
Implement reaper for orphaned integration-test instances

### LINKED ISSUE #1958: Implement reaper for orphaned integration-test instances
There are currently failure states that cause the chef-server integration tests to leave orphaned instances running in AWS.  With the number of instances the integration-test pipeline launches this can result in significant costs being incurred with no added value to the company.

The pipeline already does its best to cleanup after itself, but since it doesn't always get every instance we need to implement a fail safe reaper mechanism that would terminate any instances in the chef-server VPC that have persisted longer that 12 hours.

christopher-snapp:

### Issues Resolved

#1958 

### Check List

- [X] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

### DIFF
--- a/.expeditor/cleanup_orphaned_test_resources.sh
+++ b/.expeditor/cleanup_orphaned_test_resources.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+terminate_instance() {
+  local instance_id
+  instance_id="$1"
+
+  if aws --profile chef-cd ec2 terminate-instances --instance-ids "$instance_id"; then
+    echo "Terminated instance $instance_id"
+  else
+    echo "Failed to terminate instance $instance_id"
+  fi
+}
+
+terminate_instances() {
+  while read -r instance_id; do
+    terminate_instance "$instance_id"
+  done
+}
+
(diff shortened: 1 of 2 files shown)


.expeditor/cleanup_orphaned_test_resources.sh
```diff
@@ -0,0 +1,27 @@
...
+
+terminate_instances() {
+  while read -r instance_id; do
+    terminate_instance "$instance_id"
+  done
+}
+
+echo "--- Setting up AWS credentials"
+export VAULT_UTIL_ACCOUNTS='{"aws":["chef-cd"]}'
+vault-util configure-accounts
```
jeremiahsnapp, line 24: Can you remove the chef-cd AWS creds setup? Running this will interfere with chef-cd creds that expeditor already has setup. In the near future expeditor will run these jobs in a more optimal environment which will require this code again in order to request temporary chef-cd creds.

christopher-snapp: should be taken care of
I left the `--profile chef-cd` to explicitly target the desired profile and simply removed the `vault-util` setup.

PR #2049, opened 2020-07-07
Cookstyle fixes for the private-chef cookbook

tas50:
Some automated fixes from Cookstyle. I haven't tested this at all so be sure to do that before shipping this.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -66,12 +66,12 @@ default['private_chef']['addons']['path'] = nil
 default['private_chef']['addons']['packages'] =
   %w(opscode-reporting opscode-manage opscode-analytics opscode-push-jobs-server chef-ha)
 default['private_chef']['addons']['ubuntu_supported_codenames'] =
-  %w(lucid precise trusty)
+  %w(xenial bionic focal)
 default['private_chef']['addons']['ubuntu_distribution'] =
   if node['private_chef']['addons']['ubuntu_supported_codenames'].include?(node['lsb']['codename'])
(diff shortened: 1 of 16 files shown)


omnibus/files/private-chef-cookbooks/private-chef/resources/pg_sqitch.rb
```diff
@@ -15,7 +15,6 @@
 #
 default_action :deploy
 
-attribute :name,           kind_of: String, required: true, name_attribute: true
```
tas50, line 18: it's possible we were abusing the name property here, but if that's the case we need to stop calling this resource with a name property specified.

PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1351

PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-integration-test/builds/194

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -68,8 +68,11 @@
 default['private_chef']['addons']['ubuntu_supported_codenames'] =
```
tas50, line 68: @PrajaktaPurohit I'm real confused by this part of the cookbook code:

Lucid: 10.04
Precise: 12.04
Trusty: 14.04

All of those are EOL and will never be encountered so this logic means we're always going to fill in Ubuntu 10.04, which seems very wrong. I'm also not seeing where this attribute is actually used in the cookbook anymore.

PrajaktaPurohit: Good catch! That should point to the latest.

PR #2050, opened 2020-07-07
Build packages for Ubuntu 20.04

tas50:
We have testers for Ubuntu 20.04 now and it has some pretty serious
performance improvements that would make it the ideal release for
customers running Chef on Ubuntu.


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1349#_

PR #2058, opened 2020-07-28
Remove the dependency for solr, rabbitmq and opscode-expander

PrajaktaPurohit:
- Remove opscode-expander, solr and rabbitmq dependency from the project
- Remove respective directories and stop service
- Solr is kept as a service to be setup to support external-solr (i.e:
elasticsearch) configuration.

https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1438#501e74a0-e1b0-4617-8d5d-e430a2dd2fea

https://buildkite.com/chef/chef-chef-server-master-integration-test/builds/223#803191f9-ce9c-4ae5-b13a-6f13076b824f


omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
```diff
@@ -129,13 +129,11 @@
   postgresql
   oc_bifrost
   oc_id
```
markan, line 131: Might be worth cleaning up the associated recipes the removed items.
* opscode-expander.rb
* rabbitmq.rb

Why does opscode-solr4 remain?

marcparadise:
Will this be the right PR to also remove related attributes?

omnibus/files/private-chef-cookbooks/private-chef/recipes/default.rb
PrajaktaPurohit: opscode-solr4 remains to support the external-solr=elasticsearch setup. That should also be changed eventually to something along the lines of external-elasticsearch.

I could potentially use this PR to remove the attributes and recipes. But if merging this in would mean that the build in current would not have the old java, that might help with a lot of pain. And I can remove the attributes, recipes and eventually the unused code as part of separate PR's, if that seems ok.

PrajaktaPurohit: #2061 and #2062

omnibus/files/private-chef-cookbooks/private-chef/recipes/rabbitmq.rb
```diff
@@ -1,300 +1 @@
...
-
-  execute "#{rmq_ctl} clear_policy -p /analytics max_length" do
-    environment rabbitmq_env
-    user opc_username
-    not_if { rabbitmq['analytics_max_length'] > 0 }
-    retries 10
-  end
-
-end
+raise "Sorry, rabbitmq is no longer used in Chef Infra Server."
```
stevendanna, line 1: I _think_ that given your other changes opscode-solr4 is the only recipe that has the possibility of being run, so you can probably remove this and expander completely.

PrajaktaPurohit: Agreed. I will merge this and create a followup PR removing the smaller bits.

Issue #1947, opened 2020-04-06
500's in erchef on upgrading chef-server


PrajaktaPurohit:
### Chef Server Version
Seen from 13.1.25 (13.1.13) not seen in 13.0.54(13.0.11)

### Platform Details
ubuntu-18.04

### Configuration
Tiered with external Elasticsearch

### Expected Result:
No 500's with sel_conn_closed

### Actual Result:

```
{<<"method=PUT; path=/organizations/foo-org/data/foo-data-bag-name/foo-data-bag-item;
status=500; ">>,{error,{error,{badmatch,{error,{error,sel_conn_closed}}},
[{oc_chef_wm_base,update_from_json,4,[{file,"/var/cache/omnibus/chef-
server/src/oc_erchef/_build/default/lib/oc_chef_wm/src/oc_chef_wm_base.erl"},{line,1111}]},
{webmachine_resource,resource_call,3,[{file,"/var/cache/omnibus/chef-
server/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_resource.erl"},{line,203}]},
{webmachine_resource,do,3,[{file,"/var/cache/omnibus/chef-
server/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_resource.erl"},{line,148}]},
{webmachine_decision_core,resource_call,1,[{file,"/var/cache/omnibus/chef-
server/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},{line,47}]},
{webmachine_decision_core,accept_helper,1,[{file,"/var/cache/omnibus/chef-
server/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},
{line,613}]},{webmachine_decision_core,decision,1,[{file,"/var/cache/omnibus/chef-
server/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},
{line,518}]},{webmachine_decision_core,handle_request,2,[{file,"/var/cache/omnibus/chef-
server/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_decision_core.erl"},{line,32}]},
{webmachine_mochiweb,loop,2,[{file,"/var/cache/omnibus/chef-
server/src/oc_erchef/_build/default/lib/webmachine/src/webmachine_mochiweb.erl"},{line,106}]}]}}}
```

This request does get retried and eventually succeed.

The error is also seen on other endpoints - /nodes, /search

[...]

Issue #1973, opened 2020-04-27
500 Internal Server Error occasionally on "knife status".


kaorihinata:
### Chef Server Version
I'm current using vanilla Chef Infra Server 13.2.0.

### Platform Details
The server is running CentOS 8.1, up to date as of a few days ago. Hardware is a virtual machine (bhyve) running on a FreeBSD 11.3 host (FreeNAS). About 3 GB of RAM has been allotted to the machine, and around 2 GB are currently free.

### Configuration
Configuration is Standalone. This server was installed as 13.1.13 and upgraded once so far to 13.2.0. chef-server.rb is as follows:

```ruby
api_fqdn "removed"
nginx["dhparam_key_length"] = 4096
nginx["non_ssl_port"] = false
nginx["ssl_certificate"] = "/etc/letsencrypt/live/removed/fullchain.pem"
nginx["ssl_certificate_key"] = "/etc/letsencrypt/live/removed/privkey.pem"
nginx["ssl_ciphers"] = [
    "ECDHE-ECDSA-AES256-GCM-SHA384",
    "ECDHE-RSA-AES256-GCM-SHA384",
    "ECDHE-ECDSA-CHACHA20-POLY1305",
    "ECDHE-RSA-CHACHA20-POLY1305",
    "DHE-RSA-AES256-GCM-SHA384",
].join(":")
nginx["ssl_protocols"] = "TLSv1.2 TLSv1.3"
```

### Scenario:
I've been trying to troubleshoot a chef-client hang on a Windows host, and while watching "knife status" for it to check in I received a 500 response from the server. "knife status" being one of the most benign things I could possibly be doing I figured I'd report it just in case someone knows of some situation where a status request may fail.

### Steps to Reproduce:
Reproducing this requires no extra resources, and doesn't involved an upload like most other occurrences of this error do. Simply run "knife status" against a server until it fails. Given it's only happened once so far I don't expect it to occur frequently.

### Expected Result:
```
% knife status
1 hour  ago, removed1, windows 10.0.18363.
53 minutes ago, removed2, freebsd 12.1-RELEASE-p2.
50 minutes ago, removed3, freebsd 12.1-RELEASE-p3.
47 minutes ago, removed4, freebsd 12.1-RELEASE-p2.
```

### Actual Result:

Command output:

```
% knife status
ERROR: Server returned error 500 for https://removed/organizations/removed/search/node?q=*:*&start=0&rows=1000, retrying 1/5 in 3s
[...]

PR #2064, opened 2020-08-03
Sel conn closed

PrajaktaPurohit:
This change pulls the version of opscoderl_httpc that takes an option to retry the ibrowse request on a new connection incase of a sel_conn_closed.
 
Fixes: #1947 and #1973

https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1472

### DIFF
--- a/oc-chef-pedant/spec/running_configs/basic_config_spec.rb
+++ b/oc-chef-pedant/spec/running_configs/basic_config_spec.rb
@@ -165,6 +165,10 @@ describe "running configs required by Chef Server and plugins", :config do
       expect(config['opscode-erchef']['solr_http_max_connection_duration'].to_s).not_to eq ''
     end
 
+    it "opscode-erchef/solr_retry_on_conn_closed" do
+      expect(config['opscode-erchef']['solr_retry_on_conn_closed'].to_s).not_to eq ''
+    end
+
     it "opscode-erchef/solr_ibrowse_options" do
       expect(config['opscode-erchef']['solr_ibrowse_options'].to_s).not_to eq ''
     end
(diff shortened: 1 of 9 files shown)


omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -456,6 +456,7 @@
 default['private_chef']['opscode-erchef']['solr_http_cull_interval'] = '{1, min}'
 default['private_chef']['opscode-erchef']['solr_http_max_age'] = '{70, sec}'
 default['private_chef']['opscode-erchef']['solr_http_max_connection_duration'] = '{70,sec}'
+default['private_chef']['opscode-erchef']['retry_on_conn_closed'] = true
```
stevendanna, line 459: I know it is still a gross misnomer, but I think perhaps we should prefix this with `solr_` still to match the other options?

PrajaktaPurohit: ha! I did go back and forth in doing that, I also though perhaps search_ would be a better option, but since we are removing solr and related things I though perhaps best to leave it without the "solr_" prefix? I dont like that is is inconsistent either.

oc-chef-pedant/spec/running_configs/basic_config_spec.rb
```diff
@@ -165,6 +165,10 @@
       expect(config['opscode-erchef']['solr_http_max_connection_duration'].to_s).not_to eq ''
     end
 
+    it "opscode-erchef/solr_retry_on_conn_closed" do
+      expect(config['opscode-erchef']['retry_on_conn_closed'].to_s).not_to eq ''
```
stevendanna, line 169:
```suggestion
      expect(config['opscode-erchef']['solr_retry_on_conn_closed'].to_s).not_to eq ''
```

PR #2071, opened 2020-08-05
Fix RHEL8 FIPS enablement in integration test

### LINKED ISSUE #2069: FIPS integration test fails with SSH timeout on RHEL 8
The FIPS scenario is now failing when run against RHEL 8 (possibly due to upstream changes to the distro).

During the initial FIPS scenario setup the newly instantiated virtual machine is modified to enable FIPS mode via the addition of `fips=1` to the GRUB boot loader and then the instance is rebooted.

Following the 60 second wait, Terraform attempts to reconnect to the virtual machine and eventually times out due to the machine not being responsive

christopher-snapp:
### Description

This PR fixes the core dump that was occurring in the `omnibus-fips` integration test scenario when run against RHEL8.

It also modifies the integration test pipeline to run `omnibus-fips` against RHEL7 due to an issue with Terraform not properly identifying the type of RSA key used to SSH onto the test instance.  In RHEL8, once FIPS mode is enabled, the machine will no longer allow the SHA1 based `ssh-rsa` algorithm access to the machine (see: https://bugs.centos.org/view.php?id=16720#c36524).

### Issues Resolved

#2069 

### Check List

### DIFF
--- a/.expeditor/integration_test.pipeline.yml
+++ b/.expeditor/integration_test.pipeline.yml
@@ -49,14 +49,14 @@ steps:
       executor:
        docker:
 
-  - label: ':terraform: fips-ipv4-rhel-8'
+  - label: ':terraform: fips-ipv4-rhel-7'
     command: .expeditor/integration_test.pipeline.sh aws apply
     artifact_paths:
       - ./integration_test.log
       - ./*capture_paths.tar.gz
     env:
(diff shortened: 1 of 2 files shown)


PrajaktaPurohit:
Looks like this is failing the azure scenario - but that is not related to this issue.

PR #2090, opened 2020-08-26
add version annotations to integration_test pipeline

### LINKED ISSUE #2087: [terraform]: Add annotations to integration_test_pipeline.
Currently the integration_test_pipeline is setup to run off master for the terraform code.

It is difficult to see what version of Chef-server it was tested with.

Adding annotation like https://buildkite.com/chef/chef-automate-master-verify-private/builds/14067

would make that easier.

Good to Include:
default install version for chef backend
default upgrade version for chef backend
default install for chef-server
default upgrade foe chef-server

christopher-snapp:
### Description

This PR adds a buildkite "info" annotation containing the versions being tested in each `integration_test` pipeline run.

A demonstration of the annotations can be found here:

https://buildkite.com/chef/chef-chef-server-master-integration-test/builds/310

### Issues Resolved

#2087

### Check List

- [X] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

### DIFF
--- a/.expeditor/integration_test.pipeline.sh
+++ b/.expeditor/integration_test.pipeline.sh
@@ -245,6 +245,20 @@ apply () {
   TF_VAR_backend_version_url=$(for channel in unstable current stable; do mixlib-install download chef-backend --url -c $channel -a x86_64 -p "$(sed 's/rhel/el/' <<<"${TF_VAR_platform%-*}")" -l "${TF_VAR_platform##*-}" -v "$BACKEND_VERSION" 2>/dev/null && break; done | head -n 1)
   export TF_VAR_backend_version_url
 
+  # update buildkite annotations
+  if [[ "$(command -v buildkite-agent)" ]]; then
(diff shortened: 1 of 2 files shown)


christopher-snapp:
The install and upgrade versions for Chef Server and the install version of Chef Backend are all required for the pipeline to function.  As a result the environment variables are populated by the pipeline's `.expeditor/integration_test.pipeline.sh` execution script before Terraform is initialized.

The Elasticsearch version isn't required and currently has its default set by Terraform which doesn't surface its value at a level the pipeline script can see.  I've added the print out of the variable value if you are just looking to see if the default version was overridden.

If you need to see the default version number printed out then I'll have to relocate where that default is set or read the default value from the Terraform code.

As for the external PostgreSQL version, we currently don't support overriding that so I'd have to implement that variable override before we can print out a meaningful value.

PR #2096, opened 2020-09-02
Harmonize Treatment of Host Header in Pedant

lbakerchef:
Harmonize treatment of Host header in Pedant.

Chef::ServerAPI always sets the Host header to HOSTNAME:PORT.
We do the same here for consistency in order to avoid sigv4
signing issues.

verify:
https://buildkite.com/chef/chef-chef-server-master-verify/builds/2323#95b54d19-1e87-45a2-9092-3e69ceb99fb7

adhoc:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1605#5a0aca9e-5a4e-49df-a761-cf18f0a27042

### DIFF
--- a/oc-chef-pedant/lib/pedant/rspec/cookbook_util.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/cookbook_util.rb
@@ -330,7 +330,7 @@ module Pedant
           http.ca_file     = Pedant::Config.ssl_ca_file     if Pedant::Config.ssl_ca_file
         end
 
-        response = http.get(uri.request_uri, {})
+        response = Pedant::Utility.get_host_port http, uri
 
         begin
           response.code.should eq expected_reponse_code.to_s
--- a/oc-chef-pedant/spec/api/cookbooks/read_spec.rb
+++ b/oc-chef-pedant/spec/api/cookbooks/read_spec.rb
@@ -325,7 +325,7 @@ describe "Cookbooks API endpoint", :cookbooks, :cookbooks_read do
               http.ca_file     = Pedant::Config.ssl_ca_file     if Pedant::Config.ssl_ca_file
             end
 
-            response = http.get(uri.request_uri, {})
+            response = Pedant::Utility.get_host_port http, uri
             response.body.should == recipe_content
           end
 
--- a/oc-chef-pedant/lib/pedant/request.rb
+++ b/oc-chef-pedant/lib/pedant/request.rb
@@ -137,11 +137,9 @@ module Pedant
       auth_headers = opts[:auth_headers] || requestor.signing_headers(method, url, payload)
 
       uri = URI.parse(url)
-      if (uri.scheme == 'http' && uri.port == 80) || (uri.scheme == 'https' && uri.port == 443)
-        host = uri.host
-      else
-        host = "#{uri.host}:#{uri.port}"
-      end
+
+      # for further details, see: get_host_port in oc-chef-pedant/lib/pedant/utility.rb
+      host = "#{uri.host}:#{uri.port}"
 
       final_headers = standard_headers.
         merge(auth_headers).
--- a/oc-chef-pedant/lib/pedant/utility.rb
+++ b/oc-chef-pedant/lib/pedant/utility.rb
@@ -81,5 +81,12 @@ module Pedant
       end
       return nil # Cannot find the fixture. Raise an error?
     end
+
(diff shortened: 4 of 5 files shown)


oc-chef-pedant/spec/api/cookbooks/read_spec.rb
```diff
@@ -325,7 +325,10 @@ def self.expects_response_of_400_with(message, _value)
               http.ca_file     = Pedant::Config.ssl_ca_file     if Pedant::Config.ssl_ca_file
             end
 
-            response = http.get(uri.request_uri, {})
+            # NOTE: Chef::ServerAPI always sets the Host
+            # header to HOSTNAME:PORT. We do the same here to avoid sigv4
+            # signing issues.
+            response = http.get(uri.request_uri, {"Host" => "#{uri.hostname}:#{uri.port}"})
```
stevendanna, line 331: I know it is silly, but maybe we can pull this out to a utility function so that we can just have this comment in one place.  Also, the comment could do more to explain what the signing issue is.  For instance:

`Chef::ServerAPI always sets the Host header to HOSTNAME:PORT. We do the same thing here since the Version 4 AWS Signature scheme uses the Host header as part of the canonically signed content.`

Also, am I correct that this isn't strictly necessary since we are also doing the Host-header fallback handling in bookshelf for compatibility?

lbakerchef: It is strictly necessary in the case of using S3 vs bookshelf.

lbakerchef: Will investigate pulling this out to a utility function.

oc-chef-pedant/lib/pedant/rspec/cookbook_util.rb
```diff
@@ -330,7 +330,11 @@ def verify_checksum_url(url, expected_reponse_code)
           http.ca_file     = Pedant::Config.ssl_ca_file     if Pedant::Config.ssl_ca_file
         end
 
-        response = http.get(uri.request_uri, {})
+        # NOTE(ssd) 2020-06-15: Chef::ServerAPI always sets the Host
+        # header to HOSTNAME:PORT. We do the same here to avoid v4
+        # signing issues.
```
markan, line 335: How far back does that behavior go? I think it's been in there a while, but it would be worth noting that if we being stricter than we used to. Pedant has been used as a guide for third party API libraries, so they may have incorporated the behavior. This is *not* intended to block this PR, but more of a thing we will have to note when the V4 signing work merges, and probably call out in the release notes, as it might break things (even if they are unsupported, its always nice to note when the contract changes).

markan (approved):

oc-chef-pedant/lib/pedant/rspec/cookbook_util.rb
stevendanna: Yup, the V4 signing is definitely going to break some clients unfortunately.  When I looked at the issue last, I don't see a way around it for S3-backed chef-server.  For bookshelf-backed chef-servers we have some flexibility because we can do the auth check using both headers and accept either.

stevendanna: Note that I _think_ all supported versions of Chef client should be OK but we should double check that, understand how far back users can be while still working, and document the issue for 3rd party clients.

stevendanna: Note, when I say "break some clients" I mean clients broadly.  I think all supported versions of Chef will be OK, but we'll definitely want to document it for 3rd party clients and confirm which versions of Chef-Client work with v4 signing.

lbakerchef: Just to provide this FYI, I did 'hand testing' of the last 3-4 Chef clients a while back (manual knife cookbook uploads and downloads and what not).  Those all passed for the testing I did.

oc-chef-pedant/spec/api/cookbooks/read_spec.rb
lbakerchef: Factored-out things into a utility function.  Let me know if it's what you had in mind.

stevendanna (approved):

stevendanna:
Please clean up the commits before merging, but this looks good to me!

lbakerchef:
Thanks!

PrajaktaPurohit (approved):

PR #2098, opened 2020-09-02
Fix fips errors in RHEL smoke tests

lbakerchef:
Fix fips errors in RHEL smoke tests.

verify:
https://buildkite.com/chef/chef-chef-server-master-verify/builds/2356

adhoc:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1624#a4a53129-f10a-4904-aaa9-f642e1c928e0

integration_test:
https://buildkite.com/chef/chef-chef-server-master-integration-test/builds/333#27b64c44-00fc-478b-ac4b-f5fc5de9eee6

### DIFF
--- a/src/bookshelf/src/bksw_sql.erl
+++ b/src/bookshelf/src/bksw_sql.erl
@@ -226,7 +226,7 @@ init_transfer_state() ->
     #file_transfer_state{
        size = 0,
        next_chunk = 0,
-       hash_context_md5 = crypto:hash_init(md5),
+       hash_context_md5 = erlang:md5_init(),
        hash_context_sha256 = crypto:hash_init(sha256),
        hash_context_sha512 = crypto:hash_init(sha512)
       }.
@@ -244,7 +244,7 @@ update_transfer_state(#file_transfer_state{
 
     TransferState#file_transfer_state{size = Size1,
                                       next_chunk = NextChunk + ChunkIncr,
-                                      hash_context_md5 = crypto:hash_update(ContextMd5, Data),
+                                      hash_context_md5 = erlang:md5_update(ContextMd5, Data),
                                       hash_context_sha256 = crypto:hash_update(ContextSha256, Data),
                                       hash_context_sha512 = crypto:hash_update(ContextSha512, Data)
                                      }.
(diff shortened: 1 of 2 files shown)


src/bookshelf/src/bksw_sql.erl
```diff
@@ -226,7 +226,7 @@ init_transfer_state() ->
     #file_transfer_state{
        size = 0,
        next_chunk = 0,
-       hash_context_md5 = crypto:hash_init(md5),
+       hash_context_md5 = erlang:md5_init(),
```
PrajaktaPurohit, line 229: This will generally be required for FIPS irrespective of sig_v4.

PrajaktaPurohit:
This change might be good to get irrespective of sig_V4 changes.

src/bookshelf/src/bksw_sql.erl
lbakerchef: Sounds good.  It passes all tests, so feel free to merge.

stevendanna (commented):
Let's get this merged.  However, this change has me nervous because it refers to an error I haven't seen.

Can we put an example of the error that this fixes in the commit message? Do we have a link to a build that exhibits this error? Why doesn't this error happen on master (or does it and I just don't know which of the pipelines it happens in)?

PrajaktaPurohit (approved):

lbakerchef:
I'm hoping @PrajaktaPurohit can chime in here with more context; she has a better memory than I do.  We started seeing failures only on the RHEL boxes.  I can't remember if it was adhoc pipeline, or integration test pipeline.  I also can't remember why the failures suddenly appeared.  But we tracked it down to being FIPs-related.

markan (approved):

markan:
@stevendanna From memory, this showed up after we tightened up the FIPS enforcement; the RHEL test https://github.com/chef/chef-server/tree/master/terraform/aws/scenarios/omnibus-fips is stricter in its setup and so finds things that our adhoc pipeline doesn't.

stevendanna:
OK.  I'm not gonna block the merge on this.  But we really need to keep track of these things.  We run the FIPS tests that @markan has pointed out, on rhe, in the integration pipline. And they are passing on the most recent nightly: https://buildkite.com/chef/chef-chef-server-master-integration-test/builds/327#77240db1-1984-45c4-932c-a36de984c254

So, while this fix looks correct and I'm down with merging it, it feels a bit suboptimal to me that we can't include some details of the failure in the commit message.

PR #2102, opened 2020-09-05
Remove support for plugins/extensions including ha-aws

stevendanna:
This fully removes support for "configuration extensions" and
"plugins". This feature was only used by the HA plugins for DRBD and
AWS. Support for the HA topology was removed in Chef Server 13.


omnibus/files/private-chef-cookbooks/private-chef/resources/chef_run.rb
```diff
@@ -1,74 +0,0 @@
-#
```
stevendanna, line 1: Removing this is the big win as it represents a source of pain on every major chef-client upgrade.

srenatus (approved):
Thanks for cleaning this up. 5/5 brooms: :broom: :broom: :broom: :broom: :broom:

PR #2111, opened 2020-09-14
Praj/minus opscode expander

PrajaktaPurohit:
Opscode Expander was a Ruby service responsible for expanding node
objects into the schema expected by our search index.

This commit removes the service as we now do this expansion directly
inside Erchef. The new indexing pipeline allows us to remove the
operational complexity of RabbitMQ for our users. The new indexing
pipeline is not as battle-tested as Expander, but we hope that over
time it can reach the levels of reliability we had with Expander.

Expander was an excellent service. It performed well in our Hosted
Chef environment and for our many users. It had remarkably few bugs in
its history and is a good example of the skill of its authors.

https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1653
https://buildkite.com/chef/chef-chef-server-master-integration-test/builds/354


doc/HOW_THINGS_WORK.md
```diff
@@ -33,24 +33,6 @@ Chef Server has 3 different "search_queue_mode"s:
...
-In the rabbitmq search_queue_mode, erchef places the object to be
-indexed on a rabbitmq queue. It them moves forward with the request
-and returns a response to the user.  Another service named
-`opscode-expander` reads the object from the queue, expands it (see
-Document Expansion for details), and posts it to solr. Note that a
-failure to write to the search index will not trigger a failure of the
-API request that wrote the data since it happens asynchronously via a
-queue. However, a failure to place the item on rabbitmq will cause a
-500 error.
-
```
stevendanna, line 53: Maybe we could keep this but move it to a section about older versions or something.

stevendanna:
We'll be sad to see this service go.  Many thanks to @danielsdeleo and its other authors who created such a reliable service.

stevendanna (approved):

PR #2112, opened 2020-09-14
Included display_name in verbose GET /users response

### LINKED ISSUE #2080: Add ability to retrieve more than just username
# Version:

```
$ chef --version
Chef Development Kit Version: 3.2.30
chef-client version: 14.4.56
delivery version: master (6862f27aba89109a9630f0b6c6798efec56b4efe)
berks version: 7.0.6
kitchen version: 1.23.2
inspec version: 2.2.70
```

# Environment:
We run these on chef sysadmin workstations (Mac and Windows) and from an AWS ECS Fargate container for nightly tasks.

# Scenario:

As a system administrator and also with a group of system administrators, it would be very helpful to be able to query the chef server for all users on the system along with certain details about their accounts, such as email address. This plugin allows us to query for this today, but the process is cumbersome and involves making multiple trips back to the chef server API.

# Steps to Reproduce:

- Run `knife opc user list` to get the entire list of chef users
- In a loop, get the user data for each user:
```
for user in `knife opc user list`; do 
  knife opc user show $user -Fj
done
```
The rest of this would be some parsing of the return values to get the username and email addresses.

# Expected Result:

I'd like to make just one API call that returns to me all of the data that I need -- complete user info for every user. Something like `knife opc user show -a -Fj`.

# Actual Result:

Today, I have to make multiple trips back to the API in order to get the data that I need. For our company, this takes a very long time. To get around this, I've concocted a direct sql query to bypass the API, but this is not future proof if the db schema changes. This also puts other undue requirements accessing the data, because our database has certain rules around inbound access (RDS postgres and AWS security group rules). It's not easily done from a sysadmin's workstation without modifying the security group rules.

antima-gupta:
Signed-off-by: antima-gupta <agupta@msystechnologies.com>

### Description

- Added display_name attribute in user list API.

### Issues Resolved

#2080 

### Check List

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_users.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_users.erl
@@ -107,7 +107,7 @@ finalize_create_body(Req, State, _Client, BodyEJ) ->
 
 to_json(Req, State) ->
     %% In the case of verbose, we cannot  use standard chef_wm_base behavior -
-    %% the client expects the fields email, first_name, last_name - while
+    %% the client expects the fields email, first_name, last_name, display_name - while
     %% the standard response tries to give us a URI.
     %% Secondary note: the original interface in opscode-account supported the combination of
     %% both email filter and verbose option, but this was unused
@@ -136,7 +136,8 @@ verbose_user(#chef_user{username = UserName, email = EMail, serialized_object = 
     EJ = chef_json:decode(SerializedObject),
     {UserName, {[ {<<"email">>, EMail},
                   {<<"first_name">>, ej:get({<<"first_name">>}, EJ, "")},
-                  {<<"last_name">>, ej:get({<<"last_name">>}, EJ, "")} ]} }.
+                  {<<"last_name">>, ej:get({<<"last_name">>}, EJ, "")} ]},
+                  {<<"display_name">>, ej:get({<<"display_name">>}, EJ, "")} }.
 
 conflict_message(_Name) ->
     {[{<<"error">>, [<<"Username or email address already in use.">>]}]}.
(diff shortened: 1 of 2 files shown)


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_users.erl
```diff
@@ -136,7 +136,8 @@ verbose_user(#chef_user{username = UserName, email = EMail, serialized_object =
     EJ = chef_json:decode(SerializedObject),
     {UserName, {[ {<<"email">>, EMail},
                   {<<"first_name">>, ej:get({<<"first_name">>}, EJ, "")},
-                  {<<"last_name">>, ej:get({<<"last_name">>}, EJ, "")} ]} }.
+                  {<<"last_name">>, ej:get({<<"last_name">>}, EJ, "")} ]},
+                  {<<"display_name">>, ej:get({<<"display_name">>}, EJ, "")} }.
```
marcparadise, line 140: You'll need to modify the query[1] to include the new field as well, because `list_users_verbose` query called from `verbose_users_as_ejson` does not return `display_name`

[1] https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_db/priv/pgsql_statements.config#L574

marcparadise (changes requested):
This is headed in the right direction,but seem comment inline. 

Before changing draft status, please also update the relevant test(s): https://github.com/chef/chef-server/blob/master/oc-chef-pedant/spec/api/user_spec.rb#L113

Note that this is also only available in the `verbose` path - from the description of the issue, I'm not sure that's the intent or if the intent is to have it retrieve this data on every GET to `/users` regardless of whether `verbose` is specified.

marcparadise:
The PR title would be more descriptive for purposes of release notes if it were something like  "include display_name in verbose GET /users response"

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_users.erl
stevendanna: @marcparadise This is pulling display_name from the serialized_object so I don't think a query change is required.  The display_name isn't otherwise part of the schema so the serialized_object is the source of truth for this field.

antima-gupta: @marcparadise I have updated the test case.

marcparadise (approved):
Looks good to me pending a rebase and CI run. I have unblocked buildkite.

marcparadise:
The rebase should fix the CI failure: 
```
===> Failed to fetch and copy dep: {git,"git://github.com/chef/erlcloud.git"}
```

markan (approved):

marcparadise (approved):

PR #2126, opened 2020-09-22
fix external elasticsearch scenario on upgrade

PrajaktaPurohit:
[chef-server-ctl] Implement callback for external elasticsearch service
                                  External service shows up as elasticsearch in chef-server-running.json
                                  We maintain the hash for opscode-solr4 to support backcompat with reporting.

 [omnibus] Cleanup all of the deprecated services in the cleanup recipe.
                    The solr, rabbit and opscode-expander recipes were being removed in
                    the elasticsearch recipe. This recipe does not get run in the
                    external-elasticsearch scenario.
                  The cleanup recipe is run on all scenarios and is the right place for
                  well.. cleanup.

https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1695

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/cleanup.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/cleanup.rb
@@ -61,3 +61,31 @@ directory '/opt/opscode/sv/opscode-expander-reindexer' do
   action :delete
   recursive true
 end
+
+
+component_runit_service 'opscode-expander' do
+  action :disable
+end
+
+directory '/opt/opscode/sv/opscode-expander' do
+  recursive true
+  action :delete
+end
+
+component_runit_service 'opscode-solr4' do
+  action :disable
+end
+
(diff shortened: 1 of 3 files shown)


omnibus/files/private-chef-upgrades/001/035_solr_to_elasticsearch_migration.rb
```diff
@@ -25,6 +25,9 @@
       log "Migrating indexed search data..."
       run_command("chef-server-ctl reindex -a -d -t -w")
 
+      # run the cleanup for post upgrade
+      run_command("chef-server-ctl cleanup")
```
stevendanna, line 29: I don't think we want to do this.  This is run by the user once they are sure the upgrade has been successful.

PrajaktaPurohit: fair enough - I went back and forth on this, will remove and add to the docs instead.

stevendanna (approved):

PR #2127, opened 2020-09-23
Elasticsearch options will not exist for versions before 13.2.0 of chef-server

PrajaktaPurohit:
12.x versions of chef-server will not have some of the options added post 13.2

https://buildkite.com/chef/chef-umbrella-master-chef-server/builds/22#_

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_indexing_validator.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_indexing_validator.rb
@@ -114,41 +114,12 @@ class IndexingPreflightValidator < PreflightValidator
     end
   end
 
-  def warn_unchanged_external_flag
-    if OmnibusHelper.has_been_bootstrapped? && backend? && previous_run
-      # ES configuration is preferred everywhere so we check that first
-      previous_external_setting = if previous_run['elasticsearch'].key?('external')
-                                    !!previous_run['elasticsearch']['external']
-                                  elsif previous_run['opscode-solr4'].key?('external')
-                                    !!previous_run['opscode-solr4']['external']
-                                  else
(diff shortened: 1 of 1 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_indexing_validator.rb
```diff
@@ -117,7 +117,9 @@ def elasticsearch_enabled?
   def warn_unchanged_external_flag
     if OmnibusHelper.has_been_bootstrapped? && backend? && previous_run
       # ES configuration is preferred everywhere so we check that first
-      previous_external_setting = if previous_run['elasticsearch'].key?('external')
+      # Check if the elasticsearch hash wiill not be present for versioons before 13.2.0
```
stevendanna, line 120:
```suggestion
      # Check if the elasticsearch hash will not be present for versioons before 13.2.0
```

stevendanna: I'm wondering...is this check even worth it?

stevendanna (approved):

omnibus/files/private-chef-cookbooks/private-chef/libraries/preflight_indexing_validator.rb
PrajaktaPurohit: Hrmm.. This check was useful when we were going to include solr as an option in 14.x. Since that is removed:
13 (internal_solr) -> 14 elasticsearch : reindex with be run on upgrade
13 (snowflake)    -> 14 elasticsearch : skip reindex on upgrade
13 (external_es) -> 14 (we do not manage that)

So I can remove this completely! Nice observation.

PR #2128, opened 2020-09-23
Update migration to exclude managing postgres services, since chef-server-ctl cannot manage external postgres

PrajaktaPurohit:

https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/1701

### DIFF
--- a/omnibus/files/private-chef-upgrades/001/035_solr_to_elasticsearch_migration.rb
+++ b/omnibus/files/private-chef-upgrades/001/035_solr_to_elasticsearch_migration.rb
@@ -11,25 +11,25 @@ define_upgrade do
 
       # Make sure API is down
       stop_services(["nginx", "opscode-erchef"])
-      # Try to call chef-server-ctl reindex
+      # Not including postgres in services here
+      # since external postgres cannot be managed
+      # via chef-server-ctl
       start_services(["elasticsearch",
                       "nginx",
                       "oc_bifrost",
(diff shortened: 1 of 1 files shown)


omnibus/files/private-chef-upgrades/001/035_solr_to_elasticsearch_migration.rb
```diff
@@ -11,25 +11,25 @@
...
       stop_services(["nginx", "opscode-erchef"])
-      # Try to call chef-server-ctl reindex
+      # Not including postgres in services here
+      # since external postgres cannot be managed
+      # via chef-server-ctl
       start_services(["elasticsearch",
                       "nginx",
                       "oc_bifrost",
                       "opscode-erchef",
-                      "postgresql",
```
stevendanna, line 19: I think we should still try to _start_ it if that command is OK since the service might be down at this point and we need it up for the migration to work.

PrajaktaPurohit: It has no effect.
```root@ip-10-0-19-247:~# chef-server-ctl start postgresql
-------------------------------------------------------------------
The service postgresql is running externally and cannot be managed
vi chef-server-ctl.  Please log into 10.0.16.245
to manage it directly.
-------------------------------------------------------------------```

PR #2142, opened 2020-10-08
Upgrade to Elasticsearch 6.8.12

### LINKED ISSUE #2140: Incorrect license in elasticsearch omnibus software definition
The [Omnibus software definition for Elasticsearch](https://github.com/chef/chef-server/blob/master/omnibus/config/software/elasticsearch.rb) states that it's Apache 2.0 licensed but the source URL is linking to the default distribution which contains code under the Elastic License.

The OSS-only version can be downloaded from https://www.elastic.co/downloads/past-releases/elasticsearch-oss-6-8-1. 

Also note that this is release is missing quite a few security patches since the latest in the 6.8 branch is 6.8.12.

stevendanna:
Fixes #2140

### DIFF
--- a/omnibus/config/software/elasticsearch.rb
+++ b/omnibus/config/software/elasticsearch.rb
@@ -15,7 +15,7 @@
 #
 
 name "elasticsearch"
-default_version "6.8.1"
+default_version "6.8.12"
 
 dependency "server-open-jre"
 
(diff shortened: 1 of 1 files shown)


PrajaktaPurohit (approved):
Thank you!

PR #2155, opened 2020-10-27
fix max elasticsearch memory

tehlers320:
Signed-off-by: Timothy Ehlers <tehlers@expedia.com>

### Description

reconfigure fails on big servers. 

### Issues Resolved

https://github.com/chef/chef-server/issues/2154

### Check List

- [x] New functionality includes tests
- [ ] All buildkite tests pass
- [ ] Full omnibus build and tests in buildkite pass
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>
- [ ] PR title is a worthy inclusion in the CHANGELOG

Aha! Link: https://chef.aha.io/features/SH-434

PR #2159, opened 2020-11-05
Tehlers320/max memory es

PrajaktaPurohit:
Rebased the changes from #2155


PrajaktaPurohit:
Merging this since #2155 is approved.

PR #2173, opened 2020-11-18
Adding version to status endpoint

vinay-satish:
### Description

Added chef-server version in status endpoint

### Issues Resolved

resolves #2147

### Check List

- [ ]  All buildkite tests pass

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_status.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/chef_wm_status.erl
@@ -32,14 +32,18 @@
 -export([init/1,
          allowed_methods/2,
          content_types_provided/2,
-         to_json/2]).
+         to_json/2,
+         init_resource_state/1]).
 
--include_lib("webmachine/include/webmachine.hrl").
+-include("oc_chef_wm.hrl").
 -define(A2B(X), erlang:atom_to_binary(X, utf8)).
--define(VERSION_PATH, '/opt/opscode/version-manifest.txt').
 
-init(_Any) ->
-    {ok, <<"{}">>}.
+init(Config) ->
+    oc_chef_wm_base:init(?MODULE, Config).
+
+init_resource_state(_) ->
+    % We're as simple as it gets, no state here.
+    {ok, undefined}.
 
 allowed_methods(Req, State) ->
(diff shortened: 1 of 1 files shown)


src/oc_erchef/apps/oc_chef_wm/src/chef_wm_status.erl
```diff
@@ -36,6 +36,7 @@
 
 -include_lib("webmachine/include/webmachine.hrl").
 -define(A2B(X), erlang:atom_to_binary(X, utf8)).
+-define(VERSION_PATH, '/opt/opscode/version-manifest.txt').
```
stevendanna, line 39: I believe we have a JSON version of this file that might be a bit more consistent to parse.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_status.erl
```diff
@@ -58,19 +59,33 @@ to_json(Req, State) ->
...
+get_version() ->
+    {ok, Device} = file:open(?VERSION_PATH, [read]),
+    %% Assuming that the first line of the file will have the version.
+    Version =
+        case file:read_line(Device) of
+            {ok, Data} -> list_to_binary(lists:subtract(Data,"chef-server \n"));
+            _ -> <<"error">>
+        end,
+    file:close(Device),
+    Version.
```
stevendanna, line 87: I wonder if we might try to do this at statup so that we don't have to read the file on every request?

stevendanna (commented):
Overall this is a good start. I left two suggestions for your consideration.

I'm wondering too if we want to add a test for this new returned data.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_status.erl
vinay-satish: will take a look at it

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_status.erl
vinay-satish: got it. Will make the necessary changes

PrajaktaPurohit:
closing in favor of #2190

PR #2176, opened 2020-11-24
Thread version in from the sup init state to avoid being called for every request

PrajaktaPurohit:

root@api:~# curl -k  https://192.168.33.100/_status
{"status":"pong","upstreams":{"chef_elasticsearch":"pong","chef_sql":"pong","oc_chef_authz":"pong"},"keygen":{"keys":10,"max":10,"max_workers":2,"cur_max_workers":2,"inflight":0,"avail_workers":2,"start_size":0},"indexing":{"mode":"batch"},"version":"14.0.67+20201111171939"}

### DIFF
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_sup.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_sup.erl
@@ -11,9 +11,12 @@
 -export([start_link/0]).
 
 %% supervisor callbacks
--export([init/1]).
+-export([init/1,
+         default_resource_init/0]).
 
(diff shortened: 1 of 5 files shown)


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_sup.erl
```diff
@@ -11,9 +11,12 @@
 -export([start_link/0]).
 
 %% supervisor callbacks
--export([init/1]).
+-export([init/1,
+         default_resource_init/0]).
```
marcparadise, line 15: Does this still need to be exported? I can't find reference to it outside of this file.

PR #2199, opened 2020-12-22
Update the label configuration in the Expeditor config

tas50:
Skip builds / version bumps when the Aspect: Documentation label is applied
Update the labels used to skip builds to use the new Expeditor labels

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -1,5 +1,6 @@
 # Documentation available at https://expeditor.chef.io/docs/getting-started/
 ---
+# The name of the product keys for this product (from mixlib-install)
 product_key: chef-server
 
 # Slack channel in Chef Software slack to send notifications about build failures, etc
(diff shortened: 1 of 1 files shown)


.expeditor/config.yml
```diff
@@ -50,28 +51,32 @@ merge_actions:
...
         - "Expeditor: Skip All"
         - "Expeditor: ACC Only"
+        - "Aspect: Documentation"
       only_if: built_in:bump_version
   - trigger_pipeline:habitat/build:
       post_commit: true
       ignore_labels:
-        - "Omnibus: Skip Build"
+        - "Expeditor: Skip Habitat"
         - "Expeditor: Skip All"
```
PrajaktaPurohit, line 77: I would have expected Skip All to handle Skipping Habitat

tas50: Hab should be skipped if someone uses Skip All or Skip Habitat.

tduffield (approved):

Issue #2219, opened 2021-01-20
Support using external JRE instead of embedded


qudongfang:
Support using external JRE instead of embedded

qudongfang:
https://chef-software.ideas.aha.io/ideas/SH-I-178

tas50:
Closing based on feedback left in the PR

PR #2220, opened 2021-01-20
SH-I-178: support using external JRE instead of the embedded one

qudongfang:
### Description

support using external JRE instead of the embedded one

### Issues Resolved

https://github.com/chef/chef-server/issues/2219

### Check List

### DIFF
--- a/docs-chef-io/content/server/v14_0/config_rb_server_optional_settings.md
+++ b/docs-chef-io/content/server/v14_0/config_rb_server_optional_settings.md
@@ -1457,6 +1457,10 @@ This configuration file has the following settings for `elasticsearch`:
 
 : Default value: `90`
 
+`elasticsearch['java_home']`
+
+: `JAVA_HOME`. Default value: `/opt/opscode/embedded/open-jre`.
+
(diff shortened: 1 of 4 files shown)


tas50:
While I appreciate the effort you've put into this issue (and PR), it's not something that we'd like to include in the Chef Infra Server. Chef Infra Server is a collection of many individual services, which interact with each other in complex ways. We take great effort to ensure that each of those components are fully tested including the bundled release of Java. Before we ship a release with a new version of Java, the entire package undergoes an extensive end to end testing process to make sure we don't introduce potentially impacting regressions. Allowing versions of Java that are not fully validated would most likely introduce hard to detect and potentially problematic failures for users.

If there are particular issues with our Java configuration or version of Java that we ship I'd highly encourage you to raise those problems, but we do want to make sure that folks consuming the Chef Infra Server package are always running a validated release of Java.

qudongfang:
Thanks for your response @tas50, I ran into a situation that requires me to update JDK due to security vulnerabilities with the embedded JDK version. I'll create another issue to give details.

PR #2225, opened 2021-01-22
Update the pedant Gemfile to track chef-15

lamont-granquist:
Seems like the right thing to do here.  There's a Gemfile.lock so
its pinned, but chef-server in general needs to track master of
the chef-15 branch now to get updates for omnibus-software.


PrajaktaPurohit:
closing in favor of #2228 2228

PR #2231, opened 2021-01-28
Update Elasticsearch to 7.9.1

PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2017

NOTE: eabee67f3b2ebeaee00cbeb07e073633660c4a88 currently removes compatibility with the older versions.
We need to support older versions as well.

### DIFF
--- a/oc-chef-pedant/lib/pedant/rspec/search_util.rb
+++ b/oc-chef-pedant/lib/pedant/rspec/search_util.rb
@@ -718,7 +718,9 @@ module Pedant
       end
 
       def get_response_count(r)
-        if r["response"].nil?
+        if (r["response"].nil?) and (r["hits"]["total"].is_a? Hash)
+          r["hits"]["total"]["value"]
+        elsif r["response"].nil?
           r["hits"]["total"]
         else
           r["response"]["numFound"]
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -173,6 +173,7 @@ elasticsearch['enable_gc_log'] = false
 elasticsearch['initial_cluster_join_timeout'] = 90
 elasticsearch['shard_count'] = 5
 elasticsearch['replica_count'] = 1
+#elasticsearch['es_version'] = '6.8.12'
 
 # each item in this list will be placed as-is into java_opts config file.
 # entries are set in chef-server.rb as
--- a/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
@@ -81,7 +81,7 @@ class OmnibusHelper
         # Decorate any JSON parsing exception or numeric exceptions when accessing and parsing the version field.
         raise "Unable to parse elasticsearch response #{e}"
       end
-      raise "Unsupported elasticsearch version of #{version}. There is current support for the major versions of 2, 5 and 6." if version != 6 && version != 5 && version != 2
+      raise "Unsupported Elasticsearch version of #{version}. There is currently support for the major versions of 2, 5, 6 and 7." unless [2, 5, 6, 7].include?(version)
       version
     else
       # Elasticsearch is disabled - this configuration setting should never be used in erlang.
@@ -91,11 +91,62 @@ class OmnibusHelper
 
   def es_index_definition
     es_version = elastic_search_major_version
-    if elastic_search_major_version == 6
+    if elastic_search_major_version == 7
+      es_7_index
+    elsif elastic_search_major_version == 6
       es_6_index
     elsif [2, 5].include?(es_version)
       es_5_or_2_index
     end
+  end
+
+  def es_7_index
+    {
+      'settings': {
+        'analysis': {
+          'analyzer': {
+            'default': {
+              'type': 'whitespace'
+            }
+          }
+        },
+        'number_of_shards':
+          node['private_chef']['elasticsearch']['shard_count'],
+        'number_of_replicas':
+          node['private_chef']['elasticsearch']['replica_count']
+      },
+      'mappings': {
+        '_source': {
+          'enabled': false
+        },
+        'properties': {
+          'X_CHEF_database_CHEF_X': {
+            'type': 'keyword',
+            'index': true,
+            'norms': false
+          },
+          'X_CHEF_type_CHEF_X': {
+            'type': 'keyword',
+            'index': true,
+            'norms': false
+          },
+          'X_CHEF_id_CHEF_X': {
+            'type': 'keyword',
+            'index': true,
+            'norms': false
+          },
+          'data_bag': {
+            'type': 'keyword',
+            'index': true,
+            'norms': false
+          },
+          'content': {
+            'type': 'text',
+            'index': true
+          }
+        }
+      }
+    }
   end
 
   def es_6_index
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch.yml.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch.yml.erb
@@ -19,6 +19,17 @@ network.host: "<%= @listen %>"
 http.port: <%= @port %>
 
 <%
+  # ES7 mandatory configuration.These settings are applicable only for the users using standalone or tiered configurations. 
+  # For BYOD(external elasticsearch) none of the omnibus configs are applicable.
+  # Note: we have not enabled bootstrap checks, so failing bootstrap checks will appear as warnings in the es logs.
+  if node['private_chef']['elasticsearch']['es_version'].to_f > 7.0
+%>
+discovery.type: single-node
+<%
+  end
+%>
+
+<%
   # ES uses secomp in production mode. This is not supported in RHEL <  7
   # Reference:
   # https://www.elastic.co/guide/en/elasticsearch/reference/master/_system_call_filter_check.html
@@ -32,3 +43,5 @@ bootstrap.system_call_filter: false
     end
   end
 %>
+
+
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch_jvm_es7.opts.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch_jvm_es7.opts.erb
@@ -1,0 +1,63 @@
+## JVM Configuration
+
+<% # Wants: log_dir, enable_gc_log, heap_size, new_size, jvm_opts, tmp_dir %>
+
+<% @jvm_opts.each do |opt| %>
+<%=opt%>
+<% end %>
+# Do not override xmx/xms/newsize in
+# jvm_options - instead set the attributes in chef-server.rb
+-Xmx<%=@heap_size%>m
+-Xms<%=@heap_size%>m
+-XX:NewSize=<%= @new_size %>M
+
+# These are defaulted based on what 7.9 ships with.
+# https://github.com/elastic/elasticsearch/blob/7.9/distribution/src/config/jvm.options
+
+## GC configuration
+8-13:-XX:+UseConcMarkSweepGC
+8-13:-XX:CMSInitiatingOccupancyFraction=75
+8-13:-XX:+UseCMSInitiatingOccupancyOnly
+
+## G1GC Configuration
+# NOTE: G1 GC is only supported on JDK version 10 or later
+# to use G1GC, uncomment the next two lines and update the version on the
+# following three lines to your version of the JDK
+# 10-13:-XX:-UseConcMarkSweepGC
+# 10-13:-XX:-UseCMSInitiatingOccupancyOnly
+14-:-XX:+UseG1GC
+14-:-XX:G1ReservePercent=25
+14-:-XX:InitiatingHeapOccupancyPercent=30
+
+## JVM temporary directory
+-Djava.io.tmpdir=<%= @tmp_dir %>
+
+## heap dumps
+
+# generate a heap dump when an allocation from the Java heap fails
+# heap dumps are created in the working directory of the JVM
+-XX:+HeapDumpOnOutOfMemoryError
+
+# specify an alternative path for heap dumps; ensure the directory exists and
+# has sufficient space
+-XX:HeapDumpPath=<%= @tmp_dir %>
+
+# specify an alternative path for JVM fatal error logs
+-XX:ErrorFile=<%= @tmp_dir %>
+
+## GC logging
+<% if @enable_gc_log %>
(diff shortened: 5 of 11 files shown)


omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -91,13 +91,64 @@ def elastic_search_major_version
...
+              'type': 'whitespace'
+            }
+          }
+        },
+        'number_of_shards':
+          node['private_chef']['elasticsearch']['shard_count'],
+        'number_of_replicas':
+          node['private_chef']['elasticsearch']['replica_count']
+      },
+      'mappings': {
```
PrajaktaPurohit, line 118: it might be ok to leave the object index here.

src/oc_erchef/apps/chef_index/src/chef_index_expand.erl
```diff
@@ -130,14 +130,13 @@ make_doc_for_del(solr, Id) ->
...
-    [<<"{\"delete\":{\"_index\":\"chef\",\"_type\":\"object\",\"_id\":\"">>,
+    [<<"{\"delete\":{\"_index\":\"chef\",\"_id\":\"">>,
      Id, <<"\" }}\n">>].
 
 make_doc_for_add(Command = #chef_idx_expand_doc{id = Id, type=Type, search_provider=elasticsearch}) ->
     MetaFieldsPL = meta_fields(Command),
     [jiffy:encode({[{<<"index">>, {[
                                    {<<"_index">>, <<"chef">>},
-                                   {<<"_type">>, <<"object">>},
                                    {<<"_id">>, Id}
```
PrajaktaPurohit, line 140: L265 of this file uses path as /chef/object. That might need to be replaced with `ok = chef_index_http:delete("/chef/_doc/" ++ Id, [], ?JSON_HEADER),`

jashaik:
Adding the pipeline status - 
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2048#83290ef9-0f2c-45fe-b076-7dc0103a4acd

omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb
```diff
@@ -81,7 +81,7 @@ def elastic_search_major_version
         # Decorate any JSON parsing exception or numeric exceptions when accessing and parsing the version field.
         raise "Unable to parse elasticsearch response #{e}"
       end
-      raise "Unsupported elasticsearch version of #{version}. There is current support for the major versions of 2, 5 and 6." if version != 6 && version != 5 && version != 2
+      raise "Unsupported elasticsearch version of #{version}. There is current support for the major versions of 2, 5 and 6." if version != 7 && version != 6 && version != 5 && version != 2
```
tas50, line 84:
```suggestion
      raise "Unsupported elasticsearch version of #{version}. There is currently support for the major versions of 2, 5 and 6." unless [2, 5, 6, 7].include?(version)
```

PrajaktaPurohit: s/There is currently support for the major versions of 2, 5 and 6./There is currently support for the major versions of 2, 5 6 and 7.

oc-chef-pedant/lib/pedant/rspec/search_util.rb
```diff
@@ -718,7 +718,9 @@ def direct_solr_query(type, query)
       end
 
       def get_response_count(r)
-        if r["response"].nil?
+        if (r["response"].nil?) and (r["hits"]["total"].is_a? Hash)
+          r["hits"]["total"]["value"]
```
PrajaktaPurohit, line 722: Is this trying to fix for the case when there is data in the system?

jashaik: get_response_count function is to return the number documents found.There is a response structure change for  total field in the ES7.
`"hits" : {
    "total" : {
      "value" : 1,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "chef",
        "_type" : "_doc",
        "_id" : "2bed8b78e2e51e0c13c82bbbf07f0ad1",
        "_score" : 1.0
      }]}`

in previous versions, Total filed is directly a number indicating the number of documents found.So Added another simple condition to check whether it is a another json or not.

src/oc_erchef/apps/chef_index/test/chef_index_expand_tests.erl
```diff
@@ -326,7 +326,19 @@ es_api_test_() ->
                                             ok
                                     end),
                         AddDoc = chef_index_expand:doc_for_index(role, <<"a1">>, <<"db1">>, MinItem),
-                        ?assertEqual(ok, chef_index_expand:send_item(AddDoc))
+                        ?assertEqual(ok, chef_index_expand:send_item(AddDoc)),
```
PrajaktaPurohit, line 329: nice!

omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch.yml.erb
```diff
@@ -18,6 +18,15 @@ network.host: "<%= @listen %>"
 #            http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-discovery.html
 http.port: <%= @port %>
 
+<%
+  # ES7 mandatory configuration
+  if node['private_chef']['elasticsearch']['es_version'].to_f > 7.0
+%>
+discovery.type: single-node
+<%
```
PrajaktaPurohit, line 26: on the similar lines there are changes in jvm opts (elasticsearch_jvm.opts.erb) fir elasticsearch 7. I wonder if it might be useful to add a separate templates for es7 and have the recipe pick the correct template based on the version of es.

PrajaktaPurohit: It might be helpful to include context for this in comments. These setting are applicable only for the users using standalone or tiered configurations. For BYOD(external elasticsearch) none of the omnibus configs are applicable.

omnibus_overrides.rb
```diff
@@ -10,6 +10,7 @@
 override :ohai, version: "v15.12.0"
 override :ruby, version: "2.6.6"
 override :perl, version: "5.18.1"
+override :elasticsearch, version: "7.9.1"
```
PrajaktaPurohit, line 13: to be removed before merging.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -173,6 +173,7 @@
 elasticsearch['initial_cluster_join_timeout'] = 90
 elasticsearch['shard_count'] = 5
 elasticsearch['replica_count'] = 1
+elasticsearch['es_version'] = '7.9.1'
```
PrajaktaPurohit, line 176: Is this required?

PrajaktaPurohit:
Have you had a chance to run the umbrella pipeline by setting ELASTIC_VERSION=7? The default setting is to run with Elasticsearch 6 for external elasticsearch.

PrajaktaPurohit:
The log files at `/var/log/opscode/elasticsearch/current` also lists the deprecation warnings. This might be a good time to  get the fixes in (if any)

omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch.yml.erb
PrajaktaPurohit: Looks like chef-server lets the user set the listen address for elasticsearch: https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb#L170 which is used in the elasticsearch.yml. So in theory this could be a multicluster setup - although I do not think that is something we advertise / support at this point.  It might be worth making a note in the docs that cluster for es will not be supported, if they want to use a cluster they can setup external es.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
jashaik: yes, For identifing the ES version before installation used in elasticsearch.yml.erb file.

jashaik:
> Have you had a chance to run the umbrella pipeline by setting ELASTIC_VERSION=7? The default setting is to run with Elasticsearch 6 for external elasticsearch.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch.yml.erb
jashaik: Sure, I will add comments and update docs.

omnibus_overrides.rb
jashaik: Yes, I have it in my TODO list

oc-chef-pedant/lib/pedant/rspec/search_util.rb
lbakerchef: With the `and`, I think parentheses aren't necessary.  But if you had used `&&`, they would be necessary:
```
irb(main):061:0> true and {}.is_a? Hash
=> true
irb(main):062:0> false and {}.is_a? Hash
=> false
irb(main):063:0> false or {}.is_a? Hash
=> true
irb(main):064:0> true && {}.is_a? Hash
Traceback (most recent call last):
        1: from C:/Ruby25-x64/bin/irb.cmd:19:in `<main>'
SyntaxError ((irb):64: syntax error, unexpected tCONSTANT, expecting end-of-input)
true && {}.is_a? Hash
                 ^~~~
irb(main):065:0> true && ({}.is_a? Hash)
=> true
```
Up to you if you consider parentheses more readable.

lbakerchef (commented):
Looks ok to me.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
PrajaktaPurohit: Here is what I was thinking about: https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/libraries/helper.rb#L58 detects the version of elasticsearch (after it is running) and then created the index accordingly. But in case of es7, we seem to require this information inorder to start es.

The auto-detection is helpful to make sure anyone running external elasticsearch does not run an unsupported version of elasticsearch. Whether is should be in that helper library or a preflight check is perhaps a question for a different time.

But since we are already doing that - I think of this parameter as redundant. Would es 2,5 and 6 continue to work if the new settings for 7 were made default or does that block from them starting up?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch.yml.erb
PrajaktaPurohit: Are there any change to elasticsearch_jvm.opts?

jashaik:
> Have you had a chance to run the umbrella pipeline by setting ELASTIC_VERSION=7? The default setting is to run with Elasticsearch 6 for external elasticsearch.

Here is the link for external elasticserach version 7 - https://buildkite.com/chef/chef-umbrella-master-chef-server/builds/276#787cb2a9-bfbd-4b1c-a4bb-ea6a63c62689

jashaik:
New pipeline reports after adding ES7 JVM options
Adhoc - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2214

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
PrajaktaPurohit: We should add this to the docs.

jashaik: Updated the docs

omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch.yml.erb
```diff
@@ -18,6 +18,16 @@ network.host: "<%= @listen %>"
 #            http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-discovery.html
 http.port: <%= @port %>
 
+<%
+  # ES7 mandatory configuration.These settings are applicable only for the users using standalone or tiered configurations. 
+  # For BYOD(external elasticsearch) none of the omnibus configs are applicable.
+  if node['private_chef']['elasticsearch']['es_version'].to_f > 7.0
+%>
+discovery.type: single-node
```
PrajaktaPurohit, line 26: It might be useful to leave a note explaining that we have not enabled bootstrap checks, so failing bootstrap checks will appear as warnings in the es logs. https://www.elastic.co/guide/en/elasticsearch/reference/current/bootstrap-checks.html

docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
```diff
@@ -1508,6 +1508,16 @@ If new_size or heap_size is also specified directly in java_opts, it will be ign
 
 {{< /note >}}
 
+`elasticsearch['es_version']`
```
PrajaktaPurohit, line 1511: This will only be supported until we finish upgrading internal es to 7, correct?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch_jvm_es7.opts.erb
```diff
@@ -0,0 +1,63 @@
+## JVM Configuration
```
marcparadise, line 1: I have not verified these, but are they based on prior version opts? And have we looked to see if we need to change anything based on what's new in ES7?

marcparadise (approved):
I'm surprised we were still doing 'norms' => { enabled: false } in es6 without problems, it looks like the syntax has been 'norms: false' since v5.0.

These changes look good. I left one question inline; assuming there is no further work to do there, I'm +1 to merge.

jashaik:
New pipelines after commenting the es_version in default.rb file
Umbrella - https://buildkite.com/chef/chef-umbrella-master-chef-server/builds/294
Adhoc - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2251

omnibus/files/private-chef-cookbooks/private-chef/templates/default/elasticsearch_jvm_es7.opts.erb
jashaik: Yes, We are referring the latest ES7 JVM configurations - https://github.com/elastic/elasticsearch/blob/7.9/distribution/src/config/jvm.options

PrajaktaPurohit (commented):
Cannot approve this PR. It looks good to me! Thanks!

Issue #2254, opened 2021-02-19
Update OpenSSL to 1.0.2y


tas50:
There is a new version of OpenSSL 1.0.2 out that resolves 3 CVEs. This will be merged into omnibus-software shortly.

tas50:
This is in the current builds now

PR #2255, opened 2021-02-19
Add Gemfile.lock for omnibus

tas50:
This way we have repeatable builds. This particular update also brings in the new openssl 1.0.2y


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2112#dc0e6f42-b797-4950-8e34-cecb9bb01c11

PrajaktaPurohit:
Thank you for taking the time to do this @tas50. While I understand the rationale for the change and dependable builds would be great, I am a little skeptical adding a Gemfile.lock for omnibus unless there is a mechanism to autoupdate the Gemfile.lock on every change to omnibus-software. 
Some of the things I am concerned about:
1. It is difficult to track failures in the pipeline, I worry that adding Gemfile.lock will provide a false sense of security and updating chef-server will start to fall behind.
2. It is often easier to fix the error in the context of the change made.
3. The other issue would be that that would mean master is not shippable.
I am open to discussing this further. As an interim it might be helpful to run a chef-server adhoc build that points to the branch in omnibus-software with changes before a PR in omnibus-software is merged.

PrajaktaPurohit:
I think we can merge this once we add the dependabot to create PR's on merge to omnibus-software. (and after rebase) :)

tas50:
This fixes https://github.com/chef/chef-server/issues/2254

tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2119#7951c31c-8fb0-4487-b229-cf849e8be735

tas50:
Merging this to prevent further changes in omnibus/omnibus-software from coming in while we troubleshoot failures

PR #2268, opened 2021-02-25
Cache gem installs for License Scout

tas50:
Let's see if we can speed up these builds

### DIFF
--- a/.expeditor/license_scout.sh
+++ b/.expeditor/license_scout.sh
@@ -33,7 +33,7 @@ bundle_install_dirs=(
 for dir in "${bundle_install_dirs[@]}"; do
   echo "--- Installing gem dependencies for $dir"
   pushd "src/$dir"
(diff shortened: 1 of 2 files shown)


tduffield (approved):

tas50:
This didn't seem to actually use the cache when I retried the job. I'll have to dig into this. The 18 minute job could easily be cut in 1/2 if we could figure this out though.

tduffield:
So the issue is that the vendor directory is actually within each of the `src/*` folders, not at the top level. So you'll either need to specify each of the nested `src/*/vendor` directories in your `cached_folders` setting, or update your `bundle install` vendor target to target the `vendor` directory in the root of the project.

tas50:
@tduffield Updated to cache to the workdir/vendor

PrajaktaPurohit (approved):

PR #2281, opened 2021-03-02
Unpin chef-cli in omnibus

tas50:
This was pinned with a note that chef-cli requires ruby 2.7. It requires 2.5+ now.


tas50:
adhoc: https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2166#65d79df3-d66b-40ef-9687-e3cb43d7bd91

PR #2283, opened 2021-03-03
Let the Chef dep in omnibus float to the latest

tas50:
Omnibus should be able to run the very latest Chef since we're just using Chef here for the actual build of the application. None of these deps impact the artifact we build directly. In general we should pin omnibus deps to hold back changes to omnibus/omnibus-software, but everything *should* be able to build at the latest. 

- Remove the chef pin
- Remove chef-cli dep entirely since we really just had this in here for chefspec
- Add Test Kitchen dep since there's a kitchen config here, but not Test Kitchen
- Cleanup the Test Kitchen config a little bit. Remove the chef version pin, the unused yum-epel dep, and old platform versions in the kitchen.yml
- Move all the test aka non-build deps into the test group. There's rspec for the cookbooks here, which we probably want to get running in CI


tas50:
Adhoc: https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2170

PR #2288, opened 2021-03-05
Speed up more tests with gem caching

tas50:
Cache our bundle installs to S3 so we don't have to compile the gems
each time


lbakerchef (approved):
That's all it takes for caching? Nice!  I know I don't have to ask, but I take it this has already been tested?

tduffield:
@lbakerchef it this in combination with https://github.com/chef/chef-server/blob/master/.expeditor/verify.pipeline.yml#L3-L4 that makes it work.

tduffield (approved):

PR #2305, opened 2021-03-06
Minor Expeditor config cleanup

tas50:
Remove the branch cleanup config that doesn't trigger anymore
Remove branch setup for 13.x since we're not going to release that
Remove a comment that doesn't apply anymore
Update the docs URL

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -8,7 +8,6 @@ slack:
   notify_channel: chef-server-notify
 
 github:
-  delete_branch_on_merge: true
   # The file where the MAJOR.MINOR.PATCH version is kept. The version in this file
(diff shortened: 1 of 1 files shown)


.expeditor/config.yml
```diff
@@ -8,7 +8,6 @@ slack:
   notify_channel: chef-server-notify
 
 github:
-  delete_branch_on_merge: true
```
tduffield, line 11: We should keep this if that's what you want to happen. I'll make sure it gets fixed next week.

tduffield (approved):

.expeditor/config.yml
PrajaktaPurohit: That would be awesome! Thanks @tduffield!

PR #2314, opened 2021-03-09
Add a post install message to the packages

tas50:
Let folks know to check out our docs post-install. This only fires on
the first install the way the logic here works.

### DIFF
--- a/omnibus/package-scripts/chef-server/postinst
+++ b/omnibus/package-scripts/chef-server/postinst
@@ -29,7 +29,7 @@ chmod o-rwx /etc/opscode/chef-server.rb
 chown -Rh 0:0 /opt/opscode
 
 if [ -e /etc/opscode/chef-server-running.json ]; then
-  echo -e "\033[1mYou have upgraded Chef Server!\033[0m"
+  echo -e "\033[1mYou have upgraded Chef Infra Server!\033[0m"
   echo
   echo "The next step in the upgrade process is to run:"
   echo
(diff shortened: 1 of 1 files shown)


omnibus/package-scripts/chef-server/postinst
```diff
@@ -52,6 +52,10 @@ if [ -e /etc/opscode/chef-server-running.json ]; then
   echo
   echo "For detailed upgrade instructions please see:"
   echo "https://docs.chef.io/server/upgrades/"
+else
+  echo -e "\033[1mThank you for installing Chef Infra Server!\033[0m"
+  echo
+  echo "For more information on getting started see https://docs.chef.io/server/"
```
PrajaktaPurohit, line 58: Should it point to the install page instead? https://docs.chef.io/server/install_server/

omnibus/package-scripts/chef-server/postinst
```diff
@@ -52,6 +52,10 @@ if [ -e /etc/opscode/chef-server-running.json ]; then
   echo
   echo "For detailed upgrade instructions please see:"
   echo "https://docs.chef.io/server/upgrades/"
+else
+  echo -e "\033[1mThank you for installing Chef Infra Server!\033[0m"
```
PrajaktaPurohit, line 56: Should we include "The next step in install process is to run: `chef-server-ctl reconfigure`" here?

tas50: We could probably say run this command and check out xyz.com for more information

PrajaktaPurohit (approved):

PR #2325, opened 2021-03-12
MTLS server issue: knife config need spaces

PrajaktaPurohit:

On enabling mtls on the Chef Infra Server, the chef-server-ctl commands running knife on local host fail with a 400.
It is a config issue fixed by replacing equals with spaces.  

Manage, Automate and Infra views still need to be tested with these changes.


omnibus/files/private-chef-cookbooks/private-chef/templates/default/pivotal.rb.erb
```diff
@@ -17,8 +17,8 @@ client_key key.path
 
  <% if node['private_chef']['nginx']['ssl_client_ca'] -%>
 ssl_ca_file = "<%= node['private_chef']['nginx']['ssl_client_ca'] %>"
```
tas50, line 19:
```suggestion
ssl_ca_file "<%= node['private_chef']['nginx']['ssl_client_ca'] %>"
```

PR #2326, opened 2021-03-12
Prevent chef.io/progress.com emails from using ZenDesk SSO

btm:
As a temporary workaround, accounts with chef.io and progress.com email addresses
cannot use the ZenDesk SSO logins. This does not prevent ZenDesk Agent
logins, which are separate. This only prevents logins using your Hosted
Chef/manage.chef.io/id.chef.io credentials.

### DIFF
--- a/src/oc-id/spec/controllers/zendesks_controller_spec.rb
+++ b/src/oc-id/spec/controllers/zendesks_controller_spec.rb
@@ -29,15 +29,40 @@ describe ZendesksController do
       end
 
       context 'when signed in' do
+        let(:testuser) do
+          User.new({
+            :username => 'jimmy',
+            :email => 'jim.kirk@federation-captains.org',
+          })
+        end
         before :each do
           allow(controller).to receive(:signed_in?).and_return(true)
-          allow(controller).to receive(:current_user).and_return('testuser')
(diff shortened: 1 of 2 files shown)


marcparadise (approved):

src/oc-id/spec/controllers/zendesks_controller_spec.rb
```diff
@@ -29,17 +29,43 @@
...
+          User.new({
+            :username => 'jimmy',
+            :email => 'jim.kirk@federation-captains.org',
+          })
+        end
         before :each do
           allow(controller).to receive(:signed_in?).and_return(true)
-          allow(controller).to receive(:current_user).and_return('testuser')
+          allow(controller).to receive(:current_user).and_return(testuser.email)
+          allow(controller).to receive(:current_user).and_return(testuser)
```
marcparadise, line 41: This seems weird - we're allowing it to receive :current_user twice; the first return is email and thee second is user object. Is that intentional?

PrajaktaPurohit: fixed! thanks!

PrajaktaPurohit (approved):

PR #2332, opened 2021-03-15
Update Elasticsearch to 6.8.14

tas50:
This resolves a CVE patched in 6.8.14 which we should get out there.

### DIFF
--- a/omnibus/config/software/elasticsearch.rb
+++ b/omnibus/config/software/elasticsearch.rb
@@ -1,5 +1,5 @@
 #
-# Copyright 2019 Chef Software, Inc.
+# Copyright:: Chef Software, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2276#95dd0f0c-5b21-4c78-9e82-649f2d21de16

omnibus/config/software/elasticsearch.rb
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2019 Chef Software, Inc.
+# Copyright:: Chef Software, Inc.
```
PrajaktaPurohit, line 2: should this be 2021?

PrajaktaPurohit (approved):
Can be merged once adhoc is green.

omnibus/config/software/elasticsearch.rb
tas50: We actually don't need dates here since the copyright is some massive length now. We've removed them in most places and I nuke them as I see them. The :: is an rdoc thing in chef that Adam sprinkled everywhere.

PR #2333, opened 2021-03-16
Added disk space warning if not enough for ES reindex

jashaik:

### Description

Raise disk space warning if disk space is not enough for ES reindexing

### Issues Resolved

Added function to verify disk space in the preflight checks

https://app.zenhub.com/workspaces/chef-infra-server-team-5fc64867d45ca500173dbbc7/issues/chef/chef-server/2197

### Check List

### DIFF
--- a/src/chef-server-ctl/lib/chef_server_ctl/helpers/du.rb
+++ b/src/chef-server-ctl/lib/chef_server_ctl/helpers/du.rb
@@ -1,0 +1,22 @@
+require 'mixlib/shellout'
+module Du
+  # Calculate the disk space used by the given path. Requires that
+  # `du` is in our PATH.
+  #
+  # @param path [String] Path to a directory on disk
+  # @return [Integer] KB used by directory on disk
+  #
+  def self.du(path)
+    command = Mixlib::ShellOut.new("du -sk #{path}")
+    command.run_command
+    if command.status.success?
+      command.stdout.split("\t").first.to_i
+    else
+      puts "du -sk #{path} failed with exit status: #{command.exitstatus}"
+      puts "du stderr: #{command.stderr}"
+      raise 'DuFailedException', command.stderr
+    end
+  rescue Errno::ENOENT
+    raise 'DuFailedException', command.stderr
+  end
+end
--- a/src/chef-server-ctl/lib/chef_server_ctl/helpers/statfs.rb
+++ b/src/chef-server-ctl/lib/chef_server_ctl/helpers/statfs.rb
@@ -1,0 +1,79 @@
+require 'ffi'
+
+class Statfs
+  #
+  # Statfs provides a simple interface to the statvfs system call.
+  # Since the statvfs struct varies a bit across platforms, this
+  # likely only works on Linux and OSX at the moment.
+  #
+  extend FFI::Library
+  ffi_lib FFI::Library::LIBC
+
+  attach_function(:statvfs, %i(string pointer), :int)
+  attach_function(:strerror, [:int], :string)
+
+  FSBLKCNT_T = if RbConfig::CONFIG['host_os'] =~ /darwin|osx|mach/i
+                 :uint
+               else
+                 :ulong
+               end
+
+  # See http://man7.org/linux/man-pages/man2/statvfs.2.html
+  class Statvfs < FFI::Struct
+    spec = [
+            :f_bsize,   :ulong,      # Filesystem block size
+            :f_frsize,  :ulong,      # Fragement size
+            :f_blocks,  FSBLKCNT_T,  # Size of fs in f_frsize units
+            :f_bfree,   FSBLKCNT_T,  # Number of free blocks
+            :f_bavail,  FSBLKCNT_T,  # Number of free blocks for unpriviledged users
+            :f_files,   FSBLKCNT_T,  # Number of inodes
+            :f_ffree,   FSBLKCNT_T,  # Number of free inodes
+            :f_favail,  FSBLKCNT_T,  # Number of free inodes for unprivilged users
+            :f_fsid,    :ulong,      # Filesystem ID
+            :f_flag,    :ulong,      # Mount Flags
+            :f_namemax, :ulong       # Max filename length
+           ]
+
+    # Linux has this at the end of the struct and if we don't include
+    # it we end up getting a memory corruption error when th object
+    # gets GCd.
+    if RbConfig::CONFIG['host_os'] =~ /linux/i
+      spec << :f_spare
+      spec << [:int, 6]
+    end
+
+    layout(*spec)
+  end
+
+  def initialize(path)
+    @statvfs = stat(path)
+  end
+
+  #
+  # @returns [Integer] Free inodes on the given filesystem
+  #
+  def free_inodes
+    @statvfs[:f_favail]
+  end
+
+  #
+  # @returns [Integer] Free space in KB on the given filesystem
+  #
+  def free_space
+    # Since we are running as root we could report f_bfree but will
+    # stick with f_bavail since it will typically be more
+    # conservative.
+    (@statvfs[:f_frsize] * @statvfs[:f_bavail]) / 1024
+  end
+
+  private
+
+  def stat(path)
+    statvfs = Statvfs.new
+    if statvfs(path, statvfs.to_ptr) != 0
+      raise 'StatvfsFailedException', command.stderr
+    end
+
+    statvfs
+  end
+end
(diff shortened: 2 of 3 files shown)


jashaik:
Pipeline information
Adhoc - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2279
Umbrella - https://buildkite.com/chef/chef-umbrella-master-chef-server/builds/300

src/chef-server-ctl/plugins/du.rb
```diff
@@ -0,0 +1,25 @@
+require 'mixlib/shellout'
```
marcparadise, line 1: This file (du.rb) should live in lib/chef_server_ctl/du.rb, under `module ChefServerCtl, because it is not a plugin.

src/chef-server-ctl/plugins/reindex.rb
```diff
@@ -63,6 +63,27 @@ def do_reindex(orgs_to_reindex, options)
...
+      puts "Free space is sufficient to start Elasticsearch reindex"
+    else
+      puts "Insufficient free space on disk to complete reindex."
+      puts "The current elasticsearch data directory contains #{data_dir_size} KB of data but only #{free_disk_space} KB is available on disk."
+      puts "The reindex process requires at least #{2*data_dir_size} KB."
+      exit 1
+    end
+  else
+    puts "Elasticserach data path does not exist, so skipping the disk space preflight validations: #{data_dir}"
+  end
```
marcparadise, line 84: This should end in a `rescue` that shows only the error message/info in case of failure.  That way  if anything in `Du.du` fails it doesn't raise a stack-trace - those are noise and generally unhelpful to the operator. 

To that end, it would be better if `du` failures raised a known type (like a new exception DuFailedException) so that we could rescue it specifically here.

src/chef-server-ctl/plugins/statfs.rb
```diff
@@ -0,0 +1,79 @@
+require 'ffi'
+
+class Statfs
```
marcparadise, line 3: This should also be in lib/chef_server_ctl.

src/chef-server-ctl/plugins/statfs.rb
```diff
@@ -0,0 +1,79 @@
...
+    # conservative.
+    (@statvfs[:f_frsize] * @statvfs[:f_bavail]) / 1024
+  end
+
+  private
+
+  def stat(path)
+    statvfs = Statvfs.new
+    if statvfs(path, statvfs.to_ptr) != 0
+      raise 'statvfs: ' + strerror(FFI.errno)
```
marcparadise, line 74: Let's also raise a new named exception type here, so that we can catch it in the caller and handle it better.

marcparadise (changes requested):
Please see comments inline.

jashaik:
New pipeline information
Adhoc - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2291
umbrella - https://buildkite.com/chef/chef-umbrella-master-chef-server/builds/304

src/chef-server-ctl/plugins/reindex.rb
```diff
@@ -63,6 +65,31 @@ def do_reindex(orgs_to_reindex, options)
...
+  begin
+    if Dir.exist?(data_dir)
+      data_dir_size = Du.du(data_dir)
+      free_disk_space = Statfs.new("#{data_dir}/..").free_space
+      if (2.2* data_dir_size) < free_disk_space #The minimum space should be double the existing data size
+        puts "Free space is sufficient to start Elasticsearch reindex"
+      else
+        $stderr.puts "Insufficient free space on disk to complete reindex."
+        $stderr.puts "The current elasticsearch data directory contains #{data_dir_size} KB of data but only #{free_disk_space} KB is available on disk."
+        $stderr.puts "The reindex process requires at least #{2*data_dir_size} KB."
```
PrajaktaPurohit, line 82: Should this multiplication factor be 2.2 since that is what we are checking on?

src/chef-server-ctl/lib/chef_server_ctl/helpers/statfs.rb
```diff
@@ -0,0 +1,79 @@
...
+    # conservative.
+    (@statvfs[:f_frsize] * @statvfs[:f_bavail]) / 1024
+  end
+
+  private
+
+  def stat(path)
+    statvfs = Statvfs.new
+    if statvfs(path, statvfs.to_ptr) != 0
+      raise 'StatvfsFailedException'
```
PrajaktaPurohit, line 74: It might be useful to include the error message here `raise StatvfsFailedException, strerror(FFI.errno)` so that rescue  can `rescue StatvfsFailedException => e` and print e.message.

Similarly for the du error

src/chef-server-ctl/plugins/reindex.rb
```diff
@@ -63,6 +65,31 @@ def do_reindex(orgs_to_reindex, options)
...
+      else
+        $stderr.puts "Insufficient free space on disk to complete reindex."
+        $stderr.puts "The current elasticsearch data directory contains #{data_dir_size} KB of data but only #{free_disk_space} KB is available on disk."
+        $stderr.puts "The reindex process requires at least #{2.2*data_dir_size} KB."
+        exit 1
+      end
+    else
+      puts "Elasticserach data path does not exist, so skipping the disk space verification: #{data_dir}"
+    end
+  rescue => exception
```
jashaik, line 88: We are giving writing the exception here. Is this what you are trying to suggest me? @PrajaktaPurohit

jashaik:
New Pipeline information - 
Adhoc - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2374
Umbrella - https://buildkite.com/chef/chef-umbrella-master-chef-server/builds/332

src/chef-server-ctl/plugins/reindex.rb
PrajaktaPurohit: yeah - I was thinking it might be of help to catch the individual exception that we are throwing in du and statsvfs instead of rescuing everything.

src/chef-server-ctl/lib/chef_server_ctl/helpers/du.rb
```diff
@@ -0,0 +1,22 @@
...
+  #
+  def self.du(path)
+    command = Mixlib::ShellOut.new("du -sk #{path}")
+    command.run_command
+    if command.status.success?
+      command.stdout.split("\t").first.to_i
+    else
+      puts "du -sk #{path} failed with exit status: #{command.exitstatus}"
+      puts "du stderr: #{command.stderr}"
+      raise 'DuFailedException', command.stderr(FFI.errno)
```
marcparadise, line 17: Does command.stderr accept an argument?

src/chef-server-ctl/plugins/reindex.rb
marcparadise: While I don't normally like catch-all rescues, I think it makes sense in this context. 

Almost any failure out of these checks would not be actionable by the user, so we'd end up just showing them an error that meant little - and then we'd leave them stuck there with it and no path forward.  (Types of errors I can think of: du command not found, broken permissions, disk failures preventing statfs reads). 

If the error was due to something like hard drive failure or permissions, we'll still fail later - and will still have the logged exception from here to give a hint if one is needed.

TL;DR - I think this should do what it says on the tin (or function name) and only fail if there is not enough disk space. Allowing exceptions to escape would mean that the user ends up with a stack trace and confusion.

src/chef-server-ctl/plugins/reindex.rb
```diff
@@ -63,6 +65,31 @@ def do_reindex(orgs_to_reindex, options)
...
+      if (2.2* data_dir_size) < free_disk_space #The minimum space should be double the existing data size
+        puts "Free space is sufficient to start Elasticsearch reindex"
+      else
+        $stderr.puts "Insufficient free space on disk to complete reindex."
+        $stderr.puts "The current elasticsearch data directory contains #{data_dir_size} KB of data but only #{free_disk_space} KB is available on disk."
+        $stderr.puts "The reindex process requires at least #{2.2*data_dir_size} KB."
+        exit 1
+      end
+    else
+      puts "Elasticserach data path does not exist, so skipping the disk space verification: #{data_dir}"
```
marcparadise, line 86: Typo in elasticsearch

marcparadise (changes requested):

src/chef-server-ctl/lib/chef_server_ctl/helpers/du.rb
jashaik: Looking like it is not accepting any argument, Do we have any docs for referring command module functions arity.I have verified this by running some smaple tests.

PrajaktaPurohit (approved):

marcparadise (approved):

PR #2346, opened 2021-03-22
Require TLSv1.2 on Chef Infra Clients by default

### LINKED ISSUE #2302: Update TLS settings to default to just TLS 1.2
To improve the out of the box security of Chef Infra Server we should default to just TLS 1.2. TLS 1.2 is supported by chef-client 10.16.4 and later so this is a pretty reasonable change to make without breaking compatibility with too many clients.

vinay-satish:

### Description

To improve the security of chef-server the TLS is defaulted to only `TLSv1.2` instead of `TLSv1 TLSv1.1 TLSv1.2`

### Issues Resolved

resolved #2302 

### Check List

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -476,7 +476,7 @@ default['private_chef']['nginx']['stub_status']['allow_list'] = ['127.0.0.1']
 # - AES added for pushy-server 1.x compatibility
 # - TLS1 and TLSv1.1 added for pushy-server 1.x compatibility
 #
-default['private_chef']['nginx']['ssl_protocols'] = 'TLSv1 TLSv1.1 TLSv1.2'
+default['private_chef']['nginx']['ssl_protocols'] = 'TLSv1.2'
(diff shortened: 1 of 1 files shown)


vinay-satish:
Adhoc build - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2300#_

tas50 (approved):
@PrajaktaPurohit Is there anything else we should be doing here other than just updating the default. We'll mention it in the release notes, but is there further documentation necessary.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -476,7 +476,7 @@
 # - AES added for pushy-server 1.x compatibility
 # - TLS1 and TLSv1.1 added for pushy-server 1.x compatibility
 #
-default['private_chef']['nginx']['ssl_protocols'] = 'TLSv1 TLSv1.1 TLSv1.2'
+default['private_chef']['nginx']['ssl_protocols'] = 'TLSv1.2'
```
PrajaktaPurohit, line 479: We should document the versions of chef-client that we will be incompatible with after this update.

tas50: I have the versions ready for the release notes and I'll make sure we get that for the docs. The docs right now actually make it sounds like we support SSL 3.0 by default. They need some work.

PrajaktaPurohit (approved):

PR #2400, opened 2021-04-07
Allow preventing user's from changing their e-mail with new `allow_email_update_only_from_manage` config

jashaik:
### Description

- Due to being an API we probably will limit permissions on the API to systems that perform email verification themselves.
- Chef-server can distinguish between the requests originating from the webui and client.
The requests from the webui will have header x-ops-request-source set.
If this is not set it implies that the request is originating from the the client / knife. If such a request has altered the email address in the json field. Send error to the effect: Use chef-manage to change user email
- Email for user creation should be case insensitive.For example 'test@chef.io' and 'TEST@chef.io' are considered to be the same email addresses in user creation and should 409 conflict.

### Issues Resolved

There are two approches for fixing the point 3-issue.
- Sql changes for UNIQUE email address. Cons:  Migration must be done for installation
- Code change in oc_erchef befor storing it to DB. Cons: It does not clean current discrepancies.

https://app.zenhub.com/workspaces/chef-infra-server-team-5fc64867d45ca500173dbbc7/issues/chef/customer-bugs/338
### Check List

### DIFF
--- a/oc-chef-pedant/spec/api/user_spec.rb
+++ b/oc-chef-pedant/spec/api/user_spec.rb
@@ -170,7 +170,7 @@ describe "users", :users do
               })
           end
 
-          context "when there's another user with that mail (different case)" do
+          context "when there's another user with that mail (different case)", :email_case_insensitive do
             let(:username_two) { "user_for_email_tests_two" }
             let(:email_two) { "useR@aOl.com" }
             let(:user_options_two) do
@@ -182,25 +182,42 @@ describe "users", :users do
                 }
               }
             end
+            let(:user_two) do
+              {
+                "username" => username_two,
+                "email" => email_two,
+                "first_name" => username,
+                "last_name" => username,
+                "display_name" => username,
+                "password" => "foobar"
+              }
+            end
 
             let(:filtered_users_body) do
               { username => "#{request_url}/#{username}",
-                username_two => "#{request_url}/#{username_two}",
               }
             end
 
             before :each do
-              @user2 = platform.create_user(username_two, user_options_two)
+              begin
+                platform.create_user(username_two, user_options_two)
+              rescue StandardError => e  
+                puts e
+                puts "Username or email address already in use."
+              end
             end
 
-            after :each do
-              platform.delete_user(@user2)
-            end
-
-            it "finds both users by this email" do
+            it "finds only one users by this email" do
               get("#{request_url}?email=#{email}", platform.superuser).should look_like({
                   :status => 200,
                   :body_exact => filtered_users_body
+              })
+            end
+
+            it "conflict when trying to create a user with duplicate case-insensitive email address." do
+              post(request_url, platform.superuser, :payload => user_two).should look_like({
+                  :status => 409,
+                  :body_exact => {"error"=>["Username or email address already in use."]}
               })
             end
           end
--- a/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
@@ -238,6 +238,7 @@ default['private_chef']['opscode-erchef']['sql_ro_user'] = 'opscode_chef_ro'
 # See default['private_chef']['postgresql']['db_connection_superuser'] for information
 default['private_chef']['opscode-erchef']['sql_connection_user'] = nil
 default['private_chef']['opscode-erchef']['enable_request_logging'] = true
+default['private_chef']['opscode-erchef']['allow_email_update_only_from_manage'] = false
 
 #
 # Reindex configurables
--- a/src/oc_erchef/apps/chef_objects/src/chef_user.erl
+++ b/src/oc_erchef/apps/chef_objects/src/chef_user.erl
@@ -316,7 +316,16 @@ common_user_validation(EJ, User, Operation) ->
     end,
     % Ensure the user can't set this - we make use of it in some cases by setting it internally
     EJ1 = ej:delete({<<"id">>}, EJ),
-    {ok, EJ1}.
+    % Setting the email to lower case for uniqueness,
+    % test@chef.io and TEST@chef.io are considered to be the same email addresses in user creation
+    EJ2 =
+    case ej:get({<<"email">>}, EJ1) of
+        undefined ->
+            EJ1;
+        Email ->
+            ej:set({<<"email">>}, EJ1, string:lowercase(Email))
+    end,
+    {ok, EJ2}.
 
 external_auth_uid(EJson, #chef_user{external_authentication_uid = ExtAuthUid}) ->
     case ej:get({<<"external_authentication_uid">>}, EJson) of
--- a/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
+++ b/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
@@ -49,6 +49,7 @@
          malformed_request_message/3,
          request_type/0,
          validate_request/3,
+         verify_update_request/5,
          finalize_update_body/3 ]).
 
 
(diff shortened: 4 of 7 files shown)


src/oc_erchef/schema/deploy/users_email_functional_index.sql
```diff
@@ -1,4 +1,4 @@
 -- Deploy enterprise_chef:users_email_functional_index to pg
 -- requires: users
 
-CREATE INDEX CONCURRENTLY users_lower_email_idx ON users (lower(email));
+CREATE UNIQUE INDEX CONCURRENTLY users_lower_email_idx ON users (lower(email));
```
PrajaktaPurohit, line 4: This will need user interaction to make sure there are no current duplicates.

jashaik: That's true

jashaik:
Pipeline Information-
Adhoc - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2422#145d962b-ebfd-4b4a-b606-b841fb7d2cd7

src/oc_erchef/apps/chef_objects/src/chef_user.erl
```diff
@@ -316,7 +316,8 @@ common_user_validation(EJ, User, Operation) ->
     end,
     % Ensure the user can't set this - we make use of it in some cases by setting it internally
     EJ1 = ej:delete({<<"id">>}, EJ),
-    {ok, EJ1}.
+    EJ2 = ej:set({<<"email">>}, EJ1, string:lowercase(ej:get({<<"email">>}, EJ1))),
```
PrajaktaPurohit, line 319: It would be helpful to add some context for this change here.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
```diff
@@ -270,5 +295,8 @@ conflict_message(Name) ->
     Msg = iolist_to_binary([<<"User '">>, Name, <<"' already exists">>]),
     {[{<<"error">>, [Msg]}]}.
 
+email_update_error_message() ->
+    {[{<<"error">>, <<"Use chef-manage to change user email">>}]}.
```
PrajaktaPurohit, line 299: s/chef-manage/Chef Manage

tas50:
For now we're going to have to default `disallow_non_manage_email_update` to be false. Many customers do not use manage and would be unable to update e-mails if we removed this ability and sent them over to manage which is listed as deprecated with an upcoming EOL date. I think we want to split this ticket into 2 different issues:

- Solving the e-mail change issue via manage
- Solving the e-mail change issue via knife

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
```diff
@@ -178,12 +178,41 @@ from_json(Req, #base_state{resource_args = invitation_response,
...
+            ej:set({<<"email">>}, UserData0, string:lowercase(Email))
+    end,
+    OrigEmail =
+    case OrigEmail0 of
+        undefined -> OrigEmail0;
+        _ -> string:lowercase(OrigEmail0)
+    end,
+    NewEmail = ej:get({<<"email">>}, UserData),
+    EmailUpdateConfig = envy:get(oc_chef_wm, disallow_non_manage_email_update, boolean),
+    RequestOrigin = wrq:get_req_header("x-ops-request-source", Req),
```
PrajaktaPurohit, line 197: %% check if the request originated from the webui or client (client /knife)

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
```diff
@@ -178,12 +178,41 @@ from_json(Req, #base_state{resource_args = invitation_response,
...
+    UserData =
+    case ej:get({<<"email">>}, UserData0) of
+        undefined ->
+            UserData0;
+        Email ->
+            ej:set({<<"email">>}, UserData0, string:lowercase(Email))
+    end,
+    OrigEmail =
+    case OrigEmail0 of
+        undefined -> OrigEmail0;
```
PrajaktaPurohit, line 192: Is the undefined here handling bad data?

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
```diff
@@ -238,6 +238,7 @@
 # See default['private_chef']['postgresql']['db_connection_superuser'] for information
 default['private_chef']['opscode-erchef']['sql_connection_user'] = nil
 default['private_chef']['opscode-erchef']['enable_request_logging'] = true
+default['private_chef']['opscode-erchef']['disallow_non_manage_email_update'] = false
```
PrajaktaPurohit, line 241: NIT: something to the effect allow_email_update_only_from_manage may read better?

PrajaktaPurohit (approved):
Nice!!

PrajaktaPurohit:
Can be merged one the chef-zero test is green (It might be a matter of merging https://github.com/chef/chef-zero/pull/311)

jashaik:
New Pipeline information -
Adhoc - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2696
Umbrella - https://buildkite.com/chef/chef-umbrella-master-chef-server/builds/462

oc-chef-pedant/spec/api/user_spec.rb
```diff
@@ -185,19 +185,19 @@
...
               { username => "#{request_url}/#{username}",
-                username_two => "#{request_url}/#{username_two}",
               }
             end
 
             before :each do
-              @user2 = platform.create_user(username_two, user_options_two)
+              begin
+                platform.create_user(username_two, user_options_two)
+              rescue StandardError => e
```
marcparadise, line 194: With this rescue in place, we aren't ever verifying what actually happens on duplicate case-insensitive email address. 

Given that this is a change in functionality, we should also add a specific example that verifies the exception we expect comes from creating a user with a duplicate email address; doing that would make it safe to do rescue this here. 

 In 'when there's another user' context, something like
```
it "raises  $exception-type when trying to create that user" do 

... 

end

File: omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb [hunk 8]
```diff
@@ -238,6 +238,7 @@
 # See default['private_chef']['postgresql']['db_connection_superuser'] for information
 default['private_chef']['opscode-erchef']['sql_connection_user'] = nil
 default['private_chef']['opscode-erchef']['enable_request_logging'] = true
+default['private_chef']['opscode-erchef']['allow_email_update_only_from_manage'] = false
```
marcparadise, line 241: The behavior of this attribute should also be tested in the pedant specs if possible.

src/oc_erchef/apps/chef_objects/src/chef_user.erl
```diff
@@ -316,7 +316,16 @@ common_user_validation(EJ, User, Operation) ->
     end,
     % Ensure the user can't set this - we make use of it in some cases by setting it internally
     EJ1 = ej:delete({<<"id">>}, EJ),
-    {ok, EJ1}.
+    % Setting the email to lower case for uniqueness,
```
marcparadise, line 319: Instead of doing this explicitly, could we let the database manage it by updating the index to be case-insensitive?

There's a risk there because it could cause a migration to fail if multiple case versions of the same email are already in the users table - but given that /is/ an error condition that we're not allowing, it's probably a better route.  We could pre-emptively identify that in upgrade preflight checks, and give the user instructions to mitigate it before proceeding.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
```diff
@@ -178,12 +178,59 @@ from_json(Req, #base_state{resource_args = invitation_response,
...
+    case {NewEmail, RequestOrigin, EmailUpdateConfig, ApiVersion} of
+        {OEmail, _, _, ?API_v0} when OEmail =:= OrigEmail; OEmail =:= undefined ->
+            oc_chef_wm_key_base:update_object_embedded_key_data_v0(Req, State, User, UserData);
+        {OEmail, _, _, _} when OEmail =:= OrigEmail; OEmail =:= undefined ->
+            oc_chef_wm_base:update_from_json(Req, State, User, UserData);
+        {_, "web", _, ?API_v0} ->
+            oc_chef_wm_key_base:update_object_embedded_key_data_v0(Req, State, User, UserData);
+        {_, "web", _, _} ->
+            oc_chef_wm_base:update_from_json(Req, State, User, UserData);
+        {_, Origin, false, ?API_v0} when Origin =/= "web" ->
```
marcparadise, line 226: We can remove the guard here - Origin can't be `web` because that would be handled in the case above  `{_, "web", _, ?API_v0}` or the next `{_, "web", _, _}.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
```diff
@@ -178,12 +178,59 @@ from_json(Req, #base_state{resource_args = invitation_response,
...
+            oc_chef_wm_key_base:update_object_embedded_key_data_v0(Req, State, User, UserData);
+        {_, Origin, false, _} when Origin =/= "web" ->
```
marcparadise, line 228: This one too - this case can't run with Origin == "web".  it  will not get executed if origin is "web", because it is caught by `{_, "web", _, _}`.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
```diff
@@ -178,12 +178,59 @@ from_json(Req, #base_state{resource_args = invitation_response,
...
+            oc_chef_wm_base:update_from_json(Req, State, User, UserData);
+        _ ->
+            {{halt, 403}, chef_wm_util:set_json_body(Req, email_update_error_message()),
```
marcparadise, line 231: 403 is debatable here - I think it technically fits because it comes down to the server refusing to the action, but in all other cases we use it to indicate that permission to do the requested thing is denied due to authz.  TO keep that consistent across the API, 412 might be better - "Precondition failed" based on the request headers is pretty accurate, even if th eoperator isn't explicitly providing the headers to us.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
```diff
@@ -178,12 +178,59 @@ from_json(Req, #base_state{resource_args = invitation_response,
...
+    % RequestOrigin - Request originated from webui or clinet
+    % EmailUpdateConfig - allow_email_update_only_from_manage config value
+    % ApiVersion - API version
+
+    % Different case clauses
+    % case 1 &  2- Orginal Email is same or undefined then Update
+    % case 3 & 4 - Request from Chef Manage/ Web then then Update
+    % case 5 & 6 - Request is not from Chef Manage/Web & allow_email_update_only_from_manage false then Update
+    % Other cases - Reject with 403
+    case {NewEmail, RequestOrigin, EmailUpdateConfig, ApiVersion} of
```
marcparadise, line 217: This is a complex case; how is coverage looking?

marcparadise (changes requested):
Comments are inline.

src/oc_erchef/apps/chef_objects/src/chef_user.erl
PrajaktaPurohit: Yeah - this was one of the approaches we considered, but since this is required more internally than for our customers, we decided to not force data migration / validation on them.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
PrajaktaPurohit: sure - I see that adding the guard makes it more readable, but might not be necessary for the code.

src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_named_user.erl
PrajaktaPurohit: https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/oc_chef_wm/test/oc_chef_wm_base_tests.erl might be a good place to add some unit tests.

src/oc_erchef/apps/chef_objects/src/chef_user.erl
marcparadise: When we first made this change (unique email requirement, a few years ago)  not having a case-insensitive index in the first place was an oversight -- the intent was to not have any duplicate emails in the system. Enforcing data constraints at the application level is more likely to let edge cases through, which overall combines to make the DB a better place for this change. 

With all that said - practically speaking email isn't used in non-hosted solutions beyond org invites, password resets, and email verification; unique emails are not critical to the operation of a self-hosted CIS and we do want to limit the upgrade impact to customers.

omnibus/files/private-chef-cookbooks/private-chef/attributes/default.rb
jashaik: I tried looking for possibilities but no luck, There is no way to change the existing components configurations via pedant tests.

oc-chef-pedant/spec/api/user_spec.rb
jashaik: Added test for convering the above suggested case

marcparadise (approved):

PrajaktaPurohit (approved):

src/oc_erchef/apps/chef_objects/src/chef_user.erl
```diff
@@ -316,7 +316,16 @@ common_user_validation(EJ, User, Operation) ->
...
+    % test@chef.io and TEST@chef.io are considered to be the same email addresses in user creation
+    EJ2 =
+    case ej:get({<<"email">>}, EJ1) of
+        undefined ->
+            EJ1;
+        Email ->
+            ej:set({<<"email">>}, EJ1, string:lowercase(Email))
```
jashaik, line 326: @lbakerchef

PR #2407, opened 2021-04-09
Add Content-Security-Policy and X-Frame-Options headers to our static content

tas50:
This came up in a security audit. The user browseable page that just points user to use knife or access our API does not set Content-Security-Policy and doesn't set `X-Frame-Options DENY` if manage is installed. This static content should be pretty locked down. I've tested this on an installation and it doesn't impact the page loading, but improved our scan scores.

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
@@ -52,13 +52,21 @@
     # Whitelist the docs necessary to serve up error pages and friendly
     # html to non-chef clients hitting this host.
     location ~ "^/[0-9]{3,3}\.(json|html)|favicon.ico|index.html$" {
+      add_header Content-Security-Policy "default-src 'self';";
+      add_header X-Frame-Options DENY;
     }
     location "/css/" {
+      add_header Content-Security-Policy "default-src 'self';";
+      add_header X-Frame-Options DENY;
     }
     location "/images/" {
+      add_header Content-Security-Policy "default-src 'self';";
+      add_header X-Frame-Options DENY;
     }
 
     location /version {
+      add_header Content-Security-Policy "default-src 'self';";
+      add_header X-Frame-Options DENY;
       types { }
       default_type text/plain;
       alias /opt/opscode/version-manifest.txt;
--- a/src/nginx/habitat/config/chef_http_lb_common
+++ b/src/nginx/habitat/config/chef_http_lb_common
@@ -48,7 +48,7 @@
     error_page 404 =404 /404.html;
     error_page 503 =503 /503.json;
 
-    # Whitelist the docs necessary to serve up error pages and friendly
+    # Allow the docs necessary to serve up error pages and friendly
     # html to non-chef clients hitting this host.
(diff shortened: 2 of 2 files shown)


omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -52,13 +52,21 @@
     # Whitelist the docs necessary to serve up error pages and friendly
```
PrajaktaPurohit, line 52: Let's change whitelist to "Allow" while we are here.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -52,13 +52,21 @@
...
     # html to non-chef clients hitting this host.
     location ~ "^/[0-9]{3,3}\.(json|html)|favicon.ico|index.html$" {
+      add_header Content-Security-Policy "default-src 'self';";
+      add_header X-Frame-Options DENY;
```
PrajaktaPurohit, line 56: https://github.com/chef/chef-server/blob/298eb078aedc2a5e1a7da9408db94116ed27f9d1/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb#L40-L41 seems to add this option only when manage is not installed.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -52,13 +52,21 @@
     # Whitelist the docs necessary to serve up error pages and friendly
     # html to non-chef clients hitting this host.
     location ~ "^/[0-9]{3,3}\.(json|html)|favicon.ico|index.html$" {
+      add_header Content-Security-Policy "default-src 'self';";
```
PrajaktaPurohit, line 55: only adding self here would mean if the request was from different sub-domain these options would not get added. Perhaps testing with manage would clarify it.

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
tas50: Wouldn't this page not display is manage was installed?

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
tas50: Fixed

tas50:
New build to test manage with: https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2481#3b0df55f-e653-4d32-80d7-1acbd5475c39

omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb
tas50: Tested with Manage. Works great still

PR #2408, opened 2021-04-09
Remove support for Ubuntu 16.04 + add Ubuntu FIPS

tas50:
We can enable FIPS on Ubuntu easily so we might as well since we're
already constrained in OpenSSL with FIPS on RHEL.

Ubuntu 16.04 is EOL at the end of April so let's remove this.

### DIFF
--- a/.expeditor/release.omnibus.yml
+++ b/.expeditor/release.omnibus.yml
@@ -5,6 +5,7 @@ install-dir: /opt/opscode
 test-path: omnibus/omnibus-test.sh
 fips-platforms:
   - el-*-x86_64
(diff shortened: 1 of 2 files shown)


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2426#_

jashaik (approved):

jashaik:
We should update our Dev VM's Vagrant file for using ubuntu 18 instead of 16 before merging this change as we are still using ubuntu 16 for development. @PrajaktaPurohit @lbakerchef @vinay-satish @tas50

PR #2428, opened 2021-04-17
Remove bundler/rubygems omnibus defs

tas50:
Ruby includes built-in rubygems and bundler now so we're just installing
these over again. It makes the packages larger and makes it so we have
to keep the bundler release in the overrides file in sync with the
version that ruby ships in the ruby package.


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2458#fed56cf2-af8c-431e-a6fc-a0fa44e80943

tas50:
Confirmed that with this change the package doesn't actually change.

PR #2436, opened 2021-04-20
Use the current version of berkshelf

tas50:
We use berkshelf in chef-server to vendor the cookbooks necessary to configure chef with `chef-server-ctl reconfigure`. It's actually a build time only dependency, but we include it with server due to the way we're building things. Previously we pinned to a very old version to avoid the depsolver dependency which was very large. The berkshelf doesn't include the depsolver dependency and using this new release avoids a ton of really old deps we were installing into the gem dir.

The next step here is to install berkshelf into the omnibus builders since this is a buildtime only dep, but this gets us closer to that and unblocks getting to chef 16


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2479#d72afa88-ab71-4898-a085-d8fa30576412

PrajaktaPurohit (approved):

PR #2440, opened 2021-04-20
Remove duplicate knife-opc omnibus definition

tas50:
This is already a direct dep in chef-server-ctl


tas50:
This should prevent redcarpet and yard from getting installed into the gem dirs and shave 30 seconds off the build time.

tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2482

PR #2458, opened 2021-05-04
Move berkshelf to a build time dep not a runtime dep

tas50:
We use berks to install the cookbooks. We don't actually need it to run
the chef-servert-ctl command. Just make sure it's in our omnibus setup
so we can use it during the building of the package


lbakerchef (approved):
Looks good.  Just adding a note to remind us to run this through the pipelines once they are back online.

tas50:
Closing this. I made it a chef-server-ctl dep and that worked which is good enough for now.

PR #2464, opened 2021-05-05
Start a document about build failure patterns

btm:
I wrote a section on license_scout failures. This may be written up
somewhere else but it was one of the first errors I found.

The list of approved licenses is from the old '1511 Chef License Policy'
Google document that we've been falling back on for years.


lbakerchef (approved):

PrajaktaPurohit (approved):
Thanks @btm!

PrajaktaPurohit (approved):
Thanks @btm!

PR #2470, opened 2021-05-06
Update all omnibus deps to the latest + pin a few omnibus deps

tas50:
Get the version of omnibus that pins ohai + a bunch of omnibus-software updates

Pin a few updates to omnibus-software that broke things. I have issues created in that repo for those.


tas50:
Adhoc is all green on this btw

PR #2492, opened 2021-05-19
Remove the need for rest-client and mixlib-install in omnibus

tas50:
Put the chef-server-ctl deps into the gemspec for that project instead of installing over the gemfile.lock and hoping it all works out.


PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2648

tas50:
The one thing to call out with this change is we no longer install mixlib-shell the CLI to the embedded/bin dir. That should not be an issue since you'd have to hit that CLI via the path before, but it's a change we should at least noodle on. The plus side is with this change we no longer end up with the artifactory gem installed into the product and a 3rd (and chef 17 versioned) chef-utils gem.

PrajaktaPurohit (approved):

PR #2503, opened 2021-05-24
Add an omnibus definition to cleanup leftover ruby files in the install

tas50:
This is 99% the same as what we're already doing in the ruby-cleanup
definition, but the paths are specific to where we install for oc-id

### DIFF
--- a/omnibus/config/software/more-ruby-cleanup-server.rb
+++ b/omnibus/config/software/more-ruby-cleanup-server.rb
@@ -1,0 +1,35 @@
+#
+# Copyright:: Copyright (c) Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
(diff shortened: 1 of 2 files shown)


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2676#2e6fbf29-8ab4-45e3-865a-43241b6a1000

PrajaktaPurohit (approved):

tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2755#2d6d15eb-efb7-439b-b792-55d728580834

tas50:
I scaled this back to a smaller cleanup, which is running fine in adhocs

PR #2504, opened 2021-05-24
Add retries to the elasticsearch/redis service starts to resolve restore issues

jashaik:

### Description

after applying backup few services are not getting started(runsv)

### Issues Resolved

There is some action issue in ES recipe which are enable & start. It was intentionally added for elasticsearch_index.rb recipe.
Verifying the impact after removing start from elasticsearch.rb recipe. There are still some more errors while restarting redis_lb & nginx.

### Check List

### DIFF
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/elasticsearch.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/elasticsearch.rb
@@ -158,8 +158,13 @@ link '/opt/opscode/embedded/elasticsearch/config' do
   to elasticsearch_conf_dir
 end
 
+# While Restoring the chef server backup, enable is always successful 
+# but start is not which is the root cause of restore issue.
+# So added retries for making sure the elasticsearch runit service is running
 component_runit_service 'elasticsearch' do
   action [:enable, :start]
+  retries 10
+  retry_delay 1
 end
 
 include_recipe 'private-chef::elasticsearch_index'
--- a/omnibus/files/private-chef-cookbooks/private-chef/recipes/redis_lb.rb
+++ b/omnibus/files/private-chef-cookbooks/private-chef/recipes/redis_lb.rb
@@ -82,6 +82,8 @@ component_runit_service 'redis_lb'
 # Restart the redis_lb runit service.
 runit_service 'redis_lb' do
   action :restart
+  retries 10
+  retry_delay 1
   only_if { is_data_master? }
 end
 


jashaik:
Pipeline Information -
Adhoc - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2681

omnibus/files/private-chef-cookbooks/private-chef/recipes/elasticsearch.rb
```diff
@@ -159,7 +159,13 @@
 end
 
 component_runit_service 'elasticsearch' do
-  action [:enable, :start]
+  action [:enable]
+end
+
+component_runit_service 'elasticsearch' do
+  action [:start]
+  retries 10
```
PrajaktaPurohit, line 167: Might be useful to add context about why this needs to be split and the reason for adding the retries here in a comment.

marcparadise (commented):
This looks fine, but would like to better understand why we need to split enable/start - is this a workaround to a chef-infra-client issue, or is there a different reason?

jashaik:
> This looks fine, but would like to better understand why we need to split enable/start - is this a workaround to a chef-infra-client issue, or is there a different reason?

 While Retoring the chef server backup, enable is always successful but start is not which is the root cause of restore issue.So It is moved to separate resource for retrying start incase of failure

marcparadise (approved):

omnibus/files/private-chef-cookbooks/private-chef/recipes/elasticsearch.rb
```diff
@@ -159,7 +159,16 @@
...
 component_runit_service 'elasticsearch' do
-  action [:enable, :start]
+  action [:enable]
+end
+
+# While Restoring the chef server backup, enable is always successful 
+# but start is not which is the root cause of restore issue.
+# So It is moved to separate resource for retrying start incase of failure 
+component_runit_service 'elasticsearch' do
+  action [:start]
```
tas50, line 169: We could actually use the the start and enable actions with the retries and keep it a single resource, which avoids some potential backend confusion in the client since both resources have the same name

jashaik:
New pipeline information -
Adhoc - https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2695
Umbrella - https://buildkite.com/chef/chef-umbrella-master-chef-server/builds/461

omnibus/files/private-chef-cookbooks/private-chef/recipes/elasticsearch.rb
jashaik: yes @tas50, That's true.

tas50 (approved):

PR #2509, opened 2021-05-26
adding error message for reindexing

### LINKED ISSUE #2457: Print better error message in case of a reindex connection failure
Currently a failure in https://github.com/chef/chef-server/blob/master/src/oc_erchef/priv/reindex-opc-organization#L142 looks like :

```
[private-chef-upgrade] - Running Command: chef-server-ctl reindex -a -d -t with options {}
Reindexing all organizations
- Disabling the Chef API.
- Reindexing.
escript: exception error: no match of right hand side value
{error,{conn_failed,{error,econnrefused}}}
Failed to reindex data for abc!
```
It would be helpful to catch this error and let the user know what is causing this failure and things to test / reasons for failure / steps to move forward.

Acceptance criteria:
Failure to connect to the erlang node provides helpful steps to move forward.

Aha! Link: https://chef.aha.io/features/SH-3045

vinay-satish:

### Description
Handling the error in case of re-indexing and print proper message.

### Issues Resolved

Resolved #2457

### Check List

### DIFF
--- a/src/oc_erchef/priv/reindex-opc-organization
+++ b/src/oc_erchef/priv/reindex-opc-organization
@@ -41,9 +41,21 @@
 %%   2 - Invalid organization name given
 %%
 main(Args) ->
-    init_network(),
-    [Command, OrgInfo, Context] = validate_args(Args),
-    perform(Command, Context, OrgInfo).
+    try
+      init_network(),
+      [Command, OrgInfo, Context] = validate_args(Args),
+      perform(Command, Context, OrgInfo)
+    catch 
+        error:{badmatch,{error,{conn_failed,{error,econnrefused}}}} ->
+            io:format("~n Unable to start new erlang node due to network issue~n Please check if the localhost is resolved to 127.0.0.1 or check if firewall is blocking the connection~n");
+        error:{badmatch,{error,{already_started,Pid}}} ->
+            io:format("~n The reindexing task is running in other process with PID - ~p~n Please kill the other process with erlang command `exit(~p, normal).` and restart the Chef Infra server~n", [Pid, Pid]);
+        error:{badmatch,{error,{{shutdown,{failed_to_start_child,net_kernel,_Error}},_Detail}}} ->
+            io:format("~n The epmd process might not have been started ~n Try running the command `epmd -names` and restart the Chef Infra server~n");
+        Type:Message ->
+            io:format("~n Caught ~p with the message ~p~n Please try again after restarting the Chef Infra Server ~n", [Type, Message])
+    end.
+
 
 %% @doc Ensure that the arguments are all valid, meaning a recognized action is specified,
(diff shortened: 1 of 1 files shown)


src/oc_erchef/priv/reindex-opc-organization
```diff
@@ -41,9 +41,14 @@
...
 main(Args) ->
-    init_network(),
-    [Command, OrgInfo, Context] = validate_args(Args),
-    perform(Command, Context, OrgInfo).
+    try
+      init_network(),
+      [Command, OrgInfo, Context] = validate_args(Args),
+      perform(Command, Context, OrgInfo)
+    catch Type:Message ->
+      io:format("~n Caught ~p with the message ~p~n Please try again after restarting the chef server ~n", [Type, Message])
```
tas50, line 49:
```suggestion
      io:format("~n Caught ~p with the message ~p~n Please try again after restarting the Chef Infra Server ~n", [Type, Message])
```

PrajaktaPurohit: Could we catch on more specific errors and provide solution accordingly?
- for example:
   - match on the error when the erlang connection cannot happen - explain this happens during the connection process, before reindex takes place and proposed solution.
   - match on the error if unable to bring up an erlang vm because the port is blocked
   - and then keep this as the more generic case.

vinay-satish: Thanks for the Input! I could not reproduce the blocked port case. Except that, the rest of the changes are updated.

src/oc_erchef/priv/reindex-opc-organization
```diff
@@ -41,9 +41,19 @@
...
-    perform(Command, Context, OrgInfo).
+    try
+      init_network(),
+      [Command, OrgInfo, Context] = validate_args(Args),
+      perform(Command, Context, OrgInfo)
+    catch 
+        error:{badmatch,{error,{conn_failed,{error,econnrefused}}}} ->
+            io:format("~n Unable to start new erlang node due to network issue~n Please check if the localhost is resolved to 127.0.0.1 or check if firewall is blocking the connection~n");
+        error:{badmatch,{error,{already_started,_PID}}} ->
+            io:format("~n The reindexing task is running in other process~n Please kill the other process and restart the Chef Infra server~n");
```
PrajaktaPurohit, line 52: Could you add the command to find and kill the other process?

vinay-satish: Updated the same

PrajaktaPurohit (approved):

tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2779#7fbaa31c-a017-4531-b5da-a3123621da1c

PR #2557, opened 2021-06-22
Add a note to help fix reindex errors seen in fips env when nginx['enable_non_ssl'] = false

PrajaktaPurohit:
point customers to export CSC_LB_URL=https://127.0.0.1 on versions prior to 14.5 of Chef Infra server when fips is enabled and nginx['enable_non_ssl'] = false


kagarmoe:
The note needs to become a shortcode.

kagarmoe (approved):
Thanks for the clarification!

PR #2576, opened 2021-07-01
Upgrade Ruby to 2.7 and embedded chef-client to 16.13

tas50:
Get on a currently supported release of chef-client within chef-server-ctl


tas50:
This is now passing builds and ready for review @PrajaktaPurohit

marcparadise (approved):

PR #2633, opened 2021-07-15
oc-id: Source gems from rubygems when we can

tas50:
There's no reason to grab these from git anymore and it makes the bundle
install / update much slower.


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2925#ff8e330c-4e6e-459d-9ed6-d239acc3f0f5

Issue #2644, opened 2021-07-16
Expand gather-logs command to better match Automate data


tas50:
Both Infra Server and Automate include a gather-logs command. We need are missing a lot of data that's included in the Automate gather logs. We should do our best to gather the same data so that we can process the data in a single pipeline more easily in the future.

tas50:
Crossing out items as they're completed on master.

tas50:
Partially resolved in https://github.com/chef/chef-server/pull/2645

tas50:
Partially resolved in https://github.com/chef/chef-server/pull/2651

PrajaktaPurohit:
Add the above mentioned information for Opensearch

PR #2645, opened 2021-07-16
Expand data gather-logs collects + better align with automate

tas50:
Ideally we should gather the same-ish system data for automate and chef server so that one day we can process these gather logs and get some value from them

- stick proc files in /proc to match Automate
- Gather fips status to match Automate
- Gather /proc/version to match Automate
- Rename free-m.txt to free_m.txt to match our standard pattern and align with Automate
- Gather df -k to align with automate
- Gather the output of Ohai which has a bunch of troubleshooting data.

### DIFF
--- a/omnibus/files/private-chef-scripts/gather-logs
+++ b/omnibus/files/private-chef-scripts/gather-logs
@@ -29,7 +29,7 @@ then
 
     if [[ ! -e "/opt/$path/bin/$ctl_cmd" ]];
     then
-        echo "ERROR: Chef Server may not be installed."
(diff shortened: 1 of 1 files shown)


tas50:
Paging @teknofire who parses this data

This addresses some of the issues in: https://github.com/chef/chef-server/issues/2644

PrajaktaPurohit (approved):

tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2941

PR #2661, opened 2021-07-23
Switch veil and omniauth-chef to gems not git sources

tas50:
We have working pipelines for both of these projects now so we can use
git releases. This reduces the amount of spam from dependabot where
every minor bump of those projects gets pulled in here and it also
speeds up the installs since we don't have to clone both of these repos
during the build.


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2971#8d09600f-4df3-48b9-b323-87a7327bb33a

tas50:
Adhoc looks good

PR #2667, opened 2021-07-25
Add lsblk output to gather logs to check for non-SSD drives

tas50:
This came up in an escalation today. Large installs need to not be on
spinning rust and we can provide this output to support so they can add
it to their parsing tool.


tas50:
this is in ohai.txt already and it turns out it's not useful on AWS where we don't get the correct data.

PR #2671, opened 2021-07-26
gather-logs: Avoid errors in console on Ubuntu

tas50:
These 2 files don't exist so they fail


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2990#1daddd93-024b-4c78-8d74-3a2831852fbf

PR #2672, opened 2021-07-26
gather-logs: Remove support for specifying the server type / chef server 11

tas50:
We ship this here. There's no need to support stuff from 2013 in this script. Also remove support for gathering data on reporting since that won't install into 14.x. Not this does change the testing syntax to be the 1st argument instead of the 2nd.


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/2989

PrajaktaPurohit (approved):

PR #2673, opened 2021-07-26
Update docs.chef.io and GitHub links to avoid redirects or 404s

tas50:
A lot of things have moved and not all these links work anymore.


PrajaktaPurohit (approved):
I might leave the mover change as is. Other than that looks good. Thanks for taking care of this!

PrajaktaPurohit (approved):

PR #2692, opened 2021-08-03
More cleanup of the configuration cookbook

tas50:
- Remove redundant `resource_name` calls
- Remove the `private_chef_FOO` resource naming
- Give all properties allowed types
- Stop using `kind_of` on the property types
- use true/false not TrueClass/FalseClass


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/3039#39835271-d6be-4950-a81d-6db14d3ac540

PrajaktaPurohit (approved):
Thank you!

PR #2704, opened 2021-08-04
Test the system connectivity to the Internet in the gather logs

tas50:
This is something we need to know when troubleshooting systems.

### DIFF
--- a/omnibus/files/private-chef-scripts/gather-logs
+++ b/omnibus/files/private-chef-scripts/gather-logs
@@ -349,18 +349,18 @@ fi
 
 # check if the system is proxied by checking if HTTPS_PROXY and HTTP_PROXY are unset or empty
 if [[ -z "$HTTPS_PROXY" || -z "$HTTP_PROXY" ]]; then # proxy not set
-  echo "0" > $tmpdir/connectivity/proxied.txt
+  echo "0" > "${tmpdir}/connectivity/proxied.txt"
 else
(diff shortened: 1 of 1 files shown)


omnibus/files/private-chef-scripts/gather-logs
```diff
@@ -345,6 +347,22 @@ then
 
 fi
 
+# check if the system is proxied by checking if HTTPS_PROXY and HTTP_PROXY are unset or empty
+if [[ -z "$HTTPS_PROXY" || -z "$HTTP_PROXY" ]]; then # proxy not set
+  echo "0" > $tmpdir/connectivity/proxied.txt
```
outlawostinato, line 352: Shellcheck complains:

```
  echo "0" > $tmpdir/connectivity/proxied.txt
             ^-- SC2086: Double quote to prevent globbing and word splitting.
Did you mean: 
  echo "0" > "$tmpdir"/connectivity/proxied.txt
```

The suggested fix is to just quote the variable, personally I might prefer to quote the whole thing a la:

```
  echo "0" > "${tmpdir}/connectivity/proxied.txt"
```

https://github.com/koalaman/shellcheck/wiki/SC2086

Shell check complains about every unquoted incantation of `$tmpdir`.

tas50: Updated. Thanks

PrajaktaPurohit (approved):

PR #2709, opened 2021-08-05
Rename the private-chef cookbook to infra-server

tas50:
This makes it so private-chef doesn't show up all over the logs when we do a reconfigure or an upgrade operation in chef-server-ctl.

This also fixes a bug where I missed a resource rename in one of the subscribes.

Note: Will also need https://github.com/chef/umbrella/pull/177

Signed-off-by: Tim Smith <tsmith@chef.io>

PR #2710, opened 2021-08-06
Remove logic for configuring ES to support RHEL < 7 

tas50:
Remove EL 6 support logic from the ES config template

This build off #2709 so that needs to be merged first


lbakerchef:
https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/3539
https://buildkite.com/chef/chef-umbrella-main-chef-server/builds/936

PrajaktaPurohit (approved):

PR #2711, opened 2021-08-06
Remove cleanup logic for older releases of Chef Infra Server

tas50:
We required stepped upgrades so we know that everyone has first upgraded to 12.17. That means this logic all ran there and we can remove it here. This greatly reduces the number of things we have to do each time the reconfigure, upgrade, or cleanup commands run, and that speeds up runs / makes troubleshooting easier.

This requires #2709 to be merged first


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/3061#c20a0b42-9ec5-4df3-ac2f-7587f8458eda

PR #2717, opened 2021-08-06
Update gather-logs to use chef-server-ctl

tas50:
Remove some variables that aren't needed now that we only support Chef
Infra Server 12+ installations.


tas50:
https://buildkite.com/chef/chef-chef-server-master-omnibus-adhoc/builds/3101#9f007910-2c1e-47e6-9ce8-54b1878a0720

PR #2718, opened 2021-08-06
Fix a notification for firing partybus

tas50:
So this has been broken for a long time and I failed to fix it because
of the way it was broken.

If the cookbook is named 'private-chef' and the resource file is named
'pg_upgrade' then the resource becomes 'private_chef_pg_upgrade'. We
were notifying 'private-chef_pg_upgrade', which is not a resource. My
change renamed the resource to just 'pg_upgrade' and I failed to update
this notify because my find/replace missed this typo.


lamont-granquist (approved):

PrajaktaPurohit (approved):
Thanks @tas50 !

PR #2827, opened 2021-09-26
Fix PostgreSQL major version detection and use for binaries & data paths

### LINKED ISSUE #2811: Postgresql upgrade from 13.3 to 13.4 is failing due to 13.3 binaries missing
### Chef Server Version

14.9.23

### Platform Details

All Platforms

### Configuration

Standalone

### Scenario
PR #2807 
https://buildkite.com/chef/chef-umbrella-main-chef-server/builds/668#a34ab76d-ef88-405b-bac2-85a58548ec91
Chef-server postgresql upgrade from 13.3 to 13.4 is not smooth with all the upgrade scenario from 12.15.17 to 14.9.22

### Steps to Reproduce

https://buildkite.com/chef/chef-umbrella-main-chef-server/builds/668#a34ab76d-ef88-405b-bac2-85a58548ec91

### Expected Result

It should be smooth and chef-server should be upgraded without any error

### Actual Result

Postgresql upgrade is failing.

apavely:
### Description

This change corrects the PostgreSQL version detection logic to allow for proper validation and upgrades between major PostgreSQL releases. It introduces a new `PgVersion` class (built off of `Gem::Version`) to properly parse and extract major versions.

This does not correct for the need to use `pg_upgrade` for any past database versions to current (since this change properly sets the data directory path to use the major version), but will address any issues with `13.x` → `13.y` minor version upgrades going forward.

Signed-off-by: Aaron Pavely <aaron@pavely.net>

### Issues Resolved
#2811 

### Check List

- [x] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/main/CONTRIBUTING.md#developer-certification-of-origin-dco>

### DIFF
--- a/omnibus/config/software/postgresql13.rb
+++ b/omnibus/config/software/postgresql13.rb
@@ -36,7 +36,7 @@ relative_path "postgresql-#{version}"
 
 build do
   env = with_standard_compiler_flags(with_embedded_path)
-  short_version = version.gsub(/^([0-9]+).([0-9]+).[0-9]+$/, '\1.\2')
+  short_version = version.gsub(/^([0-9]+).[0-9]+$/, '\1')
 
   update_config_guess(target: "config")
 
--- a/omnibus/config/software/server-complete.rb
+++ b/omnibus/config/software/server-complete.rb
@@ -29,7 +29,7 @@ dependency "runit"
 dependency "openssl-fips-config" if fips_mode?
 
 # the backend
-dependency "postgresql96-bin" # for upgrading 9.6 -> 13.3
+dependency "postgresql96-bin" # for upgrading 9.6 -> 13
 dependency "postgresql13"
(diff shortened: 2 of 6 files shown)


apavely:
Updated to correct `pg_upgrade` paths for `--old-bindir=`.

Same major (i.e., minor-to-minor) PostgreSQL upgrades don't require older binaries, so pg_upgrade's `--old-bindir=` can be set to the same path as `--new-bindir=`.

One important consequence is that older minor PG releases (e.g., 13.3) do _not_ need to be built. Older binaries are only required for major-to-major version changes (i.e., 9.6.x → 13.x, or 13.x → 14.x ... when that time comes).

This should save both some build time and keep from bloating the resulting build.

lbakerchef:
See https://github.com/chef/chef-server/pull/2830

PR #2955, opened 2021-11-18
Enable cookbook version response caching

marcparadise:
Most of our time post-depsolve in the `cookbook_versions/` endpoint is spent assembling the results for the client to consume.    On a heavily loaded server since OTP 20, the same
call can take 60+ seconds to complete and significantly degrade performance of the entire system.   Prior to that, it could still take 3-6 seconds though without systemic performance degradation.

This change introduces an ets-based cache that:
- allows callers to cache the results 
- has overload protection, allowing the caller to force client retry (via 503 response) instead of waiting for the cache to become available
- has synchronization capabilities, so that one caller can determine that another process is already calculating results for a given key and retry the request instead of allowing multiple callers to do the same results calculation. 

Some numbers, using sample org data that tends towards generating large
responses to cookbook_version endpoint, because of the total number of
cookbook files:

7k clients/15min interval - cache disabled - CPU 1800-2000%, LA 25. cookbook_versions rough average time range 2-5s
7k clients/15min interval - cache enabled - CPU 150-200%, LA 5-6. cookbook_versions rough average time range 0.2-1.1s (edited)
40k clients/15min interval - cache enabled - CPU 1000-1200%, LA 19 . CBV time range: 0.2-3s with most on the < 0.5 end.
40k clients/15min interval - cache disabled - server fails within a minute

For more details, see the commit messages in this PR. 

### DIFF
--- a/docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
+++ b/docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
@@ -1456,6 +1456,23 @@ This configuration file has the following settings for `opscode-erchef`:
 
 :   The virtual IP address. Default value: `127.0.0.1`.
 
+`opscode_erchef['cbv_cache_enabled']`
+
+:   Enable cookbook versions caching by setting this to `true`. If you are frequently seeing
+    very long response times from `cookbook_versions` when under load, this is worth enabling.
+    Default value: `false`.
+
+`opscode_erchef['cbv_cache_item_ttl']`
+
+:   The minimum time in milliseconds that given cookbook versions list will be expired
+    when cbv_cache_enabled is enabled.
+    Default value: `30000`.
+{{< note >}}
+
+Be careful if increasing this number - requests for a given set of cookbook_versions will be stale if the resolved cookbook versions are updated before the cache entry times out.Not an issue if you are incrementing cookbook version with every change.
+
+{{< /note >}}
+
 ### Elasticsearch
 
 This configuration file has the following settings for `elasticsearch`:
--- a/docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
+++ b/docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
@@ -974,7 +974,7 @@ This configuration file has the following settings for `oc_chef_authz`:
 
 `oc_chef_authz['max_connection_request_limit']`
 
-:   The max number of requests allowed per connections.
+:   The maximum number of requests allowed per connection.
     Default value: `100`.
 
 ### oc-chef-pedant
@@ -1458,22 +1458,22 @@ This configuration file has the following settings for `opscode-erchef`:
 
 `opscode_erchef['cbv_cache_enabled']`
 
-:   Enable cookbook version response caching by setting this to `true`. If you are frequently seeing
+:   Enable cookbook version response caching by setting this to `true`. If you frequently see
     very long response times from `cookbook_versions` when under load, this is worth enabling.
-    Enabling this does open a window for a client to receive stale results. When a cookbook is updated
-    in place (without incrementing the version), and the old response has not been expired from cache,
-    the old response will be given to the client.  Subsequent client runs will receive the updated response.
-    Default value: `false`.
+    Enabling this makes it possible for a client to receive stale results. When a cookbook is updated
+    in place (without incrementing the version), and the old response has not expired from the cache,
+    the Infra Server will give the old response to the client. Subsequent client runs will receive the
+    updated response. Default value: `false`.
 
 `opscode_erchef['cbv_cache_item_ttl']`
 
-:   The minimum time in milliseconds that any given cookbook version response will be expired from the cache when
-    when cbv_cache_enabled is enabled.
+:   The minimum time in milliseconds that Chef Infra Server will keep any given cookbook version response in the cache when
+    when `cbv_cache_enabled` is enabled.
     Default value: `30000`.
 {{< note >}}
 
-Be careful if increasing this number - requests for a given set of cookbook_versions will be stale if the resolved cookbook versions are updated before the cache entry times out. This will
-not occur if cookbook version is incremented with every cookbook update, which is the recommended approach to updating cookbooks.
+Be careful if increasing this number - requests for a given set of cookbook versions will be stale if the resolved cookbook versions are updated before the cache entry times out. This will
+not occur if you increment the version of a cookbook with every cookbook update, which is the recommended approach to updating cookbooks.
 
 {{< /note >}}
 
--- a/docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
+++ b/docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
@@ -974,7 +974,7 @@ This configuration file has the following settings for `oc_chef_authz`:
 
 `oc_chef_authz['max_connection_request_limit']`
 
-:   The max number of requests allowed per connections. 
+:   The max number of requests allowed per connections.
     Default value: `100`.
 
 ### oc-chef-pedant
@@ -1458,18 +1458,22 @@ This configuration file has the following settings for `opscode-erchef`:
 
 `opscode_erchef['cbv_cache_enabled']`
 
-:   Enable cookbook versions caching by setting this to `true`. If you are frequently seeing
+:   Enable cookbook version response caching by setting this to `true`. If you are frequently seeing
     very long response times from `cookbook_versions` when under load, this is worth enabling.
+    Enabling this does open a window for a client to receive stale results. When a cookbook is updated
+    in place (without incrementing the version), and the old response has not been expired from cache,
+    the old response will be given to the client.  Subsequent client runs will receive the updated response.
     Default value: `false`.
 
 `opscode_erchef['cbv_cache_item_ttl']`
 
-:   The minimum time in milliseconds that given cookbook versions list will be expired
+:   The minimum time in milliseconds that any given cookbook version response will be expired from the cache when
     when cbv_cache_enabled is enabled.
     Default value: `30000`.
 {{< note >}}
 
-Be careful if increasing this number - requests for a given set of cookbook_versions will be stale if the resolved cookbook versions are updated before the cache entry times out.Not an issue if you are incrementing cookbook version with every change.
+Be careful if increasing this number - requests for a given set of cookbook_versions will be stale if the resolved cookbook versions are updated before the cache entry times out. This will
+not occur if cookbook version is incremented with every cookbook update, which is the recommended approach to updating cookbooks.
 
 {{< /note >}}
 
--- a/omnibus/files/server-ctl-cookbooks/infra-server/attributes/default.rb
+++ b/omnibus/files/server-ctl-cookbooks/infra-server/attributes/default.rb
@@ -353,6 +353,16 @@ default['private_chef']['opscode-erchef']['strict_search_result_acls'] = false
 default['private_chef']['opscode-erchef']['ssl_session_caching']['enabled'] = false
 default['private_chef']['opscode-erchef']['include_x_ops_api_info'] = false
 
+# Enable cookbook versions caching by setting this to true. If you are frequently seeing
+# very long response times from `cookbook_versions` when under load, this is worth enabling.
+default['private_chef']['opscode-erchef']['cbv_cache_enabled'] = false
+# When CBV caching is enabled, this indicates the minimum time in milliseconds that given
+# CBV list will be expired. Be careful if increasing this number - requests for a given
+# set of cookbook_versions will be stale if the resolved cookbook versions
+# are updated before the cache entry times out.
+# Not an issue if you are incrementing cookbook version with every change.
+default['private_chef']['opscode-erchef']['cbv_cache_item_ttl'] = 30000
+
 # The amount of milliseconds before we timeout and assume an endpoint is down for
 # the /_status endpoint.
 
--- a/src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
+++ b/src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
@@ -58,19 +58,38 @@
          terminate/2
         ]).
 
-
-
+-spec start_link() -> {ok, pid()} | ignore | {error, term()}.
 start_link() ->
     Enabled = envy:get(chef_objects, cbv_cache_enabled, false, boolean),
     TTL = envy:get(chef_objects, cbv_cache_item_ttl, ?DEFAULT_TTL, integer),
     gen_server:start_link({local, ?SERVER}, ?MODULE, [Enabled, TTL], []).
 
+%% @doc this function retrieves a stored value from the cache.
+%% If the value is missing, it will return undefined.
+%% If another caller has claimed intent to populate the key, this will return
+%% 'retry' indicating that the caller should try again after a brief delay.
+%% If it returns 'busy' the process is not available to service the request, and
+%% the caller should fail without retrying.
+-spec get(any()) -> retry | busy | undefined | term().
 get(Key) ->
     send_if_available({get, Key}).
 
+%% @doc claim intent to populate Key into the cache.
+%% When the claim is accepted, returns 'ok'.
+%% When another process has already claimed intent, this returns 'retry' indicating
+%% that the caller should retry the original 'get' request after a brief delay.
+%% If it returns 'busy' the process is not available to service the request, and
+%% the caller should fail without retrying.
+-spec claim(any()) -> ok | retry | busy.
 claim(Key) ->
     send_if_available({claim, Key}).
 
+%% @doc store a value in the cache.  Caller must have first claimed intent to populate
+%% this key by invoking claim/1.
+%% If the caller has not first invoked claim/1, this will return {error, noclaim}.
+%% If another process has invoked claim/1, this will return {error, already_claimed}.
+%% returns 'ok' when the value is stored successfully
+-spec put(any(), term()) -> {error, already_claimed} | {error, noclaim} | ok.
 put(Key, Value) ->
     % We do not protect this with the breaker - this function is only to be called
     % after a successful 'claim', and we want to allow caller to put the claim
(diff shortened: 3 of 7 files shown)


omnibus/files/server-ctl-cookbooks/infra-server/attributes/default.rb
```diff
@@ -353,6 +353,16 @@
 default['private_chef']['opscode-erchef']['ssl_session_caching']['enabled'] = false
 default['private_chef']['opscode-erchef']['include_x_ops_api_info'] = false
 
+# Enable cookbook versions caching by setting this to true. If you are frequently seeing
+# very long response times from `cookbook_versions` when under load, this is worth enabling.
+default['private_chef']['opscode-erchef']['cbv_cache_enabled'] = false
```
marcparadise, line 358: For now, do we want to enable by default or leave it to support to direct people to enable it when appropriate?

src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
```diff
@@ -0,0 +1,228 @@
...
+stop() ->
+    gen_server:call(?SERVER, stop).
+
+% gen_server implementation
+
+init([Enabled, TTL]) ->
+    process_flag(trap_exit, true),
+    spawn_breaker(),
+    {ok, #state{enabled = Enabled,
+                ttl = round(TTL/2), % Splay is TTL/2 + (0-100% of TTL) for max of TTL.
```
marcparadise, line 96: s/b 0-50% of TTL

src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
```diff
@@ -0,0 +1,228 @@
...
+    end.
+
+insert_into_cache(Tid, Key, Value, TTL) ->
+    case ets:insert_new(Tid, {Key, Value}) of
+        true ->
+            % - we see that often the pattern is that many requests for new keys
+            % come in clusters when traffic is first directed to the server.
+            % We can't avoid the initial cluster and corresponding CPU spike
+            % when we have nothing cached, but staggering the expirations after that
+            % will reduce CPU/VM utilization spikes that can impact the overall system.
```
marcparadise, line 157: considering removing this behavior - now that we have synchronization to prevent a flood of callers from doing the same work, it may no longer be necessary. (On the other hand, it probalby doesn't hurt...)

src/oc_erchef/apps/chef_objects/test/chef_cbv_cache_tests.erl
```diff
@@ -0,0 +1,235 @@
...
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+
+-module(chef_cbv_cache_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+
```
marcparadise, line 25: Oops - missing test s for responses when the service is disabled.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_depsolver.erl
```diff
@@ -278,22 +281,64 @@ wm_halt(Code, Req, State, ErrorData, LogMsg) ->
...
-                    [ { CBV#chef_cookbook_version.name,
-                       chef_cookbook_version:minimal_cookbook_ejson(CBV, chef_wm_util:base_uri(Req), ApiVersion) }
-                      || CBV <- CookbookVersions ]
-                    },
-            CBMapJson = chef_json:encode(JsonList),
-            {true, wrq:append_to_response_body(CBMapJson, Req), State};
+            case make_json_list(CookbookVersions, chef_wm_util:base_uri(Req), ApiVersion) of
+                {error, busy} ->
+                  % Force backoff until the cache catches up with demand. Occurs when caching is enabled
+                  % and the cache message queue is overloaded.
```
marcparadise, line 290: Also occurs when we give up on waiting for another process that has claimed a given cache key to complete its work.

omnibus/files/server-ctl-cookbooks/infra-server/attributes/default.rb
tas50: It might be best to burn this in a bit first and enable it in a major later

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_depsolver.erl
lbakerchef: Do we want to add this^^ to the comment?

lbakerchef:
```
                  % Force backoff until the cache catches up with demand. Occurs when caching is enabled
                  % and the cache message queue is overloaded. Also occurs when we give up on waiting for
                  % another process that has claimed a given cache key to complete its work.
```

lbakerchef (approved):
Looks good.  Just some general questions and thoughts:

1) Do you have some kind of conceptual diagram or diagrams showing the design?  I don't know if this even applies (thus, why I'd find such a thing useful :) ), but what I'm imagining is a a square or block of squares representing ETS cells, a circle representing a gen_server interacting with both callers and the ETS cell(s), callers calling the gen_server, and (if this even happens) callers interacting with the ETS cell(s) directly, bypassing the gen_server.

2) Is the gen_server necessary so that we aren't duplicating or repeating work all over the place? Is it more write coordination, read coordination, or both?  Or neither?

3) Do callers have to go through the gen_server to read the cache, or do they read it directly (I noticed you set `read_concurrency` to `true`)?  From my experience years ago optimizing erlang concurrency, I found that the less a gen_server is a bottleneck to anything, the better.  Having said that, whatever the case is here, this should hopefully be a tremendous improvement to the current state of affairs.

Good job.

marcparadise:
>    Do you have some kind of conceptual diagram or diagrams showing the design? I don't know if this even applies (thus, why I'd find such a thing useful :) ), but what I'm imagining is a a square or block of squares representing ETS cells, a circle representing a gen_server interacting with both callers and the ETS cell(s), callers calling the gen_server, and (if this even happens) callers interacting with the ETS cell(s) directly, bypassing the gen_server.

I have some things I sketched out by hand as I was thinking it through.  It does seem worthwhile to update that and make it into a shareable doc, will see what I can do there.  

  >  Is the gen_server necessary so that we aren't duplicating or repeating work all over the place? Is it more write coordination, read coordination, or both? Or neither?

Yes, we're using it to coordinate writes, and to allow callers to retry reads when a write is pending for the solution/key they're waiting for. Though that might be worth looking closer at. The flow could look something like: 

1) caller attempts direct lookup in ets, if it fails..
2) try to 'claim' the solution in chef_cbv_cache; if it succeeds it calculates the solution and stores it, as it does now. If it fails, it does the same retry behavior it has now. 

This would still prevent multiple callers from assembling the same results on a cache miss; and further reduces the possibility that the gen_server gets overloaded.  But (thinking this through as I type...) this approach does open a small gap where a caller (A) might miss on ETS , and another caller (B) completes its own cache put.  When (A) attempts to claim intent on the key, the gen_server will allow it because (B) already finished, causing (A) to recalculate the response that (B) just populated.  If that happens for multiple "A"s at the same time, things get messy.

>    Do callers have to go through the gen_server to read the cache, or do they read it directly (I noticed you set read_concurrency to true)? 

They do go through the gen_server for reads. I set read_concurrency true because this table is very read-heavy; and I was initially considering a pool of workers which would enable concurrent access.  The flag is left over from that, I'll remove it. 

> From my experience years ago optimizing erlang concurrency, I found that the less a gen_server is a bottleneck to anything, the better. Having said that, whatever the case is here, this should hopefully be a tremendous improvement to the current state of affairs.

That was my main concern as well - and it actually did happen in early testing under load, which is what led to adding the `circuit breaker` behavior.   It works well both in our load testing and for a customer seeing this issue.  But I've since  added the claim/intent behavior to prevent dup work.  Testing with that change has shown that  even under heaviest load the breaker was rarely 'tripped'. I think this is for a couple of reasons: 

a) multiple callers were no longer building the same solution, which reduces load by a surprising amount, and leaves the cbv_cache gen server more cycles to continue processing messages 
b) multiple callers were no longer sending the same /large/ `put` message of `{Key, ReallyBigSolutionTerm}` to the cbv_cache as they finished their work. 

The breaker still does get tripped under burst load when the server is already under load, but I've only seen it a handful of times for a very short period. More often I would see 503s because of callers giving up on being able to claim the key, but even those were rare: About 3-5% of cookbook_version endpoint calls would respond with 503, when the server was under max sustainable load of 40k run/15min [about 520 total API requests per second].

src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
```diff
@@ -0,0 +1,253 @@
...
+    case dict:take(Key, Claims) of
+        error ->
+            % Mis-usage. Claim must be called before put.
+            {reply, {error, no_claim}, State};
+        { From, Claims1 } ->
+            % It is claimed and by this caller, they can go ahead and put.
+            % Update claims dictionary in state to remove the claim on this key
+            insert_into_cache(Tid, Key, Value, TTL),
+            {reply, ok, State#state{claims = Claims1}};
+        { _OtherPid, _Claims1 } ->
```
jashaik, line 168: This caluse doesn't get invoked as 163 matches always.I read the comment but is it still required ?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_depsolver.erl
```diff
@@ -278,22 +281,65 @@ wm_halt(Code, Req, State, ErrorData, LogMsg) ->
...
-            %% if the authz check has succeeded.  We use a minimal version of the
-            %% cookbook which has just enough information for chef-client to run
-            JsonList = {
-                    [ { CBV#chef_cookbook_version.name,
-                       chef_cookbook_version:minimal_cookbook_ejson(CBV, chef_wm_util:base_uri(Req), ApiVersion) }
-                      || CBV <- CookbookVersions ]
-                    },
-            CBMapJson = chef_json:encode(JsonList),
-            {true, wrq:append_to_response_body(CBMapJson, Req), State};
+            case make_json_list(CookbookVersions, chef_wm_util:base_uri(Req), ApiVersion) of
```
jashaik, line 287: Would it help if we reduce the number of calls to chef_cbv_cache.erl module in the case of 'cbv_cache_enabled = false'
Flag = envy:get(chef_objects, cbv_cache_enabled, false, boolean),
case make_json_list(CookbookVersions, chef_wm_util:base_uri(Req), ApiVersion, Flag) of

Function Defination:
make_json_list(CookbookVersions, URI, ApiVersion, false) ->
    get_json_list(CookbookVersions, URI, ApiVersion);
make_json_list(CookbookVersions, URI, ApiVersion, true) ->
    Hash = erlang:phash2(CookbookVersions, 134217728),
    make_json_list(CookbookVersions, URI, ApiVersion, Hash, 0).
    
get_json_list(CookbookVersions, URI, ApiVersion) ->
       {
                      [ { CBV#chef_cookbook_version.name,
                          chef_cookbook_version:minimal_cookbook_ejson(CBV, URI, ApiVersion) }
                        || CBV <- CookbookVersions ]
          }.

We can use get_json_list(CookbookVersions, URI, ApiVersion) at line 323.

src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
```diff
@@ -0,0 +1,253 @@
...
+
+%% @doc intended for testing.
+stop() ->
+    gen_server:call(?SERVER, stop).
+
+% gen_server implementation
+
+init([Enabled, TTL]) ->
+    process_flag(trap_exit, true),
+    spawn_breaker(),
```
jashaik, line 117: Would it be possible to enable spawn_breaker()/breaker() process only if 'cbv_cache_enabled = true' as 'erlang:send_after(?QUEUE_LEN_REFRESH_INTERVAL, self(), update)' makes this process busy unnecessarily if 'cbv_cache_enabled = false'.

src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
```diff
@@ -0,0 +1,253 @@
...
+      self() ! update,
+      breaker(Count, Disabled);
+    update ->
+      case whereis(chef_cbv_cache) of
+        undefined ->
+          Count2 = Count;
+        Pid ->
+          {_, Count2} = process_info(Pid, message_queue_len)
+      end,
+      erlang:send_after(?QUEUE_LEN_REFRESH_INTERVAL, self(), update),
```
jashaik, line 251: Would it be possible to enable spawn_breaker()/breaker() process only if 'cbv_cache_enabled = true' as 'erlang:send_after(?QUEUE_LEN_REFRESH_INTERVAL, self(), update)' makes this process busy unnecessarily if 'cbv_cache_enabled = false'.

lbakerchef:
> > Do callers have to go through the gen_server to read the cache, or do they read it directly (I noticed you set read_concurrency to true)?
> 
> They do go through the gen_server for reads. I set read_concurrency true because this table is very read-heavy; and I was initially considering a pool of workers which would enable concurrent access. The flag is left over from that, I'll remove it.

Actually, I wasn't suggesting it be removed (it's generally good to have, but use your own judgement in this case).  I was merely hoping it was the case that processes didn't have to go through the gen_server for reads.  But if they do, no worries, and either way this is a substantial improvement to what we have, and it solves a customer issue nicely.

src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
marcparadise: Testing shows that this is still helpful under load, and does allow the server fail point to be at a higher rate of traffic.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_depsolver.erl
marcparadise: In my testing with service disabled, there was no significant impact to having the additional call in place -it failed at the same load that it failed without these changes in place.  Too, this way the chef_object configuration remains visible only within the module that needs it.

src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
marcparadise: This can only be done if we do go ahead and move the config check to the caller process - otherwise the breaker needs to be running for `get` to check its status.

jashaik (approved):

marcparadise:
Will rebase and merge this after a final round of perf/load testing today.

marcparadise:
@lbakerchef For a future refinement, I've just about talked myself into allowing callers to read the ets directly without going through the gen_server.  I think we'd  need a minor modification to handle the race scenario I talked about above: 

>  this approach does open a small gap where a caller (A) might miss on ETS , and another caller (B) completes its own cache put. When (A) attempts to claim intent on the key, the gen_server will allow it because (B) already finished, causing (A) to recalculate the response that (B) just populated. If that happens for multiple "A"s at the same time, things get messy.

Modifying the cache to check for the key already existing on `claim` would be a small change - and responding `retry` if it is present - /should/ still ensure that only one proc does the work of assembling the solution in that situation too.

docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
```diff
@@ -1456,6 +1456,27 @@ This configuration file has the following settings for `opscode-erchef`:
...
 :   The virtual IP address. Default value: `127.0.0.1`.
 
+`opscode_erchef['cbv_cache_enabled']`
+
+:   Enable cookbook version response caching by setting this to `true`. If you are frequently seeing
+    very long response times from `cookbook_versions` when under load, this is worth enabling.
+    Enabling this does open a window for a client to receive stale results. When a cookbook is updated
+    in place (without incrementing the version), and the old response has not been expired from cache,
+    the old response will be given to the client.  Subsequent client runs will receive the updated response.
+    Default value: `false`.
```
IanMadd, line 1466:
```suggestion
:   Enable cookbook version response caching by setting this to `true`. If you frequently see
    very long response times from `cookbook_versions` when under load, this is worth enabling.
    Enabling this makes it possible for a client to receive stale results. When a cookbook is updated
    in place (without incrementing the version), and the old response has not expired from the cache,
    the Infra Server will give the old response to the client. Subsequent client runs will receive the
    updated response. Default value: `false`.
```

docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
```diff
@@ -1456,6 +1456,27 @@ This configuration file has the following settings for `opscode-erchef`:
...
+
+`opscode_erchef['cbv_cache_item_ttl']`
+
+:   The minimum time in milliseconds that any given cookbook version response will be expired from the cache when
+    when cbv_cache_enabled is enabled.
+    Default value: `30000`.
```
IanMadd, line 1472:
```suggestion
:   The minimum time in milliseconds that Chef Infra Server will keep any given cookbook version response in the cache when
    when `cbv_cache_enabled` is enabled.
    Default value: `30000`.
```

docs-chef-io/content/server/v14/config_rb_server_optional_settings.md
```diff
@@ -1456,6 +1456,27 @@ This configuration file has the following settings for `opscode-erchef`:
...
+{{< note >}}
+
+Be careful if increasing this number - requests for a given set of cookbook_versions will be stale if the resolved cookbook versions are updated before the cache entry times out. This will
+not occur if cookbook version is incremented with every cookbook update, which is the recommended approach to updating cookbooks.
+
+{{< /note >}}
```
IanMadd, line 1478:
```suggestion
{{< note >}}

Be careful if increasing this number - requests for a given set of cookbook versions will be stale if the resolved cookbook versions are updated before the cache entry times out. This will
not occur if you increment the version of a cookbook with every cookbook update, which is the recommended approach to updating cookbooks.

{{< /note >}}
```

IanMadd (commented):
Docs suggestions

lbakerchef:
> @lbakerchef For a future refinement, I've just about talked myself into allowing callers to read the ets directly without going through the gen_server. I think we'd need a minor modification to handle the race scenario I talked about above

@marcparadise In general, if something is good as-is (or in this case, great as-is), I'm all for going with what we've got right now, and reserving any refinements for the future if and when we get around to it, and if and when we deem it practicable to do, the reward outweighs the risk, etc.  In this case, maybe a couple line comment somewhere in the source (assuming it's not already there) briefly describing at a high level the optimization that could be looked at, and the race condition to be solved, would be more than adequate :)

lbakerchef:
@marcparadise Notes for potential future improvements look great!

PR #3031, opened 2021-12-22
Improve security of the default site with more HTTP headers

tas50:
Our previous headers applied to the actual product endpoints, but not the static website. This was somewhat intentional since the static website doesn't matter and doesn't load any content so there's no security risk there. Not securing the additional endpoints did cause security scanners to incorrectly see the application as insecure. This gets us a passing grade now at securityheaders.com/

### DIFF
--- a/omnibus/files/server-ctl-cookbooks/infra-server/templates/default/nginx/nginx_chef_api_lb.conf.erb
+++ b/omnibus/files/server-ctl-cookbooks/infra-server/templates/default/nginx/nginx_chef_api_lb.conf.erb
@@ -42,7 +42,7 @@
     add_header Content-Security-Policy "default-src 'self';";
     add_header X-Content-Type-Options nosniff;
     add_header Permissions-Policy "camera=(); payment=(); microphone=(); gyroscope=(); magnetometer=(); midi=(); geolocation=()";
-    add_header Referrer-Policy 'origin';
+    add_header Referrer-Policy 'strict-origin';
       <% if @server_proto == 'https' %>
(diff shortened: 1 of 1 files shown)


omnibus/files/server-ctl-cookbooks/infra-server/templates/default/nginx/nginx_chef_api_lb.conf.erb
```diff
@@ -37,10 +37,12 @@
     proxy_send_timeout      300;
     proxy_read_timeout      300;
 
-    <% if !node['packages']['chef-manage'] %>
+    <% unless node['packages']['chef-manage'] %>
     add_header X-Frame-Options DENY;
     add_header X-Content-Type-Options nosniff;
     add_header X-XSS-Protection 1;
+    add_header Permissions-Policy "camera=(); payment=(); microphone=(); gyroscope=(); magnetometer=(); midi=(); geolocation=()";
+    add_header Referrer-Policy 'origin';
```
PrajaktaPurohit, line 45: do we want this to be strict-origin instead of origin?

tas50: I updated to strict-origin to avoid warnings

tas50:
With this config we now get an A on https://securityheaders.com. The key was making sure that we not just enabled headers on the API endpoints but also no index.html and images/css since that's all a scanner is going to ever see.

jashaik (approved):

PR #3053, opened 2022-01-06
Source sprockets-rails from rubygems in oc-id

tas50:
At some point we switched this to github. Probably to get a dep that wasn't released yet. There's no need to do this now and it slows down the bundle update/install. It also means we're pulling in pre-release changes which is pretty bad.


tas50:
https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/3713

PrajaktaPurohit:
https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/3778

PR #3063, opened 2022-01-07
Updated PG upgrade steps with analyze commands

jashaik:

### Description

Updated chef server upgrade steps with PG ANALYZE commands 

### Issues Resolved
Identified while upgrading PG 9.6 to 13.4 in hosted chef environment.

### Check List

### DIFF
--- a/docs-chef-io/content/server/upgrades.md
+++ b/docs-chef-io/content/server/upgrades.md
@@ -76,22 +76,7 @@ Set the `postgresql['pg_upgrade_timeout']` attribute in [chef-server.rb]({{< rel
        vacuumdb: vacuuming database "template1"
     ```
 
-1. Run `ANALYZE` on the PostgreSQL database.This process gathers statistics for the query planner to create the most efficient query execution paths. Accurate statistics will help the planner to choose the most appropriate query plan, and thereby improve the speed of query processing.
-    ```bash
-       sudo su - opscode-pgsql
-       /opt/opscode/embedded/bin/vacuumdb --all --analyze-only
-    ```
-
-   You should then see output like:
-   
-    ```bash
-       vacuumdb: vacuuming database "bifrost"
-       vacuumdb: vacuuming database "oc_id"
-       vacuumdb: vacuuming database "opscode-pgsql"
-       vacuumdb: vacuuming database "opscode_chef"
-       vacuumdb: vacuuming database "postgres"
-       vacuumdb: vacuuming database "template1"
-    ```
+{{% server_analyze_postgresql_db %}}
 
 1. Back up the PostgreSQL database before upgrading so you can restore the full database to a previous release in the event of a failure. See [Backup and Restore]({{< relref "server_backup_restore" >}}) for more information.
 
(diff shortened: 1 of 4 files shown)


docs-chef-io/content/server/upgrades.md
```diff
@@ -76,6 +76,23 @@ Set the `postgresql['pg_upgrade_timeout']` attribute in [chef-server.rb]({{< rel
...
+
+   You should then see output like:
+   
+    ```bash
+       vacuumdb: vacuuming database "bifrost"
+       vacuumdb: vacuuming database "oc_id"
+       vacuumdb: vacuuming database "opscode-pgsql"
+       vacuumdb: vacuuming database "opscode_chef"
+       vacuumdb: vacuuming database "postgres"
+       vacuumdb: vacuuming database "template1"
```
PrajaktaPurohit, line 93: Interesting - the changing flags do not seem to impact the output? It still looks like the db is being vacuumed.

docs-chef-io/content/server/upgrades.md
```diff
@@ -218,6 +235,23 @@ If you are running a Chef Infra Server release before 12.3.0, please contact Che
...
```
PrajaktaPurohit, line 252: We have suggested the vacuum full at the beginning of the process, but depending on the size of the data compared to the system that can also be done just before running analyze, optionally.

docs-chef-io/content/server/upgrades.md
```diff
@@ -471,6 +505,22 @@ The following External PostgreSQL upgrade steps are provided as a courtesy only.
...
+    ```
```
PrajaktaPurohit, line 522: Is there a section that links the commands to vacuum before this analyze? (I think this is the external PG section)

docs-chef-io/content/server/upgrades.md
jashaik: `/opt/opscode/embedded/bin/vacuumdb` is tool which accepts options like `--analyze-only` or `--full`. So the output is same if we use vacuumdb script.Both commands can be run on the psql console but needs to give the database names.

docs-chef-io/content/server/upgrades.md
jashaik: You mean just a link to Vacuum command ?

docs-chef-io/content/server/upgrades.md
PrajaktaPurohit: yes, the suggestion here would be that vacuum at the start is recommended to make the process faster, but the customers can also do that after the upgrade.

docs-chef-io/content/server/upgrades.md
PrajaktaPurohit: yes somewhere wee should mention even for external postgres that vacuum should be performed before running analyze

docs-chef-io/content/server/upgrades.md
vinay-satish: Have added them

vinay-satish (approved):
LGTM

PR #3072, opened 2022-01-10
Update address ranges in use for vbox compat

marcparadise:
Newer versions of VirtualBox no lnoger default to allowing
hosts to be assigned arbitrary IP addressed.  Instead, the
range 192.168.56.0/24 is permitted.

This change updates our dev vm environment and related tools/scripts
to use addresses in acceptable range.


dev/Vagrantfile
```diff
@@ -56,7 +56,6 @@ end
 
 Vagrant.configure("2") do |config|
   attributes = load_settings
-  config.omnibus.chef_version = :latest
```
marcparadise, line 59: 'omnibus' is not a supported attr of `config` anymore. This has no functional effect on dev vm, because we provision using the chef client included in chef server.

PrajaktaPurohit (approved):

PR #3079, opened 2022-01-12
Cookbook version caching updates 

### LINKED ISSUE #3077: Prevent cross-org collisions of CBV cached data
Currently, we cache cookbook version results based on a hash of the original cookbook version terms, before they are decorated with file paths. 

Given  multiple organizations have the same cookbooks/dependencies at the same version: 

1. Org 1 posts to cookbook_versions. Hash computes to 'ABC', and the decorated results with org-specific URLs, etc is cached under 'ABC'
2. Org 2 posts to cookbook_versions, and the resolved hash also computes to 'ABC'. We perform a cache lookup and get a hit on 'ABC' - but the data is the decorated cookbook versions out of 'Org A'. 

To avoid this possibility, include the OrgId in the cache key  -  `{OrgId, Hash}` instead of just `Hash`.

### LINKED ISSUE #3078: /cookbook_versions endpoint sometimes returns 'busy' as body response under heavy load
When cookbook version cache is enabled, and the server is shedding load under heavy usage, we can return a body response of `busy` which causes chef client runs to fail.  This should be a 503 to signal the client to retry.

 We'll fix this by making sure usage of `{error, busy}` and `{error, retry}` are consistently returned by `chef_cbv_cache` and and consistently handled by `chef_wm_depsolver`.

marcparadise:
Fixes #3077 by including the OrgId as part of the CBV cache key, eliminating the chance of cross-org cache pollution 
Fixes #3078 by ensuring consistent responses from `chef_cbv_cache` for `{error, busy}` and `{error, retry}`
 
#### Pending: 
* [x] pipeline run (blocked on pipeline/license failure) 
* [x] functional testing with cache enabled (in progress)

### DIFF
--- a/src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
+++ b/src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
@@ -131,7 +131,7 @@ enabled() ->
     envy:get(chef_objects, cbv_cache_enabled, false, boolean).
 
 %% Internal function that checks in with the process message queue for chef_cbv_cache intermittently
-%% and tracks whether its mailbox has grown to large. If that happens, response with '{error, busy}' instead
+%% and tracks whether its mailbox has grown too large. If that happens, response with '{error, busy}' instead
 %% of sending the request through.
 send_if_available(Msg) ->
     case enabled() of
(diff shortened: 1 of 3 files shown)


src/oc_erchef/apps/chef_objects/src/chef_cbv_cache.erl
```diff
@@ -130,13 +131,13 @@ enabled() ->
     envy:get(chef_objects, cbv_cache_enabled, false, boolean).
 
 %% Internal function that checks in with the process message queue for chef_cbv_cache intermittently
-%% and tracks whether its mailbox has grown to large. If that happens, response with 'busy' instead
+%% and tracks whether its mailbox has grown to large. If that happens, response with '{error, busy}' instead
```
lbakerchef, line 134:
```suggestion
%% and tracks whether its mailbox has grown too large. If that happens, response with '{error, busy}' instead
```
Minor nitpick, but while you're already in there...

lbakerchef:
Looks reasonable to me.  I assume you tested it already.

jashaik (approved):

PR #3103, opened 2022-01-20
change find command to ignore /var/log/lastlog

y-me-y:
The find tends to detect the `lastlog` because it's been recently modified. the `lastlog` can report as very large on disk and when copying to the temp directory, erroneously creates a large tar.gz file.

Note
The lastlog file is a database which contains info on the last login of each user. You should not rotate it. It is a sparse file, so its size on the disk is usually much smaller than the one shown by "ls -l" (which can indicate a really big file if you have in passwd users with a high UID). You can display its real size with "ls -s".

### Description

The find tends to detect the `lastlog` because it's been recently modified. the `lastlog` can report as very large on disk and when copying to the temp directory, erroneously creates a large tar.gz file. this change just ignores the lastlog file.

### Issues Resolved

N/A but had this bug come up a couple times for existing customers.

### Check List

### DIFF
--- a/omnibus/files/private-chef-scripts/gather-logs
+++ b/omnibus/files/private-chef-scripts/gather-logs
@@ -50,7 +50,7 @@ for i in /opt/{chef,opscode}*/version-manifest.txt \
     fi
 done
 
-for i in `find -L /var/log/${path}* -type f -mmin -"$modified_within_last_x_minutes"`; do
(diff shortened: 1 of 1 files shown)


PrajaktaPurohit (approved):

PrajaktaPurohit:
Thank you very much for the detailed description and the change PR! Much appreciated!

tas50 (approved):

PR #3129, opened 2022-01-29
Update a pile of misc deps & let dependabot bump more

tas50:
- Fix  src/oc_bifrost/oc-bifrost-pedant/Gemfile  to include the branch. This has never worked since we moved to main
- Use any chef 16 in chef-server-ctl and let dependabot do its things here


tas50:
https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/3812

PR #3130, opened 2022-01-29
Remove unused dependencies in chef-server-ctl

tas50:
- pry is not a runtime dep here
- rb-readline doesn't get required anywhere
- bunder is part of ruby now


tas50:
https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/3813

PrajaktaPurohit (approved):

PR #3193, opened 2022-02-22
Updating status and response header tests

vinay-satish:

### Description

We were using chef-server-running.json for testing Status and header response. In automate environment we do not have chef-server-running.json. Therefore we were skipping these test in automate. This PR updates the test to use the pedant_config.json instead of chef-server-running.json .

### Issues Resolved

JIRA - https://chefio.atlassian.net/browse/INFS-51
Updating chef-server in automate - https://github.com/chef/automate/pull/6726

### Check List

### DIFF
--- a/oc-chef-pedant/pedant_config.rb
+++ b/oc-chef-pedant/pedant_config.rb
@@ -204,3 +204,6 @@ old_runlists_and_search true
 
 # Log HTTP Requests
 log_file "/var/log/opscode/oc-chef-pedant/http-traffic.log"
+
+include_version_in_status false
+include_x_ops_api_info false
--- a/oc-chef-pedant/spec/api/status_spec.rb
+++ b/oc-chef-pedant/spec/api/status_spec.rb
@@ -1,12 +1,12 @@
 
-describe "Testing Status API with config", :config do
-  let (:config) { JSON.parse(IO.read("/etc/opscode/chef-server-running.json"))['private_chef'] }
+describe "Testing Status API" do
   context "Status API", :status, :smoke do
     it "GET /_status should respond with valid health data" do
       r = get("#{platform.server}/_status", superuser)
       r.should look_like(status_for_json: 200)
       data = JSON.parse(r)
-      data.has_key?("server_version").to_s.should eql config['opscode-erchef']['include_version_in_status'].to_s
+      include_version_in_status = Pedant::Config.include_version_in_status
+      data.has_key?("server_version").should eql include_version_in_status
       data.has_key?("status").should eql true
       data.has_key?("upstreams").should eql true
       data.has_key?("keygen").should eql true
--- a/src/chef-server-ctl/habitat/config/pedant_config.rb
+++ b/src/chef-server-ctl/habitat/config/pedant_config.rb
@@ -199,3 +199,6 @@ required_recipe_enabled false
 reindex_endpoint "https://127.0.0.1"
 
 chef_pgsql_collector true
+
+include_version_in_status false
+include_x_ops_api_info false
--- a/oc-chef-pedant/spec/api/header_spec.rb
+++ b/oc-chef-pedant/spec/api/header_spec.rb
@@ -6,7 +6,6 @@
 describe "Headers", :headers do
   let (:request_url) { api_url("users") }
   let (:requestor) { platform.admin_user }
-  let (:config) { JSON.parse(IO.read("/etc/opscode/chef-server-running.json"))['private_chef'] }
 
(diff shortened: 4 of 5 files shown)


vinay-satish:
Adhoc build - https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/3939#facfa621-7ca6-4dd0-84e5-be5eb4eb9cee

vinay-satish:
Hab build - https://buildkite.com/chef/chef-chef-server-main-habitat-build/builds?branch=zei%2Ffixing_test_for_automate

vinay-satish:
Umbrella - https://buildkite.com/chef/chef-umbrella-main-chef-server/builds/1073

oc-chef-pedant/spec/api/status_spec.rb
```diff
@@ -1,12 +1,12 @@
...
-  let (:config) { JSON.parse(IO.read("/etc/opscode/chef-server-running.json"))['private_chef'] }
+describe "Testing Status API" do
   context "Status API", :status, :smoke do
     it "GET /_status should respond with valid health data" do
       r = get("#{platform.server}/_status", superuser)
       r.should look_like(status_for_json: 200)
       data = JSON.parse(r)
-      data.has_key?("server_version").to_s.should eql config['opscode-erchef']['include_version_in_status'].to_s
+      include_version_in_status = Pedant::Config.include_version_in_status
+      data.has_key?("server_version").to_s.should eql include_version_in_status.to_s
```
marcparadise, line 9: I know we were previously doing this, but should we be forcing to string? The config option is a boolean value.

oc-chef-pedant/pedant_config.rb
```diff
@@ -204,3 +204,6 @@
 
 # Log HTTP Requests
 log_file "/var/log/opscode/oc-chef-pedant/http-traffic.log"
+
+include_version_in_status false
```
marcparadise, line 208: Do these tests pass when we set these two values to `true`?

marcparadise (changes requested):
This looks good, with a couple of questions inline. 


Overall, this does subtly change the thing we are testing from "is the API  behaving correctly given its configuration" to "is the API behaving correctly given that our test expectations for its configuration are correct".  I think that's OK (or at least consistent with other tests like ldap), but it seems like exposing the running configuration via erchef API would have the potential to simplify this and other testing (like ldap).  That's something we could look into outside of this change.

jashaik:
LGTM.

oc-chef-pedant/pedant_config.rb
vinay-satish: It fails because the default value is false. If we set the variables in the chef-server.rb to true, and set here also true, then it will pass.

The default values in default.rb
`default['private_chef']['opscode-erchef']['include_version_in_status'] = false`
`default['private_chef']['opscode-erchef']['include_x_ops_api_info'] = false`

oc-chef-pedant/spec/api/status_spec.rb
vinay-satish: Ya I think forcing to string is not needed. I will try this out and see.

marcparadise (approved):

oc-chef-pedant/spec/api/status_spec.rb
```diff
@@ -1,12 +1,12 @@
 
-describe "Testing Status API with config", :config do
-  let (:config) { JSON.parse(IO.read("/etc/opscode/chef-server-running.json"))['private_chef'] }
+describe "Testing Status API" do
```
PrajaktaPurohit, line 2: The :config tag would let these tests run with the config flag, that might be ok to leave in?

PrajaktaPurohit (approved):
Few questions/ suggestions, mostly looking good.

oc-chef-pedant/spec/api/status_spec.rb
jashaik: I feel it is ok.

src/chef-server-ctl/habitat/config/pedant_config.rb
```diff
@@ -199,3 +199,6 @@
 reindex_endpoint "https://127.0.0.1"
 
 chef_pgsql_collector true
+
```
marcparadise, line 202: I start to wonder if we shouldn't do some rendering of this config based on the state of the server config since there are a few places where a mismatch will break things. (Not for this PR)

marcparadise (approved):

PR #3209, opened 2022-02-28
Don't source veil from git in oc-bifrost-pedant

tas50:
We shipped the latest to rubygems here


jashaik:
https://buildkite.com/chef/chef-umbrella-main-chef-server/builds/1096

PR #3256, opened 2022-04-04
Remove unsupported migrations (< 12.17.15)

marcparadise:
The oldest version we support upgrading from is 12.17.15.

Migrations 020-034 are present in 12.17.15, which means that they will
not be used in versions later than this unless the customer is on <
12.17.15

In those cases, the customer must upgrade to 12.17.15 before proceeding
with additional upgrades.


PrajaktaPurohit (approved):

jashaik:
Pipeline Information-
Adhoc -https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/4198

jashaik:
Adhoc - https://buildkite.com/chef/chef-umbrella-main-chef-server/builds/1188

PR #3354, opened 2022-06-07
Revise gem update process

lbakerchef:
Inward facing docs revision.

`oc-id` can now be updated without any special `Rails` steps.

### DIFF
--- a/dev-docs/FrequentTasks.md
+++ b/dev-docs/FrequentTasks.md
@@ -64,9 +64,6 @@ out and stable.
 - omnibus/Gemfile.lock
 - oc-chef-pedant/Gemfile.lock
 - src/oc-id/Gemfile.lock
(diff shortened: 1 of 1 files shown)


dev-docs/FrequentTasks.md
```diff
@@ -64,10 +64,6 @@ out and stable.
 - omnibus/Gemfile.lock
 - oc-chef-pedant/Gemfile.lock
 - src/oc-id/Gemfile.lock
-  DO NOT update this file unless Rails is upgraded.
-  Due to the Rails version, we've locked a lot of dependencies. Make
-  sure we can build and run oc-id before merging to main.
-  You'll need the libsqlite3-dev library if doing this on Ubuntu.
```
PrajaktaPurohit, line 70: Looks like the last line might be a comment for depsolver? The rest of it can be removed.

PR #3206, opened 2022-02-24
Upgrade Erlang to 24.2

lbakerchef:
Signed-off-by: Lincoln Baker <lbaker@chef.io>

### Description

[Please describe what this change achieves]

### Issues Resolved

[List any existing issues this PR resolves, or any Discourse or
StackOverflow discussions that are relevant]

### Check List

- [ ] New functionality includes tests
- [ ] All buildkite tests pass
- [ ] Full omnibus build and tests in buildkite pass
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/main/CONTRIBUTING.md#developer-certification-of-origin-dco>
- [ ] PR title is a worthy inclusion in the CHANGELOG

PR #3368, opened 2022-06-13
Upgrade dependencies 2

lbakerchef:
After extensive consultation with @PrajaktaPurohit, this PR was created as a "debug PR" to remove as many commits as possible from the "update Erlang" PR (https://github.com/chef/chef-server/pull/3206), as long as builds and pipelines remain green.  This is being done so that there are fewer commits in the "update Erlang" PR to debug, and so we can more easily know where the issues are in that remaining PR. These are the commits we know are green and don't have any issues.  These commits have been debugged and tested, and result in green builds and pipelines on main.

Any other commits in the other PR (besides the license_scout commit, which was originally here but was moved there) will break main if moved here.

This PR depends on:
https://github.com/chef/mini_s3/pull/39


lbakerchef:
https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/4659
https://buildkite.com/chef/chef-umbrella-main-chef-server-full/builds/189

lbakerchef:
https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/4674
https://buildkite.com/chef/chef-umbrella-main-chef-server-full/builds/207

lbakerchef:
Do we want these commits squashed?

src/bookshelf/rebar.lock
```diff
@@ -37,8 +43,16 @@
   {git,"https://github.com/chef/erlware_commons",
        {ref,"f511ed875bea87212ba9ee40ade1c9f53498ce79"}},
   0},
- {<<"fs">>,{pkg,<<"fs">>,<<"6.1.1">>},1},
- {<<"goldrush">>,{pkg,<<"goldrush">>,<<"0.1.9">>},1},
+% {<<"fs">>,{pkg,<<"fs">>,<<"6.1.1">>},1},
+ {<<"fs">>,
```
PrajaktaPurohit, line 47: is this here because of license_scout issues? Is license scout making it so that we cannot pull the hex packages as erlang dependencies?

PrajaktaPurohit: Do we know how this worked with the previous versions of the dependencies?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.app.src
```diff
@@ -23,6 +23,7 @@
   {applications, [kernel,
                   stdlib,
                   stats_hero,
+                  chef_objects,
```
PrajaktaPurohit, line 26: Could you please add more context related to this?

lbakerchef: There's context in the commit.  Would you rather have the context in the code itself?
```
    ===> Compiling oc_chef_authz
    ===> Compiling apps/oc_chef_authz/src/oc_chef_policy_group.erl failed
    apps/oc_chef_authz/src/oc_chef_policy_group.erl:26:2: behaviour chef_object undefined
```

src/bookshelf/rebar.config
```diff
@@ -10,26 +10,41 @@
     %% lager has to come first since we use its parse transform
     {lager, ".*",
         {git, "https://github.com/erlang-lager/lager",              {branch, "master"}}},
+    % this is here because of license scout issues
+    {base16, ".*",
+        {git, "https://github.com/goj/base16",                      {tag,    "1.0.0"}}},
```
sreepuramsudheer, line 15: Could you please let us know where are we using this library in chef-server code.

src/bookshelf/rebar.config
```diff
@@ -10,26 +10,41 @@
...
     {cf, "",
         {git, "https://github.com/project-fifo/cf",                 {branch, "master"}}},
     {chef_secrets, ".*",
         {git, "https://github.com/chef/chef_secrets",               {branch, "main"}}},
+    % this is here because of license scout issues
+    {eini, ".*",
+        {git, "https://github.com/erlcloud/eini",                   {tag,    "1.2.6"}}},
```
sreepuramsudheer, line 22: Could you please let us know where are we using this library in chef-server code.
Are we using this to resolve licence scout issues.

src/bookshelf/rebar.lock
```diff
@@ -51,19 +65,26 @@
...
+       {ref,"8e34985a3cd0ac2a7fc2a88a041554c64d33e74b"}},
+  1},
  {<<"meck">>,
   {git,"https://github.com/eproxus/meck",
        {ref,"06192a984750070ace33b60a492ca27ec9bc6806"}},
   0},
  {<<"mini_s3">>,
   {git,"https://github.com/chef/mini_s3",
-       {ref,"8a6b80ec60be799c5f69d9b2f22eb0af55b4dc77"}},
+       {branch,"lbaker/upgrade-erlang-24.2"}},
```
sreepuramsudheer, line 87: Would it be possible to change it to reference/tag instead of branch name.

src/bookshelf/rebar.lock
jashaik: @lbakerchef -  I feel, We need to understand why the license_scout behaviour is different in this case ?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.app.src
jashaik: Is it the same before erlang upgarde 24 ?

src/bookshelf/src/bksw_io_names.erl
```diff
@@ -17,6 +17,8 @@
 
 -module(bksw_io_names).
 
+-compile([nowarn_deprecated_function]).
```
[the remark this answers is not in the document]
jashaik: It would be better If we update the deprecated fucntions than nowarn flag.

src/bookshelf/rebar.config
jashaik: @sreepuramsudheer Is trying to test by removing newly added deps for licence_scout to understand the actual issue.

src/bookshelf/rebar.lock
lbakerchef: @PrajaktaPurohit asked me to create an issue about it.  It's in the backlog now.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.app.src
lbakerchef: @jashaik No errors before the erlang 24 upgrade.

src/bookshelf/src/bksw_io_names.erl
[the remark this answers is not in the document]
lbakerchef: @sreepuramsudheer @jashaik We could remove the change, but then the change would come again when we upgrade to erlang 24, which we want to do as soon as possible.

The bigger context is, this issue is fixed in the sigv4 changes.  So this is a bandaid until sigv4 lands.  That was my idea, anyway.  That's why I did nowarn instead of investing time to fix it.

src/bookshelf/rebar.config
lbakerchef: Maybe transitive.  See comment above.

src/bookshelf/rebar.lock
lbakerchef: @sreepuramsudheer The mini_s3 changes have not been merged yet, that's why we are referencing the special branch.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
```diff
@@ -115,10 +115,22 @@ handle_undefined_start(undefined, Req) ->
 handle_undefined_start(Start, _Req) ->
     Start.
 
+% NOTE(Lincoln Baker):
+% http_uri:decode became deprecated, however there didn't appear to be a
+% direct drop-in library function to replace it for how we were using it.
+% uri_string:dissect_query/1 returns key-value pairs, and for our inputs
+% the value component will often or always be true based on my testing.
```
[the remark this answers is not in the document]
marcparadise: But I'm also trying to track down the reason we approach this in this way.  In this case, we want to convert the value we receive for 'start' , which comes out of chef_index:search.  For all search providers, it's extracted from the response JSON - and the docs for both ES and solr state this value will be an int. 

For GET/POST requests, we hit this path when the value is for any reason `undefined` coming out of the search API call - in that case, we'll use the query value.  We pass the query value through this function to  convert it to an int, which then gets included in the search response body. 

It does look like the value of the query string could be URL encoded (webmachine docs aren't clear, and I didn't dive too deeply).  It is not possible to get the multi-key-value-pairs because we're not looking at a full query string - only at the resolved value. In addition, integers don't get any encoding treatment - so if the string doesn't directly convert to an int, it's not valid for our purposes. 

Separately, we hit it when we invoke `from_params` to build the query record (applies to both GET/POST), via `chef_index:query_from_params`, which does the same thing with validation and will raise if the value is non-integer. 

TL;DR:
- the replacement function does not work the same as the original, though it seems close. But we may not need this, because: 
  - the things we're calling unencode on are query string values, extracted via `wrq:get_qs_value`. 
  - these things are not full URL-encoded strings - they are a potentially URL-encoded portion of a value, not a series of KVPs 
  - any integer will not get url encoded; anything that does get url encoded is not a valid integer and can't be accepted. 

If that logic sounds sane and you agree, maybe we could simplify by  doing an int conversion of the original  string list in both places.

marcparadise (changes requested):

src/bookshelf/src/bksw_io_names.erl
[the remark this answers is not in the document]
lbakerchef: > Would it be possible to remove this change as we are still on erlang 22.

I would be happy to do this, but if I understood @PrajaktaPurohit correctly, the purpose of this PR was to move as many commits from the Erlang 24x upgrade as possible (if the commits don't break anything, pipelines remain green, etc) onto main.  You didn't have this context, so I'm just providing it for you :)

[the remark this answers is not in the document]
lbakerchef: > It would be better If we update the deprecated fucntions than nowarn flag.

Fixed in upcoming sigv4 changes, if that helps your thoughts on this.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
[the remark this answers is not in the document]
lbakerchef: Greetings.  Just to provide context which may or may not be useful, since I didn't have any information or context to go on when I wrote the replacement (including what this function was originally designed to do or why, what inputs it was designed to transform and the outputs it was supposed to generate, etc), I simply took all of the inputs that we use during integration tests (obtained by simply inserting `io:format` statements into appropriate locations), and made sure the replacement function I wrote gave the same outputs as those inputs.

Having said that, that may not be what we want :)

> If that logic sounds sane and you agree, maybe we could simplify by doing an int conversion of the original string list in both places.

If you could explain the above a little more (what you mean by int conversion, and what you mean by 'both places,' I'll be happy to get to work.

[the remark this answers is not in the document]
lbakerchef: EDIT:
A little more context.  IIRC, during testing, every value arriving into line 124's KVs (unencode function) was of the form `{"some-key", true}`, so that's basically what I designed the function to deal with.  I have no idea if that changes anything (or even if it should), I'm just throwing it out there.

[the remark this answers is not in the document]
lbakerchef: Ok, I made the changes you requested (if I understood you correctly).  Adhoc and umbrella are green (see comment below).

src/bookshelf/rebar.lock
lbakerchef: license_scout changes have been removed from this PR, and put in the final PR for the Erlang upgrade.

src/bookshelf/rebar.config
lbakerchef: license_scout changes have been removed from this PR, and put in the final PR for the Erlang upgrade.

src/bookshelf/rebar.config
lbakerchef: license_scout changes have been removed from this PR, and put in the final PR for the Erlang upgrade.

src/bookshelf/src/bksw_io_names.erl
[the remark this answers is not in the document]
jashaik: We are not updating the Erlang in this PR So please remove this line.This can be part of your Erlang upgrade PR

src/oc_erchef/rebar.lock
```diff
@@ -86,7 +86,7 @@
   1},
  {<<"mini_s3">>,
   {git,"https://github.com/chef/mini_s3",
-       {ref,"8a6b80ec60be799c5f69d9b2f22eb0af55b4dc77"}},
+       {branch,"lbaker/upgrade-erlang-24.2"}},
```
jashaik, line 89: Could you please use the references If you would like to merge this to main as lock should always have the reference which doesn’t behave if you commit something into "lbaker/upgrade-erlang-24.2 branch

jashaik:
@lbakerchef you could also add the below commits If possible to make one dep PR
https://github.com/chef/chef-server/commit/f46c52cd6e9b931afba1803db77752b39edbd085
https://github.com/chef/chef-server/commit/41114dd52ddc833633e36879f826333e5208a6d4
https://github.com/chef/chef-server/commit/c63c60b9425364e48f6665c0669ca5f16f375da0
https://github.com/chef/chef-server/commit/5120fd7e8452d3110feaa3f2816f3ddc9d73edd2

src/oc_erchef/apps/chef_index/src/chef_index_query.erl
```diff
@@ -106,7 +106,7 @@ decode({nonneg_int, Key}, Val, Default) ->
                 {Default, default};
             Value ->
                 try
-                    {list_to_integer(http_uri:decode(Value)), Value}
```
jashaik, line 109: Any reason for removing the http_uri:decode completely ? Does it effect any functionality ?

src/bookshelf/rebar.lock
```diff
@@ -63,7 +63,7 @@
...
```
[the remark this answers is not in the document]
lbakerchef: Currently, that mini_s3 branch hasn't been merged yet, so we can't use the reference yet (as far as I know).  After that mini_s3 PR is merged, I will update the reference here.

src/oc_erchef/rebar.lock
lbakerchef: See comment above.

src/oc_erchef/apps/chef_index/src/chef_index_query.erl
lbakerchef: The http_uri:decode function was either deprecated or removed.  I fixed the issue, and then made further changes that @marcparadise recommended in this PR.

It builds and passes tests.

lbakerchef:
> @lbakerchef you could also add the below commits If possible to make one dep PR
> [f46c52c](https://github.com/chef/chef-server/commit/f46c52cd6e9b931afba1803db77752b39edbd085)
> [41114dd](https://github.com/chef/chef-server/commit/41114dd52ddc833633e36879f826333e5208a6d4)
> [c63c60b](https://github.com/chef/chef-server/commit/c63c60b9425364e48f6665c0669ca5f16f375da0)
> [5120fd7](https://github.com/chef/chef-server/commit/5120fd7e8452d3110feaa3f2816f3ddc9d73edd2)

If those commits are added, it will break main.  Those commits only work after Erlang is upgraded.  That's why they are in the other PR.

src/bookshelf/src/bksw_io_names.erl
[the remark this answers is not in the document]
lbakerchef: I understand why you are asking for this.  However the full context of this PR is, after extensive consultation with @PrajaktaPurohit, this is a "debug PR" to move as many commits as possible from the "update Erlang" PR into this PR (as long as builds and pipelines remain green) so there are fewer commits in that PR to debug, and so we more easily know where the issues are in that remaining PR.  These are the commits we know are green and don't have any issues.

It's not your fault that you don't have all the context - I made this PR quickly and there was no write-up or context.  I will add it now, and perhaps it will help.

Let me know your thoughts.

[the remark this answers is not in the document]
lbakerchef: @jashaik Maybe it is possible that we merge both PRs (this one, and the "upgrade Erlang" PR) at the same time, after both are approved.  If we do that, all the changes land at once.

src/oc_erchef/apps/chef_index/src/chef_index_query.erl
marcparadise: While it is safe/correct to skip the decode in casese where we're just doing a `list_to_integer` anyway, I did not see our usage of http_uri:decode in the context of `transform_query` where I think we still need that functionality for complex/multipart queries.  

Maybe our best answer for this case is to pull in the original http_uri implementation of `decode`:
 
https://github.com/erlang/otp/blob/maint-21/lib/inets/src/http_lib/http_uri.erl#L143-L161

lbakerchef: @mp Done.  Feel free to have a look.

lbakerchef:
7/5/22
https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/4732
https://buildkite.com/chef/chef-umbrella-main-chef-server-full/builds/221

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
```diff
@@ -118,7 +118,7 @@ handle_undefined_start(Start, _Req) ->
 decode(undefined) ->
     0;
 decode(Val) ->
-    list_to_integer(http_uri:decode(Val)).
```
vinay-satish, line 121: don't we want to use http_uri_decode here?

vinay-satish (approved):
LGTM!

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
lbakerchef: @marcparadise Thought it was safe to remove it, which also simplifies the code.

src/bookshelf/rebar.lock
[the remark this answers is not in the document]
PrajaktaPurohit: let us get the mini_s3 reference locked before merging this, left some comments on the mini_s3 PR.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.app.src
PrajaktaPurohit: does this make sense to be included as a part of the Erlang bump PR then?

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
PrajaktaPurohit: since we have the taken the implementation from OTP, it might be the best to leave the original calls to encode and decode as is?

src/oc_erchef/apps/chef_index/src/chef_index_query.erl
PrajaktaPurohit: Since we have pulled in the original implementation, should we leave the calls to decode and unencode as is?

src/bookshelf/src/bksw_io_names.erl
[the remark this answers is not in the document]
lbakerchef: @PrajaktaPurohit Thanks, and no worries.  I'll have a look at the `nowarn`s to see how much work it would be to duplicate the work.

src/bookshelf/rebar.lock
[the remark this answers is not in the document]
lbakerchef: @PrajaktaPurohit Will do.  I've responded to the mini_s3 comments -have a look when you can.

src/oc_erchef/apps/chef_index/src/chef_index_query.erl
lbakerchef: @PrajaktaPurohit Marc thought it would be safe, plus it would simplify things, if we eliminated some of the calls to decode.  But others are left alone.

src/oc_erchef/apps/oc_chef_wm/src/chef_wm_search.erl
lbakerchef: After discussion we agreed to leave it as it currently is.

Issue #1896, opened 2020-01-29
Additional authz_time, authz_count, url in the logs


PrajaktaPurohit:
It would be useful while debugging to get the same request times specifically the authz_time and authz_count and the url being used by the http request in the logs for failed requests as the successful requests.

A successful request:
```
2019-04-16T22:35:20Z erchef@127.0.0.1 method=GET; path=/organizations/support/groups/admins; status=200; req_id=g3IAA2QAEGVyY2hxMjcuMC4wLjECAAAJrQAAAAIAAAAA; org_name=support; couchdb_groups=false; couchdb_organizations=false; couchdb_containers=false; couchdb_acls=false; 503_mode=false; couchdb_associations=false; couchdb_association_requests=false; req_time=21; rdbms_time=10; rdbms_count=3; authz_time=6; authz_count=1; user=pivotal; req_api_version=1;
2019-04-16T22:35:20Z erchef@127.0.0.1 method=PUT; path=/organizations/support/groups/admins; status=200; req_id=g3IAA2QAEGVyY2hlZMjcuMC4wLjECAAAJ1wAAAAIAAAAA; org_name=support; couchdb_groups=false; couchdb_organizations=false; couchdb_containers=false; couchdb_acls=false; 503_mode=false; couchdb_associations=false; couchdb_association_requests=false; req_time=41; rdbms_time=29; rdbms_count=4; authz_time=6; authz_count=1; user=pivotal; req_api_version=1;
```
Failed request:
```
2020-01-28T15:42:39Z erchef@127.0.0.1 method=GET; path=/organizations/explms/nodes/chexfhredis016.ch.expeso.com; status=500; req_id=g3IAA2QAEGVyY2hlZkAxMjcuMC4wLjECAAA54gAAAAIAAAAA; org_name=explms; couchdb_groups=false; couchdb_organizations=false; couchdb_containers=false; couchdb_acls=false; 503_mode=false; couchdb_associations=false; couchdb_association_requests=false; req_time=4008; rdbms_time=1; rdbms_count=3; user=chexfhredis016.ch.expeso.com; req_api_version=1;
```

teknofire:
For the URL that would only apply during a failed request that either timed out or had some error while being performed.  For example with the following crash log

```
[...]

PR #3408, opened 2022-07-13
introduced error cases to supress case_clause errors because of bifrost delayed response.

sreepuramsudheer:
### Description

Code changes to solve issue raised https://github.com/chef/chef-server/issues/1896.
### Issues Resolved

(https://github.com/chef/chef-server/issues/1896)

### Check List

### DIFF
--- a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.erl
+++ b/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.erl
@@ -123,6 +123,7 @@ create_entity_if_authorized(Context, OrgId, CreatorAId, ObjectType) ->
             {error, forbidden};
         true ->
             create_entity_with_container_acl(CreatorAId, ContainerAId, ObjectType)
+
     end.
 
 %%%
@@ -318,7 +319,7 @@ get_data_for_id([Id | _Ids] = OIds , [{RId, RData} | Resources], Denied, Permitt
 %
 -spec is_authorized_on_resource(requestor_id(), resource_type(), object_id(),
                                 'actor'|'group', actor_id(), access_method())
-                               -> true|false|{error,server_error}.
+                               -> true|false|{error,any()}.
 is_authorized_on_resource(RequestorId, ResourceType, ResourceId, ActorType, ActorId, AccessMethod)
   when is_atom(ResourceType) and is_atom(ActorType) and is_atom(AccessMethod) ->
     Url = make_url([pluralize_resource(ResourceType), ResourceId, <<"acl">>,
@@ -328,7 +329,7 @@ is_authorized_on_resource(RequestorId, ResourceType, ResourceId, ActorType, Acto
         %% This api returns not found for any missing permission
         {error, not_found} -> false;
         %% Otherwise, we expect server_error; not forbidden
-        {error, server_error} -> {error, server_error}
+        {error, Error} -> {error, Error}
     end.
 
 %


sreepuramsudheer:
Adhoc build: https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/4767

jashaik:
LGTM. Nice work @sreepuramsudheer

PrajaktaPurohit:
Thanks for getting the additional error reporting started. Could you please look at the verify test currently failing?

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.erl
```diff
@@ -328,7 +329,9 @@ is_authorized_on_resource(RequestorId, ResourceType, ResourceId, ActorType, Acto
         %% This api returns not found for any missing permission
         {error, not_found} -> false;
         %% Otherwise, we expect server_error; not forbidden
-        {error, server_error} -> {error, server_error}
+        {error, server_error} -> {error, server_error};
```
marcparadise, line 332: Let's change this to have any other error tuple returned as-is, eg `{error, Any} -> {error, Any}`; the shape of the tuple is more important to capture than each individual possible error type.

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.erl
```diff
@@ -123,6 +123,7 @@ create_entity_if_authorized(Context, OrgId, CreatorAId, ObjectType) ->
             {error, forbidden};
```
marcparadise, line 123: This will still fail with a 500 in case of a non-forbidden error return, which I think is correct - but it does mean that we'll potentially lose logging/error info here.  Let's create a new issue to confirm & fix that if needed.

marcparadise (changes requested):

src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.erl
PrajaktaPurohit: Yeah - I think it would be useful to add the matcher for the {error, Error} case from https://github.com/chef/chef-server/blob/170469aeff9eb81eb98784a2d033f9361ee5012a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.erl#L332

PrajaktaPurohit: and similarly a matcher for {error, Error} in https://github.com/chef/chef-server/blob/170469aeff9eb81eb98784a2d033f9361ee5012a/src/oc_erchef/apps/oc_chef_authz/src/oc_chef_authz.erl#L118

PrajaktaPurohit: That error would also need to propogate all the way up through https://github.com/chef/chef-server/blob/6d31841ecd73d984d819244add7ad6ebac284323/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_organizations.erl#L253, https://github.com/chef/chef-server/blob/6d31841ecd73d984d819244add7ad6ebac284323/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_associations.erl#L212, https://github.com/chef/chef-server/blob/afd0ec44666facb2c9c87d937fcbce7bb57ad10f/src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl#L367 into the modules calling those functions.

marcparadise (approved):

marcparadise:
We can merge after rebase

PR #3637, opened 2023-04-03
Updating s3 docs for Automate and A2HA

vinay-satish:
### Description

[Please describe what this change achieves]

### Issues Resolved

### Check List

### DIFF
--- a/docs-chef-io/content/server/_index.md
+++ b/docs-chef-io/content/server/_index.md
@@ -229,6 +229,14 @@ modification when using an external S3 provider:
 </tbody>
 </table>
 
+{{< note >}}
+
+For automate or A2HA environment the `chef-server.rb` will not be present. In that case, we need to update
+the `s3_url`, `s3_external_url` and `s3_platform_bucket_name` with appropriate value in the `sys.config`
+present at the location `$(hab pkg path chef/automate-cs-oc-erchef)/config/sys.config`
+
+{{< /note >}}
+
 ## External PostgreSQL
 
 The following diagram highlights the specific changes that occur when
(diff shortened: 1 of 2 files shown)


docs-chef-io/content/server/_index.md
```diff
@@ -229,6 +229,13 @@ modification when using an external S3 provider:
 </tbody>
 </table>
 
+{{< note >}}
+
+For automate or A2HA environment the `chef-server.rb` will not be present. In that case, we need to update
+the `s3_url`, `s3_external_url` and `s3_platform_bucket_name` with appropriate value in the `sys.config`
```
vinay-satish, line 235: Add the location - '$(hab pkg path chef/automate-cs-oc-erchef)/config/sys.config'

IanMadd:
Why should this be in the Server docs? This seems like content that should be in an Automate config doc. I also don't see any of those config settings in the server docs or anywhere else.

IanMadd:
Maybe I'm missing something.

First, we shouldn't be talking about A2HA in Server docs. 

None of those settings are referenced in the Server docs, so if users can and should modify or set them in Server, then we should document what they are first. 

From the context here, it looks like something that only applies to Automate users, so this would fall under a doc for configuring Server to work with Automate in the Automate docs.

If chef-server.rb isn't present, then what is the alternative? Why do they care about sys.config if their issue is the chef-server.rb. The relationship between the two isn't clear here. Can the settings that normally fall in the chef-server.rb be passed from the sys.config, or is that file stored on S3 and the path is in the sys.config?

Also, it looks like you're not specifying a location, but a command to run to return a file path.

IanMadd:
Closing because there hasn't been a response to my questions.

PR #3832, opened 2024-07-08
Shahid/keydb

jashaik:
### Description

[Please describe what this change achieves]

https://chefio.atlassian.net/browse/CHEF-9582

### Issues Resolved

### Check List


jashaik:
https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/6418#019090a4-6fd1-477d-b4b2-a59d0aacd073

jashaik:
Keydb working ubuntu build - https://buildkite.com/chef/chef-chef-server-main-omnibus-adhoc/builds/6477

omnibus/files/server-ctl-cookbooks/infra-server/attributes/default.rb
```diff
@@ -452,6 +452,36 @@
   '60' => '1000',
 }
```
marcparadise, line 454: we should be able to get rid of most of the redis_lb entries - only hte stuff that runit cares about needs to remain in order to remove it (enable = false, dir, data_dir, log_*)

omnibus/files/server-ctl-cookbooks/infra-server/libraries/private_chef.rb
```diff
@@ -309,8 +313,8 @@ def gen_frontend
       PrivateChef[:role] = 'frontend'
       PrivateChef['bookshelf']['enable'] ||= false
       PrivateChef['bookshelf']['vip'] ||= PrivateChef['backend_vips']['ipaddress']
-      PrivateChef['redis_lb']['enable'] ||= false
```
marcparadise, line 312: I think we'll want to keep redis_lb.enable , and change this to '= false' instead of ||=false

marcparadise (commented):
Other than needing a squash, this looks good. 

If the intent is also to clean up the original service, we should remove most of the redis_lb attributes (for reference you can look in default attributes for the comment "Legacy attributes required for cleanup", which shows which ones we should keep) . I think it's mostly data_dir, dir, and log_directory. 

Then we'll add a section to the `omnibus/files/server-ctl-cookbooks/infra-server/recipes/cleanup.rb` recipe to clean up redis_lb. 

Also add redislb to the 'removed_services' attribute for consistency - though it doesn't seem like it's actually used...

PR #3927, opened 2024-10-17
[CHEF-1685] WIP - allow org-admins to modify organizations

marcparadise:
In addition to give org-admins permissions to CRUD organizations, this removes the requirement that in order to modify an organization, the actor must be a member of the organization.

However, they must still have appropriate permissions to perform any CRUD action related to an organization.

This supports the multi-tenancy case where a customer has many organizations to manage but does not necessarily need to admins to be a part of those organizations. The primary use case is SaaS offering , in which customers have full control over a chef server installation but do not have local/chef-server-ctl access, and must keep the pivotal key locked down for security purposes.

This functionality is already available using the pivotal/superuser key, but the pivotal key should not be widely distributed. This functionality was also originally intended to be available to org-admins but the completion of that work was never prioritized.


src/oc_erchef/apps/oc_chef_wm/src/oc_chef_wm_base.erl
```diff
@@ -441,7 +443,23 @@ is_authorized(Req, State, Extractor) ->
             end;
         {false, ReqOther, StateOther} ->
             %% FIXME: the supported version is determined by the chef_authn application
-            %% also, see: https://wiki.corp.chef.io/display/CORP/RFC+Authentication+Version+Negotiation
```
marcparadise, line 444: ^ removed because this link is dead, and I can find no replacement for it in the new wiki.

PR #3943, opened 2024-11-09
Disable busywaits

feld:
On VMs this often leads to increased CPU usage and in cloud environments increased consumption of CPU credits ($$).

The main offender is oc-bifrost which has been observed consuming 100% CPU making calls for CLOCK_MONOTONIC. With this change it only consumes 1% CPU.


jashaik:
@feld Thanks for the pull request, We will take a look and update you.

PR #4012, opened 2025-05-15
make pipelines public: verify

jaymzh:
Signed-off-by: Phil Dibowitz <phil@ipom.com>


sean-sype-simmons:
Will have to check with the team on the access to the environment secrets and if we can source those from somewhere else to make this public test work. Or if those parts of the file are even needed for a public test.

jaymzh:
@sean-sype-simmons What are you talking about? The line I'm removing in this file doesn't do anything. It's a hash. So the second line completely overwrites the first line.

PR #4152, opened 2026-02-24
[CHEF-29745]  local license check, default off 

marcparadise:
This changes adds a license check using a local license file. 

This behavior is controlled by a build-time compilation macro, OC_LICENSE_PATH
By default, this will have a value of 'cli', which will preserve the
existing license check behavior of using the automate cli when present.
Existing builds (including downstream) will therefore not be impacted. 

If the macro is set to a file path (this can be done by setting the
environment variable OC_LICENSE_PATH to the target location, prior to
build) then at run-time, erchef will expect to find a license file in
this location.

If the specified the license file is missing or invalid, it is treated as
a 90 day trial license from time of upgrade to the version that
implements this change.

When the file is present, the expiration is pulled from the file, based
on the entitlement end time furthest in the future. This was chosen
because the license content does not directly contain expiration date.


src/oc_erchef/apps/chef_license/test/chef_license_worker_test.erl
marcparadise, line 1: Something weird happened with the unit tests in one of my internal rebases,  so I have excluded them. I willl get them pushed up tomorrow.

src/oc_erchef/apps/chef_license/src/chef_license_worker.erl
```diff
@@ -79,35 +87,147 @@ code_change(_OldVsn, State) ->
...
+                   ]}}
+  ].
+
+
+
+%%%
 %%% =====================
 %%% Internal functions
 %%% =====================
-check_license(State) ->
```
marcparadise, line 85: This is just formatting changes.. but check_license is otherwise unchanged.

jashaik (approved):

kalroy (approved):

